PR #3, opened 2019-02-01
bindata/v3.11.0/etcd: update ports and volumes

hexfusion:
This PR updates ports to use etcd defaults for client and peer. I also adjusted the volumes to consume what is being created by [machine-config-operator]( https://github.com/openshift/machine-config-operator/blob/71ace53d656266d73656bbccaccec2b3bb6c07c2/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml)

### DIFF
--- a/bindata/v3.11.0/etcd/pod.yaml
+++ b/bindata/v3.11.0/etcd/pod.yaml
@@ -21,10 +21,22 @@ spec:
         memory: 200Mi
         cpu: 100m
     ports:
-      - containerPort: 10257
+      - containerPort: 2379
+        name: client
+        protocol: TCP
+    ports:
+      - containerPort: 2380
+        name: peer
+        protocol: TCP
     volumeMounts:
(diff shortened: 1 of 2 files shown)


bindata/v3.11.0/etcd/pod.yaml
```diff
@@ -21,10 +21,22 @@ spec:
...
+        protocol: TCP
     volumeMounts:
-    - mountPath: /etc/kubernetes/static-pod-resources
-      name: resource-dir
+    - name: discovery
+      mountPath: /run/etcd/
+    - name: certs
+      mountPath: /etc/ssl/etcd/
+    - name: data-dir
+      mountPath: /var/lib/etcd/
```
deads2k: I recommend against this.  You should be managing it yourself in that staticpod resource dir

hexfusion: ack

hexfusion:
going to rethink this per comments

PR #4, opened 2019-02-04
Fix deps for new CI

tnozicka:
We need https://github.com/openshift/library-go/pull/218 to get a fix for validating vendor/ directory to pass CI.

/cc @deads2k @hexfusion


tnozicka:
/hold
seems like we need to deal with glide not properly managing mercurial, will use this as a playground for a while :)

hexfusion:
Just noticed a few text files in the vendor is that something we generally vet?

tnozicka:
@hexfusion vendor/ is whatever `glide` puts there, we shouldn't touch it - and there is now a CI check to enforce it. also we don't prune vendor to used files only otherwise you'd have to be changing deps with every second change in /pkg ;)

hexfusion:
@tnozicka ok that is fine I manage/vet these non go files in other packages just checking protocol here, thanks.

PR #5, opened 2019-02-04
Add manifests for installation

deads2k:
Adds manifests for installation

### DIFF
--- a/manifests/0000_12_etcd-operator_01_config.crd.yaml
+++ b/manifests/0000_12_etcd-operator_01_config.crd.yaml
@@ -1,0 +1,16 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: etcds.operator.openshift.io
+spec:
+  scope: Cluster
+  group: operator.openshift.io
+  version: v1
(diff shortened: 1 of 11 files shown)


manifests/0000_12_etcd-operator_01_config.crd.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: etcds.operator.openshift.io
+spec:
+  scope: Cluster
+  group: operator.openshift.io
```
hexfusion: I just wonder if we should make this EtcdCluster isn't Cluster `clusters.cluster.k8s.io`?

deads2k: > I just wonder if we should make this EtcdCluster isn't Cluster `clusters.cluster.k8s.io`?

`oc get etcd.operator.openshift.io`

deads2k:
`oc apply -f ../etcd-man/manifests/`kind of works

deads2k:
update the deployment to point to the images you built and pushed.

PR #6, opened 2019-02-04
fix the operator config type

deads2k:
Fix the bootstrapping code required to ensure that operator starts


hexfusion:
looking at unit test failure

hexfusion:
Do we need to adjust the test data to reflect new CRD?
https://github.com/openshift/cluster-etcd-operator/blob/472095a447a1a611424bf03f936679ed5b8f171b/pkg/cmd/render/testdata/configs/kube-controller-manager-config-overrides.yaml

hexfusion:
Sort of confusing the role KubeControllerManagerConfig plays in this.

hexfusion:
taking a stab at resolution https://github.com/openshift/cluster-etcd-operator/pull/7

PR #7, opened 2019-02-04
bindata/bootkube/config: update naming to EtcdConfig

hexfusion:
It appears this change is a necessary bit left over from migration from boilerplate code.

### DIFF
--- a/bindata/bootkube/config/bootstrap-config-overrides.yaml
+++ b/bindata/bootkube/config/bootstrap-config-overrides.yaml
@@ -1,23 +1,3 @@
 apiVersion: kubecontrolplane.config.openshift.io/v1
-kind: KubeControllerManagerConfig
+kind: EtcdConfig
 extendedArguments:
-  root-ca-file:
-  - "/etc/kubernetes/secrets/root-ca.crt"
-  service-account-private-key-file:
-  - "/etc/kubernetes/secrets/service-account.key"
-  cluster-signing-cert-file:
-  - "/etc/kubernetes/secrets/kube-ca.crt"
-  cluster-signing-key-file:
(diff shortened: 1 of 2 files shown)


bindata/bootkube/config/bootstrap-config-overrides.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: kubecontrolplane.config.openshift.io/v1
-kind: KubeControllerManagerConfig
+kind: EtcdConfig
 extendedArguments:
```
deads2k: delete the content and you can tag and merge

bindata/bootkube/config/bootstrap-config-overrides.yaml
```diff
@@ -1,23 +1,3 @@
...
```
deads2k: this can go too.  I don't think etcd has a config file, does it?

bindata/bootkube/config/config-overrides.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: kubecontrolplane.config.openshift.io/v1
-kind: KubeControllerManagerConfig
+kind: EtcdConfig
 extendedArguments:
```
deads2k: you would be deleting it all in here too

bindata/bootkube/config/bootstrap-config-overrides.yaml
hexfusion: The server can have a [config file](https://github.com/etcd-io/etcd/blob/master/etcd.conf.yml.sample)  but I am not sure it would be useful here. I understand.

deads2k:
brought this in in a different pull

PR #9, opened 2019-04-22
Use CVO to manage operator config resource

damemi:
Ditch the EnsureOperatorConfigExists helper method in favor of tagging the operator CR to be managed by the CVO


deads2k:
this isn't used, just keep it open.

damemi:
no longer relevant, closing

PR #11, opened 2019-06-11
Include in release

beekhof:
As per https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/operators.md , we need to add this label to our container so that our manifests are included in the result of `oc adm release new`.


beekhof:
Wrong target

PR #30, opened 2019-11-05
manifests: Make cluster-etcd-operator Managed

retroflexer:
With the latest checkins on MCO, CEO, and installer, we are in a position to turn cluster-etcd-operator as `Managed` in 4.3.


hexfusion:
cleaning up old PRs this is no longer necessary

/close

PR #34, opened 2019-11-26
*: add support for IP based bootstrap endpoint

hexfusion:
This PR adds support for etcd-bootstrap to have a client address that is IP vs FQDN. The reason is that customers do not want to add additional DNS records as a dependency for building clusters.

### DIFF
--- a/pkg/operator/configobservation/etcd/observe_etcd.go
+++ b/pkg/operator/configobservation/etcd/observe_etcd.go
@@ -168,7 +168,6 @@ func ObservePendingClusterMembers(genericListers configobserver.Listers, recorde
 
 	if len(observer.ObservedPending) > 0 {
 		if err := unstructured.SetNestedField(observedConfig, observer.ObservedPending, observer.pendingPath...); err != nil {
-			klog.Errorf("etcdURLs > 0 ERRRPRRRRRR: %v", errs)
 			return existingConfig, append(errs, err)
 		}
 	}
@@ -369,9 +368,9 @@ func (e *etcdObserver) setBootstrapMember() error {
 	for _, subset := range endpoints.Subsets {
 		for _, address := range subset.Addresses {
 			if address.Hostname == "etcd-bootstrap" {
-				name := address.Hostname
-				peerURLs := fmt.Sprintf("https://%s.%s:2380", name, e.ClusterDomain)
-				clusterMember, err := setMember(name, []string{peerURLs}, ceoapi.MemberUnknown)
+				ip := address.IP
+				peerURLs := fmt.Sprintf("https://%s:2380", ip)
+				clusterMember, err := setMember(ip, []string{peerURLs}, ceoapi.MemberUnknown)
 				if err != nil {
 					return err
 				}
--- a/pkg/operator/hostetcdendpointcontroller/members.go
+++ b/pkg/operator/hostetcdendpointcontroller/members.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net"
+	"net/url"
 	"strings"
 	"time"
 
@@ -41,19 +43,33 @@ func (h *healthyEtcdMemberGetter) GetHealthyEtcdMembers() ([]string, error) {
 	}
 	hostnames := make([]string, 0)
 	for _, m := range member {
-		hostname := getEtcdName(m.PeerURLS)
+		hostname, err := getEtcdName(m.PeerURLS[0])
+		if err != nil {
+			return nil, err
+		}
 		hostnames = append(hostnames, hostname)
 	}
 	return hostnames, nil
 }
 
-func getEtcdName(peerURLs []string) string {
-	for _, peerURL := range peerURLs {
-		if strings.Contains(peerURL, "etcd-") {
-			return strings.TrimPrefix(strings.Split(peerURLs[0], ".")[0], "https://")
-		}
+// getEtcdName returns the name of the peer from a valid peerURL
+func getEtcdName(peerURL string) (string, error) {
+	if peerURL == "" {
+		return "", fmt.Errorf("getEtcdName: peerURL is empty")
 	}
-	return ""
+	if strings.Contains(peerURL, "etcd-") {
+		return strings.TrimPrefix(strings.Split(peerURL, ".")[0], "https://"), nil
+	}
+	u, err := url.Parse(peerURL)
+	if err != nil {
+		return "", err
+	}
+	host, port, _ := net.SplitHostPort(u.Host)
+	//TODO peer port should be a global constant
+	if IsIP(host) && port == "2380" {
+		return "etcd-bootstrap", nil
+	}
+	return "", fmt.Errorf("getEtcdName: peerURL %q is not properly formatted", peerURL)
 }
 
 func (h *healthyEtcdMemberGetter) EtcdList(bucket string) ([]ceoapi.Member, error) {
@@ -162,4 +178,12 @@ func (h *healthyEtcdMemberGetter) Endpoints() ([]string, error) {
 	}
 
 	return endpoints, nil
+}
+
+//TODO add to util
+func IsIP(addr string) bool {
+	if ip := net.ParseIP(addr); ip != nil {
+		return true
+	}
+	return false
 }
--- a/pkg/operator/hostetcdendpointcontroller/members_test.go
+++ b/pkg/operator/hostetcdendpointcontroller/members_test.go
@@ -18,25 +18,42 @@ func Test_getHostname(t *testing.T) {
 		peerURLs []string
 	}
 	tests := []struct {
-		name string
-		args args
-		want string
+		name    string
+		args    args
+		want    string
+		wantErr bool
 	}{
 		{
 			name: "valid test case for etcd member",
-			args: args{peerURLs: []string{"https://etcd-0.foouser.tests.com"}},
+			args: args{peerURLs: []string{"https://etcd-0.foouser.tests.com:2380"}},
 			want: "etcd-0",
 		},
 		{
 			name: "valid test case for etcd bootstrap node",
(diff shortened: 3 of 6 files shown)


pkg/operator/hostetcdendpointcontroller/members.go
```diff
@@ -41,19 +43,31 @@ func (h *healthyEtcdMemberGetter) GetHealthyEtcdMembers() ([]string, error) {
...
+			return "", err
+		}
+		host, port, _ := net.SplitHostPort(u.Host)
+		//TODO peer port should be a global constant
+		if isIP(host) && port == "2380" {
+			return "etcd-bootstrap", nil
 		}
 	}
-	return ""
+	return "", nil
```
hexfusion: is this an error? not found?

hexfusion:
/retest 

etcd came up ..

pkg/operator/hostetcdendpointcontroller/members.go
alaypatel07: yes, reaching that point is very unhealthy. I would think a lot of things are not going right for us to reach that state, it would be fair to panic at that point?

pkg/operator/hostetcdendpointcontroller/members.go
```diff
@@ -41,19 +43,31 @@ func (h *healthyEtcdMemberGetter) GetHealthyEtcdMembers() ([]string, error) {
...
 	}
 	return hostnames, nil
 }
 
-func getEtcdName(peerURLs []string) string {
+func getEtcdName(peerURLs []string) (string, error) {
 	for _, peerURL := range peerURLs {
 		if strings.Contains(peerURL, "etcd-") {
-			return strings.TrimPrefix(strings.Split(peerURLs[0], ".")[0], "https://")
+			return strings.TrimPrefix(strings.Split(peerURLs[0], ".")[0], "https://"), nil
```
alaypatel07: this existed in old code, my bad in not catching it earlier, but can we do?
```
strings.TrimPrefix(strings.Split(peerURL, ".")[0], "https://"), nil
```

pkg/operator/hostetcdendpointcontroller/members.go
```diff
@@ -163,3 +177,11 @@ func (h *healthyEtcdMemberGetter) Endpoints() ([]string, error) {
 
 	return endpoints, nil
 }
+
```
alaypatel07: question: would it be fair to add a TODO here to util this function for using it across the projects?

alaypatel07:
Few of small things:
1. we probably need to change IP address in discovery container, https://github.com/openshift/machine-config-operator/commit/a56735d1ede127255a774a0caa79310146be0c15#diff-f26a1d5dc0cdd6c1f8863a9a3d7a3630R177
2. we should also change the storage URLs as per https://github.com/openshift/cluster-kube-apiserver-operator/pull/673/files, although it will not break things, for consistency
3. squash the commits into a single logical commit?

pkg/operator/hostetcdendpointcontroller/members.go
hexfusion: agreed

hexfusion:
> Few of small things:
> 
> 1. we probably need to change IP address in discovery container, [openshift/machine-config-operator@a56735d#diff-f26a1d5dc0cdd6c1f8863a9a3d7a3630R177](https://github.com/openshift/machine-config-operator/commit/a56735d1ede127255a774a0caa79310146be0c15#diff-f26a1d5dc0cdd6c1f8863a9a3d7a3630R177)
> 2. we should also change the storage URLs as per https://github.com/openshift/cluster-kube-apiserver-operator/pull/673/files, although it will not break things, for consistency
> 3. squash the commits into a single logical commit?

these will be addressed

pkg/operator/hostetcdendpointcontroller/members.go
alaypatel07: s/peerURLs[0]/peerURL/

hexfusion: yeah gotcha

pkg/operator/configobservation/etcd/observe_etcd.go
```diff
@@ -369,9 +369,9 @@ func (e *etcdObserver) setBootstrapMember() error {
 	for _, subset := range endpoints.Subsets {
 		for _, address := range subset.Addresses {
 			if address.Hostname == "etcd-bootstrap" {
-				name := address.Hostname
-				peerURLs := fmt.Sprintf("https://%s.%s:2380", name, e.ClusterDomain)
-				clusterMember, err := setMember(name, []string{peerURLs}, ceoapi.MemberUnknown)
+				ip := address.IP
```
alaypatel07: drop the extra variable?

hexfusion: it's used more than once I think its fine IMO

pkg/operator/hostetcdendpointcontroller/members.go
hexfusion: done

pkg/operator/hostetcdendpointcontroller/members.go
hexfusion: put an error for now but we need to think about how we handle this.

pkg/operator/hostetcdendpointcontroller/members.go
hexfusion: done

hexfusion:
@alaypatel07 updated PTAL

hexfusion:
/hold 

adding IP address for bootstrap pod

hexfusion:
> Container test exited with code 2, reason OOMKilled

odd
/test unit

pkg/operator/hostetcdendpointcontroller/members.go
```diff
@@ -41,19 +43,31 @@ func (h *healthyEtcdMemberGetter) GetHealthyEtcdMembers() ([]string, error) {
...
+			return "", err
+		}
+		host, port, _ := net.SplitHostPort(u.Host)
+		//TODO peer port should be a global constant
+		if IsIP(host) && port == "2380" {
+			return "etcd-bootstrap", nil
 		}
 	}
-	return ""
+	return "", fmt.Errorf("peerURL %s is not properly formatted", peerURLs[0])
```
alaypatel07: s/peerURL/peerURLs?

pkg/operator/hostetcdendpointcontroller/members.go
```diff
@@ -163,3 +177,12 @@ func (h *healthyEtcdMemberGetter) Endpoints() ([]string, error) {
 
 	return endpoints, nil
 }
+
+//TODO add to util
+func IsIP(addr string) bool {
+	ip := net.ParseIP(addr)
```
alaypatel07: Can we please use the variable in the if block below?

pkg/operator/hostetcdendpointcontroller/members.go
hexfusion: that was intended as it was in single context `peerURLs[0]`

pkg/operator/hostetcdendpointcontroller/members_test.go
```diff
@@ -24,19 +24,19 @@ func Test_getHostname(t *testing.T) {
...
-			args: args{peerURLs: []string{"https://etcd-0.foouser.tests.com"}},
+			args: args{peerURLs: []string{"https://etcd-0.foouser.tests.com:2380"}},
 			want: "etcd-0",
 		},
 		{
 			name: "valid test case for etcd bootstrap node",
-			args: args{peerURLs: []string{"https://etcd-bootstrap.foouser.tests.com"}},
+			args: args{peerURLs: []string{"https://10.0.139.142:2380"}},
 			want: "etcd-bootstrap",
 		},
```
alaypatel07: Can we add couple of more test cases for errors? malformed url and empty peerURLs?

pkg/operator/hostetcdendpointcontroller/members.go
alaypatel07: This case will be hit only if the function does not return in the loop right? so it should be something like

`return "", fmt.Errorf("peerURLs %s is not properly formatted or empty", peerURLs)`

pkg/operator/hostetcdendpointcontroller/members_test.go
hexfusion: sure done

pkg/operator/hostetcdendpointcontroller/members.go
hexfusion: done

pkg/operator/hostetcdendpointcontroller/members.go
hexfusion: changed the function a little as we don't need to loop at all as we only handle [0]. this makes the function more explicit and error handling is more clear.

pkg/operator/hostetcdendpointcontroller/members.go
```diff
@@ -41,19 +43,33 @@ func (h *healthyEtcdMemberGetter) GetHealthyEtcdMembers() ([]string, error) {
 	}
 	hostnames := make([]string, 0)
 	for _, m := range member {
-		hostname := getEtcdName(m.PeerURLS)
+		hostname, err := getEtcdName(m.PeerURLS[0])
```
alaypatel07: Is it reasonable to assume that `len(m.PeerURLS) > 0`? if not this will be a nasty panic to catch

hexfusion: I think so we check that here https://github.com/openshift/cluster-etcd-operator/blob/f14a3303f3ff4807ffc3f8bd888b15786008e592/pkg/operator/hostetcdendpointcontroller/members.go#L109

alaypatel07: Perfect, thanks!

PR #35, opened 2019-11-26
*: add subcommand wait-for-ceo

alaypatel07:
This subcommand will be used in bootkube.sh to signal it is 
safe to destroy bootstrap vm and related resources.

It re-organises the teardown logic to reuse the same function
for removing etcd-bootstrap from etcd-membership

### DIFF
--- a/pkg/cmd/waitforceo/waitforceo.go
+++ b/pkg/cmd/waitforceo/waitforceo.go
@@ -1,0 +1,59 @@
+package waitforceo
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/bootstrapteardown"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
+)
+
+type waitForCeoOpts struct {
+	errOut     io.Writer
+	kubeconfig string
+}
+
+// NewWaitForCeoCommand waits for cluster-etcd-operator to bootstrap.
+func NewWaitForCeoCommand(errOut io.Writer) *cobra.Command {
+	waitForCeoOpts := &waitForCeoOpts{
+		errOut: errOut,
+	}
+	cmd := &cobra.Command{
+		Use:   "wait-for-ceo --FLAGS",
+		Short: "waits for cluster-etcd-operator to finish bootstrap etcd cluster",
+		Long:  "This command makes sure that the cluster-etcd-operator is available before signaling bootstrap is done",
+		Run: func(cmd *cobra.Command, args []string) {
+			must := func(fn func() error) {
+				if err := fn(); err != nil {
+					if cmd.HasParent() {
+						klog.Fatal(err)
+						fmt.Fprint(waitForCeoOpts.errOut, err.Error())
+					}
+				}
+			}
+			must(waitForCeoOpts.Run)
+		},
+	}
+
+	waitForCeoOpts.AddFlags(cmd.Flags())
+	return cmd
+}
+
+func (w *waitForCeoOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&w.kubeconfig, "kubeconfig", "", "kubeconfig for the cluster it is bootstrapping")
+}
+
+func (w *waitForCeoOpts) Run() error {
+	config, err := clientcmd.BuildConfigFromFlags("", w.kubeconfig)
+	if err != nil {
+		klog.Errorf("error loading kubeconfig: %#v", err)
+	}
+	if err := bootstrapteardown.WaitForEtcdBootstrap(context.TODO(), config); err != nil {
+		klog.Errorf("error waiting for cluster-etcd-operator %#v", err)
+	}
+	return nil
+}
--- a/pkg/operator/bootstrapteardown/teardown.go
+++ b/pkg/operator/bootstrapteardown/teardown.go
@@ -4,59 +4,21 @@ import (
 	"context"
 	"time"
 
-	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
-	configclient "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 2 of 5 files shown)


pkg/cmd/waitforceo/waitforceo.go
```diff
@@ -0,0 +1,59 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
+)
+
+type waitForCeoOpts struct {
+	errOut     io.Writer
+	kubeconfig string
+}
+
+// NewMountCommand creates a staticsync controller.
```
hexfusion: nit naming

pkg/cmd/waitforceo/waitforceo.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (w *waitForCeoOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&w.kubeconfig, "kubeconfig", "", "kubeconfig for the cluster it is bootstrapping")
+}
+
+func (w *waitForCeoOpts) Run() error {
+	config, err := clientcmd.BuildConfigFromFlags("", w.kubeconfig)
+	if err != nil {
+		klog.Errorf("error loadign kubeconfig: %#v", err)
```
retroflexer: minor nit: typo: s/loadign/loading/

pkg/cmd/waitforceo/waitforceo.go
alaypatel07: addressed, thanks!

pkg/cmd/waitforceo/waitforceo.go
alaypatel07: changed, thanks

alaypatel07:
/hold cancel 

added unit tests

alaypatel07:
@hexfusion rebased and cherry-picked the commits, PTAL, thanks!

pkg/operator/bootstrapteardown/teardown.go
```diff
@@ -4,61 +4,19 @@ import (
...
-			klog.Infof("Still waiting for the cluster to initialize...")
-			return false, nil
-		},
-	)
-
-	if err != nil {
-		klog.Errorf("error in watching clusterversions: %#v", err)
-	}
-
+	err := WaitForEtcdBootstrap(context.TODO(), config)
```
hexfusion: why do we disregard these errors?

pkg/operator/bootstrapteardown/waitforceo.go
```diff
@@ -0,0 +1,71 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	clientwatch "k8s.io/client-go/tools/watch"
+	"k8s.io/klog"
+)
+
+func WaitForEtcdBootstrap(ctx context.Context, config *rest.Config) error {
+	operatorConfigClient, err := operatorversionedclient.NewForConfig(config)
+	if err != nil {
+		return err
```
hexfusion: since we disregard this error should we print it as level 4 for debug?

pkg/operator/bootstrapteardown/waitforceo.go
```diff
@@ -0,0 +1,71 @@
...
+					return false, nil
+				}
+				return done(etcd)
+			}
+			klog.Infof("Still waiting for the cluster-etcd-operator to bootstrap...")
+			return false, nil
+		},
+	)
+
+	if err == nil {
```
hexfusion: I am not that familiar with `UntilWithSync` but what case will we actually error != nil?

alaypatel07: One case that comes to my mind is if the context passed is cancelled, it might lead to a non-nil error. Also, I don't see any reason for this to not follow the usual convention of Go in the if blocks being `if err != nil {..}` I will address this. Thanks

pkg/operator/bootstrapteardown/teardown.go
alaypatel07: +1, good catch

No reason to ignore the error should catch it with an if block

pkg/operator/bootstrapteardown/waitforceo.go
alaypatel07: ack, will add logging

alaypatel07:
@hexfusion my assumption is that if we don't go to `Available: True, Progressing; False and Degraded: False` state by 30 mins, we let the installer timeout and declare this a failed install attempt?

Irrespective of whether the installer has succeeded or failed, the CEO will try to remove etcd-bootstrap. We can add a timeout of 30 mins, but I was not sure if we can reliably coordinate between installer and CEO on when the counter starts. Keeping it async and disjoint was the simplist

pkg/operator/bootstrapteardown/teardown.go
alaypatel07: done

alaypatel07:
@hexfusion added a todo [here](https://github.com/openshift/cluster-etcd-operator/pull/35/files#diff-0da85577dd85ec13e04069679d5f9b1eR24) to see if we can time this function out in future and addressed other comments. PTAL, thanks

alaypatel07:
Prometheus and marketplace failing, unrelated to the code changes

/test e2e-aws

pkg/operator/bootstrapteardown/teardown.go
```diff
@@ -4,59 +4,21 @@ import (
...
 	operatorv1 "github.com/openshift/api/operator/v1"
-	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	etcdv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/clustermembercontroller"
-	cov1helpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/apimachinery/pkg/watch"
+
```
alaypatel07: ugh, goimports needs to be fix bad :(

PR #36, opened 2019-11-28
pkg/operator/etcdcertsigner: this fix only updates secret if empty

alaypatel07:
Before this fix every time the reconciliation loop ran, secret data
was updated, which was not ideal

### DIFF
--- a/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
+++ b/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
@@ -172,13 +172,13 @@ func (c *EtcdCertSignerController) sync() error {
 		return err
 	}
 
-	err = ensureCASecret(etcdCASecret)
+	err = ensureTLSData(etcdCASecret)
 	if err != nil {
 		klog.Errorf("etcd-signer ca secret invalid: %v", err)
 		return err
 	}
 
-	err = ensureCASecret(etcdMetricCASecret)
+	err = ensureTLSData(etcdMetricCASecret)
 	if err != nil {
 		klog.Errorf("etcd-metric-signer ca secret invalid: %v", err)
 		return err
@@ -192,7 +192,7 @@ func (c *EtcdCertSignerController) sync() error {
 
 	err = c.populateSecret(getSecretName(peerOrg, scaling.PodFQDN), secretNamespace, pCert, pKey)
 	if err != nil {
-		klog.Errorf("unable to create peer secret %#v", err)
+		klog.Errorf("unable to popopulate peer secret %#v", err)
 		return err
 	}
 
@@ -202,7 +202,7 @@ func (c *EtcdCertSignerController) sync() error {
 
 	err = c.populateSecret(getSecretName(serverOrg, scaling.PodFQDN), secretNamespace, sCert, sKey)
 	if err != nil {
-		klog.Errorf("unable to create server secret %#v", err)
+		klog.Errorf("unable to populate server secret %#v", err)
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -367,12 +371,19 @@ func (c *EtcdCertSignerController) populateSecret(secretName, secretNamespace st
...
+			"tls.key": key.Bytes(),
+		}
+		klog.Warningf("secret %s/%s does not have valid data", secretNamespace, secretName)
+		klog.Infof("attempting to update secret %s/%s with valid certs", secretNamespace, secretName)
+		_, err = c.clientset.CoreV1().Secrets(secretNamespace).Update(secretCopy)
+		return err
 	}
-	_, err = c.clientset.CoreV1().Secrets(secretNamespace).Update(secret)
-	return err
+	klog.Infof("secret %s/%s has valid certs", secretNamespace, secretName)
```
hexfusion: how chatty is this logging?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -367,12 +371,19 @@ func (c *EtcdCertSignerController) populateSecret(secretName, secretNamespace st
...
-		"tls.key": key.Bytes(),
+	if err := ensureTLSData(secret); err != nil {
+		secretCopy := secret.DeepCopy()
+		secretCopy.Data = map[string][]byte{
+			"tls.crt": cert.Bytes(),
+			"tls.key": key.Bytes(),
+		}
+		klog.Warningf("secret %s/%s does not have valid data", secretNamespace, secretName)
+		klog.Infof("attempting to update secret %s/%s with valid certs", secretNamespace, secretName)
+		_, err = c.clientset.CoreV1().Secrets(secretNamespace).Update(secretCopy)
```
hexfusion: this error would overwrite the original is that intended?

alaypatel07: yes, I think so. I will log the err getting overridden in line 380

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
alaypatel07: This will only be enabled if the operator is started with INFO, honestly I am not sure about it.

IMO the convention is to log the errors with Errorf, and other things with a combination of Debugf, Infof and Warningf. This felt like an Infof to me.

I will read up on the log levels of klog and get back with a more precise answer

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -367,12 +371,19 @@ func (c *EtcdCertSignerController) populateSecret(secretName, secretNamespace st
...
-	secret.Data = map[string][]byte{
-		"tls.crt": cert.Bytes(),
-		"tls.key": key.Bytes(),
+	if err := ensureTLSData(secret); err != nil {
+		secretCopy := secret.DeepCopy()
+		secretCopy.Data = map[string][]byte{
+			"tls.crt": cert.Bytes(),
+			"tls.key": key.Bytes(),
+		}
+		klog.Warningf("secret %s/%s does not have valid data", secretNamespace, secretName)
```
alaypatel07: need to log the error here

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: I think we can adjust if it's too chatty.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
alaypatel07: done

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
alaypatel07: sure, marking it resolved for now

alaypatel07:
```
2019/12/03 20:00:23 Container test in pod e2e-gcp-upgrade failed, exit code 1, reason Error
2019/12/03 20:00:48 No custom metadata found and prow metadata already exists. Not updating the metadata.
```

PR #38, opened 2019-12-04
WIP cmd/cluster-etcd-operator: enable certsigner command

hexfusion:
Moving cert signer from kubecsr to ceo.


hexfusion:
We will want to do this but in 4.5

PR #39, opened 2019-12-04
*: minor changes to bootstrap successfully

alaypatel07:
- added fix to set Degraded status so bootstrap teardown works
- added storageURLs back
- fix observe_etcd to take ip address for etcd-bootstrap

### DIFF
--- a/pkg/operator/clustermembercontroller/clustermembercontroller.go
+++ b/pkg/operator/clustermembercontroller/clustermembercontroller.go
@@ -239,34 +239,44 @@ func (c *ClusterMemberController) sync() error {
 		c.eventRecorder.Warning("ScalingFailed", rerr.Error())
 		return rerr
 	}
-	klog.Infof("All cluster members observed, scaling complete!")
-	// report available
-	condAvailable := operatorv1.OperatorCondition{
-		Type:   operatorv1.OperatorStatusTypeAvailable,
-		Status: operatorv1.ConditionTrue,
-	}
-	condUpgradable := operatorv1.OperatorCondition{
-		Type:   operatorv1.OperatorStatusTypeUpgradeable,
-		Status: operatorv1.ConditionTrue,
-	}
-	condProgressing := operatorv1.OperatorCondition{
-		Type:   operatorv1.OperatorStatusTypeProgressing,
-		Status: operatorv1.ConditionFalse,
-	}
-	condDegraded := operatorv1.OperatorCondition{
-		Type:   operatorv1.OperatorStatusTypeDegraded,
-		Status: operatorv1.ConditionFalse,
+
+	originalSpec, _, _, err := c.operatorConfigClient.GetOperatorState()
+	if err != nil {
+		klog.Errorf("unable to get operator spec: %#v", err)
+		return err
 	}
 
-	if _, _, updateError := v1helpers.UpdateStatus(c.operatorConfigClient,
-		v1helpers.UpdateConditionFn(condAvailable),
-		v1helpers.UpdateConditionFn(condUpgradable),
-		v1helpers.UpdateConditionFn(condProgressing),
-		v1helpers.UpdateConditionFn(condDegraded)); updateError != nil {
-		klog.Infof("Error updating status %#v", err)
-		return updateError
-	}
+	if isClusterEtcdOperatorReady(originalSpec) {
+		// report available
+		condAvailable := operatorv1.OperatorCondition{
+			Type:   operatorv1.OperatorStatusTypeAvailable,
+			Status: operatorv1.ConditionTrue,
+		}
+		condUpgradable := operatorv1.OperatorCondition{
+			Type:   operatorv1.OperatorStatusTypeUpgradeable,
+			Status: operatorv1.ConditionTrue,
+		}
+		condProgressing := operatorv1.OperatorCondition{
+			Type:   operatorv1.OperatorStatusTypeProgressing,
+			Status: operatorv1.ConditionFalse,
+		}
+		condDegraded := operatorv1.OperatorCondition{
+			Type:   operatorv1.OperatorStatusTypeDegraded,
+			Status: operatorv1.ConditionFalse,
+		}
 
+		if _, _, updateError := v1helpers.UpdateStatus(c.operatorConfigClient,
+			v1helpers.UpdateConditionFn(condAvailable),
+			v1helpers.UpdateConditionFn(condUpgradable),
+			v1helpers.UpdateConditionFn(condProgressing),
+			v1helpers.UpdateConditionFn(condDegraded)); updateError != nil {
+			klog.Infof("Error updating status %#v", err)
+			return updateError
+		}
+		klog.Infof("All cluster members observed, scaling complete!")
+		return nil
+	}
+	klog.Infof("Wait for cluster-etcd-operator to get ready")
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -62,7 +62,7 @@ func NewConfigObserver(
 				),
 			},
 			//TODO enable after migratation from KAO
```
hexfusion: maybe drop TODO here?

pkg/operator/configobservation/etcd/observe_etcd.go
```diff
@@ -368,9 +375,8 @@ func (e *etcdObserver) setBootstrapMember() error {
 	for _, subset := range endpoints.Subsets {
 		for _, address := range subset.Addresses {
 			if address.Hostname == "etcd-bootstrap" {
-				ip := address.IP
-				peerURLs := fmt.Sprintf("https://%s:2380", ip)
-				clusterMember, err := setMember(ip, []string{peerURLs}, ceoapi.MemberUnknown)
+				peerURLs := fmt.Sprintf("https://%s:2380", address.IP)
```
hexfusion: do we check if IP has value?

alaypatel07: I think we check in hostendpoint controller quite exhaustively if IP is present. Will add a check to be extra safe like in KAO

hexfusion: thanks

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -248,11 +253,16 @@ func (c *ClusterMemberController) sync() error {
 		Type:   operatorv1.OperatorStatusTypeProgressing,
 		Status: operatorv1.ConditionFalse,
 	}
+	condDegraded := operatorv1.OperatorCondition{
```
alaypatel07: Theses conditions should only be applied if all the members are in ready state

alaypatel07:
/wip 

adding test cases for isClusterEtcdOperatorReady

pkg/operator/clustermembercontroller/clustermembercontroller.go
alaypatel07: addressed

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
alaypatel07: sure

alaypatel07:
/wip cancel

Added a couple of test cases

pkg/operator/clustermembercontroller/clustermembercontroller_test.go
```diff
@@ -1,13 +1,19 @@
 package clustermembercontroller
 
 import (
+	"bytes"
+	"encoding/json"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
-	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
alaypatel07: needs to go with other k8s imports

pkg/operator/clustermembercontroller/clustermembercontroller_test.go
```diff
@@ -88,3 +94,88 @@ func TestClusterMemberController_RemoveBootstrapFromEndpoint(t *testing.T) {
 		})
 	}
 }
+
+func getBytes(obj interface{}) ([]byte, error) {
+	var buf bytes.Buffer
+	enc := json.NewEncoder(&buf)
+	err := enc.Encode(obj)
```
alaypatel07: use the err variable in the if statement

pkg/operator/configobservation/etcd/observe_etcd.go
alaypatel07: added

alaypatel07:
I will squash the import changes into minor changes commit along with other nits I pointed out

pkg/operator/hostetcdendpointcontroller/hostendpointcontroller.go
```diff
@@ -184,9 +184,30 @@ func (h *HostEtcdEndpointController) getNewAddressSubset(addresses []corev1.Endp
...
+		return
+	}
+	for index, addr := range addresses {
+		if addr.Hostname == "etcd-bootstrap" && index != len(addresses)-1 {
+			e := addr
+			addresses = append(addresses[0:index], addresses[index+1:]...)
+			addresses = append(addresses, e)
+			return
+		}
+	}
```
retroflexer: I tested this change and looks good.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -573,3 +593,34 @@ func (c *ClusterMemberController) getResyncName(pods *corev1.PodList) (string, e
...
+		klog.Errorf("error reading members: %#v", err)
+		return false
+	}
+	klog.Infof("observed members: %#v", members)
+	pendingMembers, _, err := unstructured.NestedSlice(existingConfig, membersPendingPath...)
+	if err != nil {
+		klog.Errorf("error reading pending members: %#v", err)
+		return false
+	}
+	if len(pendingMembers) > 0 {
```
hexfusion: Question: what if we are missing a node somehow? Don't we want to verify N are healthy? Just want to make sure this is enough to say we have succeeded.

hexfusion: I know we have this issue of N state do we need to deal with that now?

alaypatel07: I would like to deal with the node issue later on with CI in place honestly, what do you think? 

even if we handle it right now how do we test it? it would open up many issues IMO, fight for another day :)

hexfusion: I think we can get this N from installer.

N := installConfig.Config.ControlPlane.Replicas+1

hexfusion: probably should be a flag into render and added to config

hexfusion: I am ok with pushing this a little bit but it should be a p0 effort as we need this state for Verizon I bet.

hexfusion: https://github.com/openshift/installer/blob/79dae77251693ad5eac0f3619efa983e589121a1/pkg/asset/manifests/operators.go#L169

alaypatel07: I agree with p0 effort

I think we should save it in `etcds.Spec` so that we survive across restarts it. Would it make sense to have the render function spit out an etcd cr yaml with managed/unmanaged and appropriate number of replicas? 

So all we have to do in the operator is read it from the Spec and compare...

hexfusion: yes

pkg/operator/hostetcdendpointcontroller/hostendpointcontroller.go
alaypatel07: thanks

PR #42, opened 2019-12-05
render: add support for parsing SDN CIDR

hexfusion:
This PR adds the start of IPv6 support for the operator.

Also added issue to track missing  tests https://github.com/openshift/cluster-etcd-operator/issues/43

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -5,16 +5,19 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/ghodss/yaml"
 	"github.com/openshift/cluster-etcd-operator/pkg/cmd/render/options"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/v430_00_assets"
 	"github.com/openshift/library-go/pkg/assets"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	yaml "gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog"
 )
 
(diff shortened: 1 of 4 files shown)


hexfusion:
Manually tested and it works very well. The rest of support for ipv6 will come from a PR to  setup-etcd-env`

pkg/cmd/render/render.go
```diff
@@ -275,6 +296,55 @@ func mustReadTemplateFile(fname string) options.Template {
...
+	if err := discoverCIDRsFromNetwork(clusterConfigFileData, renderConfig); err != nil {
+		if err = discoverCIDRsFromClusterAPI(clusterConfigFileData, renderConfig); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func discoverCIDRsFromNetwork(clusterConfigFileData []byte, renderConfig *TemplateData) error {
+	configJson, err := yaml.YAMLToJSON(clusterConfigFileData)
```
alaypatel07: the next 11 lines are the same for both discoverCIDRsFromNetwork and discoverCIDRsFromClusterAPI, why not a separate function that returns and *Unstructured and error?

pkg/cmd/render/render.go
```diff
@@ -275,6 +296,55 @@ func mustReadTemplateFile(fname string) options.Template {
...
+func discoverCIDRs(clusterConfigFileData []byte, renderConfig *TemplateData) error {
+	if err := discoverCIDRsFromNetwork(clusterConfigFileData, renderConfig); err != nil {
+		if err = discoverCIDRsFromClusterAPI(clusterConfigFileData, renderConfig); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func discoverCIDRsFromNetwork(clusterConfigFileData []byte, renderConfig *TemplateData) error {
```
alaypatel07: Why does this function take in a pointer to *TemplateData rather than returning []string, []string, error?

returning the addresses and assigning them separately will save us from creating a blank struct in testing [here](https://github.com/openshift/cluster-etcd-operator/pull/42/files#diff-ab2090bd61b2cf366b708a2ac6dd52adR109). IMO it is a better way of testing

alaypatel07:
Overall lgtm, I added some comments inline, they are not blockers for merging!!

pkg/cmd/render/render.go
hexfusion: this is taken from upstream KAO

PR #46, opened 2019-12-09
*: add wait-for-kube command

hexfusion:
When the static pod starts we need to wait for kube until any containers that require a k8s client start. This ensures the container has been scheduled with kube and will contain required ENV for client creation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,6 @@ VERSION_OVERRIDE:=$(shell git describe --abbrev=8 --dirty --always)
 HASH:=$(shell git rev-parse --verify 'HEAD^{commit}')
 BUILD_DATE:=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 GLDFLAGS=-X $(REPO)/pkg/version.versionFromGit=$(VERSION_OVERRIDE) -X $(REPO)/pkg/version.commitFromGit=$(HASH) -X $(REPO)/pkg/version.buildDate=$(BUILD_DATE)
-GOFALGS=
 
 all: build
 .PHONY: all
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -7,7 +7,6 @@ VERSION_OVERRIDE:=$(shell git describe --abbrev=8 --dirty --always)
 HASH:=$(shell git rev-parse --verify 'HEAD^{commit}')
 BUILD_DATE:=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 GLDFLAGS=-X $(REPO)/pkg/version.versionFromGit=$(VERSION_OVERRIDE) -X $(REPO)/pkg/version.commitFromGit=$(HASH) -X $(REPO)/pkg/version.buildDate=$(BUILD_DATE)
-GOFALGS=
```
hexfusion: this was giving a warning obviously not needed cleaned up here.

pkg/cmd/waitforkube/waitforkube.go
```diff
@@ -0,0 +1,57 @@
...
+			must(waitForKubeOpts.Run)
+		},
+	}
+
+	return cmd
+}
+
+func (w *waitForKubeOpts) Run() error {
+	wait.PollInfinite(retryDuration, func() (bool, error) {
+		if _, err := os.Stat(saTokenPath); os.IsNotExist(err) {
```
alaypatel07: is it sufficient enough to check that the env variables are populated with this?

Should we have an equivalent of `inCluster` function here and check for env variable to be safe?

hexfusion: > is it sufficient enough to check that the env variables are populated with this?

If the SA token has synced that means kube scheduled staticsync so yes.

>Should we have an equivalent of inCluster function here and check for env variable to be safe?

Where? we are waiting.

hexfusion: to be clear static sync is a daemonset

hexfusion: our container will not change its ENV unless it restarts right?

alaypatel07: yes ideally it will not change, but want to be sure with something like

```
if !inCluster {
    fmt.Errorf("kube env not populated")
}
```

Where inCluster is 
```
func inCluster() bool {
	if os.Getenv("KUBERNETES_SERVICE_HOST") == "" || os.Getenv("KUBERNETES_SERVICE_PORT") == "" {
		return false
	}
	return true
}
```

We can either do it after the wait loop, so the initContainer will crashlooping until the env is populated.

if the env is not populated, it is not a bug against us

PR #52, opened 2020-01-04
Bug 1788231: pkg/cmd/render: improve IPv6 support

hexfusion:
Second iteration of IPv6 support includes a refactor of render and improved unit tests.

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
@@ -43,41 +43,38 @@ spec:
 
       source /run/etcd/environment
 
-      # TODO remove when this is handeled in setup-etcd-env
-      ETCD_IP_ADDRESS=${ETCD_IPV4_ADDRESS:-$ETCD_IPV6_ADDRESS}
-
-      [ -e /etc/ssl/etcd/system:etcd-server:${ETCD_IP_ADDRESS}.crt -a \
-        -e /etc/ssl/etcd/system:etcd-server:${ETCD_IP_ADDRESS}.key ] || \
+      [ -e /etc/ssl/etcd/system:etcd-server:{{ .Hostname }}.crt -a \
+        -e /etc/ssl/etcd/system:etcd-server:{{ .Hostname }}.key ] || \
         kube-client-agent \
           request \
             --kubeconfig=/etc/kubernetes/kubeconfig \
             --orgname=system:etcd-servers \
             --assetsdir=/etc/ssl/etcd \
             --dnsnames={{.EtcdServerCertDNSNames}} \
-            --commonname=system:etcd-server:${ETCD_IP_ADDRESS} \
-            --ipaddrs=${ETCD_IPV4_ADDRESS},{{ .EtcdAddress.LocalHost }} \
+            --commonname=system:etcd-server:{{ .Hostname }} \
+            --ipaddrs=${ETCD_ESCAPED_IP_ADDRESS},{{ .EtcdAddress.LocalHost }} \
 
-      [ -e /etc/ssl/etcd/system:etcd-peer:${ETCD_IP_ADDRESS}.crt -a \
-        -e /etc/ssl/etcd/system:etcd-peer:${ETCD_IP_ADDRESS}.key ] || \
+      [ -e /etc/ssl/etcd/system:etcd-peer:{{ .Hostname }}.crt -a \
+        -e /etc/ssl/etcd/system:etcd-peer:{{ .Hostname }}.key ] || \
         kube-client-agent \
           request \
             --kubeconfig=/etc/kubernetes/kubeconfig \
             --orgname=system:etcd-peers \
             --assetsdir=/etc/ssl/etcd \
             --dnsnames={{.EtcdPeerCertDNSNames}} \
-            --commonname=system:etcd-peer:${ETCD_IP_ADDRESS} \
-            --ipaddrs=${ETCD_IPV4_ADDRESS} \
+            --commonname=system:etcd-peer:{{ .Hostname }} \
+            --ipaddrs=${ETCD_ESCAPED_IP_ADDRESS} \
 
-      [ -e /etc/ssl/etcd/system:etcd-metric:${ETCD_IP_ADDRESS}.crt -a \
-        -e /etc/ssl/etcd/system:etcd-metric:${ETCD_IP_ADDRESS}.key ] || \
+      [ -e /etc/ssl/etcd/system:etcd-metric:{{ .Hostname }}.crt -a \
+        -e /etc/ssl/etcd/system:etcd-metric:{{ .Hostname }}.key ] || \
         kube-client-agent \
           request \
             --kubeconfig=/etc/kubernetes/kubeconfig \
             --orgname=system:etcd-metrics \
             --assetsdir=/etc/ssl/etcd \
             --dnsnames={{.EtcdServerCertDNSNames}} \
-            --commonname=system:etcd-metric:${ETCD_IP_ADDRESS} \
-            --ipaddrs=${ETCD_IP_ADDRESS} \
+            --commonname=system:etcd-metric:{{ .Hostname }} \
+            --ipaddrs=${ETCD_ESCAPED_IP_ADDRESS} \
     terminationMessagePolicy: FallbackToLogsOnError
     securityContext:
       privileged: true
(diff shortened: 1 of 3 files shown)


pkg/cmd/render/render.go
```diff
@@ -181,142 +177,114 @@ func (r *renderOpts) Run() error {
...
+			ListenMetricServer: net.JoinHostPort(net.IPv6zero.String(), "9978"),
+			ListenMetricProxy:  net.JoinHostPort(net.IPv6zero.String(), "9979"),
 		}
-		data, err := ioutil.ReadFile(file.source)
-		if err != nil {
-			return nil, err
+	} else {
+		// IPv4
+		etcdAddress = options.EtcdAddress{
+			ListenClient:       net.JoinHostPort(net.IPv4zero.String(), "2379"), // 0.0.0.0:2370
```
danwinship: (typo in the port number in the comment)

danwinship:
This is currently totally blocking IPv6-on-4.4 testing so it would be good to get it in soon even if it's not perfect yet.
/retest

hexfusion:
/hold

this still needs testing

hexfusion:
But as dan noted it is blocking

hexfusion:
/test e2e-aws

using payload for manual testing

bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
```diff
@@ -43,41 +43,38 @@ spec:
...
+        -e /etc/ssl/etcd/system:etcd-server:${ETCD_ESCAPED_IP_ADDRESS}.key ] || \
         kube-client-agent \
           request \
             --kubeconfig=/etc/kubernetes/kubeconfig \
             --orgname=system:etcd-servers \
             --assetsdir=/etc/ssl/etcd \
             --dnsnames={{.EtcdServerCertDNSNames}} \
-            --commonname=system:etcd-server:${ETCD_IP_ADDRESS} \
+            --commonname=system:etcd-server:${ETCD_ESCAPED_IP_ADDRESS} \
             --ipaddrs=${ETCD_IPV4_ADDRESS},{{ .EtcdAddress.LocalHost }} \
```
deads2k: since this env var no longer exists, I don't understand how this works.

bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
```diff
@@ -43,41 +43,38 @@ spec:
 
       source /run/etcd/environment
 
-      # TODO remove when this is handeled in setup-etcd-env
-      ETCD_IP_ADDRESS=${ETCD_IPV4_ADDRESS:-$ETCD_IPV6_ADDRESS}
-
-      [ -e /etc/ssl/etcd/system:etcd-server:${ETCD_IP_ADDRESS}.crt -a \
-        -e /etc/ssl/etcd/system:etcd-server:${ETCD_IP_ADDRESS}.key ] || \
+      [ -e /etc/ssl/etcd/system:etcd-server:${ETCD_ESCAPED_IP_ADDRESS}.crt -a \
```
deads2k: Why are we using IP addresses?  Did it just seem convenient for uniqueness?  How about node name instead which eliminates this use of IP addresses?

Migration could have a step that copies a file (if present) to the new name.

bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
```diff
@@ -43,41 +43,38 @@ spec:
...
+        -e /etc/ssl/etcd/system:etcd-server:${ETCD_ESCAPED_IP_ADDRESS}.key ] || \
         kube-client-agent \
           request \
             --kubeconfig=/etc/kubernetes/kubeconfig \
             --orgname=system:etcd-servers \
             --assetsdir=/etc/ssl/etcd \
             --dnsnames={{.EtcdServerCertDNSNames}} \
-            --commonname=system:etcd-server:${ETCD_IP_ADDRESS} \
+            --commonname=system:etcd-server:${ETCD_ESCAPED_IP_ADDRESS} \
```
deads2k: is this the common name of the certificate that was signed above?  If so, I think you should retrieve it from the cert, not count on the cert to be matched in a particular way. This constrains future migration choice.

bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
```diff
@@ -43,41 +43,38 @@ spec:
...
             --assetsdir=/etc/ssl/etcd \
             --dnsnames={{.EtcdPeerCertDNSNames}} \
-            --commonname=system:etcd-peer:${ETCD_IP_ADDRESS} \
-            --ipaddrs=${ETCD_IPV4_ADDRESS} \
+            --commonname=system:etcd-peer:${ETCD_ESCAPED_IP_ADDRESS} \
+            --ipaddrs=${ETCD_ESCAPED_IP_ADDRESS} \
 
-      [ -e /etc/ssl/etcd/system:etcd-metric:${ETCD_IP_ADDRESS}.crt -a \
-        -e /etc/ssl/etcd/system:etcd-metric:${ETCD_IP_ADDRESS}.key ] || \
+      [ -e /etc/ssl/etcd/system:etcd-metric:${ETCD_ESCAPED_IP_ADDRESS}.crt -a \
```
deads2k: I think my lack of bash knowledge is showing here.  What is the difference between this stanza and the one above?

bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
hexfusion: > Why are we using IP addresses? Did it just seem convenient for uniqueness?

right

> How about node name instead which eliminates this use of IP addresses?

sounds reasonable as this is the bootstrap node. The -e flag is just checking the file exists but it also serves as validation that the TLS certs will match what is expected in SAN. On the bootstrap node this is not a concern.

bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
hexfusion: > is this the common name of the certificate that was signed above? If so, I think you should retrieve it from the cert, not count on the cert to be matched in a particular way. This constrains future migration choice.

The agent is requesting is sending a CSR based on the commonname defined. So this action against the`request` command is before certs exist.

bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
hexfusion: We are checking if the file exists. In doing so we are also validating that the values have not changed. In which case the cert would be invalid as $ETCD_ESCAPED_IP_ADDRESS is used in the IP SAN. On the master nodes, this is FQDN [1].

But logically this reads: if crt and key exist no-op (certs exist and SAN is probably still valid) or we need to request new certs.

[1] https://github.com/openshift/machine-config-operator/blob/2810704cf1e74e9a598fed158886c1fead39b33e/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml#L71

danwinship:
What needs to be done to get this to merge? If parts of it are not ready, is it possible to split out just the "make IPv6 work" part without any of the refactoring and get that merged? This is currently blocking us from being able to have IPv6 CI on 4.4. (It's not the *only* thing blocking that, or else @eparis would be standing over your shoulder asking "are we there yet?" over and over until you merged, but it will soon be the only bug *that we know of* that blocks IPv6 CI.)

pkg/cmd/render/render.go
```diff
@@ -181,142 +177,110 @@ func (r *renderOpts) Run() error {
...
+	allAddresses := "0.0.0.0"
+	localhost := "127.0.0.1"
 
-	file := new(File)
-	if err := yaml.Unmarshal(data, file); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal file into struct: %v", err)
+	// IPv6
+	if t.SingleStackIPv6 {
+		allAddresses = "[::]"
+		localhost = "[::1]"
```
danwinship: You're using `net.JoinHostPort` below, so the brackets here are redundant (and result in eventually generating `--listen-client-urls https://[[::]]:2379`)

hexfusion: nice, thanks

hexfusion:
> What needs to be done to get this to merge? If parts of it are not ready, is it possible to split out just the "make IPv6 work" part without any of the refactoring and get that merged? This is currently blocking us from being able to have IPv6 CI on 4.4. (It's not the _only_ thing blocking that, or else @eparis would be standing over your shoulder asking "are we there yet?" over and over until you merged, but it will soon be the only bug _that we know of_ that blocks IPv6 CI.)

point taken :), will resolve today.

deads2k:
I doubt this works in a week.

/lgtm

hexfusion:
@danwinship this solves some but not all the problems. we need this anyway so consider this a down payment.

PR #55, opened 2020-01-25
observe_etcd.go: fix numberOfInitContainers

alaypatel07:


alaypatel07:
Cannot believe!! First day after feature freeze and CI passes without a single flaky retest!

PR #56, opened 2020-01-25
clustermembercontroller.go: fix operator degraded condition, log message

alaypatel07:
The cluster-etcd-operator should not be reporting degraded when
it is scaling.

Fix the log message in IsStatus function and error message


alaypatel07:
/hold 

CI is down, don't want the bot to go nuts on re-runs!!

PR #67, opened 2020-01-30
hostendpointcontroller.go: add a hack to reduce timeouts on KAS rollout

alaypatel07:
Depends on #66


alaypatel07:
the gcp timeouts I think will settle down with #68 

/test e2e-gcp

Issue #24, opened 2019-10-28
Add MemberId and ClusterId to etcd observations.


hexfusion:
One very important part about CEO working correctly is that we can validate the members that we observe are actually the members that are expected.

By this I mean it could be possible for an admin to manually add an etcd member that is not actually part of the cluster. For example, nothing currently would stop an admin from adding compute to vpc and installing etcd and manually using the certs to use the Cluster API and add the member to the cluster.

While there is not much we can do about this as an admin who has access to the certs to do this action would have the permissions to do pretty much anything.

But CEO should not allow this. If a member is added to cluster that was not added by the operator it should be removed and or the operator should become degraded and fire an alert of some kind. We not only want this etcd in the cluster but we want to make a big enough problem for the admin that they understand this will not be accepted.

This is not the same problem as an admin manually adding a member to a node but I feel that this should also be discouraged. Anytime something is done to membership that is not actionable by CEO should cause a degraded state until that is resolved. By resolved that would mean the member is removed and added by CEO.

To manage this we need to know our memberids and that information when possible should be the output of MemberAdd. The only time this becomes problematic is scaling from 1->2 in which case we lose quorum briefly. I think a window could exist where we would not be able to write this to state before controller goes down. For this reason I think we need an API call to MemberList to allow us to backfill this data.

My feels are that we should add this to the ClusterMemberController. So that we can verify the source. But I think it could also be considered in observations but I am less keen on that idea.

`</brainDump>`

PR #73, opened 2020-02-01
Bug 1797897: pkg/cmd/waitforkube: don't wait if etcd has previously been initialed as member

hexfusion:
When we added `wait-for-kube` we made poor assumptions that we always want to wait. In certain circumstances such as disaster recovery or lights-out situations where all master nodes go down kube will never be available before etcd starts.

This PR resolves this chicken and egg by not waiting if an etcd datafile exists. In this case we can assume that etcd has been previously initialized. Thus we no longer should depend on kube to start.

blocks https://github.com/openshift/machine-config-operator/pull/1428

### DIFF
--- a/pkg/cmd/waitforkube/waitforkube.go
+++ b/pkg/cmd/waitforkube/waitforkube.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -14,15 +15,17 @@ import (
 const (
 	retryDuration = 10 * time.Second
 	saTokenPath   = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+	dataDir       = "/var/lib/etcd" //TODO should be flag or ENV
+	dbPath        = "member/snap/db"
 )
 
-type waitForKubeOpts struct {
+type waitForKube struct {
 	errOut io.Writer
 }
 
 // NewWaitForKubeCommand waits for kube to come up before continuing to start the rest of the containers.
 func NewWaitForKubeCommand(errOut io.Writer) *cobra.Command {
-	waitForKubeOpts := &waitForKubeOpts{
+	waitForKube := &waitForKube{
 		errOut: errOut,
 	}
 	cmd := &cobra.Command{
@@ -34,29 +37,51 @@ func NewWaitForKubeCommand(errOut io.Writer) *cobra.Command {
 				if err := fn(); err != nil {
 					if cmd.HasParent() {
 						klog.Fatal(err)
-						fmt.Fprint(waitForKubeOpts.errOut, err.Error())
+						fmt.Fprint(waitForKube.errOut, err.Error())
 					}
 				}
 			}
-			must(waitForKubeOpts.Run)
+			must(waitForKube.Run)
 		},
 	}
 
 	return cmd
 }
 
-func (w *waitForKubeOpts) Run() error {
+func (w *waitForKube) Run() error {
+	dataFilePath := filepath.Join(dataDir, dbPath)
+	if CheckEtcdDataFileExists(dataFilePath) {
+		klog.Infof("etcd has previously been initialized as a member")
+		return nil
+	}
 	wait.PollInfinite(retryDuration, func() (bool, error) {
 		if _, err := os.Stat(saTokenPath); os.IsNotExist(err) {
 			klog.Infof("waiting for kube service account resources to sync: %v", err)
 			return false, nil
 		}
+		klog.Infof("kube dependencies verified, service account resources found")
 		return true, nil
 	})
 	if !inCluster() {
-		return fmt.Errorf("kube env not populated")
+		return fmt.Errorf("container missing required env KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT")
 	}
 	return nil
+}
+
+// CheckEtcdDataFileExists checks for the existance of an etcd db file. If the db exists we can assume this etcd
+// has previously started or is currently being recovered. If true we can not assume kube exists yet or will
+// during the life of this container. For example disaster recovery.
+// TODO organize into a general util
+func CheckEtcdDataFileExists(dataFilePath string) bool {
+	if _, err := os.Stat(dataFilePath); err != nil {
+		if os.IsNotExist(err) {
+			klog.Infof("CheckEtcdDataFileExists(): %s does not exist, waiting for kube", dataFilePath)
+			return false
+		}
+		// print error for debug but technically we can't say the file does not exist
+		klog.Errorf("CheckEtcdDataFileExists(): %v", err)
+	}
+	return true
 }
 
 //TODO: add to util


hexfusion:
/hold

for testing

hexfusion:
```
I0201 18:45:08.958350       1 waitforkube.go:78] CheckEtcdDataFileExists(): /var/lib/etcd/member/snap/db does not exist, waiting for kube
I0201 18:45:18.959014       1 waitforkube.go:58] waiting for kube service account resources to sync: stat /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
[...]
I0201 18:47:58.959095       1 waitforkube.go:61] kube dependencies verified, service account resources found
```
logs are a lot more sane

pkg/cmd/waitforkube/waitforkube.go
```diff
@@ -34,31 +37,53 @@ func NewWaitForKubeCommand(errOut io.Writer) *cobra.Command {
...
 		},
 	}
 
 	return cmd
 }
 
-func (w *waitForKubeOpts) Run() error {
+func (w *waitForKube) Run() error {
+	dataFilePath := filepath.Join(dataDir, dbPath)
+	if CheckEtcdDataFileExists(dataFilePath) {
```
alaypatel07: How will this affect the CEO use case of recovering single-member corrupted data directory?

Ideally the workflow is: crashloop detected by the operator -> member pushed to Remove state -> staticpodcontroller will remove the data directory -> operator will scale down -> the pod will restart -> CheckEtcdDataFileExists returns false, and scale up happens successfully

Is there a possibility of hitting CheckEtcdDataFileExists before operator detects crashloop or before the staticpodcontroller blows up the directory? if so we risk hitting a possible race condition.

Regardless we should create a test against openshift/origin to check single-member recover on ever PR, will create the corresponding issue!

hexfusion: So workflow should be as you noted.

* `etcd-member` or `etcd-metrics` in crashloop backoff is observed by observation controller and status us set to MemberRemove[1]

* staticpodcontroller then stops the static pod and removes the directory[2].

> Is there a possibility of hitting CheckEtcdDataFileExists before operator detects crashloop or before the staticpodcontroller blows up the directory? if so we risk hitting a possible race condition.

If for some reason we hit `CheckEtcdDataFileExists` before we check for crashloop then the result is still valid. A peer exists until its data-dir is destroyed. we continue to etcd-member container and it crashloops controller removes after observation.

I think we can better handle the open areas of this potential race once we land https://github.com/openshift/cluster-etcd-operator/issues/24 as we can defer the status of the member more accurately.

So, for now, I believe this is needed for a few circumstances described https://github.com/openshift/cluster-etcd-operator/pull/73#issue-369909426

> Regardless we should create a test against openshift/origin to check single-member recover on ever PR, will create the corresponding issue!

I think this could fall into the disruptive test.

[1] https://github.com/openshift/cluster-etcd-operator/blob/6dcaeec687252d9bebe6cf97f6b31ca2a1273f66/pkg/operator/configobservation/etcd/observe_etcd.go#L449

[2] https://github.com/openshift/cluster-etcd-operator/blob/6dcaeec687252d9bebe6cf97f6b31ca2a1273f66/pkg/cmd/staticpodcontroller/staticpodcontroller.go#L233

[3] https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/cmd/staticpodcontroller/staticpodcontroller.go#L221

hexfusion: > How will this affect the CEO use case of recovering single-member corrupted data directory?

I will write up the various failure cases in UML and how they would be handled to make this more visually understandable.

pkg/cmd/waitforkube/waitforkube.go
```diff
@@ -34,31 +37,53 @@ func NewWaitForKubeCommand(errOut io.Writer) *cobra.Command {
...
 		if _, err := os.Stat(saTokenPath); os.IsNotExist(err) {
 			klog.Infof("waiting for kube service account resources to sync: %v", err)
 			return false, nil
 		}
+		klog.Infof("kube dependencies verified, service account resources found")
 		return true, nil
 	})
 	if !inCluster() {
-		return fmt.Errorf("kube env not populated")
+		return fmt.Errorf("container missing required env KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT")
```
tnozicka: if this is running as a static pods these envs are not present and this was blocking the pod startup as it is an init container

hexfusion: When kube comes up and the container restarts on this error these are then available. I have been using this as a quick sanity check. But I am open to other ideas.

hexfusion: If it didn't work we would never scale :)

hexfusion: An example where these are valid without a restart.

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_installer/3051/pull-ci-openshift-installer-master-e2e-aws/9952/artifacts/e2e-aws/pods/openshift-etcd_etcd-member-ip-10-0-130-38.us-east-2.compute.internal_wait-for-kube.log

hexfusion: But yes in recovery or lights out this will never happen that's is why we need the changes in this PR

pkg/cmd/waitforkube/waitforkube.go
alaypatel07: > If for some reason we hit CheckEtcdDataFileExists before we check for crashloop then the result is still valid. A peer exists until its data-dir is destroyed. we continue to etcd-member container and it crashloops controller removes after observation.

Makes sense, thanks for the explanation.

alaypatel07:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/73/pull-ci-openshift-cluster-etcd-operator-master-e2e-gcp-upgrade/198/artifacts/e2e-gcp-upgrade/must-gather/quay-io-openshift-origin-must-gather-sha256-dd03a9e0ac7b6e710037d8f0d0a5001b47c27bf859e49b0256e8ac54f5bd8198/cluster-scoped-resources/operator.openshift.io/etcds/cluster.yaml

```
      pending:
      - name: etcd-member-ci-op-qvxgm-m-1.c.openshift-gce-devel-ci.internal
        peerURLs: https://etcd-1.ci-op-fr818h9s-e4498.origin-ci-int-gce.dev.openshift.com:2380
        status: Unknown
      - name: etcd-member-ci-op-qvxgm-m-2.c.openshift-gce-devel-ci.internal
        peerURLs: https://etcd-2.ci-op-fr818h9s-e4498.origin-ci-int-gce.dev.openshift.com:2380
        status: Unknown
      - name: etcd-member-ci-op-qvxgm-m-0.c.openshift-gce-devel-ci.internal
        peerURLs: https://etcd-0.ci-op-fr818h9s-e4498.origin-ci-int-gce.dev.openshift.com:2380
        status: Ready
```

gcp upgrade failed because of a known bug, retrying

/test e2e-gcp-upgrade

hexfusion:
needs a bug I will update

hexfusion:
/hold
we are gutting this init lets see how long that takes.

PR #75, opened 2020-02-02
pkg/cmd: add membership command

hexfusion:

### DIFF
--- a/pkg/cmd/membership/membership.go
+++ b/pkg/cmd/membership/membership.go
@@ -1,0 +1,158 @@
+package membership
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/pkg/transport"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+
+	"github.com/openshift/cluster-etcd-operator/pkg/cmd/waitforkube"
+)
+
+const (
+	retryDuration     = 2 * time.Second
+	dialTimeout       = 5 * time.Second
+	etcdCertFile      = "/var/run/secrets/etcd-client/tls.crt"
+	etcdKeyFile       = "/var/run/secrets/etcd-client/tls.key"
+	etcdTrustedCAFile = "/var/run/configmaps/etcd-ca/ca-bundle.crt"
+	dbPath            = "member/snap/db"
+)
+
+type membership struct {
+	errOut        io.Writer
+	EtcdDnsName   string
+	EtcdEndpoints string
+	etcdDataDir   string
+}
+
+// NewWaitForKubeCommand waits for kube to come up before continuing to start the rest of the containers.
+func NewMembershipCommand(errOut io.Writer) *cobra.Command {
+	membership := &membership{
+		errOut: errOut,
+	}
+	cmd := &cobra.Command{
+		Use:   "membership",
+		Short: "cluster membership validate",
+		Long:  "This command makes sure that the local etcd member is part of the cluster.",
+		Run: func(cmd *cobra.Command, args []string) {
+			must := func(fn func() error) {
+				if err := fn(); err != nil {
+					if cmd.HasParent() {
+						klog.Fatal(err)
+						fmt.Fprint(membership.errOut, err.Error())
+					}
+				}
+			}
+			must(membership.Complete)
+			must(membership.Validate)
+			must(membership.Run)
+		},
+	}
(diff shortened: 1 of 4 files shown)


pkg/cmd/membership/membership.go
```diff
@@ -0,0 +1,158 @@
...
+	"strings"
+	"time"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/pkg/transport"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+	"path/filepath"
```
hexfusion: moving this up to system-level import.

pkg/cmd/membership/membership.go
```diff
@@ -0,0 +1,168 @@
...
+			return fmt.Errorf("membership validation requires etcd-dns-name and endpoints")
+		}
+	}
+	return nil
+}
+
+func (m *membership) Complete() error {
+	m.EtcdDataFile = filepath.Join(m.etcdDataDir, dbPath)
+
+	etcdDnsName, etcdEndpoints := getENV()
```
alaypatel07: why does `getENV` have to be a separate function, replace it with `x, y := os.Getenv("x"), os.Getenv("y")` construct?

pkg/cmd/membership/membership.go
```diff
@@ -0,0 +1,168 @@
...
+	etcdTrustedCAFile = "/var/run/configmaps/etcd-ca/ca-bundle.crt"
+	dbPath            = "member/snap/db"
+)
+
+type membership struct {
+	errOut         io.Writer
+	EtcdDnsName    string
+	EtcdEndpoints  string
+	etcdDataDir    string
+	EtcdDataFile   string
```
alaypatel07: does this have to be exported variable?

alaypatel07: can we have a function instead that calculates data file from data directory and dbfile constant? I think that we make unit testing this easier

pkg/cmd/membership/membership.go
```diff
@@ -0,0 +1,168 @@
...
+	return nil
+}
+
+func (m *membership) Run() error {
+	if m.IsExistingMember() {
+		klog.Infof("etcd has previously been initialized as a member")
+		return nil
+	}
+
+	cli, err := getEtcdClient(strings.Split(m.EtcdEndpoints, ","))
```
alaypatel07: I think we should have a thin wrapper in our repo that deals with etcd client and use that for membership controller and here, we already have an issue tracking this, will update it with a reference

pkg/cmd/membership/membership.go
```diff
@@ -0,0 +1,168 @@
...
+				return true, nil
+			}
+		}
+		klog.Infof("%s is currently not a member of the cluster %+v", m.EtcdDnsName, l.Members)
+		return false, nil
+	})
+
+	return nil
+}
+func (m *membership) IsExistingMember() bool {
```
alaypatel07: why have this function, it simply calls the underlying function, can we use `waitforkube.CheckEtcdDataFileExists(m.EtcdDataFile)` instead?

alaypatel07: and probably import waitforkube pack as `util` 📦 so it reads better

hexfusion: just readability is all, yes good idea on util

pkg/cmd/membership/membership.go
hexfusion: yeah, in general, we have a lot of technical assets that can be centralized and resused.

pkg/cmd/membership/membership.go
hexfusion: just readability

pkg/cmd/membership/membership.go
hexfusion: > does this have to be exported variable?

To keep this simple it needs to be exported because the value is populated in Complete() after struct is initialized. We could create another struct and initialize it but it seemed a bit much for this little script so I just used membership.

>can we have a function instead that calculates data file from data directory and dbfile constant? I think that will make unit testing this easier

OK we can do that but the field is exported so that it could be injected directly as well right?

pkg/cmd/membership/membership.go
hexfusion: We are essentially just doing this https://github.com/kubernetes/client-go/blob/master/rest/config.go#L456

retroflexer: In my opinion, I think it is more readable if it is `x, y := os.Getenv("x"), os.Getenv("y")` than as a separate function, unless we intend to do more in that function in future.

hexfusion:
@alaypatel07 I believe I have updated per your concerns.

PR #78, opened 2020-02-04
bug 1798219: use dynamic client instead of taking an MCO dependency

deads2k:
mco prevents go mod from working and this removes the one dep


mfojtik:
LGTM

hexfusion:
Nice to drop the dep here thanks @deads2k 

/lgtm

hexfusion:
cc @alaypatel07

abhinavdahiya:
/cc @runcom 

because he might get stuck supporting something when they change for igntion 3.

PR #83, opened 2020-02-04
[wip]: bug 1798586: pkg/*: Remove support for Unmanaged

retroflexer:
Remove any obedience of being `Unmanaged`. We are always `Managed`.


sttts:
Add a restriction to the config API. Compare how kube-apiserver does that via https://github.com/openshift/api/blob/master/operator/v1/0000_20_kube-apiserver-operator_01_config.crd.yaml-merge-patch.

manifests/0000_12_etcd-operator_01_config.crd.yaml
```diff
@@ -14,3 +14,10 @@ spec:
     - coreoperators
   subresources:
     status: {}
+  validation:
```
retroflexer: Based on @sttts' suggestion here:
https://github.com/openshift/cluster-etcd-operator/pull/83#issuecomment-582351754

sttts: From our slack discussion:

The CRD should not be in the ceo operator, but in openshift/api. Then you also get the schema generation for free (there is a makefile target for that). And you have to follow the approach kube-apiserver, i.e. using a yaml patch.

sttts: cc @damemi – maybe you can help @retroflexer with moving the CRD manifest over.

retroflexer: I will move the CRD manifest to openshift/api in a separate PR. I would let this PR go through without the CRD changes (I reverted the above change).

retroflexer:
Opened a new PR to use the latest rebase. Closing this.
https://github.com/openshift/cluster-etcd-operator/pull/89

PR #63, opened 2020-01-28
*: make teardown bootstrap controller

alaypatel07:
Before this, the bootstrap teardown function runs a single function
sync loop, i.e run once and return after done. This changes it to a 
controller. This introduces two status conditions: 
the `BootstrapSafeToRemove` and `BootstrapRemoved`. The
installer will wait for `BootstrapRemoved` condition to be `True`
before destroying the bootstrap vm. The `BootstrapRemoved` 
will only be `True` after the `BootstrapSafeToRemove` has gone 
`True`.

PR #84, opened 2020-02-05
Sixtythree

deads2k:
rebase https://github.com/openshift/cluster-etcd-operator/pull/63


deads2k:
aws passed, gcp flaked on a known OLM alert problem, nothing has merged in the interim.  We expect this to improve stability because we will achieve a level instead of firing once.  merging.

PR #88, opened 2020-02-05
bug 1798642: add basic static pod operator

deads2k:
If this goes green, we should merge it as a starting point.


deads2k:
@hexfusion this looks nice and green.  I plan to rebase tomorrow and get it in.

deads2k:
actually, this appears to be clean now.  I'll tag and give it a chance overnight and deal with more feedback tomorrow.

PR #89, opened 2020-02-05
Bug 1798586: pkg/*: Remove support for Unmanaged

retroflexer:
Remove any obedience of being `Unmanaged`. We are always `Managed`.


retroflexer:
@deads2k, @hexfusion, @alaypatel07  needed rebasing again after latest merges. Can you tag again? Thanks.

PR #90, opened 2020-02-05
Bug 1798696: etcd operator does not reference etcd images for cvo substitution

sanchezl:


sanchezl:
/hold cancel
no target controller yet

deads2k:
@sanchezl bootstrapping still broken.

PR #91, opened 2020-02-05
bug 1798740: stop setting global conditions to allow unions

deads2k:
some conditions in the etcd operator are too generic and are resulting in collisions with unioned conditions


deads2k:
merged in another PR

PR #92, opened 2020-02-05
*: refactor clustermembercontroller

alaypatel07:
This sync loop is a giant monolith, first attempt in refactoring this.
Probably needs to break into two controllers in the future, but trying to
get there organically.

### DIFF
--- a/pkg/operator/clustermembercontroller/clustermembercontroller.go
+++ b/pkg/operator/clustermembercontroller/clustermembercontroller.go
@@ -22,7 +22,6 @@ import (
 	"k8s.io/client-go/informers"
 	corev1client "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 )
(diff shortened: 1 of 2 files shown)


alaypatel07:
wip because I need to review adding events

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -38,6 +37,8 @@ const (
 	EtcdEndpointName               = "etcd"
 	ConditionBootstrapSafeToRemove = "BootstrapSafeToRemove"
 	ConditionBootstrapRemoved      = "BootstrapRemoved"
+	waitForKubeContainerNumber     = 0
+	EtcdClusterSize                = 3
```
hexfusion: This feels like it should be based on the expected number of master nodes as reported by machine api or similar? I understand for testing this works but a use-case for Verizon is installing a single master. Then a short time afterward reprovision the bootstrap node as a master and scale up to 3. Just something we should consider.

alaypatel07: yes, I planned to have a todo, that number needs to come from somewhere, maybe the installer. For now, is it okay if we make it equal to the number of masters (querying from kube-api and not hardcoded as it currently is)?

alaypatel07:
closing because big pictures remove this controller

PR #94, opened 2020-02-06
Bug 1800303: *: fix logs and events from cmc and cert signer

alaypatel07:


deads2k:
/hold

Units are broken

alaypatel07:
@deads2k fixed the broken units, PTAL

deads2k:
we will either fix this by merging the static pod operator approach or by reverting the operator, building the static pod approach and trying in 4.5.  either way, this is dead.

PR #97, opened 2020-02-06
[WIP]: etcdcertsignercontroller.go: refactor sync loop with status updates

alaypatel07:


alaypatel07:
closed because no point in refactoring when big pictures are tending towards removing the old controllers

PR #98, opened 2020-02-07
bug 1801769: use etcd static pod status to build host endpoints

sanchezl:
* List of etcd members inferred from etcd member pod readiness instead of querying etcd directly.
* `etcd-bootstap` host preserved if found in existing `host-etcd` endpoint.
* endpoint addresses always sorted by hostname, but `etcd-bootstrap` always last in list.
* endpoint address IP assigned sequentially. The IP of a specific member might change when the list of members changes, except for `etc-bootstrap`, whose IP is always preserved.
* `dns-suffix` annotation is always set to etcd discovery domain.


pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -31,7 +31,7 @@ func NewConfigObserver(
 	interestingNamespaces := []string{
 		operatorclient.GlobalUserSpecifiedConfigNamespace,
 		operatorclient.GlobalMachineSpecifiedConfigNamespace,
-		operatorclient.TargetNamespace,
+		"openshift-etcd",
```
hexfusion: Just curious why not use the constant?

https://github.com/openshift/cluster-etcd-operator/blob/2ffe45d0e01405f7e7e742e0b39a3292a678eb5f/pkg/operator/operatorclient/interfaces.go#L7

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -0,0 +1,342 @@
...
+
+	// list etcd member pods
+	etcdPodRequirement, err := labels.NewRequirement("k8s-app", selection.In, []string{"etcd"})
+	if err != nil {
+		return err
+	}
+	etcdPodSelector := labels.NewSelector().Add(*etcdPodRequirement)
+	pods, err := c.podLister.List(etcdPodSelector)
+
+	// get dns names of ready etc member pods
```
hexfusion: s/etc/etcd

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+
```
deads2k: only standard lib is separate.

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -0,0 +1,342 @@
...
+const (
+	workQueueKey = "key"
+)
+
+// HostEndpointsController maintains an Endpoints resource with
+// the dns names of the current etcd cluster members for use by
+// components unable to use the etcd service directly.
+type HostEndpointsController struct {
+	eventRecorder       events.Recorder
```
deads2k: nit: prefer eventrecorder, queue, cachestosync in a separate stanza below the rest

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	queue               workqueue.RateLimitingInterface
+	cachesToSync        []cache.InformerSynced
+	operatorClient      v1helpers.OperatorClient
+	machineConfigLister cache.GenericLister
```
deads2k: if this can come from infrastructures.config.openshift.io, use that instead.

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	if err != nil {
+		_, _, updateErr := v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:    "HostEndpointsDegraded",
+			Status:  operatorv1.ConditionTrue,
+			Reason:  "ErrorUpdatingHostEndpoints",
+			Message: err.Error(),
+		}))
+		if updateErr != nil {
+			c.eventRecorder.Warning("HostEndpointsErrorUpdatingStatus", updateErr.Error())
+			return updateErr
```
deads2k: return the original error.

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	if len(addresses) == 0 {
+		return fmt.Errorf("no etcd member pods are ready")
+	}
+
+	required := hostEndpointsAsset()
+
+	if required.Annotations == nil {
+		required.Annotations = map[string]string{}
+	}
+	required.Annotations["alpha.installer.openshift.io/dns-suffix"] = discoveryDomain
```
deads2k: wow, I did some ugly back then, eh?

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	for _, pod := range pods {
+		var ready bool
+		for _, condition := range pod.Status.Conditions {
+			if condition.Type == corev1.PodReady {
+				ready = condition.Status == corev1.ConditionTrue
+				break
+			}
+		}
+		if ready {
+			dnsName, err := c.getEtcdDNSName(discoveryDomain, pod.Status.HostIP)
```
hexfusion: with the possibility of transient DNS failures should we consider a retry here? If we error would that result in setting operator Degraded? Since we are within the service network this lookup is against the internal resolver right? The actual DNS lookups for this by kube-aspiserver would happen on the host network DNS resolver. I am looking for ways to mitigate DNS flakes that would set operator degraded when it is actually operating fine.

deads2k:
@hexfusion After we get ourselves fully onto static pods, I see next steps in this control loop as
 1. set actual IP addresses for the nodes
 2. always set *all* the master nodes, regardless of whether they have etcd pods on them or not
 3. re-use the logic to set a second endpoints object that only uses valid IPs for all masters
 4. update the kas-o to use this second endpoints object in preference to the original
 5. in 4.5, actively delete the original endpoints.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
deads2k: We don't generally create constants that are used in a few places and cannot change.  These four namespaces are generally the exception, but none of the resource names for instance.

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -0,0 +1,334 @@
...
+	}))
+	if updateErr != nil {
+		c.eventRecorder.Warning("HostEndpointsErrorUpdatingStatus", updateErr.Error())
+		return updateErr
+	}
+	return nil
+}
+
+func (c *HostEndpointsController) syncHostEndpoints() error {
+	discoveryDomain, err := c.getEtcdDiscoveryDomain()
```
deads2k: Wild idea Luis.  Does this still work if you put a bunch of bad values in?  If so, can we simply put the internal IP addresses of all the master nodes and the dns name of the bootstrap node and just always keeps it like that?

sanchezl: There should be no problem putting the actual IP addresses in there. As long as kube-apiserver's etcd client handle failing over non-ready nodes.

PR #100, opened 2020-02-07
bug 1800708: add credential pusher daemonset

deads2k:
add a daemonset to get certificates off of the node for migration purposes.


deads2k:
/hold

this may not necessary if we have the signers in cluster already in openshift-config

deads2k:
prefer https://github.com/openshift/cluster-etcd-operator/pull/101
/close

PR #101, opened 2020-02-07
bug 1800708: add new cert signer for new static pods

deads2k:
Adds a new means to create certificates that we can leverage to create our static pods.


deads2k:
included in https://github.com/openshift/cluster-etcd-operator/pull/103

PR #103, opened 2020-02-07
bug 1800313: the etcd process 

deads2k:
This wires our secrets and env vars to our etcd process

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -20,32 +20,37 @@ spec:
         #!/bin/sh
         set -euo pipefail
 
-        echo "NODE__NAME=NODE_NAME"
-        echo "NODE__ENVVAR_NAME=NODE_ENVVAR_NAME"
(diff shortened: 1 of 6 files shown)


deads2k:
This pushes the arc forward and gives everyone something to test with on Monday.  If we have problems, we'll push through.  This is very close overall now.

bindata/etcd/pod.yaml
```diff
@@ -20,47 +20,93 @@ spec:
...
-        #        --peer-client-cert-auth=true \
-        #        --advertise-client-urls=https://${NODE_NODE_ENVVAR_NAME_IP}:2379 \
-        #        --listen-client-urls=https://${LISTEN_ON_ALL_IPS}:2379 \
-        #        --listen-peer-urls=https://${LISTEN_ON_ALL_IPS}:2380 \
-        #        --listen-metrics-urls=https://${LISTEN_ON_ALL_IPS}:9978 \
+        exit 0
+
+        # add logic here to confirm that we are part of the etcd members (the controller added us).
+        # this is probably a golang command that tries to confirm for two minutes before exiting
+        # and prints nothing except for the ETCD_INITIAL_CLUSTER
```
hexfusion: can we put this logic in an init container to the Pod so we can have access to CEO commands?

PR #104, opened 2020-02-07
Bug 1800810: pkg/operator/clustermembercontroller: add WithBlock() to dial options

hexfusion:
New etcd clients that use gRPC 1.23+ require the use of `grpc.WithBlock()`or connection can close before call is made.

Also bumping timeout to use apiserver default and explicitly calling cancel() on ctx after every call.


hexfusion:
/hold

need to update this everywhere

hexfusion:
flake 
/test e2e-gcp-upgrade

hexfusion:
etcd came up I am not sure the failure here.

/test e2e-aws

PR #108, opened 2020-02-10
add code for cluster-member-controller-2

alaypatel07:
This adds the cluster-member-controller-2 according to #105 

Enhancements added:
- Report DNS errors, and set degraded conditions accordingly
- Check for etcd membership in the actual etcd API instead of
observations, with the restriction of scaling only one member 
at a time

The new controller is not instantiated yet.

PR #111, opened 2020-02-10
Remove MemberRemove from ConfigObserver

retroflexer:
Removing code for setting status to MemberRemove (ceoapi.MemberRemove). When a node is not available or Pod is crash looping, it needs Disaster Recovery procedures to recover.

### DIFF
--- a/pkg/operator/configobservation/etcd/observe_etcd.go
+++ b/pkg/operator/configobservation/etcd/observe_etcd.go
@@ -5,7 +5,6 @@ import (
 	"net"
 	"reflect"
 	"strings"
-	"time"
 
 	"github.com/cloudflare/cfssl/log"
 	ceoapi "github.com/openshift/cluster-etcd-operator/pkg/operator/api"
@@ -107,15 +106,7 @@ func ObserveClusterMembers(genericListers configobserver.Listers, recorder event
 
 			_, err := observer.listers.NodeLister.Get(nodeName)
 			if errors.IsNotFound(err) {
-				// if the node is no londer available we use the endpoint observatiopn
-				klog.Warningf("error: Node %s not found: adding remove status to %s ", nodeName, previousMember.Name)
-				clusterMember, err := setMember(previousMember.Name, previousMember.PeerURLS, ceoapi.MemberRemove)
-				if err != nil {
-					return existingConfig, append(errs, err)
-
-				}
-				observer.ObservedMembers = append(observer.ObservedMembers, clusterMember)
-				continue
+				klog.Warningf("error: Node %s not found", nodeName)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/operator/configobservation/etcd/observe_etcd.go
```diff
@@ -446,8 +439,17 @@ func (e *etcdObserver) setObservedEtcdFromEndpoint(bucket string) error {
 			status := ceoapi.MemberUnknown
 			switch {
 			case isPodCrashLoop(bucket, pod):
```
hexfusion: lets just make remove this case for now IMO.

We check if we have a member pending here. I believe that would be true if the readiness probe was failing on the crashloop pod.

func (c *ClusterMemberController) isClusterEtcdOperatorReady() bool {

pkg/operator/configobservation/etcd/observe_etcd.go
```diff
@@ -107,15 +106,8 @@ func ObserveClusterMembers(genericListers configobserver.Listers, recorder event
...
-				klog.Warningf("error: Node %s not found: adding remove status to %s ", nodeName, previousMember.Name)
-				clusterMember, err := setMember(previousMember.Name, previousMember.PeerURLS, ceoapi.MemberRemove)
-				if err != nil {
-					return existingConfig, append(errs, err)
-
-				}
-				observer.ObservedMembers = append(observer.ObservedMembers, clusterMember)
-				continue
+				klog.Warningf("error: Node %s not found", previousMember.Name)
+				return existingConfig, append(errs, err)
```
hexfusion: what does this do and do we want to do it? Let's use an example of a node being terminated by mistake.

retroflexer: We are returning the existingConfig. So, clustermember thinks nothing changed.

retroflexer: We could just continue so that observedMembers be sent without including the terminated node.

hexfusion: if a previous members pod was not found, we check the node if the node was not found we removed the member from cluster with MemberRemove. I think your idea makes sense.

retroflexer: How about appending err to errs? 
```append(errs, err)```

File: pkg/operator/configobservation/etcd/observe_etcd.go [hunk 3]
```diff
@@ -107,15 +106,7 @@ func ObserveClusterMembers(genericListers configobserver.Listers, recorder event
...
-				// if the node is no londer available we use the endpoint observatiopn
-				klog.Warningf("error: Node %s not found: adding remove status to %s ", nodeName, previousMember.Name)
-				clusterMember, err := setMember(previousMember.Name, previousMember.PeerURLS, ceoapi.MemberRemove)
-				if err != nil {
-					return existingConfig, append(errs, err)
-
-				}
-				observer.ObservedMembers = append(observer.ObservedMembers, clusterMember)
-				continue
+				klog.Warningf("error: Node %s not found", nodeName)
```
retroflexer: Should we append `err` to `errs` here?

```append(errs, err)```

deads2k:
I expected to see code deleted that reacted to this config observation.

/assign @alaypatel07

retroflexer:
> I expected to see code deleted that reacted to this config observation.
@deads2k Remember, Alay said he would take care of the cluster member controller changes in the PR he was working on: https://coreos.slack.com/archives/C851TKLLQ/p1581021645292500?thread_ts=1581021342.289700&cid=C851TKLLQ

alaypatel07:
@deads2k With the big picture commits[1][2], all traces of MemberRemove were removed. Do you want a separate PR that takes care of removing the MemberRemove?

@retroflexer I made sure that the code is not present PR[2] in the new clustermembercontroller2

1. https://github.com/openshift/cluster-etcd-operator/pull/126/commits/3f180c786251bcbcb661868c18fa9799f0c4123d
2. https://github.com/openshift/cluster-etcd-operator/pull/126/commits/822cd85d5ff4634cbd7c5e27d2b6307e11b64f90
3. https://github.com/openshift/cluster-etcd-operator/pull/108

PR #112, opened 2020-02-10
bug 1801397: use a static pod manager

deads2k:
This PR bring all the disparate parts together to see if we can get the whole up and running.

PR #117, opened 2020-02-11
add all etcd nodes to host-etcd resource

sanchezl:


pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -112,52 +112,38 @@ func (c *HostEndpointsController) sync() error {
...
 	}
 	required.Annotations["alpha.installer.openshift.io/dns-suffix"] = discoveryDomain
 
-	sort.Strings(addresses)
-	for i, address := range addresses {
-		required.Subsets[0].Addresses = append(required.Subsets[0].Addresses, corev1.EndpointAddress{
-			Hostname: strings.TrimSuffix(address, "."+discoveryDomain),
-			IP:       net.IPv4(byte(192), byte(0), byte(2), byte(i+1)).String(),
+	// create endpoint addresses for each node
+	pods, err := c.podLister.List(labels.Set{"k8s-app": "etcd"}.AsSelector())
```
deads2k: don't use pods.  Get the nodes

sanchezl: Done

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -112,52 +113,51 @@ func (c *HostEndpointsController) sync() error {
...
 		})
 	}
 
+	sort.Slice(*endpointAddresses, func(i, j int) bool {
+		return (*endpointAddresses)[i].Hostname < (*endpointAddresses)[j].Hostname
+	})
+
 	// if etcd-bootstrap exists, keep it (at the end of the list)
 	existing, err := c.endpointsLister.Endpoints("openshift-etcd").Get("host-etcd")
 	if err != nil && !errors.IsNotFound(err) {
```
deads2k: update this to the switch/case style from https://github.com/openshift/cluster-etcd-operator/pull/112

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
```diff
@@ -112,52 +113,51 @@ func (c *HostEndpointsController) sync() error {
...
-		var ready bool
-		for _, condition := range pod.Status.Conditions {
-			if condition.Type == corev1.PodReady {
-				ready = condition.Status == corev1.ConditionTrue
+	// create endpoint addresses for each node
+	nodes, err := c.nodeLister.List(labels.Set{"node-role.kubernetes.io/master": ""}.AsSelector())
+	if err != nil {
+		fmt.Errorf("unable to list expected etcd member nodes: %v", err)
+	}
+	endpointAddresses := &required.Subsets[0].Addresses
```
deads2k: instead of this, create an empty one and then do an assignment at the end.  it will be easier for slow people like me.

pkg/operator/hostendpointscontroller/host_endpoints_controller.go
deads2k: > update this to the switch/case style from #112

Lack of this results in stomping out the the bootstrap etcd and we fail to start.

deads2k:
also a verify failure.

deads2k:
fixed in https://github.com/openshift/cluster-etcd-operator/pull/123

PR #118, opened 2020-02-11
bug 1801877: simplify the bootstrapteardown controller

deads2k:
Make the bootstrap teardown controller stand on it's own without relying on the clustermember controller

### DIFF
--- a/pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
+++ b/pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
@@ -1,6 +1,7 @@
 package bootstrapteardown
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -9,10 +10,13 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 	operatorv1informers "github.com/openshift/client-go/operator/informers/externalversions"
 	operatorv1listers "github.com/openshift/client-go/operator/listers/operator/v1"
-	"github.com/openshift/cluster-etcd-operator/pkg/operator/clustermembercontroller"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+	"go.etcd.io/etcd/clientv3"
+	"go.etcd.io/etcd/pkg/transport"
+	"google.golang.org/grpc"
 	corev1 "k8s.io/api/core/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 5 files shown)


pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -246,15 +345,24 @@ func (c *BootstrapTeardownController) Run(stopCh <-chan struct{}) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	l, err := cli.MemberList(ctx)
+	cancel()
+	if err != nil {
+		return false, err
 	}
-	c.eventRecorder.Eventf("WaitingForEtcdToBeAvailable", "waiting for %s to be true", clustermembercontroller.ConditionBootstrapSafeToRemove)
-	return false
+
+	if len(l.Members) > 2 {
```
alaypatel07: This condition is not enough, we not only need to check for 2 members but also need to check if both of them are healthy.

deads2k: > This condition is not enough, we not only need to check for 2 members but also need to check if both of them are healthy.
> /cc @hexfusion

if we add one at a time, this is better than where we are.

Where can I see health?

alaypatel07:
```
	for _, m := range l.Members {
		statusResp, err := cli.Status(context.Background(), m.ClientURLs[0])
		if err != nil {
			klog.Warningf("member %s with etcdName %s is not healthy", m.Name,
				getEtcdName(m.PeerURLs))
			klog.Errorf("error reading %s health: %#v", m.Name, err)
			continue
		}
		klog.Infof("member %s is healtly with %d index", m.Name, statusResp.RaftIndex)
        }
```

whithout this the member remove calls will fail, but I can see how temporary we will go to degraded and as soon as 2nd member is healthy the remove calls will succeed and well not be degraded anymore!

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -23,64 +27,65 @@ import (
...
+		endpointLister:      kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().Endpoints().Lister(),
 
 		cachesToSync: []cache.InformerSynced{
-			operatorConfigClient.Informer().HasSynced,
+			operatorClient.Informer().HasSynced,
 			operatorInformers.Operator().V1().Etcds().Informer().HasSynced,
 			operatorInformers.Operator().V1().KubeAPIServers().Informer().HasSynced,
 			openshiftKubeAPIServerNamespacedInformers.Core().V1().ConfigMaps().Informer().HasSynced,
+			openshiftKubeAPIServerNamespacedInformers.Core().V1().Endpoints().Informer().HasSynced,
+			kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().Endpoints().Informer().HasSynced,
```
alaypatel07: out of curiosity, why don't we have informers for pods registered in TargetNamespace?

PR #119, opened 2020-02-11
Bug 1802219: add etcdcli that can be used across the project.

alaypatel07:


deads2k:
please create an API more like https://github.com/openshift/cluster-etcd-operator/pull/122

alaypatel07:
Thanks for the POC, I will follow through and add other wrappers

pkg/etcdcli/etcdcli.go
```diff
@@ -0,0 +1,180 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	membersResp, err := cli.MemberList(ctx)
+	cancel()
+	if err != nil {
+		return false, err
+	}
+
+	for _, member := range membersResp.Members {
+		if len(member.PeerURLs) > 0 && member.PeerURLs[0] == peerURL {
+			return true, err
```
deads2k: true, nil?

pkg/etcdcli/etcdcli.go
```diff
@@ -0,0 +1,180 @@
...
+func getInternalIPAddressForNodeName(node *corev1.Node) (string, error) {
+	for _, currAddress := range node.Status.Addresses {
+		if currAddress.Type == corev1.NodeInternalIP {
+			return currAddress.Address, nil
+		}
+	}
+	return "", fmt.Errorf("node/%s missing %s", node.Name, corev1.NodeInternalIP)
+}
+
+func (g *etcdClientGetter) MemberAdd(peerURL string) (bool, error) {
```
deads2k: what is the bool?

pkg/etcdcli/etcdcli.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	_, err = cli.MemberAdd(ctx, []string{peerURL})
+	if err != nil {
+		return false, err
+	}
+	return true, err
+}
+
+func (g *etcdClientGetter) MemberRemove(member string) (bool, error) {
```
deads2k: go doc the bool

alaypatel07: whether or not the operation was successful or not, it can be assumed with `err != nil`, removing it

pkg/etcdcli/etcdcli.go
```diff
@@ -0,0 +1,182 @@
...
+func (g *etcdClientGetter) getEtcdClient() (*clientv3.Client, error) {
+	if !g.nodeListerSynced() {
+		return nil, fmt.Errorf("node lister not synced")
+	}
+	if !g.endpointsListerSynced() {
+		return nil, fmt.Errorf("node lister not synced")
+	}
+
+	etcdEndpoints := []string{}
+	nodes, err := g.nodeLister.List(labels.Set{"node-role.kubernetes.io/master": ""}.AsSelector())
```
hexfusion: So we blindly add any master node as an endpoint? Why not just what etcd.svs reports as healthy? Depending on what we do down the road with the client we should try to dial in the backend.

What about if nothing reports healthy from svs endpoints we fall back on nodes and see if we have any luck? Just an observation.

pkg/etcdcli/etcdcli.go
```diff
@@ -0,0 +1,182 @@
...
+	cli, err := clientv3.New(*cfg)
+	if err != nil {
+		return nil, err
+	}
+	return cli, err
+}
+
+func getInternalIPAddressForNodeName(node *corev1.Node) (string, error) {
+	for _, currAddress := range node.Status.Addresses {
+		if currAddress.Type == corev1.NodeInternalIP {
```
hexfusion: NodeInternalIP type do we want to be so explicit here? Unless we document this as a requirement for OpenShift IE IPI. I am nervous to do this. This typing can lead to strange bugs.

hexfusion: cc @deads2k

hexfusion: This needs to be like a global config or something scattering these types around is frightening. Who is the authority?

PR #123, opened 2020-02-11
Bug 1801984: host-etcd endpoints should always contain all etcd nodes

deads2k:
fix minor comments in https://github.com/openshift/cluster-etcd-operator/pull/117


sanchezl:
/retitle Bug 1801984: host-etcd endpoints should always contain all etcd nodes

PR #126, opened 2020-02-12
[wip] Big picture 2

deads2k:


deads2k:
"don't remove the bootstrap etcd until the cluster has bootstrapped" doesn't work because it creates a cycle between removing the bootstrap and not removing it.

PR #137, opened 2020-02-13
Bug 1802748: old clustermembercontroller: use EtcdClient interface

alaypatel07:
This PR 
- refactors the clustermembercontroller to use EtcdClient
interface.
- it also fixes the etcd client prematurely cancelling context


deads2k:
see last commit in https://github.com/openshift/cluster-etcd-operator/pull/140.  you're using a cancelled context

PR #140, opened 2020-02-13
[wip] Big picture 5

deads2k:
bad context cancel


deads2k:
cleared bootstrapping, failed on image registry

/retest

deads2k:
doesn't pass upgrades because we cannot get our pods to start.

PR #147, opened 2020-02-14
Bug 1803269: render: fix listening addresses in IPv6 case

danwinship:
I pointed this out in #52 and @hexfusion marked the comment resolved but then I guess accidentally didn't actually fix it...

(The addresses are passed to `net.JoinHostPort` which will add its own brackets to them, so we don't want to add our own brackets.)

/assign @hexfusion

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -230,8 +230,8 @@ func (t *TemplateData) setEtcdAddress() {
 
 	// IPv6
 	if t.SingleStackIPv6 {
-		allAddresses = "[::]"
-		localhost = "[::1]"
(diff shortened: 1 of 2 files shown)


hexfusion:
> I pointed this out in #52 and @hexfusion marked the comment resolved but then I guess accidentally didn't actually fix it...
> 
> (The addresses are passed to `net.JoinHostPort` which will add its own brackets to them, so we don't want to add our own brackets.)
> 
bah sorry @danwinship please update the test as well, thanks!

hexfusion:
I will also point out that the new operator will break other areas of this, we are working to address.

PR #148, opened 2020-02-14
fix upgrade path for static pod manager

alaypatel07:
This adds a commit make sure the etcd pods rendered by static
pod manager sleeps until the MCO pods is not evicted. Since, 
CEO is the first to get upgraded and MCO is the last, we need to
wait in the interim. 

This also updates the readiness probe. The static pod controllers
in CEO will only progress if etcd pods are ready. With certs now 
available in the pods, we can use the `etcdctl` to check the health
of local etcd member.

This will break the installs until openshift/machine-config-operator#1465
does not land

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -72,9 +72,21 @@ spec:
         initial_cluster="${initial_cluster::-1}"
         echo $initial_cluster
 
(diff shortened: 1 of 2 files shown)


alaypatel07:
FTR this will only upgrade from 4.3 to 4.4.

Upgrades from 4.4 (old CEO) to 4.4(static pod manager CEO) will break

PR #150, opened 2020-02-17
bug 1801397: static pod manager

deads2k:
A possibly cleaner test history

/hold


deads2k:
install of master didn't finish with the cluster monitoring resource issue again.

PR #158, opened 2020-02-18
[test]: make mco quorum guard deployment unmanaged

alaypatel07:


alaypatel07:
the test https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/158/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws/705 failed because it took 2 mins for quorum guard pods to die when scaling down. Added a commit for explicitly killing the pods without any graceful period

alaypatel07:
all test bootstrapped, gcp and aws fired alerts checking if I hit them again, upgrade passed
/test e2e-aws
/test e2e-gcp
/test e2e-gcp-upgrade

alaypatel07:
/lgtm

so that boot keeps firing the runs for folks to look at in the morning

manifests/0000_12_etcd-operator_00_pdb-rolebinding.yaml
```diff
@@ -11,6 +11,9 @@ rules:
 - apiGroups: ["apps"]
   resources: ["deployments"]
   verbs: ["get", "watch", "list", "delete", "create", "update"]
+- apiGroups: ["config.openshift.io"]
+  resources: ["clusterversions"]
+  verbs: ["get", "watch", "list", "delete", "create", "update"]
```
deads2k: why?

pkg/operator/guardbudget/guard_budget_controller.go
```diff
@@ -106,9 +119,23 @@ func (c *GuardBudgetController) checkGuardBudget() error {
 		deploymentScaledDown = true
 	}
 
+	clusterVersion, err := c.clusterVersionLister.Get("version")
+	if err != nil {
+		return err
+	}
+	for _, override := range clusterVersion.Spec.Overrides {
```
deads2k: ah, yeah, we don't offer that option anymore. and you don't need delete powers

pkg/operator/guardbudget/guard_budget_controller.go
```diff
@@ -141,6 +168,27 @@ func (c *GuardBudgetController) checkGuardBudget() error {
...
+		clusterVersion.Spec.Overrides = []v1.ComponentOverride{
+			{
+				Name:      "etcd-quorum-guard",
+				Namespace: "openshift-machine-config-operator",
+				Group:     "apps/v1",
+				Kind:      "Deployment",
+				Unmanaged: true,
+			},
+		}
+		_, updateErr := c.configClient.ConfigV1().ClusterVersions().Update(clusterVersion)
```
deads2k: you can't modify the clusterversion resource.

pkg/operator/guardbudget/guard_budget_controller.go
```diff
@@ -111,11 +112,11 @@ func (c *GuardBudgetController) sync() error {
 func (c *GuardBudgetController) checkGuardBudget() error {
 	deploymentScaledDown := false
 	quorumGuardUnmanaged := true
-	etcdQGScale, mcoDeployErr := c.kubeClient.AppsV1().Deployments("openshift-machine-config-operator").GetScale("etcd-quorum-guard", metav1.GetOptions{})
+	etcdQG, mcoDeployErr := c.kubeClient.AppsV1().Deployments("openshift-machine-config-operator").Get("etcd-quorum-guard", metav1.GetOptions{})
```
deads2k: don't we want this to be deleted in the near future?  I can see it exists on both, but it will fail soon.

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -254,6 +254,9 @@ func (c *BootstrapTeardownController) isKASConfigValid(kasOperator *operatorv1.K
 	// appropriate storageConfig
 	for _, revision := range uniqueRevisions {
 		configMapNameWithRevision := fmt.Sprintf("%s-%d", configMapName, revision)
+		if revision == 0 {
+			configMapNameWithRevision = configMapName
```
deads2k: I think I prefer the "notfound" error I would expect.  This shouldn't do work until the revision is present, right?  You could choose to skip at a higher level.

deads2k:
nicely green though.

pkg/operator/guardbudget/guard_budget_controller.go
alaypatel07: removed

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
alaypatel07: modified it to skip if revision is 0, check for other node's revision, if all are 0, it would return false at the bottom, PTAL @deads2k

PR #160, opened 2020-02-18
bug 1801397: pkg/operator/guardbudget: scale down MCO quorum-guard

hexfusion:


hexfusion:
> level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"

/retest

PR #164, opened 2020-02-18
bug 1804439: clean up packages

deads2k:
This also proves that we're still working.

### DIFF
--- a/pkg/operator/clustermembercontroller/clustermembercontroller.go
+++ b/pkg/operator/clustermembercontroller/clustermembercontroller.go
@@ -1,4 +1,4 @@
-package clustermembercontroller2
+package clustermembercontroller
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -1,4 +1,4 @@
-package clustermembercontroller2
+package clustermembercontroller
```
sanchezl: Also:
* workqueue name
* controller starting / stopping log entry
* event recorder component
* condition types

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -1,4 +1,4 @@
-package etcdcertsigner2
+package etcdcertsigner
```
sanchezl: Same comment as per clustermembercontroller

PR #167, opened 2020-02-18
bug 1801938: use IPs only to configure new peers

deads2k:
This removes most of our DNS dependencies for building an etcd cluster.


retroflexer:
Assume it works well for IP6 as well, but we need to test (but that's a different effort).

deads2k:
/hold
/retest

Need to think about what happens when a node gets a new up address

alaypatel07:
@deads2k I have ideas regarding how to handle the change of IP node addresses. The assumption I am making is that the IP address of only 1 node will change at a time so we still have the quorum on 2 etcd members working. If this assumption breaks, we would have to resort to the DR restore script to get quorum back.

For handling the change in IP address, the CMC can look like:
1. get the list of all unhealthy etcd member.
2. if any unhealthy member found with`peerURL != https://<node-ip>:2380`, update the member using etcd client `MemberUpdate` function setting peerURL to `http://<node-ip>:2380` and return nil. We will check in the next sync loop if that improved the count of unhealthy members.
3. if len(unhealthyMembers) == 0, get the IP address of a new unready pod to add
4. add the pod to etcd membership if it is not already added.  


### Upgrade Strategy:
This steps above will not break upgrades from a version with DNS to a version with IP because the pods are rendered 1 node at a time. The new pods will have its corresponding etcd member go unhealthy, CMC will update them in step 2 and they will go ready.

### Downgrade Strategy:
In order to support the downgrade we need to add step 2 to current CMC so that when downgrade happens to DNS enabled cluster, CMC can update the peerURL to peer FQDN, one unhealthy member at a time.

PR #139, opened 2020-02-13
Bug 1802664: *: add controller to report etcd members status conditions

alaypatel07:
This PR adds a controller that triggers every minute to report
the status of etcd members. It queries the actual etcd membership
to get this information.

PR #168, opened 2020-02-18
Bug 1802664: *: fix etcdmemberscontroller

alaypatel07:
picked up the changes missed from #139

### DIFF
--- a/pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
+++ b/pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
@@ -6,18 +6,15 @@ import (
 	"strings"
 	"time"
 
+	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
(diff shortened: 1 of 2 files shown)


pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
```diff
@@ -129,44 +110,44 @@ func (c *EtcdMembersController) reportEtcdMembers() error {
 		}))
 		if updateErr != nil {
 			c.eventRecorder.Warning("EtcdMembersErrorUpdatingStatus", updateErr.Error())
-			return err
+			return updateErr
```
deads2k: don't return these errors, just keep going.

deads2k: you may aggregate them if you wish, but set everything

deads2k:
/lgtm
/hold 

holding for comment

pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
alaypatel07: done

PR #172, opened 2020-02-19
Bug 1804921: Exclude etcd operator from hosted control plane deployments

csrwng:
Prevents cvo from instantiating this operator in a hosted control plane deployment. See https://github.com/openshift/enhancements/pull/202


hexfusion:
@csrwng I assume this needs to be backported to 4.4 as well? I will start the process in case.

/cherrypick release-4.4

PR #175, opened 2020-02-20
pkg/cmd/render: populate bootstrapIP from local interfaces

hexfusion:
This PR removes the `discovery` init container from the bootstrap etcd instance and replaces its functionality with observations of the local systems interfaces.

We also move to etcd-host-2 service for populating endpoints for etcdcli

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
@@ -25,7 +25,7 @@ spec:
             --assetsdir=/etc/ssl/etcd \
             --dnsnames={{.EtcdServerCertDNSNames}} \
             --commonname=system:etcd-server:{{ .Hostname }} \
-            --ipaddrs={{ .EtcdAddress.BootstrapIP }},{{ .EtcdAddress.LocalHost }} \
+            --ipaddrs={{ .EtcdAddress.EscapedBootstrapIP }},{{ .EtcdAddress.LocalHost }} \
 
       [ -e /etc/ssl/etcd/system:etcd-peer:{{ .Hostname }}.crt -a \
         -e /etc/ssl/etcd/system:etcd-peer:{{ .Hostname }}.key ] || \
@@ -36,7 +36,7 @@ spec:
             --assetsdir=/etc/ssl/etcd \
             --dnsnames={{.EtcdPeerCertDNSNames}} \
             --commonname=system:etcd-peer:{{ .Hostname }} \
-            --ipaddrs={{ .EtcdAddress.BootstrapIP }} \
+            --ipaddrs={{ .EtcdAddress.EscapedBootstrapIP }} \
 
       [ -e /etc/ssl/etcd/system:etcd-metric:{{ .Hostname }}.crt -a \
         -e /etc/ssl/etcd/system:etcd-metric:{{ .Hostname }}.key ] || \
@@ -47,7 +47,7 @@ spec:
             --assetsdir=/etc/ssl/etcd \
             --dnsnames={{.EtcdServerCertDNSNames}} \
             --commonname=system:etcd-metric:{{ .Hostname }} \
-            --ipaddrs={{ .EtcdAddress.BootstrapIP }} \
+            --ipaddrs={{ .EtcdAddress.EscapedBootstrapIP }} \
     terminationMessagePolicy: FallbackToLogsOnError
     securityContext:
       privileged: true
@@ -69,7 +69,7 @@ spec:
       set -euo pipefail
 
       exec etcd \
-        --initial-advertise-peer-urls=https://{{ .EtcdAddress.BootstrapIP }}:2380 \
+        --initial-advertise-peer-urls=https://{{ .EtcdAddress.EscapedBootstrapIP }}:2380 \
         --cert-file=/etc/ssl/etcd/system:etcd-server:{{ .Hostname }}.crt \
         --key-file=/etc/ssl/etcd/system:etcd-server:{{ .Hostname }}.key \
         --trusted-ca-file=/etc/ssl/etcd/ca.crt \
@@ -78,7 +78,7 @@ spec:
         --peer-key-file=/etc/ssl/etcd/system:etcd-peer:{{ .Hostname }}.key \
         --peer-trusted-ca-file=/etc/ssl/etcd/ca.crt \
         --peer-client-cert-auth=true \
-        --advertise-client-urls=https://{{ .EtcdAddress.BootstrapIP }}:2379 \
+        --advertise-client-urls=https://{{ .EtcdAddress.EscapedBootstrapIP }}:2379 \
         --listen-client-urls=https://{{ .EtcdAddress.ListenClient }} \
         --listen-peer-urls=https://{{ .EtcdAddress.ListenPeer }} \
         --listen-metrics-urls=https://{{ .EtcdAddress.ListenMetricServer }} \
@@ -121,7 +121,7 @@ spec:
       set -euo pipefail
 
       exec etcd grpc-proxy start \
-        --endpoints https://{{ .EtcdAddress.BootstrapIP }}:9978 \
+        --endpoints https://{{ .EtcdAddress.EscapedBootstrapIP }}:9978 \
         --metrics-addr https://{{ .EtcdAddress.ListenMetricProxy }}\
         --listen-addr {{ .EtcdAddress.LocalHost }}:9977 \
         --key /etc/ssl/etcd/system:etcd-peer:{{ .Hostname }}.key \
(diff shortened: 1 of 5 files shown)


pkg/cmd/render/render.go
```diff
@@ -223,14 +235,45 @@ func (r *renderOpts) Run() error {
...
+
 func (t *TemplateData) setEtcdAddress() {
 	// IPv4
 	allAddresses := "0.0.0.0"
 	localhost := "127.0.0.1"
 
 	// IPv6
 	if t.SingleStackIPv6 {
-		allAddresses = "[::]"
+		allAddresses = "::"
```
deads2k: didnt' this just break us?

deads2k:
I'm ok with this. Update the `pkg/etcdcli` to use host-etcd-2 or this won't work.

pkg/cmd/render/render.go
hexfusion: we reverted localhost which was wrong

hexfusion:
> I'm ok with this. Update the `pkg/etcdcli` to use host-etcd-2 or this won't work.

Or just this?

https://github.com/openshift/installer/pull/3148/files#diff-ce82c1d8a44f7dfc41dfc024085ccfeeR136

hexfusion:
or not

bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
```diff
@@ -7,31 +7,6 @@ metadata:
     k8s-app: etcd
 spec:
   initContainers:
-  - name: discovery
```
deads2k: glad to see these go

pkg/cmd/render/render.go
```diff
@@ -190,6 +195,13 @@ func newTemplateData(opts *renderOpts) (*TemplateData, error) {
 	if err := templateData.setSingleStackIPv6(); err != nil {
 		return nil, err
 	}
+
+	// derive bootstrapIP from local interfaces if not passed by flag
+	if templateData.BootstrapIP == "" {
+		if err := templateData.setBootstrapIP(); err != nil {
```
deads2k: I'd like to have a goal of removing this.  I think it is reasonable for the installer coordinate the bootstrap IP.

pkg/etcdcli/etcdcli.go
```diff
@@ -61,8 +63,16 @@ func (g *etcdClientGetter) getEtcdClient() (*clientv3.Client, error) {
...
 		if addr.Hostname == "etcd-bootstrap" {
-			// etcd-bootstrap has a valid IP in host-etcd
-			etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", addr.IP))
+			// etcd-bootstrap has a valid IP in host-etcd-2
+			hostEtcd2, err := g.endpointsLister.Endpoints(operatorclient.TargetNamespace).Get("host-etcd-2")
+			if err != nil {
+				return nil, err
+			}
+			ip, ok := hostEtcd2.Annotations[BootstrapIPAnnotationKey]
+			if !ok || ip == "" {
```
deads2k: this is optional. If it's not present then you skip it.  Otherwise you will break during upgrades.

pkg/etcdcli/etcdcli.go
```diff
@@ -61,8 +63,16 @@ func (g *etcdClientGetter) getEtcdClient() (*clientv3.Client, error) {
...
+				return nil, fmt.Errorf("service/host-etcd-2 annotation %s missing IP address", BootstrapIPAnnotationKey)
+			}
+			etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", ip))
```
deads2k: need escape if ipv6.  Look at my ipv6 PR and dnshelpers

pkg/etcdcli/etcdcli.go
```diff
@@ -61,8 +63,16 @@ func (g *etcdClientGetter) getEtcdClient() (*clientv3.Client, error) {
 	}
 	for _, addr := range hostEtcd.Subsets[0].Addresses {
 		if addr.Hostname == "etcd-bootstrap" {
-			// etcd-bootstrap has a valid IP in host-etcd
-			etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", addr.IP))
+			// etcd-bootstrap has a valid IP in host-etcd-2
+			hostEtcd2, err := g.endpointsLister.Endpoints(operatorclient.TargetNamespace).Get("host-etcd-2")
```
deads2k: no, make this edit up above and just get the correct host-etcd-2 to start with.

pkg/etcdcli/etcdcli.go
hexfusion: we should check it to be sure but I escape the annotation value if IPv6

/home/remote/sbatsche/projects/openshift/installer/bin/openshift-install

pkg/etcdcli/etcdcli.go
hexfusion: yeah makes sense

pkg/etcdcli/etcdcli.go
deads2k: Don't escape the annotation.

pkg/etcdcli/etcdcli.go
```diff
@@ -55,14 +58,22 @@ func (g *etcdClientGetter) getEtcdClient() (*clientv3.Client, error) {
 		etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", internalIP))
 	}
 
-	hostEtcd, err := g.endpointsLister.Endpoints(operatorclient.TargetNamespace).Get("host-etcd")
+	hostEtcd, err := g.endpointsLister.Endpoints(operatorclient.TargetNamespace).Get("host-etcd-2")
 	if err != nil {
 		return nil, err
 	}
 	for _, addr := range hostEtcd.Subsets[0].Addresses {
 		if addr.Hostname == "etcd-bootstrap" {
```
deads2k: We never put bootstrap in endpoints for host-etcd-2

pkg/cmd/render/render.go
```diff
@@ -227,15 +236,48 @@ func (r *renderOpts) Run() error {
...
 	allAddresses := "0.0.0.0"
 	localhost := "127.0.0.1"
+	bootstrapIP := t.BootstrapIP
 
 	// IPv6
 	if t.SingleStackIPv6 {
-		allAddresses = "[::]"
+		allAddresses = "::"
 		localhost = "[::1]"
+		bootstrapIP = "[" + t.BootstrapIP + "]"
```
deads2k: Name it something else if it is escaped

deads2k: host-etcd-2 should have an unescaped value

hexfusion: yeah got too late I will step through this today,

hexfusion:
moved to 4.4 to see if CI will work for azure..

pkg/etcdcli/etcdcli.go
```diff
@@ -55,22 +58,26 @@ func (g *etcdClientGetter) getEtcdClient() (*clientv3.Client, error) {
...
-			etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", addr.IP))
-			break
+	bootstrapIP, ok := hostEtcd.Annotations[BootstrapIPAnnotationKey]
+	if !ok {
+		klog.V(2).Infof("service/host-etcd-2 is missing annotation %s", BootstrapIPAnnotationKey)
+	}
+	if bootstrapIP != "" {
+		// escape if IPv6
+		if net.ParseIP(bootstrapIP).To4() == nil {
+			bootstrapIP = "[" + bootstrapIP + "]"
```
deads2k: bad practice to have a variable named one thing, then mutated to mean something different.  After escaping, my understanding is that this is no longer an IP.

pkg/cmd/render/render.go
```diff
@@ -227,15 +237,47 @@ func (r *renderOpts) Run() error {
 	return WriteFiles(&r.generic, &templateData.FileConfig, templateData)
 }
 
+// setBootstrapIP gets a list of IPs from local interfaces and returns the first IPv4 or IPv6 address.
+func (t *TemplateData) setBootstrapIP() error {
```
deads2k: style-wise, we produce getters for this information and then assign the values.  We don't make mutating functions in general.

pkg/cmd/render/render.go
```diff
@@ -227,15 +237,47 @@ func (r *renderOpts) Run() error {
...
+
+	t.BootstrapIP = bootstrapIP
+	return nil
+}
+
 func (t *TemplateData) setEtcdAddress() {
 	// IPv4
 	allAddresses := "0.0.0.0"
 	localhost := "127.0.0.1"
+	bootstrapIP := t.BootstrapIP
```
deads2k: this is actually the escapedBootstrapIP

deads2k:
several things that should be fixed. nothing fatal that prevents us from unblocking now.

Please make a followup and if this requires a repush, fix here

/lgtm

hexfusion:
aws is blowing up after 2 mins..

PR #177, opened 2020-02-20
remove unused files from bootstrap

deads2k:
Trying to remove some confusing cruft that is no longer used.


deads2k:
@ironcladlou will probably do something smarter.

PR #179, opened 2020-02-20
bug 1805421: allow bootstrap server to terminate before removing etcd-bootstrap fr…

deads2k:
…om membership

After https://github.com/openshift/cluster-kube-apiserver-operator/pull/772 merges, we no longer need to have a complicated check for etcd endpoints being present in the kube-apiserver configuration because they will *always* be present in the kube-apiserver configuration or the static-pod kube-apiserver will not be running.

This allows the bootstrap teardown to simply wait until
1. bootstrap is finished
1. etcd-bootstrap is present
1. there are at least 4 etcd members
1. all members besides etcd-bootstrap are healthy
if all those conditions true, then the etcd-bootstrap can be removed.


deads2k:
/retest

matches our discussion in slack.  tagging.

PR #181, opened 2020-02-20
bug 1804960: Merge 4.5 into 4.4

sanchezl:
Bug 1805241 - ROKS: prevent etcd operator from being applied by CVO 
Bug 1804682 - existing PDB needs certs on disk
Bug 1805096 - Cluster etcd operator remove guard budget controller 
Bug 1802664: *: fix etcdmemberscontroller
Bug 1804439: add debugging to clustermembercontroller and clean up add bug
Bug 1804960: Revert: render: fix listening addresses in IPv6 case


deads2k:
we have a lot of catch up. this moves us there quickly.  Confirmed with eparis on the approach

PR #173, opened 2020-02-19
bug 1804913: make ipv6 support a little better

deads2k:
Updates the etcd operator to

1. sign for all internal node IP addresses. This will fix issues with nodes having multiple IPs and not being trusted
2. choose the correct preferred ip address based on the networks.config.openshift.io
3. properly escape ipv6 addresses in the static pod and in the internal etcd client

PR #182, opened 2020-02-20
Ipv6 4.4

deads2k:
https://github.com/openshift/cluster-etcd-operator/pull/173 on 4.4


deads2k:
The picks back are merging

PR #183, opened 2020-02-21
Skip member check in case of lights-out and upgrade

retroflexer:
Skip the wait for the member list to show up in case the cluster is already installed.

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -48,62 +48,65 @@ spec:
         #!/bin/sh
         set -euo pipefail
 
-        ETCDCTL="etcdctl --cacert=/etc/kubernetes/static-pod-resources/configmaps/etcd-serving-ca/ca-bundle.crt \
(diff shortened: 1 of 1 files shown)


bindata/etcd/pod.yaml
```diff
@@ -48,62 +48,65 @@ spec:
...
-                           --cert=/etc/kubernetes/static-pod-resources/secrets/etcd-all-peer/etcd-peer-NODE_NAME.crt \
-                           --key=/etc/kubernetes/static-pod-resources/secrets/etcd-all-peer/etcd-peer-NODE_NAME.key \
-                           --endpoints=${ALL_ETCD_ENDPOINTS}"
-        ${ETCDCTL} member list
-
-        echo "waiting for member ${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}..."
-        COUNT=30
-        while [ $COUNT -gt 0 ]; do
-          echo "current member list is..."
+        if [[ ! -d /var/lib/etcd/member/ ]]; then
```
retroflexer: This is the only check added in this PR. Rest of the code is identical (but tabbed under if condition).

deads2k:
Replaced by go code

PR #185, opened 2020-02-21
bug 1805807: call the discovery-etcd-initial-cluster

deads2k:
replace bash logic with a command.

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -54,56 +54,22 @@ spec:
                            --endpoints=${ALL_ETCD_ENDPOINTS}"
         ${ETCDCTL} member list
 
-        echo "waiting for member ${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}..."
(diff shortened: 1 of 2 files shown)


bindata/etcd/pod.yaml
```diff
@@ -54,56 +54,20 @@ spec:
...
         ${ETCDCTL} member list
 
-        echo "waiting for member ${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}..."
-        COUNT=30
-        while [ $COUNT -gt 0 ]; do
-          echo "current member list is..."
-          ${ETCDCTL} member list
-          echo ""
-          echo ""
+        export ETCD_INITIAL_CLUSTER=$(discover-etcd-initial-cluster \
```
deads2k: @sttts when `discover-etcd-intial-cluster` returns non-zero, I'd like to print it's output and exit

deads2k:
note to self, cherrypick will be bug 1805809

PR #184, opened 2020-02-21
bug 1805519: Merge master into release-4.4

sanchezl:
bug 1801938: add host-etcd-2 which will never list the bootstrap etcd as an endpoi…
bug 1805421: allow bootstrap server to terminate before removing etcd-bootstrap fr…

PR #189, opened 2020-02-21
get master and release-4.4 back in sync

sanchezl:
Sync `master` and `release-4.4` branches to avoid the need to create more merge commits, until we really want to split. 
* Merged master into release-4.4 (https://github.com/openshift/cluster-etcd-operator/pull/184).
* Now, merging release-4.4 into master (this PR). 
This is a zero-diff PR.


deads2k:
zero diff change.

@hexfusion don't merge out of order

/lgtm

PR #191, opened 2020-02-21
pkg/cmd/render: use machineNetwork CIDR to validate bootstrap IP

hexfusion:
This PR adds validation for the IPs we collect from local interfaces of bootstrap node. Instead of taking the first IPv4 address or IPv6 (single stack). We attempt to parse the MachineCIDR from installer assets. We then validate if the IP address is part of this network.

To do this we are adding a new flag  `--cluster-network-file` currently we pass the network-config 
 to the `--cluster-config` [1] This change will require installer update.

enhancements https://github.com/openshift/enhancements/pull/228

[1] https://github.com/openshift/installer/blob/98773b31eca6002e6f44375118d9eae8467cd016/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L135

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -1,6 +1,7 @@
 package render
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -10,10 +11,11 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/openshift/cluster-etcd-operator/pkg/cmd/render/options"
+	"github.com/openshift/cluster-etcd-operator/pkg/dnshelpers"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/etcd_assets"
 
 	"github.com/ghodss/yaml"
-	"github.com/openshift/cluster-etcd-operator/pkg/cmd/render/options"
 	"github.com/openshift/library-go/pkg/assets"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -36,6 +38,7 @@ type renderOpts struct {
 	setupEtcdEnvImage        string
 	kubeClientAgentImage     string
 	clusterConfigFile        string
+	clusterNetworkFile       string
 	bootstrapIP              string
 }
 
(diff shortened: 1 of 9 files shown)


hexfusion:
Going to restructure this

hexfusion:
/hold

needs enhancement and a few tweaks I am trying to piviot this etcd spec to something more inline with what a revision from installer would lay down

hexfusion:
I think the static pod might have to wait depending on id this is blocking IPv6 for metal.

pkg/cmd/render/render.go
```diff
@@ -312,8 +401,47 @@ func (t *TemplateData) getClusterConfigFromFile(clusterConfigFile string) (*unst
...
+			return nil
+		}
+		// IPv6
+		if !isIPV4 && t.SingleStackIPv6 {
+			t.MachineCIDR = networkString
+			return nil
+		}
+	}
+
+	// depricated
```
hexfusion: deprecated ..

pkg/cmd/render/render.go
```diff
@@ -81,7 +85,8 @@ func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&r.kubeClientAgentImage, "manifest-kube-client-agent-image", r.kubeClientAgentImage, "kube-client-agent manifest image")
 	fs.StringVar(&r.setupEtcdEnvImage, "manifest-setup-etcd-env-image", r.setupEtcdEnvImage, "setup-etcd-env manifest image")
 	fs.StringVar(&r.etcdDiscoveryDomain, "etcd-discovery-domain", r.etcdDiscoveryDomain, "etcd discovery domain")
-	fs.StringVar(&r.clusterConfigFile, "cluster-config-file", r.clusterConfigFile, "Openshift Cluster API Config file.")
+	fs.StringVar(&r.clusterNetworkFile, "cluster-config-file", r.clusterNetworkFile, "Openshift Cluster API Config file.") //TODO move back to normal values after installer change.
+	fs.StringVar(&r.clusterNetworkFile, "cluster-network-file", r.clusterNetworkFile, "Openshift Cluster Network API Config file.")
```
hexfusion: this is intended so we can update the installer and still pass CI.

hexfusion:
/hold

I am afraid to merge with without IPv6 test

hexfusion:
work continued https://github.com/openshift/cluster-etcd-operator/pull/375/files

PR #199, opened 2020-02-24
[Bug 1807493]: add the restore command to restore pod

alaypatel07:
This builds on top of #197 to restore the single-member cluster. 
It can be used to make the incantation of `etcd-restore-script.sh`
with just one parameter. The INITIAL_CLUSTER will be calculated,
by the restore pod and it will perform the actual restore to be
consumed by the etcd command to follow.

Makes the steps in [here](https://github.com/openshift/enhancements/pull/226#discussion_r383055434) implementable

### DIFF
--- a/bindata/etcd/restore-pod.yaml
+++ b/bindata/etcd/restore-pod.yaml
@@ -32,23 +32,17 @@ spec:
           exit 1
         fi
 
-        # check if we have backup file to be restored and if the backup file is not currently
-        # being copied, exit non-zero if copying continues for more than 5 seconds
+        # check if we have backup file to be restored
+        # if the file exist, check if it has not changed size in last 5 seconds
         if [ ! -f /var/lib/etcd-backup/snapshot.db ]; then
           echo "please make a copy of the snapshot db file, then move that copy to /var/lib/etcd-backup/snapshot.db"
           exit 1
         else
           filesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
-          for i in `seq 1 5`;
-          do
-            newfilesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
-            if [ "$filesize" == "$newfilesize" ]; then
-              break
-            fi
-            sleep 1
-          done
+          sleep 5
+          newfilesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
           if [ "$filesize" != "$newfilesize" ]; then
-            echo "file size is changing for more than 5 seconds"
+            echo "file size has changed since last 5 seconds, retry sometime after copying is complete"
             exit 1
           fi
         fi
@@ -62,7 +56,7 @@ spec:
          --initial-advertise-peer-urls $ETCD_NODE_PEER_URL \
          --data-dir="/var/lib/etcd/restore-{$UUID:0:10}"
 
-        mv /var/lib/etcd/restore-{$UUID:0:10}/ /var/lib/etcd/
+        mv /var/lib/etcd/restore-{$UUID:0:10}/* /var/lib/etcd/
 
         set -x
         exec etcd \
--- a/bindata/etcd/restore-pod.yaml
+++ b/bindata/etcd/restore-pod.yaml
@@ -24,6 +24,45 @@ spec:
         export ETCD_NAME=${NODE_NODE_ENVVAR_NAME_ETCD_NAME}
         export ETCD_INITIAL_CLUSTER="${ETCD_NAME}=https://${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}:2380"
         env | grep ETCD | grep -v NODE
+        export ETCD_NODE_PEER_URL=https://${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}:2380
+
+        # checking if data directory is empty, if not etcdctl restore will fail
+        if [ ! -z $(ls -A "/var/lib/etcd") ]; then
+          echo "please delete the contents of data directory before restoring, running the restore script will do this for you"
+          exit 1
+        fi
+
+        # check if we have backup file to be restored and if the backup file is not currently
+        # being copied, exit non-zero if copying continues for more than 5 seconds
+        if [ ! -f /var/lib/etcd-backup/snapshot.db ]; then
+          echo "please make a copy of the snapshot db file, then move that copy to /var/lib/etcd-backup/snapshot.db"
+          exit 1
+        else
+          filesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
+          for i in `seq 1 5`;
+          do
+            newfilesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
+            if [ "$filesize" == "$newfilesize" ]; then
+              break
+            fi
+            sleep 1
+          done
+          if [ "$filesize" != "$newfilesize" ]; then
+            echo "file size is changing for more than 5 seconds"
+            exit 1
+          fi
+        fi
+
+        UUID=$(uuidgen)
+        echo "restoring to a single node cluster"
+        ETCDCTL_API=3 /usr/bin/etcdctl snapshot restore /var/lib/etcd-backup/snapshot.db \
+         --name  $ETCD_NAME \
+         --initial-cluster=$ETCD_INITIAL_CLUSTER \
+         --initial-cluster-token "openshift-etcd-{$UUID:0:10}" \
+         --initial-advertise-peer-urls $ETCD_NODE_PEER_URL \
+         --data-dir="/var/lib/etcd/restore-{$UUID:0:10}"
+
+        mv /var/lib/etcd/restore-{$UUID:0:10}/ /var/lib/etcd/
 
         set -x
         exec etcd \
@@ -70,6 +109,8 @@ ${COMPUTED_ENV_VARS}
         name: cert-dir
       - mountPath: /var/lib/etcd/
         name: data-dir
+      - mountPath: /var/lib/etcd-backup/
+        name: backup-dir
   hostNetwork: true
   priorityClassName: system-node-critical
   tolerations:
@@ -91,4 +132,8 @@ ${COMPUTED_ENV_VARS}
         path: /var/lib/etcd
         type: ""
       name: data-dir
+    - hostPath:
+        path: /var/lib/etcd-backup
+        type: ""
+      name: backup-dir
 
--- a/pkg/operator/etcd_assets/bindata.go
+++ b/pkg/operator/etcd_assets/bindata.go
@@ -402,23 +402,17 @@ spec:
           exit 1
         fi
 
-        # check if we have backup file to be restored and if the backup file is not currently
-        # being copied, exit non-zero if copying continues for more than 5 seconds
+        # check if we have backup file to be restored
+        # if the file exist, check if it has not changed size in last 5 seconds
         if [ ! -f /var/lib/etcd-backup/snapshot.db ]; then
           echo "please make a copy of the snapshot db file, then move that copy to /var/lib/etcd-backup/snapshot.db"
           exit 1
         else
           filesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
-          for i in ` + "`" + `seq 1 5` + "`" + `;
-          do
-            newfilesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
-            if [ "$filesize" == "$newfilesize" ]; then
-              break
-            fi
-            sleep 1
-          done
+          sleep 5
+          newfilesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
           if [ "$filesize" != "$newfilesize" ]; then
-            echo "file size is changing for more than 5 seconds"
+            echo "file size has changed since last 5 seconds, retry sometime after copying is complete"
             exit 1
           fi
         fi
@@ -432,7 +426,7 @@ spec:
          --initial-advertise-peer-urls $ETCD_NODE_PEER_URL \
          --data-dir="/var/lib/etcd/restore-{$UUID:0:10}"
 
-        mv /var/lib/etcd/restore-{$UUID:0:10}/ /var/lib/etcd/
+        mv /var/lib/etcd/restore-{$UUID:0:10}/* /var/lib/etcd/
 
         set -x
         exec etcd \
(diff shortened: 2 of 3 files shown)


bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,27 @@ spec:
...
+        fi
+
+        # checking if we have backup file to be restored
+        if [ ! -f var/lib/etcd-backup/snapshot.db ]; then
+            echo "please copy the backup file to /var/lib/etcd-backup/snapshot.db before restoring"
+            exit 0
+        fi
+
+        echo "restoring to a single node cluster"
+        ETCDCTL_API=3 etcdctl snapshot restore /var/lib/etcd-backup/snapshot.db \
```
alaypatel07: I am not sure if we would have access to etcdctl in PATH, probably will have to modify with the path to binary for preciseness

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,27 @@ spec:
         export ETCD_NAME=${NODE_NODE_ENVVAR_NAME_ETCD_NAME}
         export ETCD_INITIAL_CLUSTER="${ETCD_NAME}=https://${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}:2380"
         env | grep ETCD | grep -v NODE
+        export ETCD_NODE_PEER_URL=https://${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}:2380
+
+        # checking if data directory is empty, if not etcdctl restore will fail
+        if [ ! "$(ls -A /var/lib/etcd)" ]; then
+          echo "please delete the data directory before restoring"
+          exit 0
```
deads2k: exit non-zero so we crash loop

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,27 @@ spec:
...
+        fi
+
+        # checking if we have backup file to be restored
+        if [ ! -f var/lib/etcd-backup/snapshot.db ]; then
```
deads2k: make sure the file size isn't changing on a five second timer.  Someone is going to use a copy like you suggest here, not a move

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,27 @@ spec:
...
+            echo "please copy the backup file to /var/lib/etcd-backup/snapshot.db before restoring"
```
deads2k: "please make a copy of the backup file, then *move* that copy to /var/lib/etcd-backup/snapshot.db".  You need the atomic switch.

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,27 @@ spec:
...
+            exit 0
```
deads2k: exit non-zero

deads2k:
This is largely plausible.  In the end, a restore shell script should take care of the actual move.

bindata/etcd/restore-pod.yaml
retroflexer: Can you check on /var/lib/etcd/member instead of /var/lib/etcd?

bindata/etcd/restore-pod.yaml
alaypatel07: added an else block to check the copying case

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,43 @@ spec:
...
+        # checking if data directory is empty, if not etcdctl restore will fail
+        if [ ! "$(ls -A /var/lib/etcd)" ]; then
+          echo "please delete the data directory before restoring"
+          exit 1
+        fi
+
+        # check if we have backup file to be restored and if the backup file is not currently
+        # being copied, exit non-zero if copying continues for more than 5 seconds
+        if [ ! -f var/lib/etcd-backup/snapshot.db ]; then
+          echo "please make a copy of the backup file, then move that copy to /var/lib/etcd-backup/snapshot.db"
```
retroflexer: Currently the backup is a directory containing static resources and a snapshot db. Here using the word backup may confuse user. Call it `snapshot db` perhaps?

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,43 @@ spec:
         export ETCD_NAME=${NODE_NODE_ENVVAR_NAME_ETCD_NAME}
         export ETCD_INITIAL_CLUSTER="${ETCD_NAME}=https://${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}:2380"
         env | grep ETCD | grep -v NODE
+        export ETCD_NODE_PEER_URL=https://${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}:2380
+
+        # checking if data directory is empty, if not etcdctl restore will fail
+        if [ ! "$(ls -A /var/lib/etcd)" ]; then
```
hexfusion: we only need to know that the dir exists right? restore will fail if /var/lib/etcd exists at all.

[ -d "/var/lib/etcd " ]

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,43 @@ spec:
...
+          echo "please delete the data directory before restoring"
+          exit 1
+        fi
+
+        # check if we have backup file to be restored and if the backup file is not currently
+        # being copied, exit non-zero if copying continues for more than 5 seconds
+        if [ ! -f var/lib/etcd-backup/snapshot.db ]; then
```
hexfusion: typo `var/` `/var/` ?

bindata/etcd/restore-pod.yaml
hexfusion: just to be clear we only move the etcd data file what about the other static pod resources.

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,43 @@ spec:
...
+          echo "please make a copy of the backup file, then move that copy to /var/lib/etcd-backup/snapshot.db"
+          exit 1
+        else
+          filesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
```
hexfusion: can you explain what you are trying to do here?

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,43 @@ spec:
...
+            exit 1
+          fi
+        fi
+
+        NEW_UUID=$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 10 | head -n 1)
+        echo "restoring to a single node cluster"
+        ETCDCTL_API=3 /usr/bin/etcdctl snapshot restore /var/lib/etcd-backup/snapshot.db \
+         --name  $ETCD_NAME \
+         --initial-cluster=$ETCD_INITIAL_CLUSTER \
+         --initial-cluster-token "openshift-etcd-$NEW_UUID" \
```
hexfusion: This needs to be part of restore for all the other members we add how does that work

bindata/etcd/restore-pod.yaml
hexfusion: why would the file size change?

bindata/etcd/restore-pod.yaml
hexfusion: actually we are doing single-member restore but we will need to make sure that add will succeed without this.

bindata/etcd/restore-pod.yaml
alaypatel07: yes, wanted to make sure that the snapshot.db is not currently copied. So I compare the file size for 5 seconds and if the filesize is constant, I break, else exit 1. 

Basically stopping ourselves for shooting in the foot.

bindata/etcd/restore-pod.yaml
alaypatel07: the restore script should handle other static pod resources, this pod only restores etcd cluster from DB file and start the etcd on restored data directory.

bindata/etcd/restore-pod.yaml
alaypatel07: yes typo, will change

bindata/etcd/restore-pod.yaml
alaypatel07: cleared over slack, we dont need to worry about the token since this is a single member etcd cluster. Other members will be added by clustermembercontroller

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,43 @@ spec:
...
+        # being copied, exit non-zero if copying continues for more than 5 seconds
+        if [ ! -f var/lib/etcd-backup/snapshot.db ]; then
+          echo "please make a copy of the backup file, then move that copy to /var/lib/etcd-backup/snapshot.db"
+          exit 1
+        else
+          filesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
+          for i in `seq 1 5`;
+          do
+            newfilesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
+            if [ "$filesize" == "$newfilesize" ]; then
```
hexfusion: can filesize be 0?

bindata/etcd/restore-pod.yaml
alaypatel07: confirmed, will change

```
[root@alpate-4jcgd-m-0 ~]# mkdir temp2
[root@alpate-4jcgd-m-0 ~]# ETCDCTL_API=3 ./assets/bin/etcdctl snapshot restore $BACKUP_FILE --name  "alpate-4jcgd-m-0.c.openshift-gce-devel.internal" --initial-cluster="alpate-4jcgd-m-0.c.openshift-gce-devel.internal=https://etcd-0.alpatel-alpatel-dr10-2020-02-24.gcp.devcluster.openshift.com:2380"              --initial-cluster-token openshift-etcd-0             --initial-advertise-peer-urls https://etcd-0.alpatel-alpatel-dr10-2020-02-24.gcp.devcluster.openshift.com:2380             --data-dir="./temp2"
2020-02-25 20:21:53.922765 I | pkg/netutil: resolving etcd-0.alpatel-alpatel-dr10-2020-02-24.gcp.devcluster.openshift.com:2380 to 10.0.0.3:2380
2020-02-25 20:21:53.923652 I | pkg/netutil: resolving etcd-0.alpatel-alpatel-dr10-2020-02-24.gcp.devcluster.openshift.com:2380 to 10.0.0.3:2380
Error: data-dir "./temp2" exists
[root@alpate-4jcgd-m-0 ~]#
```

Thanks, @hexfusion I was unaware of this feature.

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,43 @@ spec:
...
+            fi
+            sleep 1
+          done
+          if [ "$filesize" != "$newfilesize" ]; then
+            echo "file size is changing for more than 5 seconds"
+            exit 1
+          fi
+        fi
+
+        NEW_UUID=$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 10 | head -n 1)
```
hexfusion: just for fun tell me how this works :)

bindata/etcd/restore-pod.yaml
alaypatel07: @retroflexer I will update the logging, but FTR the user will never have to see that logging if restore.sh is run. The script will make sure that the file exists.

bindata/etcd/restore-pod.yaml
alaypatel07: not if it is copied by the script.

bindata/etcd/restore-pod.yaml
alaypatel07: updated with less fun stuff :)

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,43 @@ spec:
         export ETCD_NAME=${NODE_NODE_ENVVAR_NAME_ETCD_NAME}
         export ETCD_INITIAL_CLUSTER="${ETCD_NAME}=https://${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}:2380"
         env | grep ETCD | grep -v NODE
+        export ETCD_NODE_PEER_URL=https://${NODE_NODE_ENVVAR_NAME_ETCD_DNS_NAME}:2380
+
+        # checking if data directory is empty, if not etcdctl restore will fail
+        if [ -d "/var/lib/etcd " ]; then
+          echo "please delete the data directory before restoring"
+          exit 1
```
retroflexer: This won't work on a pod. "/var/lib/etcd" is a mount point, and it will always be there. :-(

alaypatel07: restored to ls -A 👍

bindata/etcd/restore-pod.yaml
```diff
@@ -24,6 +24,45 @@ spec:
...
+        UUID=$(uuidgen)
+        echo "restoring to a single node cluster"
+        ETCDCTL_API=3 /usr/bin/etcdctl snapshot restore /var/lib/etcd-backup/snapshot.db \
+         --name  $ETCD_NAME \
+         --initial-cluster=$ETCD_INITIAL_CLUSTER \
+         --initial-cluster-token "openshift-etcd-{$UUID:0:10}" \
+         --initial-advertise-peer-urls $ETCD_NODE_PEER_URL \
+         --data-dir="/var/lib/etcd/restore-{$UUID:0:10}"
+
+        mv /var/lib/etcd/restore-{$UUID:0:10}/ /var/lib/etcd/
```
retroflexer: This is not correct. You need to move the contents of `/var/lib/etcd/restore-{$UUID:0:10}/` into `/var/lib/etcd`.

Typically, it will have a `member` directory.

```mv /var/lib/etcd/restore-{$UUID:0:10}/* /var/lib/etcd/```

File: bindata/etcd/restore-pod.yaml [hunk 15]
```diff
@@ -24,6 +24,45 @@ spec:
...
+        fi
+
+        # check if we have backup file to be restored and if the backup file is not currently
+        # being copied, exit non-zero if copying continues for more than 5 seconds
+        if [ ! -f /var/lib/etcd-backup/snapshot.db ]; then
+          echo "please make a copy of the snapshot db file, then move that copy to /var/lib/etcd-backup/snapshot.db"
+          exit 1
+        else
+          filesize=$(stat --format=%s "/var/lib/etcd-backup/snapshot.db")
+          for i in `seq 1 5`;
```
deads2k: no, just sleep

deads2k:
simplify, then this looks plausible

deads2k:
/lgtm

tagging based on slack conversation.

alaypatel07:
/hold

performing manual verification will unhold as soon as verified

alaypatel07:
/hold cancel

I got the restore pod to pivot at etcd-pod.yaml's place and restore. FTR the cluster only had 1 member even on forced redeployment, debugging how I reached there but this works as intended.

PR #203, opened 2020-02-25
[Bug 1807054]: fix recovery of master nodes restarting at the same time

alaypatel07:
The ` ${ETCDCTL} member list` will fail when all the masters
are restarted at the same time. It will be good to have the 
member list, but if we don't, we should not fail and continue
to the discovery.


alaypatel07:
FTR I was able to bring etcd pods back up in one of the failed cluster runs https://coreos.slack.com/archives/C027U68LP/p1582641688035800?thread_ts=1582598465.022300&cid=C027U68LP

PR #206, opened 2020-02-25
WIP: Proof metric observer to observe fsync prometheus metrics

mfojtik:


pkg/operator/starter.go
```diff
@@ -198,12 +210,35 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		controllerContext.EventRecorder,
 	)
 
+	metricsObserver, err := observer.NewPrometheusMetricObserver("EtcdMetricsObserver",
+		[]observer.QueryHandler{
+			{
+				Name:  "EtcdFSyncDurationSeconds",
+				Query: "histogram_quantile(0.99, rate(etcd_disk_wal_fsync_duration_seconds_bucket[5m])",
+				Handler: func(ctx context.Context, recorder events.Recorder, currentValue string, valueType string) error {
```
deads2k: I'll find the other PR, but I expected to have one handler passed the current values for all metrics.

mfojtik: can you elaborate? how you want to pass all metrics? I think we want to treat individual metrics separately... also we might watch alerts instead of using queries, dunno which one we would prefer for this one.

mfojtik: i changed the handle so you have now full access to prometheus client (so you can do query/alert/etc.).

pkg/operator/metricshandler/fsync_handler.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	if len(warns) > 0 {
+		klog.Warningf("Prometheus warnings received: %#v", warns)
+	}
+	matrix, ok := result.(model.Matrix)
+	if !ok {
+		return fmt.Errorf("result is not Matrix")
+	}
+	for _, s := range matrix {
+		// TODO: Do something useful with these values
```
mfojtik: @hexfusion @deads2k these are the numbers I got in AWS, we should start thinking what are the tolerable numbers per provider (and I need to wire getting the provider into this handler, so probably it will turn into a struct or something)

mfojtik: actually, I found `cluster_infrastructure_provider` so we can query provider via prometheus :D

hexfusion:
I want this for autotune. Basically we are going to watch on I/O metrics and when we hit threshold X we are going bump some defaults to add resiliency against I/O latencies and Alert cluster.

deads2k:
@hexfusion You'll want this soon.  A basic 4.4 backport will save you extra scrutiny and difficulty as you refine it.

mfojtik:
Close in favor of https://github.com/openshift/cluster-etcd-operator/pull/313

PR #207, opened 2020-02-25
Bug 1805981: oc rsh to an etcd pod needs to have etcdctl "just work" 

sanchezl:


bindata/etcd/pod.yaml
```diff
@@ -37,6 +37,31 @@ spec:
         - mountPath: /etc/kubernetes/static-pod-certs
           name: cert-dir
   containers:
+  - name: etcdctl
```
deads2k: comment explaining why this container must be first or we'll move it.

pkg/operator/targetconfigcontroller/etcd_env.go
```diff
@@ -72,11 +73,24 @@ func getFixedEtcdEnvVars(envVarContext envVarContext) (map[string]string, error)
...
+	envVars, err := getAllClusterMembers(envVarContext)
+	if err != nil {
+		return nil, err
+	}
+	return map[string]string{
+		"ETCDCTL_API":       "3",
+		"ETCDCTL_CACERT":    "/etc/kubernetes/static-pod-resources/configmaps/etcd-serving-ca/ca-bundle.crt",
+		"ETCDCTL_CERT":      "/etc/kubernetes/static-pod-resources/secrets/etcd-all-peer/etcd-peer-NODE_NAME.crt",
+		"ETCDCTL_KEY":       "/etc/kubernetes/static-pod-resources/secrets/etcd-all-peer/etcd-peer-NODE_NAME.key",
+		"ETCDCTL_ENDPOINTS": envVars["ALL_ETCD_ENDPOINTS"],
```
deads2k: This has no impact on the `etcd` process itself?

pkg/operator/targetconfigcontroller/etcd_env.go
```diff
@@ -72,11 +73,24 @@ func getFixedEtcdEnvVars(envVarContext envVarContext) (map[string]string, error)
 	return map[string]string{
 		"ETCD_DATA_DIR":              "/var/lib/etcd",
 		"ETCD_QUOTA_BACKEND_BYTES":   "7516192768", // 7 gig
-		"ETCDCTL_API":                "3",
 		"ETCD_INITIAL_CLUSTER_STATE": "existing",
 	}, nil
 }
 
+func getEtcdctlEnvVars(envVarContext envVarContext) (map[string]string, error) {
+	envVars, err := getAllClusterMembers(envVarContext)
```
deads2k: no.  Refactor into a method that gives just the string and use it in both.

pkg/operator/targetconfigcontroller/etcd_env.go
hexfusion: any var that is prefixed with `ETCDCTL_` is only consumed by the client binary. Where `ETCD_` is consumed by server instance.

PR #210, opened 2020-02-26
[Bug 1807493]: add scripts to files laid down to host

deads2k:
A starting point.

/assign @retroflexer

### DIFF
--- a/pkg/operator/scriptcontroller/scriptcontroller.go
+++ b/pkg/operator/scriptcontroller/scriptcontroller.go
@@ -1,0 +1,174 @@
+package scriptcontroller
+
+import (
+	"fmt"
+	"path/filepath"
+	"time"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/etcd_assets"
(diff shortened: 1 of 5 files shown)


pkg/operator/scriptcontroller/scriptcontroller.go
```diff
@@ -0,0 +1,174 @@
...
+
+	return utilerrors.NewAggregate(errors)
+}
+
+func (c *ScriptController) manageScriptConfigMap() (*corev1.ConfigMap, bool, error) {
+	scriptConfigMap := resourceread.ReadConfigMapV1OrDie(etcd_assets.MustAsset("etcd/scripts-cm.yaml"))
+	// TODO get the env vars to produce a file that we write
+
+	for _, filename := range []string{
+		"etcd/etcd-restory-backup.sh",
```
alaypatel07: typo?

pkg/operator/starter.go
```diff
@@ -242,6 +252,7 @@ var RevisionSecrets = []revision.RevisionResource{
 
 var CertConfigMaps = []revision.RevisionResource{
 	{Name: "restore-etcd-pod"},
+	{Name: "etcd-scripts"},
```
retroflexer: I called them "recovery-scripts". This is already etcd pod, so ...

deads2k: > I called them "recovery-scripts". This is already etcd pod, so ...

I think these may contain more than recovery scripts.

PR #213, opened 2020-02-26
Bug 1807632: fix readiness probe

alaypatel07:
This adds readiness probes that check for the health of local etcd endpoint

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -99,7 +99,7 @@ ${COMPUTED_ENV_VARS}
         command:
           - /bin/sh
           - -ec
-          - "lsof -n -i :2380 | grep LISTEN"
+          - "etcdctl --cacert=\"/etc/kubernetes/static-pod-resources/configmaps/etcd-peer-client-ca/ca-bundle.crt\" --cert=\"/etc/kubernetes/static-pod-resources/secrets/etcd-all-peer/etcd-peer-NODE_NAME.crt\" --key=\"/etc/kubernetes/static-pod-resources/secrets/etcd-all-peer/etcd-peer-NODE_NAME.key\" --endpoints=https://localhost:2379 endpoint health"
       failureThreshold: 3
       initialDelaySeconds: 3
       periodSeconds: 5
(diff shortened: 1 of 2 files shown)


bindata/etcd/pod.yaml
```diff
@@ -99,7 +99,7 @@ ${COMPUTED_ENV_VARS}
         command:
           - /bin/sh
           - -ec
-          - "lsof -n -i :2380 | grep LISTEN"
+          - "etcdctl --cacert=\"/etc/kubernetes/static-pod-resources/configmaps/etcd-peer-client-ca/ca-bundle.crt\" --cert=\"/etc/kubernetes/static-pod-resources/secrets/etcd-all-peer/etcd-peer-NODE_NAME.crt\" --key=\"/etc/kubernetes/static-pod-resources/secrets/etcd-all-peer/etcd-peer-NODE_NAME.key\" --endpoints=https://localhost:2379 endpoint health"
```
abhinavdahiya: I think https://yaml-multiline.info/ and help see if `>` or `|` can make this readable?

deads2k:
@hexfusion we need to make sure that in cases with uncoordinated restart of etcd pods across the cluster, we become stable.  I think this is a maybe for 4.4

/hold

hold until @hexfusion thinks this will make us more reliable.

hexfusion:
I think we want to fast track availability if we are listening on 2380 which means we have started and can accept requests which is a reasonable definition of health. In the case of disruption should we create watch streams with 2 endpoints because of a network blip? We have a client balancer to manage this IMO.

>This adds readiness probes that check for the health of local etcd endpoint

I really don't like it because we are not just checking the health of the member but of the cluster[1]. So if this member is ready to serve traffic but cluster is temporary unquorate is it accurate that no endpoints are healthy? My feeling is this battle should be fought on another layer (client balancer). But if you have a clear example where this would improve the cluster I will consider the argument.

[1] https://github.com/etcd-io/etcd/blob/master/etcdctl/ctlv3/command/ep_command.go#L126

ironcladlou:
Replaced by https://github.com/openshift/cluster-etcd-operator/pull/474

/close

PR #214, opened 2020-02-26
[WIP]: fix EtcdMembersProgressing condition

alaypatel07:
this fixes EtcdMembersProgressing condition so that it is
true only when all etcd members are available.

### DIFF
--- a/pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
+++ b/pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
@@ -116,12 +116,12 @@ func (c *EtcdMembersController) reportEtcdMembers() error {
 		}
 	}
 
-	if len(notStartedMembers) != 0 {
+	if len(notStartedMembers) != 0 || len(availableMembers) != 3 {
 		_, _, updateErr := v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
 			Type:    "EtcdMembersProgressing",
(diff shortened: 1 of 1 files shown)


pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
```diff
@@ -116,12 +116,12 @@ func (c *EtcdMembersController) reportEtcdMembers() error {
 		}
 	}
 
-	if len(notStartedMembers) != 0 {
+	if len(notStartedMembers) != 0 || len(availableMembers) != 3 {
```
alaypatel07: this will break adding 4th master, we don't support it, but still doesn't feel good breaking it in control loop

cgruver: I believe that statically setting this to `3` will also further complicate issues that have cropped up building a single node cluster for code ready containers.  

See:  https://bugzilla.redhat.com/show_bug.cgi?id=1805034 

If three is the magic number for production ready clusters, (which I wholeheartedly believe that it is), then perhaps we need a switch that can be set when invoking the installer to override off this restriction for single node clusters.  Or, even a way to override it completely for those who are brave, or just really want 4+ nodes in their control plane...

cgruver: FWIW, we're also trying to get CRC working for OKD 4.X with FCOS as well.

https://github.com/openshift/okd/issues/88

PR #217, opened 2020-02-27
[Bug 1808493]: add ca-bundles to fixed locations and use the for restore pod

alaypatel07:
This will make the restore pod work without requiring revision
specific resources from backup.


deads2k:
I expectd to see updates to the etcd-pod itself, to the recovery pod paths, and the env vars that get created and current point to revisions.

PR #220, opened 2020-02-28
Bug 1810469: The disaster recovery scripts need simplification

retroflexer:
With the introduction of cluster-etcd-operator, much of the functionality in the old scripts is no longer needed. Therefore the scripts needed to be trimmed down and simplified. This PR attempts to achieve that goal.


bindata/etcd/etcd-member-remove.sh
```diff
@@ -17,21 +17,40 @@ if [ "$1" == "" ]; then
...
-ASSET_DIR_TMP="$ASSET_DIR/tmp"
-ETCDCTL=$ASSET_DIR/bin/etcdctl
-ETCD_DATA_DIR=/var/lib/etcd
-CONFIG_FILE_DIR=/etc/kubernetes
+source etcd.env
+
+# download etcdctl from upstream release assets
+function dl_etcdctl() {
+  ETCDCTL_BIN="${ASSET_DIR}/bin/etcdctl"
+  local etcdimg="registry.svc.ci.openshift.org/ci-op-l7y8kffx/stable@sha256:c576980bdce812e5d000e6f96fc238d75688da01f767cdaed6cbf04c40c600fa"
```
retroflexer: This is wrong. The image should come from `local etcdimg="{{.Images.etcdKey}}"`

deads2k: > This is wrong. The image should come from `local etcdimg="{{.Images.etcdKey}}"`

ok, you'll need to do that in the golang code.

bindata/etcd/etcd-member-remove.sh
```diff
@@ -17,21 +17,40 @@ if [ "$1" == "" ]; then
...
-ASSET_DIR=/home/core/assets
-ASSET_DIR_TMP="$ASSET_DIR/tmp"
-ETCDCTL=$ASSET_DIR/bin/etcdctl
-ETCD_DATA_DIR=/var/lib/etcd
-CONFIG_FILE_DIR=/etc/kubernetes
+source etcd.env
+
+# download etcdctl from upstream release assets
+function dl_etcdctl() {
+  ETCDCTL_BIN="${ASSET_DIR}/bin/etcdctl"
```
deads2k: can we just add this to the path for the duration of this script?

bindata/etcd/etcd-member-remove.sh
```diff
@@ -17,21 +17,40 @@ if [ "$1" == "" ]; then
...
+  local etcdimg="registry.svc.ci.openshift.org/ci-op-l7y8kffx/stable@sha256:c576980bdce812e5d000e6f96fc238d75688da01f767cdaed6cbf04c40c600fa"
+  local etcdctr=$(podman create ${etcdimg})
+  local etcdmnt=$(podman mount "${etcdctr}")
+  cp ${etcdmnt}/bin/etcdctl $ASSET_DIR/bin
+  umount "${etcdmnt}"
+  podman rm "${etcdctr}"
+  $ETCDCTL_BIN version
+}
+
+function etcd_member_remove() {
```
deads2k: not a function please

bindata/etcd/etcd-member-remove.sh
```diff
@@ -17,21 +17,40 @@ if [ "$1" == "" ]; then
...
+  NAME="$1"
 
-source "/usr/local/bin/openshift-recovery-tools"
+  ID=$($ETCDCTL_WITH_TLS member list | awk -F "," "/\s${NAME}\,/"'{print $1}') || true
+  if [ "$?" -ne 0 ] || [ -z "$ID" ]; then
+    echo "could not find etcd member $NAME to remove."
+    exit 1
+  fi
+
+  $ETCDCTL_BIN member remove $ID
```
deads2k: Does this command already fail if the member isn't present?  if so, let's just call this

bindata/etcd/etcd-member-remove.sh
```diff
@@ -17,21 +17,40 @@ if [ "$1" == "" ]; then
...
+}
+
+function etcd_member_remove() {
+  NAME="$1"
 
-source "/usr/local/bin/openshift-recovery-tools"
+  ID=$($ETCDCTL_WITH_TLS member list | awk -F "," "/\s${NAME}\,/"'{print $1}') || true
+  if [ "$?" -ne 0 ] || [ -z "$ID" ]; then
+    echo "could not find etcd member $NAME to remove."
+    exit 1
```
deads2k: we normally run our scripts with something that makes them fail if the command fails.  I'd prefer the script have that

retroflexer: Yes. I will add `-e`.

bindata/etcd/etcd-member-remove.sh
retroflexer: But the difference is ETCDCTL_BIN accepts ID, whereas our script accepts NAME or ID.

bindata/etcd/etcd-member-remove.sh
```diff
@@ -8,30 +12,31 @@ if [[ $EUID -ne 0 ]]; then
...
-  etcd_member_remove $ETCD_NAME
-}
+NAME="$1"
 
-run
+source etcd.env
 
+dl_etcdctl
+ID=$($ETCDCTL_WITH_TLS member list | awk -F "," "/\s${NAME}\,/"'{print $1}')
+etcdctl member remove $ID
```
deads2k: Just make them pass the ID directly.  Any reason not to?

bindata/etcd/etcd-restore-backup.sh
```diff
@@ -11,77 +12,50 @@ if [[ $EUID -ne 0 ]]; then
...
-    usage
-  fi
-elif [ -d "$1" ]; then
-  BACKUP_FILE=$(ls -vd "$1"/static_kuberesources*.tar.gz | tail -1) || true
-  SNAPSHOT_FILE=$(ls -vd "$1"/snapshot*.db | tail -1) || true
-  if [ ! -f ${BACKUP_FILE}  -o ! -f ${SNAPSHOT_FILE} ]; then
-    usage
-  fi
-else
+if [ "$1" == "" ] || [ ! -d "$1" ]; then
```
deads2k: please comment the bash conditionals for bash illiterates like me

bindata/etcd/etcd-restore-backup.sh
```diff
@@ -11,77 +12,50 @@ if [[ $EUID -ne 0 ]]; then
   exit 1
 fi
 
-usage () {
-    echo 'Path to the directory containing backup files is required: ./etcd-snapshot-restore.sh <path-to-backup>'
-    exit 1
+function usage {
+  echo 'Path to the directory containing backup files is required: ./etcd-snapshot-restore.sh <path-to-backup>'
```
deads2k: this isn't descriptive enough for me. What content is expected here

bindata/etcd/etcd-restore-backup.sh
```diff
@@ -11,77 +12,50 @@ if [[ $EUID -ne 0 ]]; then
...
+fi
 
-run
+ls -1d ${MANIFEST_DIR}/* | while read STATIC_POD; do
+  echo "... stopping $STATIC_POD"
+  mv ${MANIFEST_DIR}/${STATIC_POD} $MANIFEST_STOPPED_DIR
+done
+
+# Restore static pod resources
+tar -C ${CONFIG_FILE_DIR} -xzf $BACKUP_FILE static-pod-resources
```
deads2k: I didn't expect us to have anything tarred.

bindata/etcd/etcd-restore-backup.sh
```diff
@@ -11,77 +12,50 @@ if [[ $EUID -ne 0 ]]; then
...
+
+# Restore snapshot
+${ETCDCTL} snapshot restore $SNAPSHOT_FILE \
```
deads2k: this is embedded in the restore static pod.

retroflexer: Yup. Will remove it.

bindata/etcd/etcd-member-remove.sh
retroflexer: Because it is easier to pass the name. When they use the name, they are sure about what they are removing. The probability of removing a wrong member is less likely when names are used.

bindata/etcd/etcd-restore-backup.sh
retroflexer: Well, our current backup is expected to be in a directory containing two components:
1. etcd snapshot
1. Bundle of Static POD resources (which includes kube-apiserver's static pod resources, that may include encryption key)

bindata/etcd/etcd-restore-backup.sh
retroflexer: We tar up the static pod resources during the backup process. We need to restore those static pod resources at the time of backup, when we are restoring the snapshot. 

This is needed for time-travel, esp. in case of encryption we need to have the encryption key match the data in the etcd snapshot.

retroflexer:
Currently, the scripts assume that the following variables come from etcd.env file:
```
${ASSET_DIR}
${CONFIG_FILE_DIR}
${ETCD_DATA_DIR_BACKUP}
${MANIFEST_DIR}
${MANIFEST_STOPPED_DIR}
${RESTORE_ETCD_POD_YAML}
${STATIC_POD}
```

bindata/etcd/etcd-restore-backup.sh
```diff
@@ -11,77 +12,48 @@ if [[ $EUID -ne 0 ]]; then
...
-if [ "$1" == "" ]; then
-    usage
+# If the argument is not passed, or if it is not a directory, print usage and exit.
+if [ "$1" == "" ] || [ ! -d "$1" ]; then
+  usage
 fi
 
-ASSET_DIR=/home/core/assets
+BACKUP_FILE=$(ls -vd "$1"/static_kuberesources*.tar.gz | tail -1) || true
+SNAPSHOT_FILE=$(ls -vd "$1"/snapshot*.db | tail -1) || true
```
hexfusion: can we bin the PARAM $1 to a variable here?

pkg/etcdenvvar/etcd_env.go
```diff
@@ -78,6 +79,12 @@ func getFixedEtcdEnvVars(envVarContext envVarContext) (map[string]string, error)
 		"ETCD_DATA_DIR":              "/var/lib/etcd",
 		"ETCD_QUOTA_BACKEND_BYTES":   "7516192768", // 7 gig
 		"ETCD_INITIAL_CLUSTER_STATE": "existing",
+		"ASSET_DIR" : "/home/core/assets",
```
deads2k: keep these local to the scripts.  They don't vary.  Some are even specific to a single script.

pkg/etcdenvvar/etcd_env.go
```diff
@@ -78,6 +79,12 @@ func getFixedEtcdEnvVars(envVarContext envVarContext) (map[string]string, error)
...
+		"CONFIG_FILE_DIR" :  "/etc/kubernetes",
+		"MANIFEST_DIR" : "${CONFIG_FILE_DIR}/manifests",
+		"ETCD_DATA_DIR_BACKUP" : "/var/lib/etcd-backup",
+		"MANIFEST_STOPPED_DIR" : "${ASSET_DIR}/manifests-stopped",
+		"RESTORE_ETCD_POD_YAML" : "${CONFIG_FILE_DIR}/static-pod-resources/etcd-certs/configmaps/restore-etcd-pod/pod.yaml",
```
deads2k: one script

bindata/etcd/etcd-member-remove.sh
deads2k: > Because it is easier to pass the name. The assumption is that the probability of removing a wrong member is less likely when names are used.

given bash complexity against human readable output, I'm against but will defer to @hexfusion

bindata/etcd/etcd-restore-backup.sh
deads2k: > Well, our current backup is expected to be in a directory containing two components:
> 
> 1. etcd snapshot
> 2. Bundle of Static POD resources (which includes kube-apiserver's static pod resources, that may include encryption key)

don't tell me, write in the usage for everyone :)

bindata/etcd/etcd-member-remove.sh
```diff
@@ -8,30 +12,31 @@ if [[ $EUID -ne 0 ]]; then
...
     echo 'The name of the etcd member to remove is required: ./etcd-member-remove.sh $etcd_name'
     exit 1
 }
 
+# download etcdctl from upstream release assets
+function dl_etcdctl {
+  local etcdimg=${IMAGE}
+  local etcdctr=$(podman create ${etcdimg})
+  local etcdmnt=$(podman mount "${etcdctr}")
+  cp ${etcdmnt}/bin/etcdctl $ASSET_DIR/bin
```
deads2k: not `ASSET_DIR`.  `ETCD_BINARY_DOWNLOAD_DIR` or some such.

bindata/etcd/etcd-restore-backup.sh
```diff
@@ -11,77 +12,49 @@ if [[ $EUID -ne 0 ]]; then
...
 
-ETCDCTL="${ASSET_DIR}/bin/etcdctl"
-ETCD_DATA_DIR=/var/lib/etcd
-ETCD_DATA_DIR_BACKUP=/var/lib/etcd-backup
-RESTORE_ETCD_POD_YAML="${CONFIG_FILE_DIR}/static-pod-resources/etcd-certs/configmaps/restore-etcd-pod/pod.yaml"
-ETCD_MANIFEST="${MANIFEST_DIR}/etcd-pod.yaml"
-ETCD_STATIC_RESOURCES="${CONFIG_FILE_DIR}/static-pod-resources/etcd-member"
-STOPPED_STATIC_PODS="${ASSET_DIR}/tmp/stopped-static-pods"
+# Move manifests and stop static pods
+if [ ! -d "$MANIFEST_STOPPED_DIR" ]; then
```
deads2k: Use a well-known working directory under static-pod-resources

bindata/etcd/etcd-member-remove.sh
deads2k: > not `ASSET_DIR`. `ETCD_BINARY_DOWNLOAD_DIR` or some such.

actually no, just use a well known location under static-pod-resources

bindata/etcd/etcd-restore-backup.sh
```diff
@@ -11,77 +12,49 @@ if [[ $EUID -ne 0 ]]; then
...
+done
 
-function run {
-  init
-  if [ ! -f "${SNAPSHOT_FILE}" ]; then
-    echo "etcd snapshot ${SNAPSHOT_FILE} does not exist."
-    exit 1
-  fi
+# Restore static pod resources
+tar -C ${CONFIG_FILE_DIR} -xzf $BACKUP_FILE static-pod-resources
```
deads2k: use well known location and directly copy instead of taring please.

bindata/etcd/etcd-restore-backup.sh
```diff
@@ -11,77 +12,49 @@ if [[ $EUID -ne 0 ]]; then
...
-STOPPED_STATIC_PODS="${ASSET_DIR}/tmp/stopped-static-pods"
+# Move manifests and stop static pods
+if [ ! -d "$MANIFEST_STOPPED_DIR" ]; then
+  mkdir $MANIFEST_STOPPED_DIR
+fi
 
-source "/usr/local/bin/openshift-recovery-tools"
+ls -1d ${MANIFEST_DIR}/* | while read STATIC_POD; do
+  echo "... stopping $STATIC_POD"
+  mv ${MANIFEST_DIR}/${STATIC_POD} $MANIFEST_STOPPED_DIR
```
deads2k: add a 30 second sleep and a TODO that says to use crictl to know when etcd is finished.

pkg/etcdenvvar/etcd_env.go
deads2k: I expected you to add the IMAGE_ETCD here.

deads2k: > I expected you to add the IMAGE_ETCD here.

nm, I see it a few lines down.

pkg/etcdenvvar/etcd_env.go
```diff
@@ -21,6 +21,7 @@ type envVarContext struct {
 	infrastructureLister configv1listers.InfrastructureLister
 	networkLister        configv1listers.NetworkLister
 	endpointLister       corev1listers.EndpointsLister
+	targetImagePullSpec, operatorImagePullSpec string
```
deads2k: i don't see `operatorImagePullSpec` used

deads2k: where is targetImagePullSpec set?

bindata/etcd/openshift-recovery-tools
```diff
@@ -1,472 +0,0 @@
...
-  if [ ! -d "$ASSET_BIN" ]; then
-    echo "Creating asset directory ${ASSET_DIR}"
-    for dir in {bin,tmp,shared,backup,templates,restore,manifests}; do
-      /usr/bin/mkdir -p ${ASSET_DIR}/${dir}
-    done
-  fi
-}
-
-# download and test etcdctl from upstream release assets
-dl_etcdctl() {
```
deads2k: looking at usage, this is the one shared function I think we have.  You can keep the one and the fixed env vars that are specific for these recovery scripts in this file.

bindata/etcd/etcd-restore-backup.sh
retroflexer: Added to the usage.

bindata/etcd/etcd-member-remove.sh
hexfusion: I really like the usability can we simplify bash as compromise
 
`etcdctl member list | awk -F, "/$NAME/{print \$1}"`

retroflexer: That is not exact match. If someone give infra1, it matches infra10, infra11, infra12 etc.

hexfusion: doh :)

`etcdctl member list | awk -F, "\$0 ~ /\s${NAME},/{print \$1}"`

hexfusion: I guess the point is we need this to be simpler or pass the ID.

bindata/etcd/etcd-restore-backup.sh
```diff
@@ -11,77 +12,52 @@ if [[ $EUID -ne 0 ]]; then
...
-ETCD_MANIFEST="${MANIFEST_DIR}/etcd-pod.yaml"
-ETCD_STATIC_RESOURCES="${CONFIG_FILE_DIR}/static-pod-resources/etcd-member"
-STOPPED_STATIC_PODS="${ASSET_DIR}/tmp/stopped-static-pods"
+# Move manifests and stop static pods
+if [ ! -d "$MANIFEST_STOPPED_DIR" ]; then
+  mkdir $MANIFEST_STOPPED_DIR
+fi
 
-source "/usr/local/bin/openshift-recovery-tools"
+ls -1d ${MANIFEST_DIR}/* | while read STATIC_POD; do
```
hexfusion: find instead of piping to while? Find can handle this natively.

```
find ${MANIFEST_DIR} \
  -maxdepth 1 \
  -type f \
  -printf '...stopping %P\n' \
  -exec mv {} $MANIFEST_STOPPED_DIR \;
```

pkg/etcdenvvar/etcd_env.go
```diff
@@ -88,10 +89,13 @@ func getEtcdctlEnvVars(envVarContext envVarContext) (map[string]string, error) {
...
-		"ETCDCTL_CACERT":    "/etc/kubernetes/static-pod-certs/configmaps/etcd-serving-ca/ca-bundle.crt",
-		"ETCDCTL_CERT":      "/etc/kubernetes/static-pod-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.crt",
-		"ETCDCTL_KEY":       "/etc/kubernetes/static-pod-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.key",
+		"ETCDCTL_CACERT":    "/etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-serving-ca/ca-bundle.crt",
+		"ETCDCTL_CACERT":    "/etc/kubernetes/static-pod-resources/etcd-certs/secrets/etcd-all-peer",
+		"ETCDCTL_CERT":      "/etc/kubernetes/static-pod-resources/etcd-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.crt",
+		"ETCDCTL_KEY":       "/etc/kubernetes/static-pod-resources/etcd-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.key",
 		"ETCDCTL_ENDPOINTS": endpoints,
+		"ETCD_IMAGE": envVarContext.targetImagePullSpec,
+		"ETCD_OPERATOR_IMAGE": envVarContext.operatorImagePullSpec,
```
deads2k: not needed

pkg/etcdenvvar/etcd_env.go
```diff
@@ -88,10 +89,13 @@ func getEtcdctlEnvVars(envVarContext envVarContext) (map[string]string, error) {
 	}
 	return map[string]string{
 		"ETCDCTL_API":       "3",
-		"ETCDCTL_CACERT":    "/etc/kubernetes/static-pod-certs/configmaps/etcd-serving-ca/ca-bundle.crt",
-		"ETCDCTL_CERT":      "/etc/kubernetes/static-pod-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.crt",
-		"ETCDCTL_KEY":       "/etc/kubernetes/static-pod-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.key",
+		"ETCDCTL_CACERT":    "/etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-serving-ca/ca-bundle.crt",
+		"ETCDCTL_CACERT":    "/etc/kubernetes/static-pod-resources/etcd-certs/secrets/etcd-all-peer",
```
deads2k: duplicate

bindata/etcd/etcd-snapshot-backup.sh
```diff
@@ -12,13 +12,33 @@ if [[ $EUID -ne 0 ]]; then
...
+function usage {
     echo 'Path to backup dir required: ./etcd-snapshot-backup.sh <path-to-backup-dir>'
     exit 1
 }
 
-ASSET_DIR=/home/core/assets
+#backup latest static pod resources for kube-apiserver
+function backup_latest_kube_static_resources {
+  echo "Trying to backup latest static pod resources.."
+  LATEST_STATIC_POD_DIR=$(ls -vd "${CONFIG_FILE_DIR}"/static-pod-resources/kube-apiserver-pod-[0-9]* | tail -1) || true
```
hexfusion: `CONFIG_FILE_DIR` is unbound here, isn't it?

hexfusion: you need to source the value I guess.

hexfusion: I see you sourced below function is called

retroflexer: The source is below at L58. It worked fine, as the shell evaluates variables at the time of invocation.
https://github.com/openshift/cluster-etcd-operator/blob/00c5f417617f9a2c008d8a3cad957c71406dae32/bindata/etcd/etcd-snapshot-backup.sh#L58

retroflexer: If you want, I can move it above.

bindata/etcd/etcd-snapshot-restore.sh
```diff
@@ -0,0 +1,74 @@
...
+SNAPSHOT_FILE=$(ls -vd "${BACKUP_DIR}"/snapshot*.db | tail -1) || true
+
+source etcd.env
+source etcd-common-tools
+
+if [ ! -f "${SNAPSHOT_FILE}" ]; then
+  echo "etcd snapshot ${SNAPSHOT_FILE} does not exist."
+  exit 1
+fi
+
```
alaypatel07: out of curiosity, why do we not care to check if `BACKUP_FILE` is not empty?

bindata/etcd/etcd-snapshot-restore.sh
```diff
@@ -0,0 +1,74 @@
...
+# //TO DO: verify using crictl that etcd and other pods stopped.
+
+# Remove data dir
+echo "Moving etcd data-dir ${ETCD_DATA_DIR}/member to ${ETCD_DATA_DIR_BACKUP}"
+[ ! -d ${ETCD_DATA_DIR_BACKUP} ]  && mkdir -p ${ETCD_DATA_DIR_BACKUP}
+mv ${ETCD_DATA_DIR}/member ${ETCD_DATA_DIR_BACKUP}/member
+
+
+# Restore static pod resources
+tar -C ${CONFIG_FILE_DIR} -xzf ${BACKUP_FILE} static-pod-resources
```
alaypatel07: again just curious: should moving kube static pod be after etcd is moved few lines below?

That way when kube-apiserver is started it will have an etcd to talk to, or we have made and effort for etcd to exist :)

bindata/etcd/etcd-snapshot-restore.sh
retroflexer: Well. We didn't have the size check in the past scripts :-). I was just trying to simplify the existing scripts.

bindata/etcd/etcd-snapshot-restore.sh
retroflexer: I can move it below. In the older scripts, we stopped kubelet, and restarted kubelet at the end. But based on David's comments, we let the kubelet be dirty.

PR #224, opened 2020-03-02
 [release-4.4] Bug 1809682: sync with master

hexfusion:
Bug 1809682: oc rsh to an etcd pod needs to have etcdctl "just work"
Bug 1809151: bump(make sh files executable)
Bug 1809142: add ca-bundles to fixed locations and use the for restore pod
Bug 1807278: pkg/operator/targetconfigcontroller: add platform raft tunables


hexfusion:
wth is up with GCP

PR #229, opened 2020-03-04
bug 1810133: remove usage of host-etcd

deads2k:
Stop maintaining host-etcd endpoints.  We will remove usage of this in 4.4.  For upgrade scenarios, the existing (and previously unmaintained) endpoints will remain exactly as-is.

In 4.5, we can remove the old endpoints entirely and stop producing them from the installer.


sanchezl:
Depends on https://github.com/openshift/cluster-kube-apiserver-operator/pull/787

PR #230, opened 2020-03-04
stop referencing DNS in env vars

deads2k:
Fixes variable names without changing behavior so we can have understandable content.


pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -310,6 +310,7 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
 		"etcd.openshift-etcd.svc.cluster.local",
 		"*." + etcdDiscoveryDomain,
 		"127.0.0.1",
+		"::1",
```
sanchezl:
```suggestion
		"0:0:0:0:0:0:0:1",
```
I know this eventually gets parsed into a net.IP, and probably makes it into the cert in the correct format, but technically, collapsed IPv6 addresses are not allowed in SAN.

PR #231, opened 2020-03-04
ensure that etcd members always have IPs

deads2k:
This adds a control loop which inspects the existing members of the etcd cluster and ensures that the first listed peer on every member is an IP address. We need this to migrate the existing members to using IPs to communicate instead of DNS names.  When we're finished, this control loop will be the only one which performs DNS lookups.

### DIFF
--- a/pkg/operator/etcdmemberipmigrator/etcdmemberipmigrator.go
+++ b/pkg/operator/etcdmemberipmigrator/etcdmemberipmigrator.go
@@ -1,0 +1,326 @@
+package etcdmemberipmigrator
+
+import (
+	"fmt"
+	"net"
+	"net/url"
+	"strings"
+	"time"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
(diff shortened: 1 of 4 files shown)


pkg/operator/etcdmemberipmigrator/etcdmemberipmigrator.go
```diff
@@ -0,0 +1,318 @@
...
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+)
+
+const (
+	workQueueKey = "key"
+)
+
+// watches etcd members and if they do not have a peer URL that has an IP address, the ip address is determined
+// and added as the first address.
```
sanchezl: What is the IP is not correct/changed?

deads2k: > What if the IP is not correct/changed?

before the join completed?  don't do that.

deads2k:
one possible theory: https://github.com/openshift/etcd/pull/38

PR #237, opened 2020-03-05
remove bootstrap annotation after bootstrapping

deads2k:
this cleans up the client usage in the operator and elsewhere after bootstrapping.


ironcladlou:
https://github.com/openshift/cluster-etcd-operator/pull/342 is rebased and supercedes this PR

/close

PR #238, opened 2020-03-05
cache the client based on the endpoints to avoid reconstruction

deads2k:
This may reduce log noise, but the etcd client has to be thread-safe


retroflexer:
/retitle Bug 1813373: cache the client based on the endpoints to avoid reconstruction

PR #240, opened 2020-03-06
Bug 1810997: pkg/operator/etcdcertsigner: always include ::1 in cert SAN

hexfusion:
We should not worry about conditionally doing this always add.


pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -305,6 +305,7 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
 		"etcd.openshift-etcd.svc.cluster.local",
 		"*." + etcdDiscoveryDomain,
 		"127.0.0.1",
+		"::1",
```
deads2k: in theory, this is the same as below.

deads2k:
eh, why not.  How much worse can it get

/lgtm

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: Yeah handshake does not seem smart enough to understand long form.

PR #244, opened 2020-03-07
require leader for membership list, make etcdmemberscontroller frequent

alaypatel07:
if the endpoint picked does not have a leader, it will fail and the
next call will be round robbined to a different etcd endpoint by client
load balancer

### DIFF
--- a/pkg/etcdcli/etcdcli.go
+++ b/pkg/etcdcli/etcdcli.go
@@ -207,7 +207,8 @@ func (g *etcdClientGetter) MemberList() ([]*etcdserverpb.Member, error) {
 	}
 	defer cli.Close()
 
-	ctx, cancel := context.WithCancel(context.Background())
+	// WithRequireLeader makes sure that the membership list is from the quorate cluster
+	ctx, cancel := context.WithCancel(clientv3.WithRequireLeader(context.Background()))
 	defer cancel()
 
 	membersResp, err := cli.MemberList(ctx)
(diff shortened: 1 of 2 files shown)


alaypatel07:
Manually verified on local cluster with the following scenario:

1. took a 3 nodes etcd cluster
2. stopped two nodes
3. queried membership list of the third

result:

```
...
...
...
{"level":"warn","ts":"2020-03-07T04:33:33.343-0500","caller":"clientv3/retry_interceptor.go:61","msg":"retrying of unary invoker failed","target":"endpoint://client-eb875d88-2739-42bf-ba84-bf026c5b58c4/localhost:32769","attempt":98,"error":"rpc error: code = Unavailable desc = etcdserver: no leader"}
{"level":"warn","ts":"2020-03-07T04:33:33.371-0500","caller":"clientv3/retry_interceptor.go:61","msg":"retrying of unary invoker failed","target":"endpoint://client-eb875d88-2739-42bf-ba84-bf026c5b58c4/localhost:32769","attempt":99,"error":"rpc error: code = Unavailable desc = etcdserver: no leader"}
error while getting member list etcdserver: no leader
```

deads2k:
This is cheap in terms of dev-time, but it still doesn't resolve the larger split-brain problem.  @hexfusion your choice about whether to take a half-solution that will make it harder for you to make the transition to the solution you'll want in the end: contact every member and make sure they all agree.

PR #246, opened 2020-03-09
The data-dir should be /var/lib/etcd instead of /var/lib/etcd/member …

retroflexer:
…to be consistent with its other uses elsewhere.


hexfusion:
/hold

https://github.com/openshift/etcd/blob/bd88bf4a80e2fd4f5e0121e3d53127bfda3674b8/openshift-tools/pkg/discover-etcd-initial-cluster/initial-cluster.go#L125

We are checking if data-dir exists would this ever fail? We are creating a bug with this change some thing needs to change for it to not.

hexfusion:
OK I see it is covered.

/hold cancel

PR #247, opened 2020-03-09
Bug 1812584: clustermembercontroller: skip if member found with hostname

alaypatel07:
In 4.3 the etcd member name was in the form of etcd-member-<node-name>.
During upgrade cluster member controller should not add a member with
<node-name> if etcd-member-<node-name> is running.

### DIFF
--- a/pkg/operator/clustermembercontroller/clustermembercontroller.go
+++ b/pkg/operator/clustermembercontroller/clustermembercontroller.go
@@ -220,9 +220,21 @@ func (c *ClusterMemberController) getEtcdPodToAddToMembership() (*corev1.Pod, er
 			continue
 		}
 
+		// check to see if this member is updating from 4.3
+		etcdMember, err := c.etcdClient.GetMember("etcd-member-" + pod.Spec.NodeName)
+		switch {
+		case apierrors.IsNotFound(err):
+			// this is not an upgrade from 4.3
+		case err != nil:
+			return nil, err
+		default:
+			klog.V(4).Infof("skipping unready pod %q because it is already an etcd member: %#v with hostname: %s", pod.Name, etcdMember, pod.Spec.Hostname)
+			return nil, nil
+		}
+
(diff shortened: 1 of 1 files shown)


pkg/etcdcli/etcdcli.go
```diff
@@ -247,7 +247,7 @@ func (g *etcdClientGetter) GetMember(name string) (*etcdserverpb.Member, error)
 		return nil, err
 	}
 	for _, m := range members {
-		if m.Name == name {
+		if strings.Contains(m.Name, name) {
```
deads2k: this is the wrong fix.  Have a `FuzzyMatchDeprecated43_memberName` or some such method. Don't break this method.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -220,9 +220,21 @@ func (c *ClusterMemberController) getEtcdPodToAddToMembership() (*corev1.Pod, er
...
 
+		// check to see if this member is updating from 4.3
+		etcdMember, err := c.etcdClient.GetMember(pod.Spec.Hostname)
+		switch {
+		case apierrors.IsNotFound(err):
+			// this is not an upgrade from 4.3
+		case err != nil:
+			return nil, err
+		default:
+			klog.Infof("skipping unready pod %q because it is already an etcd member: %#v with hostname: %s", pod.Name, etcdMember, pod.Spec.Hostname)
```
deads2k: will this be constantly printed in the upgrade cases?  if so, V(4)

alaypatel07: I think it will only be temporary and very rare to find. Can still push it to V(4)

alaypatel07:
/hold 

walking through with @hexfusion to vet if this will work

hexfusion:
continued on #253

PR #248, opened 2020-03-10
Manifest stopped dir should be created with "p" flag

retroflexer:
We cannot assume /home/core/assets directory exists.


hexfusion:
@retroflexer thanks for fixing this.
/lgtm

PR #249, opened 2020-03-10
Bug 1812236: etcdctl cannot find cert

sanchezl:


retroflexer:
This is not right.The paths on the node and pod are different because of the mount points are different. This change breaks our DR scripts.

PR #251, opened 2020-03-10
Fix etcdctl env var paths

cybertron:
These env vars are referencing paths that don't exist in the etcd
containers. When running etcdctl, the following error occurs:

Error: open /etc/kubernetes/static-pod-resources/etcd-certs/secrets/etcd-all-peer/etcd-peer-master-0.crt: no such file or directory

This is because the /etc/kubernetes/static-pod-resources mount in the
etcd containers is /etc/kubernetes/static-pod-resources/etcd-pod-X[0],
which contains the secrets directory with no interventing etcd-certs
directory. For example:

$ ls /etc/kubernetes/static-pod-resources/etcd-pod-3
configmaps  etcd-pod.yaml  secrets

This just removes the etcd-certs part of the path to make the env
vars point at the files as intended.

0: https://github.com/openshift/cluster-etcd-operator/blob/173b86cba63c98b2c5237cdab6e94994340957c9/bindata/etcd/pod.yaml#L202

### DIFF
--- a/pkg/etcdenvvar/etcd_env.go
+++ b/pkg/etcdenvvar/etcd_env.go
@@ -89,9 +89,9 @@ func getEtcdctlEnvVars(envVarContext envVarContext) (map[string]string, error) {
 	}
 	return map[string]string{
 		"ETCDCTL_API":       "3",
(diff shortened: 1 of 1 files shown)


cybertron:
Oops, https://github.com/openshift/cluster-etcd-operator/pull/249 beat me to it.

PR #256, opened 2020-03-12
restore-pod.yaml: fix temporary restore directory name

alaypatel07:


bindata/etcd/restore-pod.yaml
```diff
@@ -48,15 +48,18 @@ spec:
         fi
 
         UUID=$(uuidgen)
+        UUID=${UUID:0:10}
```
hexfusion: just use the whole ID or another unique string

alaypatel07: done

hexfusion:
/close

continued on #257

PR #257, opened 2020-03-12
bindata/etcd: backup and restore all static pods

hexfusion:
Disaster recovery involves restoring to previous state. This state is defined not only by etcd but also by the static-pod resources on disk. In order to traverse time properly both need to match the actual state.

The PR does the following.

- takes a snapshot of etcd state file and backup of the last modified revision of static pod resources:
**kube-apiserver-pod**
**kube-controller-manager-pod**
**kube-scheduler-pod**
**etcd-pod**

**Assumptions**: we need to make the assumption that the latest revision is last modified . We will read the revision from static manifest itself in future z-stream. The reason for this is if we go back in time we dont want to worry about the revisions on disk for all nodes. Meaning if we have a backup of rev 2 but the cluster is currently at 4. We will not remove the old revisions on each node. Instead, we restore etcd state to rev 2, make sure rev 2 is on disk and then force a new rev for each static pod operator.

### DIFF
--- a/bindata/etcd/etcd-snapshot-restore.sh
+++ b/bindata/etcd/etcd-snapshot-restore.sh
@@ -28,12 +28,42 @@ if [ "$1" == "" ] || [ ! -d "$1" ]; then
   usage
 fi
 
+function restore_static_pods() {
+  STATIC_PODS=("$@")
+
+  for POD_FILE_NAME in "${STATIC_PODS[@]}"; do
+    BACKUP_POD_PATH=$(tar -tvf ${BACKUP_FILE} "*${POD_FILE_NAME}" | awk '{ print $6 }') || true
+    if [ -z "${BACKUP_POD_PATH}" ]; then
+      echo "${POD_FILE_NAME} does not exist in ${BACKUP_FILE}"
+      exit 1
+    fi
+
+    echo "starting ${POD_FILE_NAME}"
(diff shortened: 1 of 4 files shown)


bindata/etcd/etcd-snapshot-restore.sh
```diff
@@ -49,24 +79,31 @@ find ${MANIFEST_DIR} \
...
 tar -C ${CONFIG_FILE_DIR} -xzf ${BACKUP_FILE} static-pod-resources
+
+# Copy snapshot to backupdir
+cp -p ${SNAPSHOT_FILE} ${ETCD_DATA_DIR_BACKUP}/snapshot.db
+
+echo "starting restore-etcd static pod"
+cp -p ${RESTORE_ETCD_POD_YAML} ${MANIFEST_DIR}/etcd-pod.yaml
+
+# start remaining static pods
+restore_static_pods "${STATIC_POD_LIST[@]}"
```
retroflexer: The STATIC_POD_LIST also has etcd-pod.yaml. Doesn't overwrite the recovery file we just copied on L196?

hexfusion: yep good catch bad copy and paste.

retroflexer:
Needs description and a BZ.

PR #262, opened 2020-03-13
[wip] Revert "Bug 1812584: fix scaling during upgrade and add unit tests"

hexfusion:
testing revert as this was last upgrade that passed


hexfusion:
This appears clear to me one more test FTR.

PR #265, opened 2020-03-16
Copy files from etcd-scripts to /usr/local/bin for convenience

retroflexer:
Copying the DR scripts from the unconventional etcd-scripts path to /usr/local/bin, so that the user can invoke it from anywhere.

### DIFF
--- a/bindata/etcd/etcd-member-remove.sh
+++ b/bindata/etcd/etcd-member-remove.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+### Created by cluster-etcd-operator. DO NOT edit.
 set -o errexit
 set -o pipefail
 set -o errtrace
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -10,7 +10,7 @@ metadata:
     revision: "REVISION"
 spec:
   initContainers:
-    - name: etcd-quorum-guard-copy
+    - name: etcd-resources-copy
       image: ${IMAGE}
       imagePullPolicy: IfNotPresent
       terminationMessagePolicy: FallbackToLogsOnError
@@ -23,6 +23,9 @@ spec:
 
           cp /etc/kubernetes/static-pod-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.crt /etc/kubernetes/etcd-backup-dir/system:etcd-peer-NODE_NAME.crt
           cp /etc/kubernetes/static-pod-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.key /etc/kubernetes/etcd-backup-dir/system:etcd-peer-NODE_NAME.key
+          rm -f $(grep -l '^### Created by cluster-etcd-operator' /usr/local/bin/*)
+          cp -p /etc/kubernetes/static-pod-certs/configmaps/etcd-scripts/*.sh /usr/local/bin
+
       resources:
         requests:
           memory: 60Mi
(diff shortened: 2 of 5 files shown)


bindata/etcd/pod.yaml
```diff
@@ -23,6 +23,11 @@ spec:
 
           cp /etc/kubernetes/static-pod-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.crt /etc/kubernetes/etcd-backup-dir/system:etcd-peer-NODE_NAME.crt
           cp /etc/kubernetes/static-pod-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.key /etc/kubernetes/etcd-backup-dir/system:etcd-peer-NODE_NAME.key
+          rm -f $(grep -l '^### Created by cluster-etcd-operator' /usr/local/bin/*)
+          ln -s /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd-member-remove.sh /usr/local/bin
+          ln -s /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd-snapshot-backup.sh /usr/local/bin
+          ln -s /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd-snapshot-restore.sh /usr/local/bin
```
hexfusion: do we really want to maintain this list? could we use find to symlink all bash.sh files?

retroflexer: We could do that too.

bindata/etcd/pod.yaml
```diff
@@ -23,6 +23,9 @@ spec:
 
           cp /etc/kubernetes/static-pod-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.crt /etc/kubernetes/etcd-backup-dir/system:etcd-peer-NODE_NAME.crt
           cp /etc/kubernetes/static-pod-certs/secrets/etcd-all-peer/etcd-peer-NODE_NAME.key /etc/kubernetes/etcd-backup-dir/system:etcd-peer-NODE_NAME.key
+          rm -f $(grep -l '^### Created by cluster-etcd-operator' /usr/local/bin/*)
+          cp -p /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/*.sh /usr/local/bin
+
```
hexfusion: If I run this I get 

> cp: target '/etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd-snapshot-restore.sh' is not a directory

bindata/etcd/etcd-member-remove.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+### Created by cluster-etcd-operator. DO NOT edit.
```
hexfusion: maybe in the future, we should have the process add this string but not something we need today.

hexfusion:
scripts come from bindata and you would need root to write so I think this is safe
```
$ ls -ld /etc/kubernetes/static-pod-resources
drwxr-xr-x. 26 root root 4096 Mar 16 17:15 /etc/kubernetes/static-pod-resources
```

alaypatel07:
not sure what was the conclusion but are we adding conf file to this? https://github.com/openshift/cluster-etcd-operator/pull/265/files#diff-f19a648fa992d810479d30657e849c6dR126

retroflexer:
> not sure what was the conclusion but are we adding conf file to this? https://github.com/openshift/cluster-etcd-operator/pull/265/files#diff-f19a648fa992d810479d30657e849c6dR126

We decided not to continue carrying etcd.conf. Sam said: "It is not an issue FTR." on slack discussions.

PR #268, opened 2020-03-17
Bug 1814479: bindata/etcd: run etcdctl inside of container

hexfusion:
This PR adds a new bash function exec_etcdctl. this allows for etcdctl commands to be run within the actual container on the host.

### DIFF
--- a/bindata/etcd/etcd-common-tools
+++ b/bindata/etcd/etcd-common-tools
@@ -20,3 +20,14 @@ function dl_etcdctl {
   podman rm "${etcdctr}"
   etcdctl version
 }
+
+# execute etcdctl command inside of running etcdctl container
+function exec_etcdctl {
+  local command="$@"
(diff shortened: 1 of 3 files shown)


hexfusion:
/hold 

lets do this right and just use the etcdctl container

bindata/etcd/etcd-common-tools
```diff
@@ -20,3 +20,14 @@ function dl_etcdctl {
   podman rm "${etcdctr}"
   etcdctl version
 }
+
+# execute etcdctl command inside of running etcdctl container
+function exec_etcdctl {
+  local command="$@"
+  local container_id=$(sudo crictl ps --label io.kubernetes.container.name=etcdctl -o json | jq -r '.containers[0].id') || true
```
retroflexer: `sudo` here is redundant. We check earlier that the script should be run with EUID as root.

PR #272, opened 2020-03-20
Bug 1815203: bindata/etcd: rename DR scripts to cluster-backup and cluster-restore

hexfusion:
This PR renames the following DR scripts. 

etcd-snapshot-backup.sh => cluster-backup.sh
etcd-snapshot-restore.sh => cluster-restore.sh

The name change serves 2 purposes.

1.) The script naming now better reflects the actions of the scripts as we are backing up not only etcd data.
2.) We cause a race in MCO when we write these files to disk with the same name. Although we could have MCD to make this work like we did with etcd-member.yaml. I feel that we should take this opportunity to remove ourselves from the MCO paths completely.

In the end the decision was made based on the stability of upgrades from 4.3.

We also remove etcd-member-remove.sh as it does not follow our expected user usage policy. We have introduced `etcdctl` container which provides very easy use of etcdctl. Users can run remove from this container vs ssh to disk of cluster.

`oc rsh ...`


retroflexer:
Looks straightforward to me.

/lgtm

fgiloux:
Mind that this kind of name change breaks the automation our users have in place. Would you consider creating a symlink and writing a deprecation notice in the release notes?

hexfusion:
@fgiloux yes this is a breaking change and should be documented. The reason we made the change was because of racing conditions with MCO that could result in upgrade failures. So I ma afraid the separation was required. 

> In the end the decision was made based on the stability of upgrades from 4.3.

I appreciate the note

hexfusion:
https://mail.google.com/mail/u/0/#inbox/FMfcgxwHMsPSwvQsglBDJgrbqbmDdGCR

PR #274, opened 2020-03-23
Stop kubelet during the cluster restore operation and restart at the end

retroflexer:
Stopping the kubelet service and restarting it after the backup is restored will flush out any dirty data kubelet may be carrying.

Also adding an extra `rm` to clean any archived directories in the etcd data-dir `/var/lib/etcd/`.

### DIFF
--- a/bindata/etcd/cluster-restore.sh
+++ b/bindata/etcd/cluster-restore.sh
@@ -96,7 +96,13 @@ if [ -d "${ETCD_DATA_DIR}/member" ]; then
   fi
   echo "Moving etcd data-dir ${ETCD_DATA_DIR}/member to ${ETCD_DATA_DIR_BACKUP}"
   mv ${ETCD_DATA_DIR}/member ${ETCD_DATA_DIR_BACKUP}/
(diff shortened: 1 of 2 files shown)


bindata/etcd/cluster-restore.sh
```diff
@@ -96,8 +96,14 @@ if [ -d "${ETCD_DATA_DIR}/member" ]; then
   fi
   echo "Moving etcd data-dir ${ETCD_DATA_DIR}/member to ${ETCD_DATA_DIR_BACKUP}"
   mv ${ETCD_DATA_DIR}/member ${ETCD_DATA_DIR_BACKUP}/
+  # Remove any other previously archived member directories
+  rm -rf ${ETCD_DATA_DIR}/*
```
retroflexer: This is needed, when one time I was restoring on a different master node, and found that there is archived member data in the directory /var/lib/etcd/restore-data-stamp-member-data. That disallows recovery pod to succeeded.

retroflexer:
We have worked out a procedure that requires no restarting of the kubelet. Closing this PR.

PR #266, opened 2020-03-17
Bug 1805034: *: allow a safe way to install with variable number of master nodes

alaypatel07:
With this CEO will read the number of expected etcd nodes
from install config stored in configmap `kube-system/install-config-v1`.
For platforms that don't have the configmap, it will default to 3.

CEO will only use the values if the bootstrap node is around. The 
installer team has warned against using the configmap after 
bootstrapping.

PR #279, opened 2020-03-26
[WIP]: allow less than 3 members if unsupported config override is set

alaypatel07:
The alternative to #266 

In order to create a cluster with less than 3 nodes, an admin would have to run the following after installer reports Kubernetes API is up.

```
$ oc patch etcd cluster -p='{"spec": {"unsupportedConfigOverrides": {"useUnsupportedUnsafeNonHANonProductionUnstableEtcd": true}}}' --type=merge
```


abhinavdahiya:
why not add a setting bootstrap_expect, that the ceo render on the bootstrap host can set to the number of control-plane replicas, but some users can patch to lower value if they seem fit?

pkg/operator/ceohelpers/unsupported_override.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func IsUnsupportedUnsafeEtcd(spec *operatorv1.StaticPodOperatorSpec) (bool, error) {
+	unsupportedConfig := map[string]interface{}{}
+	if err := json.NewDecoder(bytes.NewBuffer(spec.UnsupportedConfigOverrides.Raw)).Decode(&unsupportedConfig); err != nil {
+		klog.V(4).Infof("decode of unsupported config failed with error: %v", err)
+		return false, err
+	}
+
+	value, found, err := unstructured.NestedBool(unsupportedConfig, "useUnsupportedUnsafeNonHANonProductionUnstableEtcd")
```
hexfusion: since there is no typing here to my knowledge on this value, seems like assuming bool might end up being problematic. IE.

```
{"useUnsupportedUnsafeNonHANonProductionUnstableEtcd": "true" }
```

Should we accept any value besides string "false" and `false` as true?

hexfusion:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/279/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws/1202

> level=error msg="Cluster operator etcd Degraded is True with BootstrapTeardown_Error::EnvVarController_Error::InstallerController_Error::RevisionController_ContentCreationError::ScriptController_Error::StaticPods_Error: StaticPodsDegraded: pods \"etcd-ip-10-0-153-205.us-east-2.compute.internal\" not found\nStaticPodsDegraded: pods \"etcd-ip-10-0-139-39.us-east-2.compute.internal\" not found\nStaticPodsDegraded: pods \"etcd-ip-10-0-140-99.us-east-2.compute.internal\" not found\nScriptControllerDegraded: \"configmap/etcd-pod\": missing env var values\nInstallerControllerDegraded: missing required resources: [configmaps: etcd-scripts,restore-etcd-pod, configmaps: config-1,etcd-metrics-proxy-client-ca-1,etcd-metrics-proxy-serving-ca-1,etcd-peer-client-ca-1,etcd-pod-1,etcd-serving-ca-1, secrets: etcd-all-peer-1,etcd-all-serving-1,etcd-all-serving-metrics-1]\nBootstrapTeardownDegraded: EOF\nEnvVarControllerDegraded: EOF\nRevisionControllerDegraded: configmaps \"etcd-pod\" not found"

> level=fatal msg="Bootstrap failed to complete: failed to wait for bootstrapping to complete: timed out waiting for the condition"

pkg/operator/ceohelpers/unsupported_override.go
alaypatel07: addressed

pkg/operator/ceohelpers/unsupported_override.go
```diff
@@ -0,0 +1,26 @@
...
+// useUnsupportedUnsafeNonHANonProductionUnstableEtcd key is set
+// to any parsable value
+func IsUnsupportedUnsafeEtcd(spec *operatorv1.StaticPodOperatorSpec) (bool, error) {
+	unsupportedConfig := map[string]interface{}{}
+	if err := json.NewDecoder(bytes.NewBuffer(spec.UnsupportedConfigOverrides.Raw)).Decode(&unsupportedConfig); err != nil {
+		klog.V(4).Infof("decode of unsupported config failed with error: %v", err)
+		return false, err
+	}
+
+	_, found, err := unstructured.NestedFieldNoCopy(unsupportedConfig, "useUnsupportedUnsafeNonHANonProductionUnstableEtcd")
```
hexfusion: cant useUnsupportedUnsafeNonHANonProductionUnstableEtcd also be false?

alaypatel07: yeah but why? we will direct users to not use a bool value and use a `""` now they presence of key will determine the config

hexfusion: ok I missed the comment

alaypatel07:
/test e2e-gcp-upgrade 

last one flaked out https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/279/pull-ci-openshift-cluster-etcd-operator-master-e2e-gcp-upgrade/1250

pkg/operator/ceohelpers/unsupported_override.go
```diff
@@ -0,0 +1,29 @@
...
+	unsupportedConfig := map[string]interface{}{}
+	if spec.UnsupportedConfigOverrides.Raw == nil {
+		return false, nil
+	}
+	if err := json.NewDecoder(bytes.NewBuffer(spec.UnsupportedConfigOverrides.Raw)).Decode(&unsupportedConfig); err != nil {
+		klog.V(4).Infof("decode of unsupported config failed with error: %v", err)
+		return false, err
+	}
+
+	_, found, err := unstructured.NestedFieldNoCopy(unsupportedConfig, "useUnsupportedUnsafeNonHANonProductionUnstableEtcd")
```
deads2k, line 24: Users hate having bool-ish field not act like bools.  Check for truthy-ness

deads2k, line 24: Add a comment explaining why this is unsafe.  I think the answer is

1. this violates operational best practices for etcd - unstable
2. this allows non-HA configurations which we cannot support in production - unsafe and non-HA
3. this allows a situation where we can get stuck unable to re-achieve quorum, resulting in cluster-death - unsafe, non-production, unstable
4. The combination of all these things makes the situation unsupportable.

I think clearly communicating these reasons for choosing such a long name makes the consequences of this choice clear to all consumers.  The name isn't offensive, it is descriptive for the state it  places etcd in.

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -126,8 +127,7 @@ func (c *BootstrapTeardownController) removeBootstrap() error {
 		return updateErr
 	}
 
-	// check to see if bootstrapping is complete
```
deads2k, line 129: don't remove this. Explain what it is better.

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -165,7 +165,12 @@ func (c *BootstrapTeardownController) canRemoveEtcdBootstrap() (bool, bool, erro
 		return false, hasBootstrap, nil
 	}
 
-	if len(members) < 4 {
+	isUnsupportedUnsafeEtcd, err := c.isUnsupportedUnsafeEtcd()
+	if err != nil {
+		return false, hasBootstrap, err
+	}
+
+	if !isUnsupportedUnsafeEtcd && len(members) < 4 {
```
deads2k, line 173: make this a switch/case.  You missed adding a condition to be sure you had at least two members.

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
alaypatel07: did not intend to remove it, bad refactor!

pkg/operator/ceohelpers/unsupported_override.go
alaypatel07: @hexfusion was concerned if we would panic if a user does a string `"true"` in the yaml instead of `bool` hence added test cases to make sure we consider both bool and string.

alaypatel07: added, thanks for drafting the reasons.

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
alaypatel07: done

alaypatel07:
infra

/test e2e-aws

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -165,17 +166,30 @@ func (c *BootstrapTeardownController) canRemoveEtcdBootstrap() (bool, bool, erro
...
 	}
 
-	if len(members) < 4 {
+	isUnsupportedUnsafeEtcd, err := c.isUnsupportedUnsafeEtcd()
+	if err != nil {
+		return false, hasBootstrap, err
+	}
+
+	switch {
+	case !isUnsupportedUnsafeEtcd && len(members) < 4:
```
deads2k, line 175: comment the cases or you'll forget what you're trying to do

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -165,17 +166,30 @@ func (c *BootstrapTeardownController) canRemoveEtcdBootstrap() (bool, bool, erro
...
 		return false, hasBootstrap, nil
+	case isUnsupportedUnsafeEtcd && len(members) < 2:
+		return false, hasBootstrap, nil
+	default:
+		// do nothing fall through
```
deads2k, line 180: explain what you're falling through to

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -165,17 +166,30 @@ func (c *BootstrapTeardownController) canRemoveEtcdBootstrap() (bool, bool, erro
...
 	unhealthyMembers, err := c.etcdClient.UnhealthyMembers()
 	if err != nil {
 		return false, hasBootstrap, nil
 	}
 
-	// the etcd-bootstrap member is allowed to be unhealthy and can still be removed
+	// the etcd-bootstrap member is allowed to be unhealthy and can still be removed with exception of isUnsupportedUnsafeEtcd
 	switch {
+	// no member is allowed to be unhealthy if isUnsupportedUnsafeEtcd
+	case isUnsupportedUnsafeEtcd && len(unhealthyMembers) != 0:
```
deads2k, line 191: I don't see why this case is here.

pkg/operator/ceohelpers/unsupported_override.go
```diff
@@ -0,0 +1,57 @@
...
+		return false, err
+	}
+	if !found {
+		return false, nil
+	}
+	switch value.(type) {
+	case bool:
+		return value.(bool), nil
+	case string:
+		return value.(string) == "true", nil
```
deads2k, line 53: strconv has methods

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
alaypatel07: @deads2k if this is a two-member cluster etcd cluster one with etcd-bootstrap and other with master node, allowing an unhealthy member will lead to quorum loss and all subsequent `MemberRemove(etcd-bootstrap)` calls will fail

deads2k: read down, the default case covers this I think.

alaypatel07: for a two-node etcd cluster (single master openshift cluster) the workflow has to be:
1. Remove etcd-bootstrap member
2. destroy bootstrap

alaypatel07: hrm thinking again, not the default case but:
```
case len(unhealthyMembers) > 1:
```
will cover it because both the member will be unhealthy.

alaypatel07: I will update restoring the previous cases

pkg/operator/ceohelpers/unsupported_override.go
alaypatel07: +1, thanks

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
alaypatel07: added the comments

deads2k:
/lgtm
/hold

hold for comment by @hexfusion

abhinavdahiya:
Can you please explain how this is going to be  or is expected to be setup during installation. like for libvirt..

hexfusion:
> Can you please explain how this is going to be or is expected to be setup during installation. like for libvirt..

We will document how to make the patch against apiserver during install as a workaround for 4.4

praveenkumar:
> We will document how to make the patch against apiserver during install as a workaround for 4.4

@hexfusion where are we documenting it, from installer side or from etcd side?

Issue #278, opened 2020-03-25
DR bug: theory and proposed solution 


alaypatel07:
The DR scenario was created by stopping two machines, the backed up data was at revision 4. After restore pods runs the control plane is up with a single master node. Revision 4 has the three nodes(1 running other two terminated) env variables that were from backup time. New nodes are created using machine API. Two new revisions are created, revision 5 and 6. Revision 5 still has old data, revision 6 has new nodes data. A pod on the new node is created with revision 5. Since it has old node data, it is not able to find the correct NODE_NODE_ENVVAR and it crashloops. This leads to full-blown outage.

In order to save CEO from this scenario I am proposing of doing the following:
1. Add the list of nodes at the time revision was created in the revision itself. Create a new configmap `nodes-<revision-number>`
2. Use the pod mutation function to get `nodes-<revision-number>` and check if the node that the installer is run on belongs to the list of the nodes in configmap. If not, this revision is not usable for the node override the `pod.Command` of the installer pod with `echo "This revision does not include the node"; exit 1` . 

This will make sure that the new nodes does not have a static pod with faulty older revision and skip to the latest available revisions

hexfusion:
```
      {
        "lastTransitionTime": "2020-03-25T12:50:07Z",
        "message": "NodeInstallerProgressing: 2 nodes are at revision 0; 1 nodes are at revision 3; 0 nodes have achieved new revision 5",
        "reason": "NodeInstaller",
        "status": "True",
        "type": "Progressing"
      },
      {
        "lastTransitionTime": "2020-03-25T12:00:34Z",
        "message": "StaticPodsAvailable: 1 nodes are active; 2 nodes are at revision 0; 1 nodes are at revision 3; 0 nodes have achieved new revision 5\nEtcdMembersAvailable: ip-10-0-131-16.us-west-1.compute.internal members are available,  have not started,  are unhealthy,  are unknown",
        "reason": "AsExpected",
        "status": "True",
        "type": "Available"
      },
```

alaypatel07:
@hexfusion yes, the cluster I was looking at had rev numbers 4, 5, and 6 instead of 3,4 and 5 like in the example you shared.

I think what happens is that since a node is on rev 3,  static pod controller tries to make sure all nodes are on rev 3. The new nodes cannot run until rev 5. Hence we need to make sure that the revision contains the list of nodes they are supposed to run on, which is in line with the proposed solution

PR #284, opened 2020-03-26
Bug 1817028: *: add an init container to stop the pod with bad revision

alaypatel07:
Add an init container that fails the pod if it has stale revision data.
This can especially occur during DR as explained in #278. 

This is a two-pronged solution:
1. Add the init container that fails on bad env data
2. Modify the clustermembercontroller to not add the member unless all 
init containers have exit 0.

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -10,6 +10,29 @@ metadata:
     revision: "REVISION"
 spec:
   initContainers:
+    - name: etcd-ensure-env-vars
+      image: ${IMAGE}
+      imagePullPolicy: IfNotPresent
+      terminationMessagePolicy: FallbackToLogsOnError
+      command:
+        - /bin/sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          : "${NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST?not set}"
+          : "${NODE_NODE_ENVVAR_NAME_ETCD_NAME?not set}"
+          : "${NODE_NODE_ENVVAR_NAME_IP?not set}"
+
+      resources:
+        requests:
+          memory: 60Mi
+          cpu: 30m
+      securityContext:
+        privileged: true
+      env:
+${COMPUTED_ENV_VARS}
     - name: etcd-resources-copy
       image: ${IMAGE}
       imagePullPolicy: IfNotPresent
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -25,6 +25,18 @@ spec:
           : "${NODE_NODE_ENVVAR_NAME_ETCD_NAME?not set}"
           : "${NODE_NODE_ENVVAR_NAME_IP?not set}"
 
+          if [ "${NODE_NODE_ENVVAR_NAME_IP}" != "${NODE_IP}" ]; then
+            # echo the error message to stderr
+            echo "Expected node IP to be ${NODE_IP} got ${NODE_NODE_ENVVAR_NAME_IP}" >&2
+            exit 1
+          fi
+
+          if [ "${NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST}" != "${NODE_IP}" ]; then
+            # echo the error message to stderr
+            echo "Expected etcd url host to be ${NODE_IP} got ${NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST}" >&2
+            exit 1
+          fi
+
       resources:
         requests:
           memory: 60Mi
@@ -33,6 +45,10 @@ spec:
         privileged: true
       env:
 ${COMPUTED_ENV_VARS}
+      - name: NODE_IP
+        valueFrom:
+          fieldRef:
+            fieldPath: status.podIP
     - name: etcd-resources-copy
       image: ${IMAGE}
       imagePullPolicy: IfNotPresent
--- a/pkg/operator/clustermembercontroller/clustermembercontroller.go
+++ b/pkg/operator/clustermembercontroller/clustermembercontroller.go
@@ -208,15 +208,17 @@ func (c *ClusterMemberController) getEtcdPodToAddToMembership() (*corev1.Pod, er
 		if strings.HasPrefix(pod.Name, "etcd-member") {
 			continue
 		}
-		ready := false
+		isEtcdContainerRunning, isEtcdContainerReady := false, false
 		for _, containerStatus := range pod.Status.ContainerStatuses {
 			if containerStatus.Name != "etcd" {
 				continue
 			}
-			ready = containerStatus.Ready
+			// set running and ready flags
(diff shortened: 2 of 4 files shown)


bindata/etcd/pod.yaml
```diff
@@ -10,6 +10,35 @@ metadata:
...
+      imagePullPolicy: IfNotPresent
+      terminationMessagePolicy: FallbackToLogsOnError
+      command:
+        - /bin/sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          if [ -z "$NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST"]; then
```
hexfusion: these are still unbound right?

These blocks can be simplified to something like below 

```
: "${NODE_NODE_ENVVAR_NAME_ETCD_NAME?not set}"
```
ref: https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion

hexfusion: @alaypatel07 if you tested this you would have seen other errors as well, you need a space after quote and before right bracket.
... `URL_HOST"]; then`
```
 [: missing `]'
```

bindata/etcd/pod.yaml
```diff
@@ -10,6 +10,35 @@ metadata:
...
+            echo "env var $NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST is not set"
+            exit 1;
+          fi
+          if [ -z "$NODE_NODE_ENVVAR_NAME_ETCD_NAME"]; then
```
retroflexer: Again needs space before the `]`

bindata/etcd/pod.yaml
```diff
@@ -10,6 +10,35 @@ metadata:
...
+            echo "env var $NODE_NODE_ENVVAR_NAME_ETCD_NAME is not set"
+            exit 1;
+          fi
+          if [ -z "$NODE_NODE_ENVVAR_NAME_IP"]; then
```
retroflexer: Space before `]` here too.

bindata/etcd/pod.yaml
alaypatel07: @retroflexer https://github.com/openshift/cluster-etcd-operator/pull/284#discussion_r399206436 gave me enough pointers :) but thanks for the review

hexfusion:
> {"component":"entrypoint","file":"prow/entrypoint/run.go:168","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Entrypoint received interrupt: terminated","time":"2020-03-27T20:04:41Z"}

prow blew up

/test all

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -220,6 +220,14 @@ func (c *ClusterMemberController) getEtcdPodToAddToMembership() (*corev1.Pod, er
 			continue
 		}
 
+		if (pod.Status.InitContainerStatuses[1].State.Terminated == nil) ||
```
alaypatel07: The intention here is to make sure that we have a:
1. Terminated state on the init containers. This will make sure that they are not in Running state
2. The exit code on the Terminated state is 0.

retroflexer: Perhaps add this comment in the code.

alaypatel07: Just realised maybe I can do simpler. Both the conditions are satisfied if "etcd" container is running, I will try to replace the if condition

hexfusion:
sigh why are we failing AWS

hexfusion:
>             s: "promQL query: count_over_time(ALERTS{alertname!~\"Watchdog|AlertmanagerReceiversNotConfigured|KubeAPILatencyHigh\",alertstate=\"firing\",severity!=\"info\"}[2h]) >= 1 had reported incorrect results:\n[{\"metric\":{\"alertname\":\"TargetDown\",\"alertstate\":\"firing\",\"job\":\"metrics\",\"namespace\":\"openshift-console-operator\",\"service\":\"metrics\",\"severity\":\"warning\"},\"value\":[1585834134.713,\"51\"]}]",

we need to get the bottom of this

bindata/etcd/pod.yaml
```diff
@@ -10,6 +10,29 @@ metadata:
...
+      imagePullPolicy: IfNotPresent
+      terminationMessagePolicy: FallbackToLogsOnError
+      command:
+        - /bin/sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          : "${NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST?not set}"
```
deads2k, line 24: I can't read this.  I expected to see a downward API injection of IP address and then a bash comparison between the computed IP and the downward API IP.

bindata/etcd/pod.yaml
```diff
@@ -10,6 +10,55 @@ metadata:
...
+          : "${NODE_NODE_ENVVAR_NAME_ETCD_NAME?not set}"
+          : "${NODE_NODE_ENVVAR_NAME_IP?not set}"
+
+          if [ "${NODE_NODE_ENVVAR_NAME_IP}" != "${POD_IP}" ]; then
+            # echo the error message to stderr
+            >&2 echo "Expected node IP to be ${NODE_IP} got ${NODE_NODE_ENVVAR_NAME_IP}";
+            exit 1;
+          fi
+
+          if [ "${NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST}" != "${NODE_NAME}" ]; then
```
retroflexer, line 34: Is this correct? I thought NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST is now going to be an IP address, and not NODE_NAME. On my cluster now, I see IP address for the URL_HOST value.

alaypatel07: goood catch, updated

bindata/etcd/pod.yaml
```diff
@@ -10,6 +10,55 @@ metadata:
...
+          if [ "${NODE_NODE_ENVVAR_NAME_IP}" != "${NODE_IP}" ]; then
+            # echo the error message to stderr
+            echo "Expected node IP to be ${NODE_IP} got ${NODE_NODE_ENVVAR_NAME_IP}" >&2
+            exit 1
+          fi
+
+          if [ "${NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST}" != "${NODE_IP}" ]; then
+            # echo the error message to stderr
+           echo "Expected etcd url host to be ${NODE_IP} got ${NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST}" >&2
+            exit 1
```
hexfusion, line 37: indenting is off here.

alaypatel07: fixed

hexfusion:
/lgtm

manual tested in DR

PR #289, opened 2020-03-31
Bug 1819472: manifests: update CVO managed configmaps to create-only

hexfusion:
This PR removes some relics of the old operator and makes sure that ConfigMaps managed by CVO are created only once and not updated.

What we saw was CVO stomping certs every 10mins or so causing controllers to sync and generate a  bunch of I/O. Also because the configmap had default values when CVO rewrote the ConfigMap we actually could fail some etcd client connections because the certs would have invalid values.


hexfusion:
/test e2e-gcp-upgrade
--

hexfusion:
> s: "Cluster did not complete upgrade: timed out waiting for the condition: Could not update job \"openshift-service-catalog-removed/openshift-service-catalog-apiserver-remover\" (492 of 565): the object is invalid, possibly due to local cluster configuration",

openshift-service-catalog

/test e2e-gcp-upgrade

hexfusion:
digging on logs 

/test e2e-gcp-upgrade

hexfusion:
infra flake 

/test e2e-gcp-upgrade

hexfusion:
upgrade test appears to be a zombie.

/test e2e-gcp-upgrade

PR #295, opened 2020-04-08
vendor: bump etcd to 3.4.7

hexfusion:
We are bumping etcd server in 4.5 this change will bump the client to match.


hexfusion:
skip

PR #297, opened 2020-04-08
Bug 1826140: add infra flag and configure raft for Azure

alaypatel07:
This PR adds the ability to parse the platform for the cluster
during bootstrap. Since the default for the flag is set appropriately
it is expected to work out of the box. The next step would be
explicitly supplying the flag with value in the installer and making it
required.

This also paves the way to check if libvirt and render the etcd CR
with unsupported config override from #279 eliminating the need
for manual run

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
@@ -104,6 +104,10 @@ spec:
       value: "/var/lib/etcd"
     - name: ETCD_NAME
       value: "etcd-bootstrap"
+    - name: ETCD_HEARTBEAT_INTERVAL
+      value: {{ .EtcdHeartbeat }}
+    - name: ETCD_ELECTION_TIMEOUT
+      value: {{ .EtcdElectionTimeout }}
     ports:
     - name: peer
       containerPort: 2380
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -160,7 +160,9 @@ type TemplateData struct {
 	BootstrapIP string
 
 	// Platform is the underlying provider the cluster is run on.
-	Platform string
+	Platform            string
+	EtcdHeartbeat       string
+	EtcdElectionTimeout string
 }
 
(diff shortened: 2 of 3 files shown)


bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
```diff
@@ -120,6 +120,8 @@ spec:
       #!/bin/sh
       set -euo pipefail
 
+      {{ if eq .Platform "Azure"}} export ETCD_HEARTBEAT_INTERVAL="500" {{ end }}
+
```
hexfusion, line 124: we also need to bump leader election timeout[1].

[1] https://github.com/openshift/cluster-etcd-operator/blob/8f7718d45c4a31235c276322365ba10432921adb/pkg/etcdenvvar/etcd_env.go#L225

alaypatel07: updated

hexfusion:
In the future, we may want something more robust to handle additional provider cases but I think this solves our problem. Thanks @alaypatel07

pkg/cmd/render/render.go
```diff
@@ -154,6 +158,11 @@ type TemplateData struct {
 
 	// BootstrapIP is address of the bootstrap node
 	BootstrapIP string
+
+	// Platform is the underlying provider the cluster is run on.
+	Platform            string
+	EtcdHeartBeat       string
```
hexfusion, line 164: nit EtcdHeartbeat

pkg/cmd/render/render_test.go
```diff
@@ -97,12 +98,28 @@ spec:
...
+	infraConfig = `
+apiVersion: config.openshift.io/v1
+kind: Infrastructure
+metadata:
+  name: cluster
+spec:
+  cloudConfig:
+    name: ""
+status:
+  platform: AWS
```
hexfusion, line 111: can we have a test case where platform is empty?

alaypatel07: added it in the Platform Unit test. PTAL

PR #304, opened 2020-04-09
Bug 1822924: The snapshot backup must be requested from a single endpoint not multiple

retroflexer:
The snapshot backup command must be requested for a single endpoint, and not multiple endpoints. The etcd versions of 3.3.x were forgiving but bumping to etcd to 3.4.7 fails with an error when ETCDCTL_ENDPOINTS contain multiple members.

### DIFF
--- a/bindata/etcd/cluster-backup.sh
+++ b/bindata/etcd/cluster-backup.sh
@@ -66,5 +66,5 @@ fi
 
 dl_etcdctl
 backup_latest_kube_static_resources "${BACKUP_RESOURCE_LIST[@]}"
-etcdctl snapshot save ${SNAPSHOT_FILE}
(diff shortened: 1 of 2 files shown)


retroflexer:
/hold

The backup script also need fixing.

alaypatel07:
Infact if I am reading this right, the restore command does not read(and hence need) `endpoints` flag 

https://github.com/etcd-io/etcd/blob/34bd797e6754911ee540e8c87f708f88ffe89f37/etcdctl/ctlv3/command/snapshot_command.go#L139-L138

alaypatel07:
/hold 

@retroflexer a nit, can you please reword the commit,? It is still pointing at restore pod. Thanks

PR #309, opened 2020-04-14
etcdcli: refactor etcdcli package to fix known issues

alaypatel07:
This PR:
- adds an EtcdEndpointGetter interface so that endpoints can be for
 overridden in tests
- fix the getEtcdCachedClient to check for a closed connection
- fix the bug with UnhealthyMember
- refactor the receiver to etcdClient as per Golang pattern

I plan to add test cases that can prove the stability of etcdcli code changes
as a follow-up PR.

### DIFF
--- a/pkg/etcdcli/etcdcli.go
+++ b/pkg/etcdcli/etcdcli.go
@@ -16,9 +16,11 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	"go.etcd.io/etcd/clientv3"
+	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
 	"go.etcd.io/etcd/etcdserver/etcdserverpb"
 	"go.etcd.io/etcd/pkg/transport"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


pkg/etcdcli/interfaces.go
```diff
@@ -41,3 +41,9 @@ type UnhealthyMemberLister interface {
 type MemberStatusChecker interface {
 	MemberStatus(member *etcdserverpb.Member) string
 }
+
+// EtcdEndpointsGetter hides the implementation for getting etcd endpoints
+// so that custom fake endpoints can be used in tests
+type EtcdEndpointsGetter interface {
+	GetEtcdEndpoints() ([]string, error)
```
mfojtik, line 48: rule of thumb is to use just Get() if the interface name is `Getter` and especially `EtcdEndpointGetter` :-)

pkg/etcdcli/etcdcli.go
```diff
@@ -101,27 +83,73 @@ func (g *etcdClientGetter) getEtcdClient() (*clientv3.Client, error) {
...
 			utilruntime.HandleError(err)
 		}
 	}
-	g.cachedClient = c
-	g.lastClientConfigKey = etcdEndpoints
+	e.cachedClient = c
+	e.lastClientConfigKey = etcdEndpoints
 
-	return g.cachedClient, nil
+	// set endpoints to existing members
```
alaypatel07, line 144: This will avoid calls to bootstrap endpoint even if it is present during original client creation.

pkg/etcdcli/interfaces.go
alaypatel07: makes sense, updated. thanks for the nit, +1

pkg/etcdcli/etcdcli.go
alaypatel07: will fix https://github.com/openshift/okd/issues/139

PR #313, opened 2020-04-17
Bug 1827585: Add controller that watches leader changes and capture disk metrics

mfojtik:
The controller added here will watch the `etcd_server_leader_changes_seen_total` metrics and get the value for last 5 minutes. If leader changes are seen (value>0), then it will query prometheus to get `etcd_disk_wal_fsync_duration_seconds_bucket` for last 5 minutes.

This is then reported as Warning event for etcd operator. In future this can become a degraded condition.

### DIFF
--- a/pkg/operator/metriccontroller/fsync_controller.go
+++ b/pkg/operator/metriccontroller/fsync_controller.go
@@ -1,0 +1,100 @@
+package metriccontroller
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog"
+
+	configclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+type FSyncController struct {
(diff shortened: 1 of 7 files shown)


mfojtik:
@deads2k @hexfusion lets finish this :-)

mfojtik:
only last commit has the change

mfojtik:
@stevekuznetsov any clues about what is happening here?

mfojtik:
@hexfusion @deads2k @alaypatel07 this is now ready for review.

pkg/operator/metriccontroller/fsync_controller.go
```diff
@@ -0,0 +1,100 @@
...
+	return factory.New().ResyncEvery(1*time.Minute).WithSync(c.sync).ToController("FSyncController", recorder)
+}
+
+func (c *FSyncController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	client, err := getPrometheusClient(ctx, c.secretClient)
+	if errors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
```
deads2k, line 42: does returning an error from this method mean the operator goes degrade?  if so, this should not error.

pkg/operator/metriccontroller/fsync_controller.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	// First check how many leader changes we see in last 15 minutes
+	etcdLeaderChangesResult, _, err := client.Query(ctx, "increase((max by (job) (etcd_server_leader_changes_seen_total) or 0*absent(etcd_server_leader_changes_seen_total))[5m:1m])", time.Now())
+	if err != nil {
+		return err
```
deads2k, line 48: we don't want to degrade if we cannot query.

pkg/operator/metriccontroller/fsync_controller.go
mfojtik: it does not in this case.

deads2k:
mfojtik says it doesn't go degraded, just gives an event and logs.  I like events.

I think at some point fsync is worth on its own, but let's start somewhere

/lgtm

PR #317, opened 2020-04-21
Bug 1825667: pkg/cmd/render: add support for s390 and s390x arch

hexfusion:
This PR replaces tolerance for upstream etcd unsupported architectures `s390` and `s390x`. Because `ETCD_UNSUPPORTED_ARCH` should only be set in ENV if it is defined we took the liberty of making the `env:`  section of the bootstrap static pod spec more inline with what we do with the etcdenvvar controller.

This will make future changes to runtime env's much easier to add and manage.

### DIFF
--- a/pkg/cmd/render/env.go
+++ b/pkg/cmd/render/env.go
@@ -1,0 +1,92 @@
+package render
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/openshift/cluster-etcd-operator/pkg/etcdenvvar"
+)
+
+var envVarFns = []envVarFunc{
+	getFixedEtcdEnvVars,
+	getHeartbeatInterval,
+	getElectionTimeout,
+	getUnsupportedArch,
+	getEtcdName,
+}
+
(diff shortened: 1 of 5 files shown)


hexfusion:
/test e2e-aws

infra flake

pkg/cmd/render/env.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	return map[string]string{
+		"ETCD_ELECTION_TIMEOUT": timeout,
+	}, nil
+
+}
+
+func getUnsupportedArch(platform, arch string) (map[string]string, error) {
+	if !strings.HasPrefix(arch, "s390") {
```
Prashanth684, line 82: can't we just use `runtime.GOARCH` directly here? instead of passing arch to all the functions ?

hexfusion: I did that for testability but I don't feel overly strongly about it.

Prashanth684: ah..i see...so this was for the purpose of the unit test....

PR #318, opened 2020-04-22
Bug 1821286: Unstarted members cannot be blank in the log/event messages

retroflexer:
etcd doesn't store the name for the unstarted members. It is initialized only when the etcd is started. So, when reporting unstarted members, currently the message appears with blank name, which can be confusing. This PR attempts to fill the name with hostname (or IP address) extracted from peer URL for the unstarted members.

### DIFF
--- a/pkg/etcdcli/etcdcli.go
+++ b/pkg/etcdcli/etcdcli.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"net/url"
 	"reflect"
 	"strings"
 	"sync"
@@ -274,6 +275,19 @@ func (g *etcdClientGetter) GetMember(name string) (*etcdserverpb.Member, error) 
 	return nil, apierrors.NewNotFound(schema.GroupResource{Group: "etcd.operator.openshift.io", Resource: "etcdmembers"}, name)
 }
 
+// If the member's name is not set, extract ip/hostname from peerURL. Useful with unstarted members.
+func GetMemberNameOrHost(member *etcdserverpb.Member) string {
+	if len(member.Name) == 0 {
+		u, err := url.Parse(member.PeerURLs[0])
+		if err != nil {
+			klog.Errorf("unstarted member has invalid peerURL: %#v", err)
+			return "NAME-PENDING-BAD-PEER-URL"
+		}
+		return fmt.Sprintf("NAME-PENDING-%s", u.Hostname())
+	}
+	return member.Name
+}
+
 func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
 	cli, err := g.getEtcdClient()
 	if err != nil {
@@ -294,7 +308,7 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
 	for _, member := range membersResp.Members {
 		if len(member.ClientURLs) == 0 {
 			unhealthyMembers = append(unhealthyMembers, member)
-			unstartedMemberNames = append(unstartedMemberNames, member.Name)
+			unstartedMemberNames = append(unstartedMemberNames, GetMemberNameOrHost(member))
 			continue
 		}
 		ctx, cancel := context.WithCancel(context.Background())
--- a/pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
+++ b/pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
@@ -84,7 +84,7 @@ func (c *EtcdMembersController) reportEtcdMembers() error {
 		case etcdcli.EtcdMemberStatusAvailable:
 			availableMembers = append(availableMembers, m.Name)
 		case etcdcli.EtcdMemberStatusNotStarted:
-			unstartedMembers = append(unstartedMembers, m.Name)
+			unstartedMembers = append(unstartedMembers, etcdcli.GetMemberNameOrHost(m))
 		case etcdcli.EtcdMemberStatusUnhealthy:
 			unhealthyMembers = append(unhealthyMembers, m.Name)
 		}


pkg/etcdcli/etcdcli.go
```diff
@@ -294,7 +295,16 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
 	for _, member := range membersResp.Members {
 		if len(member.ClientURLs) == 0 {
 			unhealthyMembers = append(unhealthyMembers, member)
-			unstartedMemberNames = append(unstartedMemberNames, member.Name)
+			unstartedName := member.Name
+			if unstartedName == "" {
```
alaypatel07, line 299: should we update the condition on L296 to this? I believe missing member name is the indication of unstarted member as used by etcdcli[1].

1. https://sourcegraph.com/github.com/etcd-io/etcd@6aea6edc341a1e15cfb86b765c9b9c0423516cc7/-/blob/etcdctl/ctlv3/command/printer.go?subtree=true#L171

pkg/etcdcli/etcdcli.go
```diff
@@ -294,7 +295,16 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
...
+				u, err := url.Parse(member.PeerURLs[0])
+				if err == nil {
+					unstartedName = u.Hostname()
+				} else {
+					unstartedName = "NO-NAME-MEMBER"
```
alaypatel07, line 304: should this error be propagated up the stack? Is it possible for an admin to add a member and get caught here? Maybe more of a design decision that needs wider discussion, but okay for now

alaypatel07:
lgtm, leaving the tag for @hexfusion if he has comments. Thanks

pkg/etcdcli/etcdcli.go
```diff
@@ -274,6 +275,20 @@ func (g *etcdClientGetter) GetMember(name string) (*etcdserverpb.Member, error)
 	return nil, apierrors.NewNotFound(schema.GroupResource{Group: "etcd.operator.openshift.io", Resource: "etcdmembers"}, name)
 }
 
+// If the member's name is not set, extract ip/hostname from peerURL. Useful with unstarted members.
+func (g *etcdClientGetter) GetMemberNameOrHost(member *etcdserverpb.Member) string {
```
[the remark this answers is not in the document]
retroflexer: Well, because it is invoked from  pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go too.

retroflexer:
> @retroflexer, in general, I think we can define an unstarted member as something like
> 
> ```
> fmt.Sprintf("UNSTARTED-NAME-PENDING.%s", hostname)
> ```
> 
> making it more clear why we see the naming?

I agree. But my only apprehension is that with IPv6 it could become very lengthy and unreadable.

pkg/etcdcli/etcdcli.go
```diff
@@ -274,6 +275,20 @@ func (g *etcdClientGetter) GetMember(name string) (*etcdserverpb.Member, error)
...
+	memberName := member.Name
+	if len(memberName) == 0 {
+		u, err := url.Parse(member.PeerURLs[0])
+		if err == nil {
+			memberName = u.Hostname()
+		} else {
+			memberName = "NO-NAME-MEMBER"
```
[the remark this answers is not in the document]
retroflexer: Added panic.

pkg/etcdcli/etcdcli.go
```diff
@@ -274,6 +275,20 @@ func (g *etcdClientGetter) GetMember(name string) (*etcdserverpb.Member, error)
 	return nil, apierrors.NewNotFound(schema.GroupResource{Group: "etcd.operator.openshift.io", Resource: "etcdmembers"}, name)
 }
 
+// If the member's name is not set, extract ip/hostname from peerURL. Useful with unstarted members.
+func (g *etcdClientGetter) GetMemberNameOrHost(member *etcdserverpb.Member) string {
+	memberName := member.Name
+	if len(memberName) == 0 {
+		u, err := url.Parse(member.PeerURLs[0])
+		if err == nil {
```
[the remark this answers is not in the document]
retroflexer: Switched to follow the suggested style.

pkg/etcdcli/etcdcli.go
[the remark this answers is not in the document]
retroflexer: Yes, it is a utility function. Made it a public to invoke from pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go

pkg/etcdcli/etcdcli.go
```diff
@@ -274,6 +275,20 @@ func (g *etcdClientGetter) GetMember(name string) (*etcdserverpb.Member, error)
 	return nil, apierrors.NewNotFound(schema.GroupResource{Group: "etcd.operator.openshift.io", Resource: "etcdmembers"}, name)
 }
 
+// If the member's name is not set, extract ip/hostname from peerURL. Useful with unstarted members.
+func GetMemberNameOrHost(member *etcdserverpb.Member) string {
+	memberName := member.Name
+	if len(memberName) == 0 {
+		u, err := url.Parse(member.PeerURLs[0])
+		if err != nil {
+			panic(fmt.Sprintf("Unstarted member has invalid peerURL: %v", err))
```
hexfusion, line 284: errors in go don't generally start with capitals can you adjust please? Did you validate this justifies panic I commented it would be very bad but also a panic is very extreme? Just making sure your confidence in the situation.

pkg/etcdcli/etcdcli.go
```diff
@@ -274,6 +275,20 @@ func (g *etcdClientGetter) GetMember(name string) (*etcdserverpb.Member, error)
...
+		} else {
+			memberName = fmt.Sprintf("NAME-PENDING.%s", u.Hostname())
```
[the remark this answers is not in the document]
retroflexer: memberName is a local variable. It originally has the `member.Name`.

https://github.com/openshift/cluster-etcd-operator/blob/eca6a3c43b36fb7ea620aeb17545124e43c66ec5/pkg/etcdcli/etcdcli.go#L280

pkg/etcdcli/etcdcli.go
retroflexer: I think it would be pretty unusual to have peerURL to be empty. But panic would be extreme. We could print an error, and return a string `NO-NAME-BAD-PEER-URL` instead.

retroflexer: Pushed a new commit removing the panic.

pkg/etcdcli/etcdcli.go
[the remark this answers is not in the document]
retroflexer: Got rid of the local variable in the new commit.

pkg/etcdcli/etcdcli.go
```diff
@@ -274,6 +275,19 @@ func (g *etcdClientGetter) GetMember(name string) (*etcdserverpb.Member, error)
 	return nil, apierrors.NewNotFound(schema.GroupResource{Group: "etcd.operator.openshift.io", Resource: "etcdmembers"}, name)
 }
 
+// If the member's name is not set, extract ip/hostname from peerURL. Useful with unstarted members.
+func GetMemberNameOrHost(member *etcdserverpb.Member) string {
+	if len(member.Name) == 0 {
+		u, err := url.Parse(member.PeerURLs[0])
+		if err != nil {
+			klog.Errorf("unstarted member has invalid peerURL: %#v", err)
+			return "NAME-PENDING-BAD-PEER-URL"
```
hexfusion, line 284: If we have a getter that can have an err it should be returned. I mean the whole idea of an error in this context is to exit and retry. I know inlining is nice but you are not covering your failure.

member, err  := GetMemberNameOrHost()

retroflexer: Since we are not the original composers of the member list (it came out of a client call), I think exiting is not required. We can display error and then on the next iteration, we get a new member list and hopefully the problem is fixed. Exiting the operator doesn't fix a problem with etcd (if it is indeed a problem with etcd's member list).

hexfusion: making the operator resync on a parse error is probably heavy-handed given the context of the error. I guess what bothers me is I can't think of a case where this parse error would happen in a sane way. This looks better in general.

PR #324, opened 2020-04-29
Bug 1829737: operator: migrate controllers to factory

mfojtik:
This migrate controllers in etcd operator to library-go controller factory.


mfojtik:
@hexfusion @alaypatel07 tests passed :)

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -68,81 +61,26 @@ func NewEtcdCertSignerController(
...
-
-	return true
+		secretClient:         v1helpers.CachedSecretGetter(kubeClient.CoreV1(), kubeInformers),
+	}
+	return factory.New().ResyncEvery(time.Minute).WithInformers(
+		kubeInformers.InformersFor("").Core().V1().Nodes().Informer(),
+		kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Secrets().Informer(),
+		kubeInformers.InformersFor(operatorclient.GlobalUserSpecifiedConfigNamespace).Core().V1().Secrets().Informer(),
+		infrastructureInformer.Informer(),
+		operatorClient.Informer(),
```
hexfusion, line 78: we added  `operatorClient.Informer()` just making sure this was intended.

pkg/operator/etcdmemberipmigrator/etcdmemberipmigrator.go
```diff
@@ -52,33 +44,26 @@ func NewEtcdMemberIPMigrator(
...
 	}
-	kubeInformers.Core().V1().Nodes().Informer().AddEventHandler(c.eventHandler())
-	networkInformer.Informer().AddEventHandler(c.eventHandler())
-	operatorClient.Informer().AddEventHandler(c.eventHandler())
 
-	return c
+	return factory.New().ResyncEvery(time.Minute).WithInformers(
+		operatorClient.Informer(),
+		kubeInformers.Core().V1().Nodes().Informer(),
+		infrastructureInformer.Informer(),
```
hexfusion, line 59: added infrastructure informer I guess we should of had that before.

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -1,66 +1,47 @@
...
+	return factory.New().ResyncEvery(time.Minute).WithInformers(
+		operatorClient.Informer(),
+		kubeInformersForNamespaces.InformersFor("kube-system").Core().V1().ConfigMaps().Informer(),
+	).WithSync(c.sync).ToController("BootstrapTeardownController", eventRecorder.WithComponentSuffix("bootstrap-teardown-controller"))
 }
 
-func (c *BootstrapTeardownController) sync() error {
-	err := c.removeBootstrap()
+func (c *BootstrapTeardownController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	err := c.removeBootstrap(syncCtx)
```
alaypatel07, line 44: Out of curiosity, why not use the .WithSyncDegraded() function of the builder, remove all the sync methods and just call the underlying action function like `removeBootstrap`? If will be more work, can be a follow-up PR..may be should add a TODO? In general IMHO it will be less code maintenance for the project

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -51,79 +43,24 @@ func NewClusterMemberController(
...
-		c.queue.Add(workQueueKey)
-	}, time.Minute, stopCh)
-
-	<-stopCh
-}
-
-func (c *ClusterMemberController) runWorker() {
-	for c.processNextWorkItem() {
 	}
+	return factory.New().ResyncEvery(time.Minute).WithInformers(
```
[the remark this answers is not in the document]
mfojtik: yeah, it is not required, but I prefer to have a trigger in case the informers are stuck.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
mfojtik: it was weird that we waited for a cache sync in this informer but we have not added event handler to it... i think we want to react when the config CR change?

pkg/operator/etcdmemberipmigrator/etcdmemberipmigrator.go
mfojtik: yeah, I think that was a bug :-)

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
mfojtik: i would take that as follow up... changing the actual logic was not the intent of this PR.

PR #325, opened 2020-04-30
Change component name in OWNERS to 'Etcd Operator'

mfojtik:


mfojtik:
@hexfusion FYI, the component name should match the repo :-)

hexfusion:
@mfojtik Ecd and Etcd operator are mixed so often I think we should just honestly have Etcd but this is fine.

PR #333, opened 2020-05-05
Fix premature context cancellation

ironcladlou:
Before this patch, a call to clientv3.Client.Status was followed immediately
by an inline call to the returned cancel() function. Although in this case
the call is benign, it's still probably better to defer for consistency.


hexfusion:
@ironcladlou need a bug

ironcladlou:
I don't think this is worth a bug as I don't know of a specific problem it causes yet. Maybe better to just wait for the 4.5 branch and let this come into master when we're open for 4.6. WDYT?

PR #334, opened 2020-05-06
WIP: Cache etcd cluster state and abstract etcd client access

ironcladlou:
The etcd cluster state is more often read than updated. State reads are chatty.

This prototype changes the controllers to access etcd cluster state through a
new cluster cache interface instead of accessing etcd directly over the the API.

The etcd cluster state is cached and updated on a fixed interval or after a
mutating operation.

The etcd client itself is more clearly demarcated and now consumed only by the
cache implementation.

Member status computations are internalized in the cache because previously it
was encapsulated in the client. It might make sense to move that out of the
cache and client and into a controller, but I didn't want to make this even more
invasive.

There are probably lots of ways to solve this or ways to improve upon what I've
started here. Locking is pretty coarse and used freely for simplicity because
I'm assuming internal contention is not that important right now. There's no
rate limiting for cache refresh because I'm assuming they don't happen that
often in the steady state.

The etcd api also provides a watch interface that could be of further use. Maybe
using k8s resource types internally ("synthesizing" k8s cluster state in memory)
would make it easier to leverage existing machinery (e.g. the cache, informer,
queue inferfaces).


ironcladlou:
@hexfusion @retroflexer @deads2k this has gone far enough I'd like some feedback on the approach before I bother spending any more time with it

pkg/etcdcli/interfaces.go
```diff
@@ -20,6 +26,8 @@ type EtcdClient interface {
 
 	GetMember(name string) (*etcdserverpb.Member, error)
 	MemberUpdatePeerURL(id uint64, peerURL []string) error
+
+	Refresh() error
```
ironcladlou, line 30: feels like an abstraction leak, maybe the impl should take care of this all by itself during construction or maybe the interface method could be `Init` or something which should be called before use?

ironcladlou:
Next idea is to remove caching entirely until proven necessary

PR #290, opened 2020-04-01
Bug 1821286: fix message reporting for etcdmemberscontroller

alaypatel07:
This PR:

1. creates a standard error message format for reporting
 etcd members
2. adds specific events for unstarted members and 
unhealthy members

Since reporting available members would be at a level, 
the choice is made not report events of available members.

PR #336, opened 2020-05-06
Bug 1823608: fix message reporting for etcdmemberscontroller

retroflexer:
Cherry picking from PRs https://github.com/openshift/cluster-etcd-operator/pull/290 and https://github.com/openshift/cluster-etcd-operator/pull/318.

### DIFF
--- a/pkg/etcdcli/etcdcli.go
+++ b/pkg/etcdcli/etcdcli.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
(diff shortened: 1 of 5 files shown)


retroflexer:
I added release 4.4 because it makes it clear to us when we are looking at the PR list, and I thought it doesn't hurt the process (although it is not mandatory anymore).

hexfusion:
I was told that we don't want the branch because of the release notes getting gummed up. cc @wking

PR #338, opened 2020-05-07
Bug 1832923: Forget the bootstrap etcd member IP after bootstrap is complete

ironcladlou:
Before this patch, after bootstrap completes, the `host-etcd-2` endpoint still
contains an annotation pointing to the bootstrap etcd member IP, causing
operator etcd clients to permanently make futile connection attempts to the
defunct member endpoint.

This patch updates the bootstrap teardown controller to clean up the endpoint by
removing the annotation, which should eliminate the continuous failing
connection attempts to the old bootstrap member.

### DIFF
--- a/pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
+++ b/pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
@@ -2,24 +2,33 @@ package bootstrapteardown
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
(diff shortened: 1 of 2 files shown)


ironcladlou:
@deads2k I accidentally duplicated the work of https://github.com/openshift/cluster-etcd-operator/pull/237 but using a different approach — why did you choose the host endpoints controller vs. the bootstrap teardown controller?

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -2,36 +2,48 @@ package bootstrapteardown
...
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+
```
deads2k, line 16: nit: one giant block.  We've found no benefit to being separate.

deads2k:
> @deads2k I accidentally duplicated the work of #237 but using a different approach — why did you choose the host endpoints controller vs. the bootstrap teardown controller?

Single writer.  We have multi-reader and are very willing to duplicate reads, but we try hard for single writer resources and fields to avoid conflicts.

ironcladlou:
/hold

I like @deads2k suggestion to enforce a single-writer pattern. I'll see if I can revive https://github.com/openshift/cluster-etcd-operator/pull/237.

ironcladlou:
Replaced by https://github.com/openshift/cluster-etcd-operator/pull/342

PR #340, opened 2020-05-07
Bug 1830510: pkg/operator/etcdcertsigner: sign peer DNS SAN with explicit FQDN

hexfusion:
Prior to 4.4 etcd was bootstrap using SRV discovery. The result was each peer used FQDN in peer <--> peer communications. In 4.4 we regressed on upgrade support due to a typo which only signed DNS SAN for `etcdDiscoveryDomain`  vs `"*."+etcdDiscoveryDomain`. Although we migrate peers to use IP addresses there is a short window where in-between upgrade to 4.4 peer TLS auth can fail.

```golang
peerHostNames := append([]string{"localhost", etcdDiscoveryDomain}, nodeInternalIPs...)
```

This PR attempts to be more mindful of a proper SAN by using data provided in the MemberList to populate the peer FQDN explicitly instead of a wildcard. In the case of failure fallback to wildcard.

## Note
This bug only affects 4.4 as we do not want or require this functionality after we upgrade for 4.3 because we drop the DNS dependency completely.

### DIFF
--- a/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
+++ b/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
@@ -6,6 +6,8 @@ import (
 	"crypto/x509/pkix"
 	"errors"
 	"fmt"
+	"net"
+	"net/url"
 	"strings"
 	"time"
 
@@ -13,11 +15,13 @@ import (
 	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/cluster-etcd-operator/pkg/dnshelpers"
+	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
 	"github.com/openshift/library-go/pkg/crypto"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	"go.etcd.io/etcd/etcdserver/etcdserverpb"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -44,6 +48,7 @@ const (
 type EtcdCertSignerController struct {
 	kubeClient           kubernetes.Interface
 	operatorClient       v1helpers.OperatorClient
+	etcdClient           etcdcli.EtcdClient
 	infrastructureLister configv1listers.InfrastructureLister
 	nodeLister           corev1listers.NodeLister
 	secretLister         corev1listers.SecretLister
@@ -64,6 +69,7 @@ type EtcdCertSignerController struct {
 func NewEtcdCertSignerController(
 	kubeClient kubernetes.Interface,
 	operatorClient v1helpers.OperatorClient,
+	etcdClient etcdcli.EtcdClient,
 
 	kubeInformers v1helpers.KubeInformersForNamespaces,
 	infrastructureInformer configv1informers.InfrastructureInformer,
@@ -72,6 +78,7 @@ func NewEtcdCertSignerController(
 	c := &EtcdCertSignerController{
 		kubeClient:           kubeClient,
 		operatorClient:       operatorClient,
+		etcdClient:           etcdClient,
 		infrastructureLister: infrastructureInformer.Lister(),
 		secretLister:         kubeInformers.SecretLister(),
 		nodeLister:           kubeInformers.InformersFor("").Core().V1().Nodes().Lister(),
@@ -296,7 +303,10 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error 
 	if err != nil {
 		return err
 	}
-	peerHostNames := append([]string{"localhost", etcdDiscoveryDomain}, nodeInternalIPs...)
+	// 4.4 defaults
+	peerHostNames := append([]string{"localhost"}, nodeInternalIPs...)
+	// 4.3 compatability
+	peerHostNames = append(peerHostNames, c.Get43CompatibilityNames(node)...)
 	serverHostNames := append([]string{
 		"localhost",
 		"etcd.kube-system.svc",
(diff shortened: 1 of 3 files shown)


pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -331,6 +343,53 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
...
+						}
+					}
+				}
+			}
+		}
+	}
+
+	switch {
+	case len(errs) > 0:
+		break // fallback to wildcard
```
retroflexer, line 383: Is it possible to have no errors but the len(hosts) is zero?

retroflexer: Never mind. Then, also it will fall back to wild-card.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -331,6 +343,53 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
...
+	hosts := []string{"localhost"}
+	for _, m := range etcdMembers {
+		// 4.3 naming
+		if strings.HasPrefix(m.Name, "etcd-member") {
+			found = true
+		}
+		if len(m.PeerURLs) > 0 && len(m.ClientURLs) > 0 {
+			for _, p := range m.PeerURLs {
+				peer, err := url.Parse(p)
+				if err != nil {
```
retroflexer, line 358: If we are adding the clinet hostname from the clientURLs, do we have to check this peerURLs too?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -331,6 +343,53 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
...
+					errs = append(errs, err.Error())
+					continue
+				}
+				if addr := net.ParseIP(peer.Hostname()); addr != nil {
+					continue
+				}
+				for _, clientUrl := range m.ClientURLs {
```
retroflexer, line 365: I don't understand this 3rd level nesting of for loop. Why do we have to loop through all clientURLs for every peer URL?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: Not sure I am following the question. Here we parse the url for peers. If this is an IP we continue. Basically, if the PeerURL is already IP then we don't need to do any more work.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: yeah and it won't ever be 0 

https://github.com/openshift/cluster-etcd-operator/pull/340/files/c1734e4befc41128038c407ac72d611a8f309347#diff-97a182619797a0137f8ca3c8b467bae4R349

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
retroflexer: OK. Got it.

retroflexer: OK. Thanks. Your test examples make it clear.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: well it might not be completely necessary I was covering the fact that it is a list. I guess it is fair to consider a manual adjustment to client-urls a fail and only attempt top match [0]? We could fall back to wildcard in the case no match for [0].

https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/configuration.md#--advertise-client-urls

retroflexer: OK. In general, we typically have only one peer URL and one client URL per member. So it is not a big deal. Even otherwise, we are looping through 3 members: the loops are not big (not in thousands). Looks good to me.

hexfusion: yeah simplified the logic and only check [0] for each. This is best-effort if the config is not default then we fallback.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -331,6 +343,49 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
...
+		}
+	}
+
+	switch {
+	case len(errs) > 0:
+		break // fallback to wildcard
+	case len(hosts) == 1 && found:
+		break // fallback to wildcard
+	case len(hosts) > 1 || !found:
+		return append(hosts, nodeInternalIPs...)
```
deads2k, line 383: just add all the things all the time, so you don't have branches. Then in 4.5, just stop adding the 4.3 compatibility ones.

sdodson:
Because this change is only relevant to the transition from 4.3 to 4.4 and introducing additional signing in 4.5 would introduce additional confusion this will go directly into 4.4 and will not be backported from master branch. Overriding bugzilla/valid-bug flags.

deads2k:
> Because this change is only relevant to the transition from 4.3 to 4.4 and additional signing in 4.5 would introduce additional confusion this will go directly into 4.4 and will not be backported from master branch. Overriding bugzilla/valid-bug flags.

The net is that if we merged into 4.5, we'd immediately revert it in 4.5 so that we don't extend our signing beyond our intent what's necessary because broader signing is bad practice and makes debugging intent of communication  harder.

/hold 

holding in case @hexfusion wants to take the appending recommendation I offered.  He can remove at his discretion.

hexfusion:
making updates based on @deads2k comments

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -296,7 +303,8 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
 	if err != nil {
 		return err
 	}
-	peerHostNames := append([]string{"localhost", etcdDiscoveryDomain}, nodeInternalIPs...)
```
deads2k, line 299: what happened to etcdDiscoveryDomain?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -331,6 +339,64 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
...
+		errs = append(errs, err.Error())
+	}
+	peerHostNames, err := getPeerDNSSubjectAlternativeNamesFromMemberList(etcdDiscoveryDomain, etcdMembers, nodeInternalIPs)
+	if err != nil {
+		errs = append(errs, err.Error())
+	}
+
+	// fallback to wildcard
+	if len(errs) > 0 {
+		return []string{"*." + etcdDiscoveryDomain}
```
deads2k, line 363: is this for 4.3 or 4.4?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -331,6 +339,64 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
...
+	}
+	return peerHostNames
```
deads2k, line 365: is this for 4.3 or 4.4?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -331,6 +339,64 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
 	return nil
 }
 
+func (c *EtcdCertSignerController) Get43CompatibilityNames(node *corev1.Node) []string {
```
deads2k, line 342: we would remove this entire method in 4.5?

deads2k:
got Sam on slack, we remove the new function and we're done for 4.5

/lgtm

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: it was invalid and a typo

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: I added more documentation to explain

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: 4.3 only

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: if we observe no 4.3 naming we return nil otherwise we return explicit DNS records for 4.3 SAN

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -331,6 +341,81 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
...
+		break // fallback to wildcard 4.3 compatability
+	case len(hosts) == 0 && found:
+		break // fallback to wildcard 4.3 compatability
+	case len(hosts) > 0:
+		return hosts, nil // 4.3 peer DNS found populate san with explicit records
+	case !found:
+		return nil, nil // 4.4
+	}
+
+	return hosts, fmt.Errorf("no peer SAN hostnames found in MemberList")
```
deads2k, line 416: it's weird to return a value and an error. and the value returned here is ignored.

hexfusion: yeah.... fixed

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: but yes

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -296,7 +303,10 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node) error
 	if err != nil {
 		return err
 	}
-	peerHostNames := append([]string{"localhost", etcdDiscoveryDomain}, nodeInternalIPs...)
+	// 4.4 defaults
+	peerHostNames := append([]string{"localhost"}, nodeInternalIPs...)
+	// 4.3 compatability
+	peerHostNames = append(peerHostNames, c.Get43CompatibilityNames(node)...)
```
hexfusion, line 309: remove this line in 4.5

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: what we wanted was a wildcard. In 4.4+ we no longer need DNS in SAN as etcd only uses IP. So new defaults are localhost to allow quorum-guard to function and  nodeInternalIPs

hexfusion: we then append compatability records for 4.3

hexfusion:
checking upgrade error
/test e2e-gcp-upgrade

PR #343, opened 2020-05-11
Bug 1832986: pkg/etcdcli: fix health checks

hexfusion:
After research, I found that the reason for failed health checks was actually related to deadlock[1] that would occur in the operator. The result is that we would see the client timeout essentially waiting for the lock to release on the shared client. This resulted in those failures be misinterpreted as client failures against etcd (failed health check).

The resolution involved using a proper linearized health check for etcd[2].  The second was to move away from the shared client for health checks as the memory profile was not really affected. The result is accurate health reporting for etcd.

With the move to non shared clients for health checks, we still saw a net improvement with RSS and CPU consumption for the operator after refactor.

[1] https://github.com/openshift/cluster-etcd-operator/pull/345

[2] https://github.com/openshift/cluster-etcd-operator/pull/343/files#diff-441a0e0f3a16f195afc90c4db047116cR353


pkg/etcdcli/etcdcli.go
```diff
@@ -329,9 +317,63 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
...
+		if len(member.ClientURLs) == 0 {
+			memberHealth.Check = append(memberHealth.Check, healthCheck{Member: member, Health: false, Started: false})
+			continue
+		}
+		wg.Add(1)
+		go func(member *etcdserverpb.Member) {
+			defer wg.Done()
+			cli, err := getEtcdClient([]string{member.ClientURLs[0]})
+			if err != nil {
+				hch <- healthCheck{Member: member, Health: false, Started: true, Error: err.Error()}
```
ironcladlou, line 347: Wrap the error with `%w`?

pkg/etcdcli/etcdcli.go
```diff
@@ -329,9 +317,63 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
...
 
+	_, unhealthyMembers := etcdMemberHealth.MemberStatus()
+
 	return unhealthyMembers, nil
 }
 
+func (g *etcdClientGetter) MemberHealth(etcdMembers []*etcdserverpb.Member) (*memberHealth, error) {
+	memberHealth, err := g.getMemberHealth(etcdMembers)
+	if err != nil {
+		return nil, err
```
ironcladlou, line 328: Wrap the error with `%w`?

pkg/etcdcli/etcdcli.go
```diff
@@ -329,9 +317,63 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
...
+			cli, err := getEtcdClient([]string{member.ClientURLs[0]})
+			if err != nil {
+				hch <- healthCheck{Member: member, Health: false, Started: true, Error: err.Error()}
+				return
+			}
+			defer cli.Close()
+			st := time.Now()
+			ctx, cancel := context.WithCancel(context.Background())
+			_, err = cli.Get(ctx, "health")
+			cancel()
```
ironcladlou, line 354: (Aside: can somebody explain this context pattern to me? I've seen it throughout and wonder how, as used, it's different than `context.TODO()` as it doesn't seem like we're actually handling cancellation asynchronously or through passthrough of an outer context)

pkg/etcdcli/etcdcli.go
```diff
@@ -329,9 +317,63 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	return memberHealth, nil
+}
+
+func (g *etcdClientGetter) getMemberHealth(etcdMembers []*etcdserverpb.Member) (*memberHealth, error) {
+	var wg sync.WaitGroup
+	memberHealth := &memberHealth{}
+	hch := make(chan healthCheck, len(etcdMembers))
```
ironcladlou, line 336: (Aside: looks like if there were a thread-safe array construct that would do the same thing but in a more obvious way? Go doesn't give us a lot to work with here... sync.Map would be okay but annoying since we just want an array...)

pkg/etcdcli/health.go
```diff
@@ -0,0 +1,50 @@
...
+	Member  *etcdserverpb.Member
+	Health  bool
+	Started bool
+	Took    string
+	Error   string
+}
+
+type memberHealth struct {
+	Check []healthCheck
+}
```
ironcladlou, line 17: Another way to represent this might be...

```go
type memberHealth []healthCheck

func (m memberHealth) Healthy() []*etcdserverpb.Member {
  // filter m by healthy
}
func (m memberHealth) Unhealthy() []*etcdserverpb.Member {
  // filter m by unhealthy
}
// ... etc
```

One possible advantage is clearer caller semantics compared to multiple return values

pkg/etcdcli/etcdcli.go
```diff
@@ -329,9 +317,63 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
...
+			defer wg.Done()
+			cli, err := getEtcdClient([]string{member.ClientURLs[0]})
+			if err != nil {
+				hch <- healthCheck{Member: member, Health: false, Started: true, Error: err.Error()}
+				return
+			}
+			defer cli.Close()
+			st := time.Now()
+			ctx, cancel := context.WithCancel(context.Background())
+			_, err = cli.Get(ctx, "health")
```
ironcladlou, line 353: Maybe worth a comment about linearization here?

pkg/etcdcli/health.go
hexfusion: yep I like that

pkg/etcdcli/etcdcli.go
```diff
@@ -329,9 +317,63 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
...
+	hch := make(chan healthCheck, len(etcdMembers))
+	for _, member := range etcdMembers {
+		if len(member.ClientURLs) == 0 {
+			memberHealth.Check = append(memberHealth.Check, healthCheck{Member: member, Health: false, Started: false})
+			continue
+		}
+		wg.Add(1)
+		go func(member *etcdserverpb.Member) {
+			defer wg.Done()
+			cli, err := getEtcdClient([]string{member.ClientURLs[0]})
```
ironcladlou, line 345: Might be worth a comment here that we're intentionally making a new client per request for simplicity/bug avoidance and that the cost has been measured and deemed insignificant.

In fact, I would extend the line of reasoning to the use of cached clients/connections generally here given the complexity already present. I don't have any numbers to justify the use of a cached client here or anywhere else, but I also don't want to suggest the caching was introduced for no reason. Maybe with other problems addressed (e.g. less noisy steady state) the problem changes.

hexfusion: I added a comment to this regard and we can continue the conversation.

pkg/etcdcli/etcdcli.go
hexfusion: added

pkg/etcdcli/etcdcli.go
hexfusion: making a true error type makes sense

pkg/etcdcli/etcdcli.go
```diff
@@ -297,39 +298,74 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
...
+			}
+			defer cli.Close()
+			st := time.Now()
+			ctx, cancel := context.WithCancel(context.Background())
+			// linearized request to verify health of member
+			_, err = cli.Get(ctx, "health")
+			cancel()
+			hc := healthCheck{Member: member, Health: false, Started: true, Took: time.Since(st).String()}
+			if err == nil || err == rpctypes.ErrPermissionDenied {
+				hc.Health = true
```
tedyu, line 351: Health is a noun.
It seems the field should be named Healthy

pkg/etcdcli/etcdcli.go
```diff
@@ -297,39 +298,74 @@ func (g *etcdClientGetter) UnhealthyMembers() ([]*etcdserverpb.Member, error) {
...
+			} else {
+				hc.Error = fmt.Errorf("health check failed: %w", err)
+			}
+			hch <- hc
+		}(member)
+	}
+
+	go func() {
+		wg.Wait()
```
tedyu, line 360: I wonder why Wait() is used in a goroutine.
Only after wg.Wait() returns, would we be able to read all healthCheck's from the channel.

pkg/etcdcli/health.go
```diff
@@ -0,0 +1,104 @@
...
+			members = append(members, etcd.Member)
+		}
+	}
+	return members
+}
+
+// HealthyMemberNames returns a list of healthy member names
+func (h *memberHealth) GetHealthyMemberNames() []string {
+	memberNames := []string{}
+	for _, etcdMember := range h.GetHealthyMembers() {
```
tedyu, line 60: It seems we can use a loop similar to the one on line 49.
For this method, we don't need the etcdserverpb.Member's since only Name field is used.

pkg/etcdcli/health.go
```diff
@@ -0,0 +1,104 @@
...
+		memberNames = append(memberNames, etcdMember.Name)
+	}
+	return memberNames
+}
+
+// Unhealthy returns unhealthy members
+func (h *memberHealth) GetUnhealthyMembers() []*etcdserverpb.Member {
```
tedyu, line 67: It seems this method and GetHealthyMembers() can be combined into one which accepts a bool flag (healthy member or not).

pkg/etcdcli/health.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, etcd := range h.Check {
+		if !etcd.Health {
+			members = append(members, etcd.Member)
+		}
+	}
+	return members
+}
+
+// UnhealthyMemberNames returns a list of unhealthy member names
+func (h *memberHealth) GetUnhealthyMemberNames() []string {
```
tedyu, line 78: This method and GetHealthyMemberNames can be combined into one method.

pkg/etcdcli/etcdcli.go
ironcladlou: Agree, this seems like an error. Because the channel is buffered, I think the `range` can cause the function to return before all items are read because the channel read will not block if the channel is empty (so the `range` consumer races with the producer goroutine).

I think the `Wait()`/`close()` pair should be synchronous.

pkg/etcdcli/health.go
ironcladlou: In this case I personally prefer explicit functions that do one thing than introducing modality through a boolean.

ironcladlou: Mainly because you can extend the combination argument to pretty much every function here (e.g. why not one function with a filter argument which could allow every filter permutation to be expressed).

pkg/etcdcli/etcdcli.go
hexfusion: Yeah, this was a typo, will address.

pkg/etcdcli/etcdcli.go
hexfusion: fair I was emulating the upstream object but you are correct.

hexfusion:
@tedyu , @ironcladlou thank you for the reviews I will address.

hexfusion:
/hold
adding additional test coverage.

pkg/etcdcli/health.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+	return members
+}
+
+// UnhealthyMemberNames returns a list of unhealthy member names
+func (h memberHealth) GetUnhealthyMemberNames() []string {
+	memberNames := []string{}
+	for _, etcdMember := range h.GetUnhealthyMembers() {
+		if len(etcdMember.ClientURLs) == 0 {
```
ironcladlou, line 127: nit: this is kind of weird, seems like it should be part of GetMemberNameOrHost or something. Seems smelly that callers have to make some decisions to compute the name of a Member. Would be nice to encapsulate the name logic in a single place (so callers always access a member name through said function) or something.

pkg/etcdcli/health.go
```diff
@@ -0,0 +1,154 @@
...
+	for _, etcd := range h {
+		if !etcd.Started {
+			members = append(members, etcd.Member)
+		}
+	}
+	return members
+}
+
+// UnstartedMemberNames returns a list of unstarted member names
+func (h memberHealth) GetUnstartedMemberNames() []string {
```
ironcladlou, line 148: I think all the *Name variant functions could be replaced by a function that generically accepts `[]*etcdserverpb.Member` and returns `[]string` of their names

ironcladlou:
@hexfusion I think you need to rebase to pick up https://github.com/openshift/cluster-etcd-operator/pull/351 so your clusters will stop killing themselves

pkg/etcdcli/etcdcli.go
ironcladlou: Actually I think I was wrong in my assessment — buffered channel reads will block when the channel is empty, so what you had would indeed work, although the async wait/close is still technically unnecessary

pkg/etcdcli/health.go
hexfusion: well we can't conclude health from only the member and I would like to only perform a  single series of health checks so names function will have the following signature.
```
 func GetHealthyMemberNames(memberHealth []healthCheck) []string {
```

hexfusion: but then I wonder why no keep them as a method?

hexfusion:
CI has thawed 

/retest

ironcladlou:
@hexfusion remove when ready

/hold

PR #347, opened 2020-05-12
go degraded with a clear message if the operator loses the race again…

deads2k:
…st bootkube

It also gives a chance for bootkube (or another actor) to put the situation right and recover.

### DIFF
--- a/pkg/operator/hostendpointscontroller2/host_endpoints_controller.go
+++ b/pkg/operator/hostendpointscontroller2/host_endpoints_controller.go
@@ -7,6 +7,15 @@ import (
 	"sort"
 	"time"
 
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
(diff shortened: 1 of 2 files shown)


pkg/operator/hostendpointscontroller2/host_endpoints_controller.go
```diff
@@ -117,6 +123,22 @@ func (c *HostEndpoints2Controller) syncHostEndpoints2(ctx context.Context, recor
...
 
+	// if the endpoint isn't present, add a check to see if we are still in a bootstrap phase (based on the etcd member).
+	// if we are, this cluster got itself in an illegal state where the endpoint is going to be created, but it won't have
+	// the bootstrap IP annotation.  Lack of this prevents successful bootstrapping and we want a better signal.  In general
+	// we want to be able to recreate this endpoint and we can do so *after* bootstrapping is complete, but not before.
+	if _, err := c.endpointsLister.Endpoints(operatorclient.TargetNamespace).Get("host-etcd-2"); errors.IsNotFound(err) {
+		members, err := c.etcdClient.MemberList()
+		if err != nil {
+			utilruntime.HandleError(err)
+		}
```
ironcladlou, line 134: Continuing past the point with a non-nil `err`

1. access a potentially unsafe return value from MemberList
2. fails to short-circuit when we haven't determined for sure whether bootstrap is complete

Seems safer to bail early on error from MemberList

deads2k: > Continuing past the point with a non-nil `err`
> 
> 1. access a potentially unsafe return value from MemberList
> 2. fails to short-circuit when we haven't determined for sure whether bootstrap is complete
> 
> Seems safer to bail early on error from MemberList

sure

hexfusion:
/hold 
lets make sure we want this now.

deads2k:
I think this got worked out by danmace differently

PR #350, opened 2020-05-14
WIP: debugging

ironcladlou:


pkg/operator/starter.go
```diff
@@ -258,9 +256,6 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	go envVarController.Run(1, ctx.Done())
 	go staticPodControllers.Start(ctx)
 
-	// TODO remove in 4.6
-	ensureServiceCleanup(ctx, kubeClient, controllerContext.EventRecorder)
```
ironcladlou, line 262: If we delete this, the service controller will delete the endpoints with it. Although the controller will re-create the endpoints, that's disruptive (and depending on when the deletion happens, maybe we're bootstrapping, and now we've lost endpoint information from annotations).

On the other hand, if we leave it in existing clusters, it should be benign and we can address removing it separately.

And on new clusters, of course we won't create it at all and don't have to worry about it.

ironcladlou: I think that it may be the case that by design kube will purge any endpoints which lack a corresponding service.

https://github.com/kubernetes/kubernetes/issues/6877
https://github.com/kubernetes/kubernetes/pull/7821
https://github.com/kubernetes/kubernetes/pull/7821#issuecomment-99344889

Unless something has changed. But the endpoints controller is definitely purging the endpoint.

```
{"kind":"Event","apiVersion":"audit.k8s.io/v1","level":"Metadata","auditID":"84a069cf-0755-4e38-8926-fe1a380d8909","stage":"ResponseComplete","requestURI":"/api/v1/namespaces/openshift-etcd/endpoints/host-etcd-2","verb":"delete","user":{"username":"system:serviceaccount:kube-system:endpoint-controller","uid":"f4261658-5859-42e6-8f20-0453474f058f","groups":["system:serviceaccounts","system:serviceaccounts:kube-system","system:authenticated"]},"sourceIPs":["::1"],"userAgent":"kube-controller-manager/v1.18.2 (linux/amd64) kubernetes/9b48758/system:serviceaccount:kube-system:endpoint-controller","objectRef":{"resource":"endpoints","namespace":"openshift-etcd","name":"host-etcd-2","apiVersion":"v1"},"responseStatus":{"metadata":{},"code":200},"requestReceivedTimestamp":"2020-05-14T21:26:34.777669Z","stageTimestamp":"2020-05-14T21:26:34.825739Z","annotations":{"authorization.k8s.io/decision":"allow","authorization.k8s.io/reason":"RBAC: allowed by ClusterRoleBinding \"system:controller:endpoint-controller\" of ClusterRole \"system:controller:endpoint-controller\" to ServiceAccount \"endpoint-controller/kube-system\""}}
```

So I'm not convinced we can get rid of the service at all. In fact, we might want to protect it and the endpoint with finalizers.

ironcladlou: cc @retroflexer @hexfusion @deads2k

ironcladlou: See https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/pkg/controller/endpoint/endpoints_controller.go#L372. Although like I said, I haven't reproduced it in an isolated fashion yet, if we're fighting a design invariant it may be beside the point.

ironcladlou: @openshift/sig-networking @smarterclayton can any of you folks verify whether we should expect endpoints without services to work as a matter of current upstream design intent? Seems like no based on https://github.com/openshift/cluster-etcd-operator/pull/350/files#r425503412 but would like to be sure.

danwinship: Currently "strays" are only deleted at kube-controller-manager startup, but it seems unsafe to rely on that

PR #351, opened 2020-05-15
Bug 1837103: Revert "remove dead host-etcd-2 service"

ironcladlou:
This reverts commit 7f142f69b1fe95ffb033526256d9498a6c8dadc5 (https://github.com/openshift/cluster-etcd-operator/pull/346).

https://github.com/kubernetes/kubernetes/issues/6877
https://github.com/kubernetes/kubernetes/pull/7821
https://github.com/kubernetes/kubernetes/pull/7821#issuecomment-99344889

Deleting the service subjects the host-etcd-2 endpoint to deletion by
the Kube endpoints controller at random times now or in the future.

The service must remain to protect the existence of the endpoint given
the current design of Kube.

By deleting the service, we risk bootstrap failure through information
loss when the endpoint is deleted during bootstrapping, and churning
unnecessary endpoint revisions by recreating an endpoint Kube wants
to delete.

cc @hexfusion @retroflexer @deads2k @smarterclayton please fact check (see discussion in https://github.com/openshift/cluster-etcd-operator/pull/350/files#r425419897)

### DIFF
--- a/bindata/bootkube/manifests/00_etcd-host-service.yaml
+++ b/bindata/bootkube/manifests/00_etcd-host-service.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: host-etcd-2
+  namespace: openshift-etcd
+  labels:
+    # this label is used to indicate that it should be scraped by prometheus
+    k8s-app: etcd
+spec:
+  clusterIP: None
+  ports:
+  - name: etcd
+    port: 2379
+    protocol: TCP
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -7,9 +7,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/klog"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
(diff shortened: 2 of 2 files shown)


deads2k:
> we risk bootstrap failure through information loss

You can't upgrade a cluster that hasn't completed bootstrapping.  Thus the information is not lost.

deads2k:
> churning
> unnecessary endpoint revisions by recreating an endpoint Kube wants
> to delete.

This may be true, but how long does it take to settle?  if it settles eventually (should settle pretty quickly), then I think we can tolerate jitter.  We just update the openshfit-apiserver-operator to have no toleration for empty.

ironcladlou:
If Kube says "all endpoints must have a service and we'll try to delete any orphan endpoints", why would we rely on the coincidence that the endpoint controller hasn't yet "caught us"?

Keeping the service is what Kube says we should do, and I'm not sure what problem the service's existence represents.

The problems with deleting it seem clear. What is the purpose of deleting the service?

deads2k:
> If Kube says "all endpoints must have a service and we'll try to delete any orphan endpoints", why would we rely on the coincidence that the endpoint controller hasn't yet "caught us"?
> 
> Keeping the service is what Kube says we should do, and I'm not sure what problem the service's existence represents.
> 
> The problems with deleting it seem clear. What is the purpose of deleting the service?

avoiding management of a thing we don't want to manage.  We're safe this release.  Leave the service out and transition to a configmap if you like, but don't bring the service back.  It brings in races and potential dual management of endpoints we don't need.

ironcladlou:
Agree this patch is desired, but right now it makes the problem far worse:

https://testgrid.k8s.io/redhat-openshift-ocp-release-4.5-blocking#release-openshift-ocp-installer-e2e-aws-4.5

https://search.apps.build01.ci.devcluster.openshift.com/chart?search=BootstrapTeardown_Error&maxAge=168h&context=1&type=junit&name=&maxMatches=5&maxBytes=20971520&groupBy=job

We must revert this pending a better fix. The periodics and CI are unstable enough without this compounding the problems.

hexfusion:
/lgtm
/approve
/hold

Holding to make sure we can't get the fix merged to prove clean signal in CI. Sorry to all affected but if this isnt fixed we have clusters that will randomly fail install among other issues.

PR #352, opened 2020-05-15
Bug 1835146: Cluster restore should not stop network pods on bare-metal

retroflexer:
Currently the cluster backup script backs up static pod resources for "kube-apiserver-pod.yaml" "kube-controller-manager-pod.yaml" "kube-scheduler-pod.yaml" and "etcd-pod.yaml". However, when cluster-restore script is run, it stops all the static pods (including network pods for the bare-metal case). 

However, after the restore operation we only restore the kube static pods and etcd, but do not restore network pods (as they are not backed up originally). This results in failed cluster restore unable to scale up fully.

This PR fixes the problem by only stopping the kube static pods, leaving the network pods to continue to run.

### DIFF
--- a/bindata/etcd/cluster-restore.sh
+++ b/bindata/etcd/cluster-restore.sh
@@ -75,11 +75,11 @@ if [ ! -d "$MANIFEST_STOPPED_DIR" ]; then
 fi
 
 # Move static pod manifests out of MANIFEST_DIR
-find ${MANIFEST_DIR} \
-  -maxdepth 1 \
-  -type f \
-  -printf '...stopping %P\n' \
-  -exec mv {} ${MANIFEST_STOPPED_DIR} \;
+for POD_FILE_NAME in "${STATIC_POD_LIST[@]}" etcd-pod.yaml; do
+  echo "...stopping ${POD_FILE_NAME}"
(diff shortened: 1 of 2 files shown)


bindata/etcd/cluster-restore.sh
```diff
@@ -75,11 +75,11 @@ if [ ! -d "$MANIFEST_STOPPED_DIR" ]; then
 fi
 
 # Move static pod manifests out of MANIFEST_DIR
-find ${MANIFEST_DIR} \
-  -maxdepth 1 \
-  -type f \
-  -printf '...stopping %P\n' \
-  -exec mv {} ${MANIFEST_STOPPED_DIR} \;
+for POD_FILE_NAME in "${STATIC_POD_LIST[@]}" etcd-pod.yaml; do
```
hexfusion, line 78: Do we need to rethink the general logic here? This feels like we are bolting the solution on a bit I want to make sure we are not glazing over anything else that should be addressed in this PR.

retroflexer: I can hold this PR, if you have a different solution. 

We definitely can't move everything out of the manifests directory and only replace 4 pods. So what we currently have definitely needs a fix.

PR #363, opened 2020-05-26
Bug 1812126: increase memory and resource limits for installer and pruner pods

vareti:
This PR picks up the memory resource and limits changes in https://github.com/openshift/library-go/pull/805


hexfusion:
manually added `bugzilla/valid` because BZ was private.

PR #367, opened 2020-06-01
Bug 1832923: Forget bootstrap etcd member IP after bootstrap

ironcladlou:
After bootstrap completes, forget the bootstrap etcd member IP by clearing
the annotation from the endpoint configmap. Otherwise the etcd client will
continue failing to connect to the dead bootstrap node forever.

### DIFF
--- a/pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
+++ b/pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
@@ -13,50 +13,50 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog"
 
+	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
 )
 
 // EtcdEndpointsController maintains a configmap resource with
 // IP addresses for etcd. It should never depend on DNS directly or transitively.
 type EtcdEndpointsController struct {
-	operatorClient  v1helpers.OperatorClient
+	operatorClient  v1helpers.StaticPodOperatorClient
 	nodeLister      corev1listers.NodeLister
 	configmapLister corev1listers.ConfigMapLister
 	configmapClient corev1client.ConfigMapsGetter
 }
 
 func NewEtcdEndpointsController(
-	operatorClient v1helpers.OperatorClient,
+	operatorClient v1helpers.StaticPodOperatorClient,
 	eventRecorder events.Recorder,
 	kubeClient kubernetes.Interface,
 	kubeInformers operatorv1helpers.KubeInformersForNamespaces,
 ) factory.Controller {
-	kubeInformersForTargetNamespace := kubeInformers.InformersFor(operatorclient.TargetNamespace)
-	configmapsInformer := kubeInformersForTargetNamespace.Core().V1().ConfigMaps()
-	kubeInformersForCluster := kubeInformers.InformersFor("")
-	nodeInformer := kubeInformersForCluster.Core().V1().Nodes()
+	nodeInformer := kubeInformers.InformersFor("").Core().V1().Nodes()
 
 	c := &EtcdEndpointsController{
 		operatorClient:  operatorClient,
 		nodeLister:      nodeInformer.Lister(),
-		configmapLister: configmapsInformer.Lister(),
+		configmapLister: kubeInformers.ConfigMapLister(),
 		configmapClient: kubeClient.CoreV1(),
 	}
 	return factory.New().ResyncEvery(time.Minute).WithInformers(
 		operatorClient.Informer(),
-		configmapsInformer.Informer(),
+		kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer(),
 		nodeInformer.Informer(),
 	).WithSync(c.sync).ToController("EtcdEndpointsController", eventRecorder.WithComponentSuffix("etcd-endpoints-controller"))
 }
 
 func (c *EtcdEndpointsController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
-	err := c.syncConfigMap(ctx, syncCtx.Recorder())
+	err := c.syncConfigMap(syncCtx.Recorder())
 
 	if err != nil {
 		_, _, updateErr := v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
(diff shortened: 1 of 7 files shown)


ironcladlou:
Latest failures are I think unrelated, and I have a successful [4.4 to 4.5 upgrade job](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/12331). I want to do a little refactoring and add test coverage for this controller.

ironcladlou:
@hexfusion @retroflexer @deads2k refactored, added a bunch of tests, re-running the Y upgrade test. Tomorrow I might go ahead and continue expanding the unit test while I'm in here to cover node=>configmap translation scenarios.

ironcladlou:
There's a problem here. In a real cluster the annotation is never cleaned up because the revision checks falsely detect a rollout because nodestates.targetrevision is empty. So, the unit tests are inaccurate. I need clarification on the semantics around these fields:

https://github.com/openshift/api/blob/master/operator/v1/types.go#L208

1. What does an empty TargetRevision mean?
2. What does an empty LatestAvailableRevision mean?

ironcladlou:
Now I'm relying on latestAvailableRevision instead of nodestatus.targetRevision.

ironcladlou:
https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/367/pull-ci-openshift-cluster-etcd-operator-master-e2e-gcp/1530

Unrelated flakes; annotation was removed and etcd related things are reporting available.

ironcladlou:
AWS failures look unrelated:

```
[sig-network-edge][Conformance][Area:Networking][Feature:Router] The HAProxy router should pass the gRPC interoperability tests [Suite:openshift/conformance/parallel/minimal]
[sig-network][Feature:Router] The HAProxy router should expose prometheus metrics for a route [Suite:openshift/conformance/parallel]

Failing tests:

[sig-network] Services should be rejected when no endpoints exist [Skipped:ibmcloud] [Suite:openshift/conformance/parallel] [Suite:k8s]
```

https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/367/pull-ci-openshift-cluster-etcd-operator-master-e2e-azure/1604

Dug into the artifacts and nothing seems amiss with etcd or the apiserver.

/retest

ironcladlou:
@hexfusion @retroflexer this is ready for review

ironcladlou:
Successful 4.4 -> 4.5 run w/ the latest commit: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/12426

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
```diff
@@ -13,50 +13,50 @@ import (
...
 		operatorClient:  operatorClient,
 		nodeLister:      nodeInformer.Lister(),
-		configmapLister: configmapsInformer.Lister(),
+		configmapLister: kubeInformers.ConfigMapLister(),
 		configmapClient: kubeClient.CoreV1(),
 	}
 	return factory.New().ResyncEvery(time.Minute).WithInformers(
 		operatorClient.Informer(),
-		configmapsInformer.Informer(),
+		kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer(),
```
hexfusion, line 53: Curious what is the difference between the former and the latter?

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
```diff
@@ -83,9 +83,30 @@ func (c *EtcdEndpointsController) sync(ctx context.Context, syncCtx factory.Sync
...
+
 	required := configMapAsset()
 
+	// If the bootstrap IP is present on the existing configmap, either copy it
+	// forward or remove it if possible so clients can forget about it.
+	if existing, err := c.configmapLister.ConfigMaps(operatorclient.TargetNamespace).Get("etcd-endpoints"); err == nil {
+		if existingIP, hasExistingIP := existing.Annotations[etcdcli.BootstrapIPAnnotationKey]; hasExistingIP {
+			if bootstrapComplete {
+				// remove the annotation
+				recorder.Eventf("EtcdEndpointsRemovedBootstrapIPAnnotation", "removed the bootstrap IP %q from configmap %s/%s", existingIP, existing.Namespace, existing.Name)
```
hexfusion, line 100: Just a general comment on event workflow it seems wrong that we event something as valid before it succeeds. For example if apply failed below event review would appear to show a success.

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
ironcladlou: functionally equivalent, both cause sync to trigger on configmap change in the openshift-etcd namespace

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
ironcladlou: Good catch, agree

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
hexfusion: OK so why change :)

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
ironcladlou: I went ahead and removed the event entirely, I'm not sure it's really useful anymore

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
ironcladlou: Because configmapsInformer was only used in one place (so what's the point of a local to begin with), and also had a misleading name (it isn't an informer for all configmaps, it's only for the openshift-etcd namespace).

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
```diff
@@ -83,9 +83,29 @@ func (c *EtcdEndpointsController) sync(ctx context.Context, syncCtx factory.Sync
...
+	bootstrapComplete, err := isBootstrapComplete(c.configmapLister, c.operatorClient)
+	if err != nil {
+		return fmt.Errorf("couldn't determine bootstrap status: %w", err)
+	}
+
 	required := configMapAsset()
 
+	// If the bootstrap IP is present on the existing configmap, either copy it
+	// forward or remove it if possible so clients can forget about it.
+	if existing, err := c.configmapLister.ConfigMaps(operatorclient.TargetNamespace).Get("etcd-endpoints"); err == nil {
```
hexfusion, line 96: do we care if the `errors.IsNotFound(err)`? looks like we check
```
err == nil
 !errors.IsNotFound(err)
```
But what if the cm does not exist shouldn't we report that error? Should this be a switch?

ironcladlou: The CM doesn't exist prior to 4.5. If it's not found, we must create it.

In a new 4.5+ cluster, the CM is created during bootstrap, so if the CM doesn't exist but bootstrapping is still in-progress, that could be construed as an error, but that doesn't seem likely at this point and we would want to smoke out that issue separately IMO

ironcladlou: Although since this patch is now targeting 4.6, I guess we could go back to the "cm must always exist" invariant. As I have it here we could feasibly backport.

hexfusion: > In a new 4.5+ cluster, the CM is created during bootstrap, so if the CM doesn't exist but bootstrapping is still in-progress, that could be construed as an error, but that doesn't seem likely at this point and we would want to smoke out that issue separately IMO

OK lets track that, thanks

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
```diff
@@ -112,15 +132,60 @@ func (c *EtcdEndpointsController) syncConfigMap(ctx context.Context, recorder ev
...
 }
+
+// isBootstrapComplete returns true if bootstrap has completed. This is used to
+// indicate whether it's safe for clients to forget about the bootstrap member IP.
+func isBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPodClient v1helpers.StaticPodOperatorClient) (bool, error) {
+	// do a cheap check to see if the annotation is already gone.
+	// check to see if bootstrapping is complete
+	bootstrapFinishedConfigMap, err := configMapClient.ConfigMaps("kube-system").Get("bootstrap")
+	if err != nil {
+		if errors.IsNotFound(err) {
```
deads2k, line 160: style: prefer
```go
if errors.IsNotFound(){
   return
}
if err != nil{
   different return
}
```

PR #368, opened 2020-06-02
Silence grpc connection log noise

ironcladlou:
A hack to eliminate the superfluous grpc connection log spam. Warnings and
errors are still emitted.

https://github.com/kubernetes/kubernetes/issues/80741
https://github.com/kubernetes/kubernetes/pull/84061

### DIFF
--- a/cmd/cluster-etcd-operator/main.go
+++ b/cmd/cluster-etcd-operator/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	goflag "flag"
 	"fmt"
+	"io/ioutil"
 	"math/rand"
 	"os"
 	"time"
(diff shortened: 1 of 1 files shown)


ironcladlou:
Examples of important but infrequent logs which are preserved:

```
WARNING: 2020/06/02 19:20:08 grpc: addrConn.createTransport failed to connect to {https://10.0.0.4:2379  <nil> 0 <nil>}. Err :connection error: desc = "transport: authentication handshake failed: context canc
eled". Reconnecting...
```

ironcladlou:
cc @hexfusion @retroflexer

retroflexer:
Looks good. 

If we find any of those logs useful, we can bring them back.

/lgtm

ironcladlou:
Just another note... gRPC is recording these connection events to its [tracing facility](https://github.com/grpc/grpc-go/blob/v1.26.0/resolver_conn_wrapper.go#L265) but I'm not sure if that maps back right now to prometheus metrics.

If these connection events seem useful, maybe we should consider figuring out how to tap into the existing event recorder to expose these as prometheus metrics (if they're not already) rather than relying on logs.

hexfusion:
/approve

I have not yet found the info logging to be of use for triaging operator bugs. We can let this soak for a bit and extend if we find the need.

PR #370, opened 2020-06-03
Bug 1835238: Exclude bootstrap in the env variable ETCDCTL_ENDPOINTS

retroflexer:
The environment variable is used by default as the set of endpoints for all etcdctl commands. Including the bootstrap node, which is removed after the successful installation, results in reporting unexpected errors. 

We do not want to include initially and later remove it because triggering a revision is costly and the simpler solution would be to not include the bootstrap node in the first place.

### DIFF
--- a/pkg/etcdenvvar/etcd_env.go
+++ b/pkg/etcdenvvar/etcd_env.go
@@ -132,18 +132,6 @@ func getEtcdGrpcEndpoints(envVarContext envVarContext) (string, error) {
 		endpoints = append(endpoints, fmt.Sprintf("https://%s:2379", endpointIP))
 	}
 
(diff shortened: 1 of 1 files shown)


ironcladlou:
Seems right to me. Aside for later discussion: is there a way to avoid persisting the bootstrap ip even in the ALL_ETCD_ENDPOINTS var without removing it and incurring another revision? Would be nice after bootstrap to purge all traces of bootstrap stuff everywhere if possible, they seem to be suspicious to end users at best and functional landmines at worst

ironcladlou:
This seems like the fix for https://bugzilla.redhat.com/show_bug.cgi?id=1835238, if so time to retitle and remove the WIP?

PR #371, opened 2020-06-04
Bug 1842431: cluster-backup.sh: error if script dependencies are not met

hexfusion:
This PR adds error handling to assist user in understanding that the backup script is not located on a node that is able to perform backup. Also updated the script to be valid against shellcheck.

### DIFF
--- a/bindata/etcd/cluster-backup.sh
+++ b/bindata/etcd/cluster-backup.sh
@@ -25,7 +25,7 @@ if [ -z "$1" ] || [ -f "$1" ]; then
 fi
 
 if [ ! -d "$1" ]; then
-  mkdir -p $1
+  mkdir -p "$1"
 fi
 
 # backup latest static pod resources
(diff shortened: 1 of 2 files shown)


bindata/etcd/cluster-backup.sh
```diff
@@ -54,17 +65,17 @@ BACKUP_TAR_FILE=${BACKUP_DIR}/static_kuberesources_${DATESTRING}.tar.gz
 SNAPSHOT_FILE="${BACKUP_DIR}/snapshot_${DATESTRING}.db"
 BACKUP_RESOURCE_LIST=("kube-apiserver-pod" "kube-controller-manager-pod" "kube-scheduler-pod" "etcd-pod")
 
-trap "rm -f ${BACKUP_TAR_FILE} ${SNAPSHOT_FILE}" ERR
+trap 'rm -f ${BACKUP_TAR_FILE} ${SNAPSHOT_FILE}' ERR
```
ironcladlou, line 68: This seems like an important change — was it broken somehow because of the expansion?

retroflexer: It works either way in this case. With double quotes the expansion happen at the definition of trap, whereas with single quote the expansion happens when signalled.

I think Sam ran through shellcheck and fixing the errors and warnings it reports.

retroflexer: For example, consider this test:
```
#!/bin/bash
tempfile="/tmp/tmpdata1"
trap 'echo $tempfile' EXIT
tempfile=/tmp/tmpdata2
```
If double quote is used, it prints `/tmp/tmpdata1`, but if single quote is used it prints `/tmp/tmpdata2`. But in our case the variables are defined once and never changed, so it doesn't matter.

hexfusion:
/hold
this is good timing to fix DR tests

PR #374, opened 2020-06-09
Bug 1812570: Restart operator when metrics serving cert is modified

ironcladlou:
Update the operator command so that when the metrics serving certificate or key
is modified, the operator restarts to pick up the new contents.

### DIFF
--- a/manifests/0000_12_etcd-operator_06_deployment.yaml
+++ b/manifests/0000_12_etcd-operator_06_deployment.yaml
@@ -33,6 +33,8 @@ spec:
         args:
         - "--config=/var/run/configmaps/config/config.yaml"
         - "-v=4"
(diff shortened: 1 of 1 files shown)


ironcladlou:
Here's what the operator logs look like when the secret contents are changed:

```
I0609 19:50:24.473573       1 observer_polling.go:120] Observed file "/var/run/secrets/serving-cert/tls.crt" has been modified (old="32fd4d69f12e3afadeeb619774cf33d0282809bb6aa6ac94316fc38301fb1fcf", new="78a
43d241e0f515e7ab7ba9fc3cf9be8e0e177907692965db400b53fd5a48fcc")
W0609 19:50:24.473628       1 builder.go:119] Restart triggered because of file /var/run/secrets/serving-cert/tls.crt was modified
```

Graceful shutdown completes and the operator restarts.

PR #375, opened 2020-06-11
[wip] Bug 1846093: pkg/cmd/render: use machineNetwork CIDR to validate bootstrap IP

hexfusion:

Note: known that unit tests will fail for now

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -267,19 +267,17 @@ func (t *TemplateData) setBootstrapIP(configPath string) error {
 		return err
 	}
 
+	//TODO single loop..
 	// check interfaces for IP address that matches the machineNetwork CIDR
-	if machineCIDR != "" {
-		for _, addr := range ips {
-			_, parsedNet, err := net.ParseCIDR(machineCIDR)
-			if err != nil {
-				return err
-			}
-			if parsedNet.Contains(net.ParseIP(addr)) {
-				t.BootstrapIP = addr
-				return nil
-			}
+	for _, addr := range ips {
+		_, parsedNet, err := net.ParseCIDR(machineCIDR)
+		if err != nil {
+			return err
 		}
-		return fmt.Errorf("no IP address matches machineNetwork CIDR %s", machineCIDR)
+		if parsedNet.Contains(net.ParseIP(addr)) {
+			t.BootstrapIP = addr
+			return nil
+		}
 	}
 
 	// fall back to returning first IPv4 or IPv6 address
@@ -445,7 +443,7 @@ func getMachineCIDR(configFile string, isSingleStackIPv6 bool) (string, error) {
 		return fmt.Sprintf("%v", networking["machineCIDR"]), nil
 	}
 
-	return "", nil
+	return "", fmt.Errorf("machineNetwork is not found in install-config")
 }
 
 func (t *TemplateData) setSingleStackIPv6() error {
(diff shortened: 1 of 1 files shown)


pkg/cmd/render/render.go
```diff
@@ -249,33 +256,48 @@ func (r *renderOpts) Run() error {
...
 		return err
 	}
 
+	machineCIDR, err := getMachineCIDR(configPath, t.SingleStackIPv6)
+	if err != nil {
+		return err
+	}
+
+	// check interfaces for IP address that matches the machineNetwork CIDR
+	if machineCIDR != "" {
```
hardys, line 271: Under what circumstances could the machineCIDR be an empty string?  

Falling back to the first-ip approach seems likely to be unreliable in most baremetal cases, so I'd like to ensure there's not a corner case which could result in that path - would it be reasonable to return err for both nil and empty string cases, and remove the fallback first-nic code?

hexfusion: good question I think I think the only way would be if it did not exist in which case I wasn't sure that would be an error as I don't believe it is required. We could just consider 
```
	machineCIDR, found, err := getMachineCIDR(configPath, t.SingleStackIPv6)
```

cc@ironcladlou

hardys: I was thinking it is required at the installer level e.g https://github.com/openshift/installer/blob/master/pkg/types/validation/installconfig.go#L249 but may be missing something wrt the usage of that network here :)

hexfusion: ok cool so if it is required then that makes this easier and we can error thanks for the heads up @hardys

romfreiman: `/openshift-baremetal-install explain installconfig.networking
KIND:     InstallConfig
VERSION:  v1

RESOURCE: <object>
  Networking is the configuration for the pod network provider in the cluster.

FIELDS:
    clusterNetwork <[]object>
      ClusterNetwork is the list of IP address pools for pods. Default is 10.128.0.0/14 and a host prefix of /23.
      ClusterNetworkEntry is a single IP address block for pod IP blocks. IP blocks are allocated with size 2^HostSubnetLength.

    clusterNetworks <[]object>
      Deprecated name for ClusterNetwork
      ClusterNetworkEntry is a single IP address block for pod IP blocks. IP blocks are allocated with size 2^HostSubnetLength.

    machineCIDR <Any>
      Deprecated name for MachineCIDRs. If set, MachineCIDRs must be empty or the first index must match.

    machineNetwork <[]object>
      MachineNetwork is the list of IP address pools for machines. This field replaces MachineCIDR, and if set MachineCIDR must be empty or match the first entry in the list. Default is 10.0.0.0/16 for all platforms other than libvirt. For libvirt, the default is 192.168.126.0/24.
      MachineNetworkEntry is a single IP address block for node IP blocks.

    networkType <string>
      Default: "OpenShiftSDN"
      NetworkType is the type of network to install. The default is OpenShiftSDN

    serviceCIDR <Any>
      Depcreated name for ServiceNetwork

    serviceNetwork <[]Any>
      ServiceNetwork is the list of IP address pools for services. Default is 172.30.0.0/16. NOTE: currently only one entry is supported.

    type <string>
      Deprecated name for NetworkType
`

hexfusion:
work continued here https://github.com/openshift/cluster-etcd-operator/pull/384

PR #381, opened 2020-06-19
remove etcdMemberIPMigrator controller

hexfusion:
After 4.4 we don't need this controller as all members have been migrated. Lets remove in 4.6 and look at 4.5 backport.


ironcladlou:
+0 -306, lovely!

/lgtm

PR #383, opened 2020-06-22
Defrag etcd on startup

ironcladlou:
This change introduces an etcd defrag on startup during pod init.

Performing the defrag on startup is a low risk way to establish a minimal
guarantee that defrag will occur automatically under at least one condition
(during an upgrade). This changes is not intended to replace the need for live
defrags during steady state operations.

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -10,6 +10,35 @@ metadata:
     revision: "REVISION"
 spec:
   initContainers:
+    - name: defrag
+      image: ${IMAGE}
+      imagePullPolicy: IfNotPresent
+      terminationMessagePolicy: FallbackToLogsOnError
+      resources:
+        requests:
+          memory: 60Mi
+          cpu: 30m
+      volumeMounts:
+      - mountPath: /var/lib/etcd/
+        name: data-dir
+      env:
+${COMPUTED_ENV_VARS}
+      command:
+        - /bin/sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          # Try and ensure etcd is really shut down.
+          echo -n "Waiting for ports 2379, 2380 and 9978 to be released."
+          while [ -n "$(lsof -ni :2379)$(lsof -ni :2380)$(lsof -ni :9978)" ]; do
+            echo -n "."
+            sleep 1
+          done
+
+          echo "Defragging etcd"
+          time etcdctl defrag --data-dir /var/lib/etcd
     - name: etcd-ensure-env-vars
       image: ${IMAGE}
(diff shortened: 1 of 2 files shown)


bindata/etcd/pod.yaml
```diff
@@ -151,6 +151,8 @@ ${COMPUTED_ENV_VARS}
         env | grep ETCD | grep -v NODE
 
         set -x
+        time etcdctl defrag
```
hexfusion, line 154: you will need to set --command-timeout not 100% sure what would be global but I think 45s should be reasonable.

hexfusion: we might need to retry in case of failure so for example if we timeout we need to retry until success.

hexfusion: also as we want to make sure we are defraging the store explicitly  which needs `-data-dir=`

hexfusion: actually disregard `--command-timeout` as we are not on the wire.

bindata/etcd/pod.yaml
```diff
@@ -10,6 +10,46 @@ metadata:
...
+        privileged: true
+      command:
+        - /bin/sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          echo -n "Waiting for ports 2379, 2380 and 9978 to be released."
+          while [ -n "$(lsof -ni :2379)$(lsof -ni :2380)$(lsof -ni :9978)" ]; do
```
tnozicka, line 27: /hold

the check needs to be in the same container as listen - if a container restarts this check wouldn't be executed again

ironcladlou: I restored the check during startup, and also duplicated the check as a precondition in the defrag container to help ensure we don't perform the operation against a live(ish) member.

Thanks!

bindata/etcd/pod.yaml
```diff
@@ -10,6 +10,35 @@ metadata:
...
+      command:
+        - /bin/sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          # Try and ensure etcd is really shut down.
+          echo -n "Waiting for ports 2379, 2380 and 9978 to be released."
+          while [ -n "$(lsof -ni :2379)$(lsof -ni :2380)$(lsof -ni :9978)" ]; do
```
tnozicka, line 35: we have found out that `lsof` doesn't output ports in TIME_WAIT and migrated to `ss` https://github.com/openshift/cluster-kube-controller-manager-operator/blob/a45783f/bindata/v4.1.0/kube-controller-manager/pod.yaml#L21 (also added it to the base image)

PR #386, opened 2020-07-08
Remove etcdDiscoveryDomain

hexfusion:
After 4.4 we no longer require DNS entries for etcd in SAN so lets clean that up.

requires https://github.com/openshift/cluster-etcd-operator/pull/381
required for https://github.com/openshift/installer/pull/3626/files


retroflexer:
DNS is no longer needed or supported. 

/lgtm

PR #387, opened 2020-07-08
hack: add shellcheck

hexfusion:
I borrowed the implementation from the cluster monitor operator. With this we can add a test to CI.

### DIFF
--- a/hack/shellcheck.sh
+++ b/hack/shellcheck.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+set -eux
+
+#TODO add make support in https://github.com/openshift/build-machinery-go
+
+if [ "${IS_CONTAINER:-}" != "" ]; then
+  TOP_DIR="${1:-.}"
+  find "${TOP_DIR}" \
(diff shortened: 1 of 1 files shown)


retroflexer:
Cool. We can manually run in our development environments but we can invoke it from CI checks as well.

/lgtm

hack/shellcheck.sh
```diff
@@ -0,0 +1,19 @@
...
+
+#TODO add make support in https://github.com/openshift/build-machinery-go
+
+if [ "${IS_CONTAINER:-}" != "" ]; then
+  TOP_DIR="${1:-.}"
+  find "${TOP_DIR}" \
+    -path "${TOP_DIR}/vendor" -prune \
+    -o -type f -name '*.sh' -exec shellcheck --format=gcc {} \+
+else
+  docker run --rm \
```
retroflexer, line 12: Do we want to use `podman` instead of `docker` here?

ironcladlou: I'd suggest making or using some existing util function to pick whichever is available for portability reasons

PR #385, opened 2020-06-24
Bug 1854402: Improve bootstrap reliability on heterogeneous UPI network configurations

ironcladlou:
Before this change, bootstrap IP discovery assumed that the first address of the
unicast interface must be the bootstrap IP. This assumption doesn't always hold
in the face of user-defined interfaces and addresses whose ordering isn't
guaranteed. When the assumptions are broken and the incorrect bootstrap IP is
selected, bootstrapping fails because quorum cannot be established.

This change improves the accuracy of bootstrap IP discovery by more flexibly
accounting for a wider variety of possible network interface configurations.

An IP is now considered the bootstrap IP if all of the following are true.

For IPv4:

The IP is contained by the machine CIDR defined in the cluster configuration
On bare metal platforms, the IP is not the API or DNS VIP in the cluster configuration
For IPv6, the same must be true in addition to the following:

The IP is not deprecated
The IP is routable according at least one non-default route
This work is adapted from https://github.com/openshift/baremetal-runtimecfg/blob/master/pkg/utils/utils.go.

Backport of https://github.com/openshift/cluster-etcd-operator/pull/384 and https://github.com/openshift/cluster-etcd-operator/pull/388

PR #388, opened 2020-07-13
Bug 1846093: Make bootstrap IP discovery backwards compatible with previous assumptions

ironcladlou:
Before this patch, the new bootstrap IP discovery mechanism would fail bootstrapping
if no IP could be intelligently discovered. A side-effect of that is effectively
validating the machine network CIDR by asserting the bootstrap IPs ability to be
discovered within it. Because there may still be edge cases where we fail to detect
but where the old assumption to choose the "first IP" would still work, we could
introduce an undue burden to fix all existing uses of machine network CIDR even
when our fallback could continue to work in those cases.

This patch adds a fallback behavior so that when intelligent discovery fails, the
first listed IP is selected with a warning, preserving the original discovery
behavior.

This does effectively mean that clusters can still fail to bootstrap if even the
first IP assumption is wrong, but we can presumably use those failures to further
improve detection.

A worthwhile future improvement would be to find a way to more loudly and clearly
surface to the user when we're blindly guessing about the IP, as the resulting
downstream failure may obfuscate the source of failure if bootkube logs are lost.

### DIFF
--- a/pkg/cmd/render/bootstrap_ip.go
+++ b/pkg/cmd/render/bootstrap_ip.go
@@ -20,6 +20,10 @@ func NetlinkBootstrapIPLocator() *bootstrapIPLocator {
 	}
 }
 
+// BootstrapIPLocator tries to find the bootstrap IP for the machine. It should
+// go through the effort of identifying the IP based on its inclusion in the machine
+// network CIDR, routability, etc. and fall back to using the first listed IP
+// as a last resort (and for compatibility with old behavior).
 type BootstrapIPLocator interface {
 	getBootstrapIP(ipv6 bool, machineCIDR string, excludedIPs []string) (net.IP, error)
 }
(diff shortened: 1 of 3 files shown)


abhinavdahiya:
The machinenetwork is always specified in install config generated by installer, but the user is not required to specify it we default whenever we are provided with that value.

So even though we can tell users that hey set the correct network if you have many interfaces attached for we can pick the right one, by default when there is no choice _which is true usually for most cases_ etcd should just pick that and not require the machine network match... as this would allow users that haven't been required to accurately set the machine network still work...

hopefully that is making sense...

dougsland:
@ironcladlou do you have a bugzilla report to track?

dougsland:
LTGM

ironcladlou:
It's a part of http://bugzilla.redhat.com/show_bug.cgi?id=1846093

sdodson:
https://bugzilla.redhat.com/show_bug.cgi?id=1856336 also filed, of course as an installer bug :grinning:

ironcladlou:
Thanks Scott, I went ahead and moved https://bugzilla.redhat.com/show_bug.cgi?id=1846093 back to POST so I could associate this PR with it given the original fix was incomplete.

ironcladlou:
https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/388/pull-ci-openshift-cluster-etcd-operator-master-e2e-metal-ipi/1283092769849479168

As far as I can tell, etcd and the self-hosted control plane all came up and significant bootstrap progress was made, and at some point there was spinning with errors like this

```
Jul 14 18:44:05 localhost bootkube.sh[2805]: "99_openshift-cluster-api_master-machines-0.yaml": failed to create machines.v1beta1.machine.openshift.io/ostest-master-0 -n openshift-machine-api: Internal error occurred: admission plugin "MutatingAdmissionWebhook" failed to complete mutation in 13s
Jul 14 18:44:05 localhost bootkube.sh[2805]: "99_openshift-cluster-api_master-machines-1.yaml": failed to create machines.v1beta1.machine.openshift.io/ostest-master-1 -n openshift-machine-api: Internal error occurred: admission plugin "MutatingAdmissionWebhook" failed to complete mutation in 13s
Jul 14 18:44:05 localhost bootkube.sh[2805]: "99_openshift-cluster-api_master-machines-2.yaml": failed to create machines.v1beta1.machine.openshift.io/ostest-master-2 -n openshift-machine-api: Internal error occurred: admission plugin "MutatingAdmissionWebhook" failed to complete mutation in 13s
Jul 14 18:44:05 localhost bootkube.sh[2805]: "99_openshift-cluster-api_worker-machineset-0.yaml": failed to create machinesets.v1beta1.machine.openshift.io/ostest-worker-0 -n openshift-machine-api: Internal error occurred: admission plugin "MutatingAdmissionWebhook" failed to complete mutation in 13s
Jul 14 18:44:19 localhost bootkube.sh[2805]: Failed to create "99_openshift-cluster-api_master-machines-0.yaml" machines.v1beta1.machine.openshift.io/ostest-master-0 -n openshift-machine-api: Internal error occurred: admission plugin "MutatingAdmissionWebhook" failed to complete mutation in 13s
Jul 14 18:44:32 localhost bootkube.sh[2805]: Failed to create "99_openshift-cluster-api_master-machines-1.yaml" machines.v1beta1.machine.openshift.io/ostest-master-1 -n openshift-machine-api: Internal error occurred: admission plugin "MutatingAdmissionWebhook" failed to complete mutation in 13s
Jul 14 18:44:45 localhost bootkube.sh[2805]: Failed to create "99_openshift-cluster-api_master-machines-2.yaml" machines.v1beta1.machine.openshift.io/ostest-master-2 -n openshift-machine-api: Internal error occurred: admission plugin "MutatingAdmissionWebhook" failed to complete mutation in 13s
```

If the failure is a result of this PR it's not really obvious how, if we can get a clean run on a retry that would probably prove there's some other flake.

ironcladlou:
https://github.com/openshift/cluster-etcd-operator/pull/385 contains the backport

PR #391, opened 2020-07-14
Bug 1846574: bindata/bootkube/bootstrap-manifests: add proper env to metrics container on bootstrap.

hexfusion:


ironcladlou:
What job will verify this?

hexfusion:
> What job will verify this?

none

hexfusion:
@ironcladlou we should consider another action item from this.

As monitoring does not actually harvest bootstrap etcd metrics.

- should/can it?
- depending on above should we just save resources and remove the container altogether.

hexfusion:
This PR is no longer necessary as we removed the metrics container from bootstrap all together in https://github.com/openshift/cluster-etcd-operator/pull/416

PR #394, opened 2020-07-15
Migrate etcd-quorum-guard from MCO.

retroflexer:
Migrating the quorum guard from machine-config-operator. 

Note that the TLS certs come from secrets and configmaps, instead of files from the host (MCO was using files on disk).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,3 +35,14 @@ $(call add-bindata,etcd,./bindata/etcd/...,bindata,etcd_assets,pkg/operator/etcd
 # See vendor/github.com/openshift/build-machinery-go/scripts/run-telepresence.sh for usage and configuration details
 export TP_DEPLOYMENT_YAML ?=./manifests/0000_12_etcd-operator_06_deployment.yaml
 export TP_CMD_PATH ?=./cmd/cluster-etcd-operator
+
+# This was copied from https://github.com/openshift/cluster-kube-apiserver-operator
+# Exclude e2e tests from unit testing
+GO_TEST_PACKAGES :=./pkg/... ./cmd/...
+
+test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
+test-e2e: GO_TEST_FLAGS += -v
+test-e2e: GO_TEST_FLAGS += -timeout 2h
+test-e2e: GO_TEST_FLAGS += -p 1
+test-e2e: test-unit
+.PHONY: test-e2e
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/openshift/build-machinery-go v0.0.0-20200512074546-3744767c4131
 	github.com/openshift/client-go v0.0.0-20200521150516-05eb9880269c
 	github.com/openshift/library-go v0.0.0-20200526124911-cd27f9384ffc
+	github.com/pkg/errors v0.8.1
 	github.com/prometheus/client_golang v1.1.0
 	github.com/prometheus/common v0.6.0
 	github.com/spf13/cobra v0.0.5
@@ -19,6 +20,7 @@ require (
 	golang.org/x/sys v0.0.0-20200323222414-85ca7c5b95cd
 	google.golang.org/grpc v1.26.0
 	k8s.io/api v0.18.3
+	k8s.io/apiextensions-apiserver v0.18.3
 	k8s.io/apimachinery v0.18.3
 	k8s.io/client-go v0.18.3
 	k8s.io/component-base v0.18.3
--- a/manifests/0000_12_etcd-operator_08_etcdquorumguard_deployment.yaml
+++ b/manifests/0000_12_etcd-operator_08_etcdquorumguard_deployment.yaml
@@ -1,0 +1,102 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: etcd-quorum-guard
+  namespace: openshift-etcd
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      k8s-app: etcd-quorum-guard
+  strategy:
+    rollingUpdate:
+      maxSurge: 0
+      maxUnavailable: 1
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        name: etcd-quorum-guard
+        k8s-app: etcd-quorum-guard
+    spec:
+      hostNetwork: true
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - labelSelector:
+                matchExpressions:
+                  - key: k8s-app
+                    operator: In
+                    values:
+                      - "etcd-quorum-guard"
+              topologyKey: kubernetes.io/hostname
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      priorityClassName: "system-cluster-critical"
+      terminationGracePeriodSeconds: 3
+      tolerations:
+        - key: node-role.kubernetes.io/master
+          effect: NoSchedule
+          operator: Exists
+        - key: node.kubernetes.io/not-ready
+          effect: NoExecute
+          operator: Exists
+          tolerationSeconds: 120
+        - key: node.kubernetes.io/unreachable
+          effect: NoExecute
+          operator: Exists
+          tolerationSeconds: 120
+        - key: node-role.kubernetes.io/etcd
+          operator: Exists
+          effect: NoSchedule
+      containers:
+        - name: guard
+          image: quay.io/openshift/origin-cli:latest
+          imagePullPolicy: IfNotPresent
+          terminationMessagePolicy: FallbackToLogsOnError
+          volumeMounts:
+            - mountPath: /var/run/secrets/etcd-client
+              name: etcd-client
+            - mountPath: /var/run/configmaps/etcd-ca
+              name: etcd-ca
+          command:
+            - /bin/bash
+          args:
+            - -c
+            - |
+              # properly handle TERM and exit as soon as it is signaled
+              set -euo pipefail
+              trap 'jobs -p | xargs -r kill; exit 0' TERM
+              sleep infinity & wait
+          readinessProbe:
+            exec:
+              command:
+                - /bin/sh
+                - -c
+                - |
+                  declare -r health_endpoint="https://localhost:2379/health"
+                  declare -r cert="/var/run/secrets/etcd-client/tls.crt"
+                  declare -r key="/var/run/secrets/etcd-client/tls.key"
+                  declare -r cacert="/var/run/configmaps/etcd-ca/ca-bundle.crt"
+                  export NSS_SDB_USE_CACHE=no
+                  [[ -z $cert || -z $key ]] && exit 1
+                  curl --max-time 2 --silent --cert "${cert//:/\:}" --key "$key" --cacert "$cacert" "$health_endpoint" |grep '{ *"health" *: *"true" *}'
+            initialDelaySeconds: 5
+            periodSeconds: 5
+            failureThreshold: 3
+            timeoutSeconds: 3
+          resources:
+            requests:
+              cpu: 10m
+              memory: 5Mi
+          securityContext:
+            privileged: true
+      volumes:
+        - name: etcd-client
+          secret:
+            secretName: etcd-client
+        - name: etcd-ca
+          configMap:
+            name: etcd-ca-bundle
--- a/test/e2e/etcquorumguard_test.go
+++ b/test/e2e/etcquorumguard_test.go
@@ -1,0 +1,272 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"github.com/pkg/errors"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/openshift/cluster-etcd-operator/test/e2e/framework"
+	corev1 "k8s.io/api/core/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+type podstatus struct {
+	node   string
+	status corev1.PodPhase
+}
+
+var nodes = make(map[string]bool)
+var pods = make(map[string]podstatus)
+
+// TestEtcdQuorumGuard tests the etcd Quorum Guard.  It assumes there
+// are exactly three master pods (as does the etcd Quorum Guard at
+// present).  The test first makes one node unschedulable and evicts
+// the EQG pod from it, ensuring that eviction succeeds.  The test
+// next makes a second node unschedulable and then attempts to evict
+// the EQG pod from it.  It checks that the pod is *not* evicted.  It
+// then makes all nodes schedulable and checks that the EQG pod is
+// present/restarted on all masters.  It then makes one node
+// unschedulable again and checks that the EQG pod is evicted.
+func TestEtcdQuorumGuard(t *testing.T) {
+	cs := framework.NewClientSet("")
+	if err := waitForEtcdQuorumGuardDeployment(cs); err != nil {
+		t.Fatalf("etcdQuorumGuard deployment not present: %s", err.Error())
+	}
+	fmt.Print("Make all schedulable\n")
+	if err := makeAllNodesSchedulable(cs); err != nil {
+		t.Errorf("Unable to make all nodes schedulable: %s", err.Error())
+	}
+	fmt.Print("Check for all running\n")
+	if err := waitForPods(cs, 3, 3, 3); err != nil {
+		t.Errorf("Unable to get all etcd-quorum-guard pods running: %s", err.Error())
+	}
+	fmt.Print("Make one unschedulable\n")
+	if err := makeOneNodeUnschedulableAndEvict(cs); err != nil {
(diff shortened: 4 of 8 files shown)


ironcladlou:
From an offline discussion worth bringing out here:

Instead of retaining CVO management of QG, this is an opportunity to bring the component under direct management of the operator. A static CVO-managed QG is simple but doesn't leave any room for us to enhance the component with dynamic behavior going forward. Is it low effort to bring QG under operator management?

retroflexer:
> From an offline discussion worth bringing out here:
> 
> Instead of retaining CVO management of QG, this is an opportunity to bring the component under direct management of the operator. A static CVO-managed QG is simple but doesn't leave any room for us to enhance the component with dynamic behavior going forward. Is it low effort to bring QG under operator management?

As we discussed on slack, I too think it is worthy idea to pursue.

But I would let this PR go through so  that the migration from MCO happens without glitches, and then, once we have the quorum guard in the etcd-operator namespace, we can move it under direct management of the operator.

ironcladlou:
>But I would let this PR go through so that the migration from MCO happens without glitches, and then, once we have the quorum guard in the etcd-operator namespace, we can move it the under direct management of the operator.

As long as we're committed to doing the operator transition within the same release, I guess I don't see the harm

hexfusion:
> As long as we're committed to doing the operator transition within the same release, I guess I don't see the harm

My understanding is that moving a CVO managed resource is easy but removing is not. If we are really going to use the operator why remove 2 CVO resources?

hexfusion:
/hold 

I would like a detailed plan to address this.

manifests/0000_12_etcd-operator_08_etcdquorumguard_deployment.yaml
```diff
@@ -0,0 +1,102 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: etcd-quorum-guard
+  namespace: openshift-etcd-operator
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+spec:
+  replicas: 3
```
vrutkovs, line 9: Could the operator control the amount of replicas? In OKD's fork we used https://github.com/openshift/machine-config-operator/commit/ddc89ca7814c38c88cfc0c27d083f82f899b5b85 in MCO - when `useUnsupportedUnsafeNonHANonProductionUnstableEtcd` is enabled etcd-quorum-guard is scaled back to 1

retroflexer:
@hexfusion  said:
> My understanding is that moving a CVO managed resource is easy but removing is not. If we are really going to use the operator why remove 2 CVO resources?
@wking Could you help me here? Is removing PDB not very easy? I was hoping just removing the pdb file is enough to get rid of it. What other implications and impacts we need to consider?

manifests/0000_12_etcd-operator_08_etcdquorumguard_deployment.yaml
wking: Operator controlling this Deployment so it can do intelligent things makes sense to me.  But can that happen after the pivot from machine-config operator to etcd operator?  We certainly don't want the etcd operator trying to do intelligent Deployment management while the machine-config operator is still trying to hand this Deployment off to the cluster-version operator, because the CVO and etcd operator would be fighting each other.

wking:
openshift/machine-config-operator#1928 is dropping some operator-specific e2e tests around this.  Does the etcd operator have a similar operator-specific e2e suite?  Or some other mechanism for ensuring that the quorum guard is operator as expected?

manifests/0000_12_etcd-operator_08_etcdquorumguard_deployment.yaml
vrutkovs: There should be any issue during transition from MCO to CEO as these are different deployments in different namespaces

wking:
> I was hoping just removing the pdb file is enough to get rid of it.

That's true for _new_ clusters.  But existing clusters will still have the resource in-cluster.  If you're just transferring ownership, that's fine, because there's no need to actually remove the in-cluster resource.  You'd want the hand-off to look something like:

1. CVO manages the resource.
2. New release, operator begins managing the resource, but with a fixed target to match the CVO.  That way the operator and CVO are both controlling the resource, but they always agree on the intended state.
3. Drop the manifest, so the CVO stops trying to manage the resource.  The operator can now do whatever it likes for intelligent management, including removing the resource entirely.

The point of the second step is that during an update, the CVO will begin reconciling the new manifests.  If you jumped straight from 1 to 3, there would be a window where the CVO had relinquished control but the etcd operator had not yet been bumped to code that picked up control.  If the manifests for the etcd operator deployment and the managed resource are close together in the update graph, maybe that's fine.  If they are far enough apart that a stuck update could leave the resource unmanaged for a significant length of time, you probably want 2's careful handoff.

manifests/0000_12_etcd-operator_08_etcdquorumguard_deployment.yaml
wking: > There should be any issue during transition from MCO to CEO as these are different deployments in different namespaces

Oh.  Why is the quorum guard not living in the same namespace as etcd?  Regardless, I'd still like the cross-repo pivot to be as boring and conservative as possible.  That way, if we flub the transition to intelligent etcd-operator management, we can easily revert the change without having to involve the MCO.  I don't think it's wrong to attempt the bigger pivot in a single PR, if folks are feeling more aggressive.  But is there much upside to moving to etcd-operator management in a single PR?

vrutkovs: >Why is the quorum guard not living in the same namespace as etcd?

That would make MCO watch events in different namespace (currently etcd-quorum-guard lives in MCO namespace). 

>  I don't think it's wrong to attempt the bigger pivot in a single PR, if folks are feeling more aggressive

Sure, I don't mind splitting replica control in a separate PR

kikisdeliveryservice:
> [openshift/machine-config-operator#1928](https://github.com/openshift/machine-config-operator/pull/1928) is dropping some operator-specific e2e tests around this. Does the etcd operator have a similar operator-specific e2e suite? Or some other mechanism for ensuring that the quorum guard is operator as expected?

Seconding that moving https://github.com/openshift/machine-config-operator/blob/master/test/e2e/etcdquorumguard_test.go is necessary and I think is necessary to have it run on this PR/change.

retroflexer:
> [openshift/machine-config-operator#1928](https://github.com/openshift/machine-config-operator/pull/1928) is dropping some operator-specific e2e tests around this. Does the etcd operator have a similar operator-specific e2e suite? Or some other mechanism for ensuring that the quorum guard is operator as expected?

 I will work on adding e2e tests on our side, once we moved to more intelligent control by operator.

test/e2e/etcquorumguard_test.go
```diff
@@ -0,0 +1,288 @@
...
+	node   string
+	status corev1.PodPhase
+}
+
+var nodes = make(map[string]bool)
+var pods = make(map[string]podstatus)
+
+type podinfo map[string]podstatus
+
+// TestEtcdQuorumGuard tests the etcd Quorum Guard.  It assumes there
```
kikisdeliveryservice, line 29: HOORAY!

retroflexer:
Update: I added e2e test framework and added a test for EtcdQuorumGuard. 

I hope we can proceed with this PR to finish cross-repo transfer.

@wking @kikisdeliveryservice

retroflexer:
@kikisdeliveryservice I need your help again. The verify-deps is failing. I ran `go mod tidy;go mod verify; go mod vendor` and it was clean.

```
go: extracting github.com/xlab/handysort v0.0.0-20150421192137-fb3537ed64a1
github.com/openshift/cluster-etcd-operator/test/e2e tested by
	github.com/openshift/cluster-etcd-operator/test/e2e.test imports
	github.com/openshift/cluster-etcd-operator/test/e2e/framework: no matching versions for query "latest"
```
I appreciate any help.

go.mod
```diff
@@ -6,10 +6,12 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
+	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
```
hexfusion, line 9: since we already have klog can we use that instead?

kikisdeliveryservice:
> @kikisdeliveryservice I need your help again. The verify-deps is failing. I ran `go mod tidy;go mod verify; go mod vendor` and it was clean.
> 
> ```
> go: extracting github.com/xlab/handysort v0.0.0-20150421192137-fb3537ed64a1
> github.com/openshift/cluster-etcd-operator/test/e2e tested by
> 	github.com/openshift/cluster-etcd-operator/test/e2e.test imports
> 	github.com/openshift/cluster-etcd-operator/test/e2e/framework: no matching versions for query "latest"
> ```
> 
> I appreciate any help.

You import "github.com/openshift/cluster-etcd-operator/test/e2e/framework" in test/e2e/etcquorumguard_test.go but you don't actually have that dir and contents anywhere :)

retroflexer:
> You import "github.com/openshift/cluster-etcd-operator/test/e2e/framework" in test/e2e/etcquorumguard_test.go but you don't actually have that dir and contents anywhere :)

Oops. Looks like I missed checking in clientset.go in framework. Thanks. It was a long day. Phew.

go.mod
retroflexer: Yes. Thanks. I got rid of glog.

Makefile
```diff
@@ -35,3 +35,15 @@ $(call add-bindata,etcd,./bindata/etcd/...,bindata,etcd_assets,pkg/operator/etcd
...
+# This was copied from https://github.com/openshift/machine-config-operator
+# Run tests
+test: test-unit test-e2e
+
+# Unit tests only (no active cluster required)
+test-unit:
+	CGO_ENABLED=0 go test -tags=$(GOTAGS) -count=1 -v ./cmd/... ./pkg/... 
+
+# This was copied from https://github.com/openshift/cluster-image-registry-operator
+test-e2e:
```
hexfusion, line 48: for e2e can we align with KAS?

https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/Makefile#L70

Also we already have `test-unit` right? 

https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-etcd-operator/openshift-cluster-etcd-operator-release-4.6.yaml#L51

retroflexer: `test-unit` was failing with e2e added. See previous test outputs:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/394/pull-ci-openshift-cluster-etcd-operator-master-unit/1285873347153891328

retroflexer:
All the required (at least the 'Required' ones) appear to be passing the tests. I am testing the y-upgrade to be working using cluster-bot (to make sure there will not be two quorum guards at the end of the upgrade).

@wking do you see any problems with going ahead with this PR?

hexfusion:
Lets step through next steps.

1.) merge this we now have 2 quorum-guard deployments.
2.) state a PR to MCO removing quorum-guard and PDB manifests.
3.) any other actions required to make sure that both Z and Y upgrades have only a single deployment?

Makefile
```diff
@@ -35,3 +35,15 @@ $(call add-bindata,etcd,./bindata/etcd/...,bindata,etcd_assets,pkg/operator/etcd
...
 export TP_DEPLOYMENT_YAML ?=./manifests/0000_12_etcd-operator_06_deployment.yaml
 export TP_CMD_PATH ?=./cmd/cluster-etcd-operator
+
+# This was copied from https://github.com/openshift/machine-config-operator
+# Run tests
+test: test-unit test-e2e
+
+# Unit tests only (no active cluster required)
+test-unit:
+	CGO_ENABLED=0 go test -tags=$(GOTAGS) -count=1 -v ./cmd/... ./pkg/...
```
wking, line 45: nit:

```console
$ git --no-pager show --oneline --check origin/pr/394
f611f340 (origin/pr/394) Add e2e test framework and add a test for EtcdQuorumGuard
Makefile:45: trailing whitespace.
+	CGO_ENABLED=0 go test -tags=$(GOTAGS) -count=1 -v ./cmd/... ./pkg/... 
```

hexfusion: @retroflexer I brought this up but I am not sure why we must redefine test-unit. KAS has both unit and e2e tests.

hexfusion:
```
vendor/github.com/openshift/build-machinery-go/make/golang.mk
```

Makefile
```diff
@@ -35,3 +35,15 @@ $(call add-bindata,etcd,./bindata/etcd/...,bindata,etcd_assets,pkg/operator/etcd
...
+# Run tests
+test: test-unit test-e2e
+
+# Unit tests only (no active cluster required)
+test-unit:
+	CGO_ENABLED=0 go test -tags=$(GOTAGS) -count=1 -v ./cmd/... ./pkg/... 
+
+# This was copied from https://github.com/openshift/cluster-image-registry-operator
+test-e2e:
+	go test -failfast -timeout 120m -v$${WHAT:+ -run="$$WHAT"} ./test/e2e/
```
wking, line 49: is `WHAT` a think here?  If so, did you want to land docs similar to what origin [used to have][1]?  Or somewhere explain to folks how it is expected to be used?

[1]: https://github.com/openshift/origin/commit/82d3e63a7e267292a28547b51bc0b79f2c7e2213#diff-b67911656ef5d18c4ae36cb6741b7965L147

retroflexer: I can removed. I think `go test` allows a regex pattern to be specified to only run a subset of tests. I can remove it, if you like.

```
-run regexp
        Run only those tests and examples matching the regular expression.
        For tests, the regular expression is split by unbracketed slash (/)
        characters into a sequence of regular expressions, and each part
        of a test's identifier must match the corresponding element in
        the sequence, if any. Note that possible parents of matches are
        run too, so that -run=X/Y matches and runs and reports the result
        of all tests matching X, even those without sub-tests matching Y,
        because it must run them to look for those sub-tests.
```

Makefile
retroflexer: @hexfusion We need to overwrite the default rules. The default test-unit is attempting to run e2e tests as well, and was failing. By specifying only ./cmd and ./pkg directories, we are excluding it from running the tests in ./e2e directory.

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/394/pull-ci-openshift-cluster-etcd-operator-master-unit/1285873347153891328

retroflexer: I deleted these overriding rules, and followed the examples from  the Makefile of https://github.com/openshift/cluster-kube-apiserver-operator. Now we have to wait for the all the tests to pass again :-).

Makefile
retroflexer: I deleted the overriding rules for test-e2e and test-unit and replaced them with definitions similar to what the Makefile from https://github.com/openshift/cluster-kube-apiserver-operator has. Let's wait for the tests to succeed again.

hexfusion:
@retroflexer has manually tested upgrades with success.
/approve

retroflexer:
Verified the PR to be working well with y-upgrades using cluster-bot on aws and gcp (tested with MCO's 1928 along with this PR).

aws:
job test upgrade 4.5 4.6,openshift/machine-config-operator#1928,openshift/cluster-etcd-operator#394 aws succeeded
(https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1286728540858880000)

gcp:
job test upgrade 4.5 4.6,openshift/machine-config-operator#1928,openshift/cluster-etcd-operator#394 gcp succeeded
(https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1286844014271664128)

wking:
[disruptive][1]:

```
Jul 25 07:41:36.324: INFO: Running AfterSuite actions on node 1
fail [github.com/openshift/origin@/test/extended/dr/quorum_restore.go:195]: Unexpected error:
    <*errors.StatusError | 0xc0023a0e60>: {
        ErrStatus: {
            TypeMeta: {Kind: "Status", APIVersion: "v1"},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "rpc error: code = Unknown desc = context deadline exceeded",
            Reason: "",
            Details: nil,
            Code: 500,
        },
    }
    rpc error: code = Unknown desc = context deadline exceeded
occurred

failed: (38m17s) 2020-07-25T07:41:36 "[sig-etcd][Feature:DisasterRecovery][Disruptive] [Feature:EtcdRecovery] Cluster should restore itself after quorum loss [Disabled:Broken] [Serial] [Suite:openshift]"
```

Any idea what's going on there?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/394/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-disruptive/1286912222311747584

retroflexer:
> fail [github.com/openshift/origin@/test/extended/dr/quorum_restore.go:195]: Unexpected error:
>    <*errors.StatusError | 0xc0023a0e60>: {

@wking This quorum restore is regarding the restoration of the backup and has nothing to do with quorum guard per se. These tests haven't passed in a while, and I agree we need to fix these tests to make them work reliably. But none of the changes in this PR, I believe, are impacting the results.

hexfusion:
/lgtm

merging with the understanding that having 2 quorum-guards for a short period of time will not cause failure.

PR #397, opened 2020-07-17
bindata/etcd: remove MCO dependency for ca.crt

hexfusion:
This change is intended to replace the existing CA on disk by MCO. If this goes right we can essentially remove all of the MCO related bits that populated the CA from MCO end.


ironcladlou:
Not familiar enough with the machinations here yet to evaluate the approach — will CI catch regressions for it?

hexfusion:
yeah, its more of a test to see if MCO will reconcile.

hexfusion:
short term this should all go away as we transition quorum-guard to openshift-etcd namespace.

PR #400, opened 2020-07-22
Set ETCD_UNSUPPORTED_ARCH on arm64

yselkowitz:


pkg/cmd/render/env.go
```diff
@@ -82,7 +81,10 @@ func getElectionTimeout(platform, arch string) (map[string]string, error) {
 }
 
 func getUnsupportedArch(platform, arch string) (map[string]string, error) {
-	if !strings.HasPrefix(arch, "s390") {
+	switch arch {
+	case "arm64":
+	case "s390x":
```
hexfusion, line 86: we need to cover s390 case as well as s390x

hexfusion:
@yselkowitz thanks for PR small knit above.

hexfusion:
can you outline the support for arm?
/approved

pkg/cmd/render/env.go
yselkowitz: Why?  Plain s390 means the very old 31-bit userspace, for which RHEL has only had multilibs in previous versions and not at all in RHEL 8.  There are certainly no plans for OCP in that configuration.

hexfusion: ok I have no context here to what s390 is outside of "both were possible" if you're saying that is not true and regression is not a concern then OK with me.

hexfusion:
do we have any plans for e2e testing to include arm base infra?

PR #405, opened 2020-07-27
Document how etcd's TLS assets are managed

markmc:
I found all of this quite bewildering to keep straight in my head, so I figured I'd try to write it down.

### DIFF
--- a/docs/etcd-tls-assets.md
+++ b/docs/etcd-tls-assets.md
@@ -1,0 +1,752 @@
+# etcd TLS Assets
+
+As of OpenShift 4.4, etcd is deployed on every control plane node as a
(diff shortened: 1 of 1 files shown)


hexfusion:
@markmc this is an amazing amount of work. Thank you, will review soon.

cgwalters:
I happened to come across this when I was trying to figure out how the etcd static pod manifest is installed.  Agreed, awesome work!  I personally would merge this as is and do other changes as followups to avoid losing it.

It seems though the second half in particular is more about the architecture, and isn't specific to certs; maybe move that into a separate `design.md`?

hexfusion:
yeah let's merge marks work and fine-tune from here thanks again.

PR #410, opened 2020-07-29
WIP: experiment with a render command for an all-in-one config

markmc:
This is just some early-stage hacking to see a render command for an all-in-one config might look like.

As such this is related to openshift/enhancements#302 but a much hackier approach to what's described there.

### DIFF
--- a/cmd/cluster-etcd-operator/main.go
+++ b/cmd/cluster-etcd-operator/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/openshift/cluster-etcd-operator/pkg/cmd/aio"
 	operatorcmd "github.com/openshift/cluster-etcd-operator/pkg/cmd/operator"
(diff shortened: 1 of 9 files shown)


hexfusion:
cc @ironcladlou

ironcladlou:
@markmc thanks for putting this together! Your work here with certificate rendering actually intersects with some tech debt we've been meaning to pay down in bootstrapping, which currently generates certificates using a potentially flaky client/server model that's now unnecessary.

I need to open a separate set of PRs for that work (in the operator and the installer) and I'll keep your PR in mind while doing so (I haven't had a chance to look closely at what you've done, it's possible you may have given me a head start).

Once I have a WIP going I'll link it to this PR. I want to make sure what I'm doing is reusable in the context of your AIO goals.

markmc:
Work along these lines is happening at openshift/enhancements#440 now

PR #412, opened 2020-07-30
Render bootstrap certificates

ironcladlou:
During render, generate the certificates necessary for the bootstrap member.

Update the bootstrap pod to use the rendered certificates.

After this change, the existing asynchronous client/server based bootstrap
certificate generation mechanism should be deleted.

Extracted from @markmc's work in https://github.com/openshift/cluster-etcd-operator/pull/410.

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
@@ -6,58 +6,6 @@ metadata:
   labels:
     k8s-app: etcd
 spec:
-  initContainers:
-  - name: certs
-    image: "{{.Images.KubeClientAgent}}"
-    command:
-    - /bin/sh
-    - -c
-    - |
-      #!/bin/sh
-      set -euxo pipefail
-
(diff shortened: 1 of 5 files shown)


bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
```diff
@@ -6,58 +6,6 @@ metadata:
   labels:
     k8s-app: etcd
 spec:
-  initContainers:
-  - name: certs
```
hexfusion, line 10: + 100

hexfusion: `+ 100`
cc @deads2k  :)

ironcladlou:
e2e should be a sufficient test here — the updated bootstrap member pod no longer refers to the certificates generated by the existing kubecsr process, so if we pivot it means our rendered certs were good enough to get us past bootstrap and the installer bits can be deleted (https://github.com/openshift/installer/pull/3995) and additional refactoring can proceed.

PR #414, opened 2020-07-31
Cleanup of old transition files and directories.

retroflexer:
etcd-member is no longer needed, as MCO doesn't depend on any etcd files.


hexfusion:
/close

in favor of https://github.com/openshift/cluster-etcd-operator/pull/469

PR #416, opened 2020-07-31
Clean up bootstrap member cert references

ironcladlou:
Adjusts the bootstrap member pod's certificate references to use the
bootkube render output paths now that the installer has been updated
to put the rendered output in the expected locations.

Remove deprecated metrics CA flag.


ironcladlou:
Depends on https://github.com/openshift/installer/pull/3995.

/hold

PR #418, opened 2020-08-06
Bug 1850057: etcd-pod: Use ionice -c2 -n0

cgwalters:
This will have no currect effect with the default I/O scheduler
of `mq-deadline`, but I think it will help once we switch to `bfq`
as part of https://github.com/openshift/machine-config-operator/pull/1946

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -151,7 +151,8 @@ ${COMPUTED_ENV_VARS}
         env | grep ETCD | grep -v NODE
 
         set -x
-        exec etcd \
+        # See https://etcd.io/docs/v3.4.0/tuning/ for why we use ionice
+        exec ionice -c2 -n0 etcd \
           --initial-advertise-peer-urls=https://${NODE_NODE_ENVVAR_NAME_IP}:2380 \
           --cert-file=/etc/kubernetes/static-pod-certs/secrets/etcd-all-serving/etcd-serving-NODE_NAME.crt \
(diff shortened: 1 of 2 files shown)


hexfusion:
I just have one outstanding question regarding ionice. https://github.com/openshift/machine-config-operator/pull/1957#issuecomment-671403646

cgwalters:
Let's get this merged in combination with https://github.com/openshift/machine-config-operator/pull/1957 ?

cgwalters:
OK, this one should be good to go!

cgwalters:
/lgtm cancel
e2e-aws-disruptive is broken by the MCO dropping e-q-c looks like

cgwalters:
> e2e-aws-disruptive is broken by the MCO dropping e-q-c looks like
Filed https://github.com/openshift/openshift-tests/issues/68 for that if someone wants to take it.

cgwalters:
Ah I think this is related to https://github.com/openshift/release/commit/151d520dc3850dd587c271581563d3898f9647bb - e2e-aws-disruptive is obsolete.

cgwalters:
OK so now that I'm not off on a wild goose chase and am looking at the right e2e-disruptive test...hm, seems generally broken right now.  There's [not many runs for that test](https://prow.ci.openshift.org/?job=*disruptive*) - do you all consider it blocking? 

I'd like to get this one in because the BZ is tracking it.

hexfusion:
> do you all consider it blocking?
generally no and certainly not for this change.

PR #419, opened 2020-08-06
Bug 1869681: handle IP migration when DNS fails on Azure IPI

ironcladlou:
Some Azure IPI clusters were observed to have DNS forwarding rules which
delegated to upstreams not containing etcd records from the cluster private DNS
zone. Such configurations thwart the IP migration logic from performing a
reverse lookup of member IP addresses, causing the migration and thus the
upgrade to fail. Technically the upstream nameserver shares some blame for not
actually being authoritative for the private zone domain, but we can work around
it to save users from intervening.

This patch introduces a fallback means of etcd member IP determination used only
when the DNS method fails on Azure. In this case, we can exploit the member
naming conventions to statically map an etcd member to a node, and then assign
the member IP based on the node IP, bypassing DNS altogether.

### DIFF
--- a/pkg/operator/etcdmemberipmigrator/etcdmemberipmigrator.go
+++ b/pkg/operator/etcdmemberipmigrator/etcdmemberipmigrator.go
@@ -7,9 +7,11 @@ import (
 	"strings"
 	"time"
 
+	"go.etcd.io/etcd/etcdserver/etcdserverpb"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/davecgh/go-spew/spew"
+	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
@@ -133,16 +135,18 @@ func (c *EtcdMemberIPMigrator) reconcileMembers() error {
 
 		c.eventRecorder.Eventf("MemberMissingIPPeer", "member %q is missing an IP in the peer list", member.Name)
 
-		// if the member does not have a peerIP, then we need to migrate it.  This approach is very heavy with dns requests
-		// but this should almost never happen and this allows us to re-use the code that we have already for DNS.
-		etcdInfo, err := c.getEtcdInfo()
+		// First, try a migration using a reverse DNS lookup.
+		requiredPeerList, err := c.getRequiredPeerListFromDNS(member)
 		if err != nil {
-			return err
+			c.eventRecorder.Warningf("MemberIPLookupFailed", "member %q IP couldn't be determined via DNS: %v; will attempt a fallback lookup", member.Name, err)
+			// If the DNS lookup fails (which could be the case if, for example, there's a problematic DNS
+			// forwarding configuration), try a non-DNS mapping of member to IP.
+			requiredPeerList, err = c.getRequiredPeerListFromFallback(member)
+			if err != nil {
+				return fmt.Errorf("failed to determine member %q IP using fallback approach: %v", member.Name, err)
+			}
 		}
-		requiredPeerList, err := getRequiredPeerList(etcdInfo, member.PeerURLs)
-		if err != nil {
-			return err
-		}
+
 		c.eventRecorder.Eventf("MemberSettingIPPeer", "member %q; new peer list %v", member.Name, strings.Join(requiredPeerList, ","))
 		if err := c.etcdClient.MemberUpdatePeerURL(member.ID, requiredPeerList); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


pkg/operator/etcdmemberipmigrator/etcdmemberipmigrator.go
```diff
@@ -152,6 +156,51 @@ func (c *EtcdMemberIPMigrator) reconcileMembers() error {
...
+	infra, err := c.infrastructureLister.Get("cluster")
+	if err != nil {
+		return nil, err
+	}
+	if infra.Status.PlatformStatus == nil {
+		return nil, fmt.Errorf("no reliable platform metadata is available")
+	}
+
+	switch infra.Status.PlatformStatus.Type {
+	case configv1.AzurePlatformType:
```
retroflexer, line 184: I assume this works too, but we have used the following code in the past to match Azure:
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/etcdenvvar/etcd_env.go#L226-L231

ironcladlou:
I'm not sure the infrastructure exists today to make an automated test for this scenario, so I'm working on manual validation at the moment.

ironcladlou:
I've got an upgrade going from 4.3.31 to a release image built from this PR with the following DNS forwarding configuration in place:

```yaml
  servers:
  - forwardPlugin:
      upstreams:
      - 1.1.1.1
    name: external-dns
    zones:
    - ci-ln-gk8wrgt-002ac.ci.azure.devcluster.openshift.com
```

The etcd rollout completes successfully and the following events are generated:

```
101s        Warning   MemberIPLookupFailed                   deployment/etcd-operator                         member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-0" IP couldn't be determined via DNS: unableto locate a node for peerURL=https://etcd-0.ci-ln-gk8wrgt-002ac.ci.azure.devcluster.openshift.com:2380, dnsErrors=lookup _etcd-server-ssl._tcp.ci-ln-gk8wrgt-002ac.ci.azure.devcluster.openshift.com on 172.30.0.10:53: no such host; will attempt a fallback lookup
101s        Normal    MemberSettingIPPeer                    deployment/etcd-operator                         member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-0"; new peer list https://10.0.0.6:2380
101s        Normal    MemberUpdate                           deployment/etcd-operator                         updating member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-0" with peers https://10.0.0.6:2380
101s        Normal    MemberMissingIPPeer                    deployment/etcd-operator                         member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-1" is missing an IP in the peer list
88s         Warning   MemberIPLookupFailed                   deployment/etcd-operator                         member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-1" IP couldn't be determined via DNS: unableto locate a node for peerURL=https://etcd-1.ci-ln-gk8wrgt-002ac.ci.azure.devcluster.openshift.com:2380, dnsErrors=lookup _etcd-server-ssl._tcp.ci-ln-gk8wrgt-002ac.ci.azure.devcluster.openshift.com on 172.30.0.10:53: no such host; will attempt a fallback lookup
88s         Normal    MemberSettingIPPeer                    deployment/etcd-operator                         member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-1"; new peer list https://10.0.0.5:2380
88s         Normal    MemberUpdate                           deployment/etcd-operator                         updating member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-1" with peers https://10.0.0.5:2380
88s         Normal    MemberMissingIPPeer                    deployment/etcd-operator                         member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-2" is missing an IP in the peer list
76s         Warning   MemberIPLookupFailed                   deployment/etcd-operator                         member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-2" IP couldn't be determined via DNS: unableto locate a node for peerURL=https://etcd-2.ci-ln-gk8wrgt-002ac.ci.azure.devcluster.openshift.com:2380, dnsErrors=lookup _etcd-server-ssl._tcp.ci-ln-gk8wrgt-002ac.ci.azure.devcluster.openshift.com on 172.30.0.10:53: no such host; will attempt a fallback lookup
76s         Normal    MemberSettingIPPeer                    deployment/etcd-operator                         member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-2"; new peer list https://10.0.0.7:2380
76s         Normal    MemberUpdate                           deployment/etcd-operator                         updating member "etcd-member-ci-ln-gk8wrgt-002ac-dmd9h-master-2" with peers https://10.0.0.7:2380
```

So this all looks good. However, I still need to do the negative test to prove it fails from 4.4.31 -> latest stable 4.4 image.

ironcladlou:
Here's the negative test proving the problem. Upgrade from 4.3.31 to 4.4.16 with this DNS configuration:

```yaml
servers:
  - forwardPlugin:
      upstreams:
      - 1.1.1.1
    name: external-dns
    zones:
    - ci-ln-2pmxhgk-002ac.ci.azure.devcluster.openshift.com
```

```
$ oc get clusterversion/version
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.3.31    True        True          10m     Unable to apply 4.4.16: the cluster operator etcd is degraded

$ oc get clusteroperators/etcd -o yaml
apiVersion: config.openshift.io/v1
kind: ClusterOperator
metadata:
  creationTimestamp: "2020-08-10T14:39:34Z"
  generation: 1
  name: etcd
  resourceVersion: "25429"
  selfLink: /apis/config.openshift.io/v1/clusteroperators/etcd
  uid: 91438145-ab1a-4717-800f-57007cda0a72
spec: {}
status:
  conditions:
  - lastTransitionTime: "2020-08-10T14:41:34Z"
    message: 'EtcdMemberIPMigratorDegraded: unable to locate a node for peerURL=https://etcd-1.ci-ln-2pmxhgk-002ac.ci.azure.devcluster.openshift.com:2380,dnsErrors=lookup _etcd-server-ssl._tcp.ci-ln-2pmxhgk-002ac.ci.azure.devcluster.openshift.com on 172.30.0.10:53: no such host'
    reason: EtcdMemberIPMigrator_Error
    status: "True"
    type: Degraded
```

pkg/operator/etcdmemberipmigrator/etcdmemberipmigrator.go
```diff
@@ -152,6 +156,51 @@ func (c *EtcdMemberIPMigrator) reconcileMembers() error {
...
+		network, err := c.networkLister.Get("cluster")
+		if err != nil {
+			return nil, err
+		}
+		nodeName := strings.ReplaceAll(member.Name, "etcd-member-", "")
```
hexfusion, line 189: Are we 100% certain this will always work even with Azure UPI? Why not use member list here to get node IP?

ironcladlou: No, but we have no data about this fallback being useful except in the following conditions:

1. Azure IPI
2. DNS forwarding enabled _for the cluster domain_
3. The upstream is misconfigured in that it excludes the etcd records or fails to delegate back to the OpenShift-defined zones containing the etcd records

So I don't really know it matters whether it breaks outside those conditions. This PR isn't intended to address every conceivable DNS forwarding problem. The answer in many cases may in fact be to ask the user to fix their nameservers.

hexfusion: OK as long as are not introducing a new failure case but looking this over the worst case is retry.

eparis:
I'm going to manually add valid-bug. This doesn't exist in 4.4->4.5.  Mace is going to include test instructions in the BZ to make sure we can have a second set of testing eyes easily and reliably.

PR #420, opened 2020-08-11
pkg/etcdenvvar: enable pprof by default

hexfusion:
This PR turns on the go profiling handler by default. My intial concern with enabling this was overhead but in speaking with Clayton and further research overhead is pretty much nothing unless the handler is used.

usage example fort debugging heap.

```
oc rsh -T -n openshift-etcd etcd-ip-10-0-155-214.us-west-1.compute.internal  sh -c 'curl --cert $ETCDCTL_CERT --key $ETCDCTL_KEY --cacert $ETCDCTL_CACERT https://localhost:2379/debug/pprof/heap' > /tmp/heap

PROF_BINARY_PATH=/tmp go tool pprof /tmp/heap
```


hexfusion:
quick perf reviewed showed no overhead in the change

ironcladlou:
Great, thanks

/lgtm

PR #434, opened 2020-09-04
Bug 1874597: static-pod-installer: remove deadlock by recreating installers that disappear

hexfusion:


hexfusion:
cc @sdodson

sdodson:
> cc @sdodson

Thanks, I'm interested in this because it's halted upgrades but I'm not patch manager right now so another patch manager will hopefully approve this Wed - Friday.

markmc:
Helpful to reference the original openshift/library-go#871 which has more background on the issue

sdodson:
> Helpful to reference the original [openshift/library-go#871](https://github.com/openshift/library-go/pull/871) which has more background on the issue

And the upstream PR for this repo too https://github.com/openshift/cluster-etcd-operator/pull/431

PR #436, opened 2020-09-08
Bug 1837540: Use restore pod yaml from the backup when restoring

retroflexer:
Currently when we restore a cluster from backup, we use restore-pod.yaml from the currently installed image. This is problematic especially when switching across releases (minor or micro release versions).

### DIFF
--- a/bindata/etcd/cluster-restore.sh
+++ b/bindata/etcd/cluster-restore.sh
@@ -7,17 +7,26 @@ set -o pipefail
 set -o errtrace
 
 # example
-# ./cluster-restore.sh $path-to-backup 
+# ./cluster-restore.sh $path-to-backup
 
 if [[ $EUID -ne 0 ]]; then
   echo "This script must be run as root"
   exit 1
 fi
 
-source /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd.env
-source /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd-common-tools
+function source_required_dependency {
+  local path="$1"
+  if [ ! -f "${path}" ]; then
+    echo "required dependencies not found, please ensure this script is run on a node with a functional etcd static pod"
+    exit 1
+  fi
+  # shellcheck disable=SC1090
+  source "${path}"
+}
+source_required_dependency /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd.env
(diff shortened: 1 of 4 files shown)


bindata/etcd/cluster-backup.sh
```diff
@@ -45,6 +45,10 @@ function backup_latest_kube_static_resources {
     LATEST_RESOURCE_DIRS+=("${LATEST_RESOURCE#${CONFIG_FILE_DIR}/}")
   done
 
+ # Add restore-pod directory the archive
```
hexfusion, line 48: could we explore adding  {Name: "restore-etcd-pod"}, to`RevisionConfigMaps`[1]

I think that way we just back it up by default. I think then we could consider dropping the ENV var `RESTORE_ETCD_POD_YAML` all together unless we have other reasons to keep it.

[1] https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/starter.go#L271

retroflexer: Thanks. Done.

hexfusion:
@retroflexer can we remove RESTORE_ETCD_POD_YAML, I don't think we need that anymore right?

https://github.com/openshift/cluster-etcd-operator/blob/68d208baa7fcef428f433d408e2dadd0fc44863f/bindata/etcd/etcd-common-tools#L8

hexfusion:
/hold 

I would like to understand disruptive failure;

```
[sig-etcd][Feature:DisasterRecovery][Disruptive] [Feature:EtcdRecovery] Cluster should restore itself after quorum loss [Serial] [Suite:openshift] expand_less	11m54s
fail [github.com/openshift/origin/test/extended/dr/common.go:299]: Unexpected error:
    <*errors.errorString | 0xc0009ed4c0>: {
        s: "failed running \"sudo -i /bin/bash -cx '/usr/local/bin/cluster-restore.sh /home/core/backup'\": <nil> (exit code 1, stderr + /usr/local/bin/cluster-restore.sh /home/core/backup\n)",
    }
    failed running "sudo -i /bin/bash -cx '/usr/local/bin/cluster-restore.sh /home/core/backup'": <nil> (exit code 1, stderr + /usr/local/bin/cluster-restore.sh /home/core/backup
    )
occurred
```

retroflexer:
> @retroflexer can we remove RESTORE_ETCD_POD_YAML, I don't think we need that anymore right?
> 
> https://github.com/openshift/cluster-etcd-operator/blob/68d208baa7fcef428f433d408e2dadd0fc44863f/bindata/etcd/etcd-common-tools#L8

hexfusion:
Based on manual testing results from @retroflexer this is working as expected.
/lgtm

PR #438, opened 2020-09-09
Bug 1877374: Render bootstrap certificates

ironcladlou:
During render, generate the certificates necessary for the bootstrap member.

Update the bootstrap pod to use the rendered certificates.

After this change, the existing asynchronous client/server based bootstrap
certificate generation mechanism should be deleted.

Backport of https://github.com/openshift/cluster-etcd-operator/pull/412

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
@@ -6,58 +6,6 @@ metadata:
   labels:
     k8s-app: etcd
 spec:
-  initContainers:
-  - name: certs
-    image: "{{.Images.KubeClientAgent}}"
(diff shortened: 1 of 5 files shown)


hexfusion:
cc @sdodson this is a really nice stability piece and eliminates etcd-signer from bootstrap I think we have vested interested in this backport

hexfusion:
cc @sdodson

hexfusion:
@bparees can you please also tag https://github.com/openshift/installer/pull/4150 which is also needed for this.

bparees:
we had a surprising number of urgent PRs come in today so i'm making room in 4.5 by untagging this.  it should be CPA'd next week.

hexfusion:
infra
/test e2e-gcp

PR #442, opened 2020-09-11
[release-4.5] Bug 1877930: Use restore pod yaml from the backup when restoring

retroflexer:
Ths is a manual cherry-pick of #436 and an additional commit to include a fallback mechanism to copy restore pod yaml from local disk, if the backup doesn't contain it. 

This is done to support backups that were taken before the change to include restore-pod yaml in the static-pod-resources of etcd.

### DIFF
--- a/bindata/etcd/cluster-restore.sh
+++ b/bindata/etcd/cluster-restore.sh
@@ -7,17 +7,26 @@ set -o pipefail
 set -o errtrace
 
 # example
-# ./cluster-restore.sh $path-to-backup 
+# ./cluster-restore.sh $path-to-backup
 
 if [[ $EUID -ne 0 ]]; then
   echo "This script must be run as root"
   exit 1
 fi
 
-source /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd.env
-source /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd-common-tools
+function source_required_dependency {
+  local path="$1"
+  if [ ! -f "${path}" ]; then
+    echo "required dependencies not found, please ensure this script is run on a node with a functional etcd static pod"
+    exit 1
+  fi
+  # shellcheck disable=SC1090
+  source "${path}"
+}
+source_required_dependency /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd.env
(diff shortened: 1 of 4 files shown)


bindata/etcd/cluster-restore.sh
```diff
@@ -59,11 +59,12 @@ function extract_and_start_restore_etcd_pod() {
    BACKUP_POD_PATH=$(tar -tvf "${BACKUP_FILE}" "*${POD_FILE_NAME}" | awk '{ print $6 }') || true
    if [ -z "${BACKUP_POD_PATH}" ]; then
      echo "${POD_FILE_NAME} does not exist in ${BACKUP_FILE}"
-     exit 1
+     return 1
```
hexfusion, line 62: using `errexit` makes this a net exit 1

retroflexer: No. I tried. It doesn't exit.

hexfusion:
cc @sdodson this bug helps in situations like 4.5 to 4.4 rollback.

bindata/etcd/cluster-restore.sh
```diff
@@ -84,28 +109,31 @@ done
...
 # Copy snapshot to backupdir
-cp -p ${SNAPSHOT_FILE} ${ETCD_DATA_DIR_BACKUP}/snapshot.db
+cp -p "${SNAPSHOT_FILE}" "${ETCD_DATA_DIR_BACKUP}"/snapshot.db
 
-echo "starting restore-etcd static pod"
-cp -p ${RESTORE_ETCD_POD_YAML} ${MANIFEST_DIR}/etcd-pod.yaml
+# Extract and start restore-etcd pod.yaml
+if ! extract_and_start_restore_etcd_pod; then
+  # If backup doesn't have restore-pod.yaml, copy it from local disk
+  cp -p ${RESTORE_ETCD_POD_YAML} ${MANIFEST_DIR}/etcd-pod.yaml
```
hexfusion, line 135: Would the error be clear to the user if this file also didn't exist? Do we need some sanity precheck to ensure the backup is complete?

hexfusion: based on conversation with @retroflexer probability of invalid backup would be low so no need to cover it outside of cp stderr.

retroflexer: I think the error would be self-explanatory. If the backup doesn't have the restore pod, and local disk doesn't have as well, then the customer has some other issues, any way.

markmc:
(patch manager) seems like this addresses an important issue for users, and the risk of regression is mostly constrained to backup/restore itself

PR #444, opened 2020-09-14
Bug 1881082: Expose etcd raft term as a metric

ironcladlou:
Piggyback on the member health checking facility to expose the current
raft term per member as a Prometheus metric to enable more granular
performance analysis.

### DIFF
--- a/pkg/etcdcli/health.go
+++ b/pkg/etcdcli/health.go
@@ -7,9 +7,28 @@ import (
 	"sync"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
 	"go.etcd.io/etcd/etcdserver/etcdserverpb"
+	"k8s.io/component-base/metrics/legacyregistry"
 )
+
+func init() {
+	legacyregistry.RawMustRegister(raftTerms)
+}
+
+const raftTermsMetricName = "etcd_debugging_raft_terms_total"
+
+var raftTerms = &raftTermsCollector{
+	desc: prometheus.NewDesc(
+		raftTermsMetricName,
+		"Number of etcd raft terms as observed by each member.",
+		[]string{"member"},
+		prometheus.Labels{},
+	),
+	terms: map[string]uint64{},
+	lock:  sync.RWMutex{},
+}
 
 type healthCheck struct {
 	Member  *etcdserverpb.Member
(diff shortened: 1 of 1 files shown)


ironcladlou:
Here's what it looks like in a cluster I've got which has experienced massive churn:

```
# HELP etcd_raft_terms_total Number of etcd raft terms as observed by each member.
# TYPE etcd_raft_terms_total counter
etcd_raft_terms_total{member="dmace-71ad-nwdm7-master-0.c.openshift-gce-devel.internal"} 22
etcd_raft_terms_total{member="dmace-71ad-nwdm7-master-1.c.openshift-gce-devel.internal"} 22
etcd_raft_terms_total{member="dmace-71ad-nwdm7-master-2.c.openshift-gce-devel.internal"} 22
```

ironcladlou:
Although I'm not seeing them being scraped by Prometheus in my cluster, so I might be missing something in terms of registration, etc.

/hold

ironcladlou:
Already spent way too much time trying to figure out why modules are screwy in the CI build — switched to the deprecated `RawMustRegister()` function which bypasses the need to import semver since we have no use for the weird k8s metrics wrapper junk anyway.

ironcladlou:
From a cluster-bot cluster running this PR:

Screen Shot 2020-09-15 at 10 59 24 AM

hexfusion:
I think in general this is a net gain in the context of historical reference. Raft terms are persisted to state thus the counter is immutable so to speak vs the current metric. As a followup, I would like to see the term added to events where reasonable as raft terms are included in all API responses thus very available.. Another benefit of adding the term to events is we can track before prom starts.

/lgtm

pkg/etcdcli/health.go
```diff
@@ -7,10 +7,29 @@ import (
...
 	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
 	"go.etcd.io/etcd/etcdserver/etcdserverpb"
+	"k8s.io/component-base/metrics/legacyregistry"
 )
 
+func init() {
+	legacyregistry.RawMustRegister(raftTerms)
+}
+
+const raftTermsMetricName = "etcd_raft_terms_total"
```
hexfusion, line 20: nit: should we rename this?

`etcd_debugging_raft_terms_total`

Per upstream

```
Note that any etcd_debugging_* metrics are experimental and subject to change.
```
This gives us the option to change remove if we feel so fit. I think any new metric should be considered experimental through one release.

ironcladlou: Good call, fixed

PR #445, opened 2020-09-16
Bug 1879637: Remove panic on cache sync

tnozicka:
When caches fail to sync we shouldn't panic but only exit 1.

Fixes a failure like https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-e2e-gcp-4.6/1303842838693285888

Bumping for https://github.com/openshift/library-go/pull/891

/cc @deads2k @hexfusion @mfojtik


vendor/github.com/openshift/library-go/pkg/controller/factory/base_controller.go
```diff
@@ -59,11 +62,38 @@ func (s *scheduledJob) Run() {
...
+		select {
+		case <-ctx.Done():
+			// Exit gracefully because the controller was requested to stop.
+			return
+		default:
+			// If caches did not sync after 10 minutes, it has taken oddly long and
+			// we should provide feedback. Since the control loops will never start,
+			// it is safer to exit with a good message than to continue with a dead loop.
+			// TODO: Consider making this behavior configurable.
+			klog.Exit(err)
```
tnozicka, line 95: the fix is here

hexfusion:
/lgtm
/approve

thanks @tnozicka 

net gain this bumps us level on 1.19

hexfusion:
> Error 400: The number of members in the policy (1,501) is larger than the maximum allowed size 1,500., badRequest

/retest

PR #448, opened 2020-09-23
[release-4.4] Bug 1876484: static-pod-installer: remove deadlock by recreating inst…

soltysh:
…allers that disappear

Picks openshift/library-go#881

/assign @hexfusion 
/hold
Sam the diff looks too big per my recollection, feel free to replace this with a proper bump. Also, I've tried 4.3 backport but that is even worse and has too many conflicts so I think I'll probably leave that up to you. 

/cc @sttts @sdodson


hexfusion:
yeah we bumped to 1.18 to accommodate metrics controller I will handle this bump. Basically, we will need to use the 4.5 pkg bumps

hexfusion:
continued by #450

PR #451, opened 2020-09-28
Bug 1881082: remove erroneously specified label drop rules

ironcladlou:
This patch removes Prometheus label drop rules which were erroneously carried
over from etcd or the apiserver. These rules caused debugging metrics exposed by
the operator to be inadvertently dropped.


hexfusion:
The removed config is essentially cruft from our fork from KAS-O. We should not be exposing any of the operand metrics directly as they scrape it independently and certainly not exposing apiserver metrics.

/lgtm

PR #453, opened 2020-09-28
Defrag etcd on pod restarts

retroflexer:
There is a periodic compaction that is run every 5minutes, but there is no auto defragmentation. This PR adds an init container that attempts to defrag on restarts.

Compacting old revisions internally fragments etcd by leaving gaps in the backend database. Fragmented space is available for use by etcd but unavailable to the host filesystem. In other words, deleting application data does not reclaim the space on disk. Defragmenting upon restart should help reclaim some of the disk space lost.

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -82,6 +82,31 @@ ${COMPUTED_ENV_VARS}
           name: cert-dir
         - mountPath: /usr/local/bin
           name: usr-local-bin
+    - name: defrag
+      image: ${IMAGE}
+      imagePullPolicy: IfNotPresent
+      terminationMessagePolicy: FallbackToLogsOnError
+      resources:
+        requests:
+          memory: 60Mi
+          cpu: 30m
+      volumeMounts:
+      - mountPath: /var/lib/etcd/
+        name: data-dir
+      env:
+${COMPUTED_ENV_VARS}
+      command:
+        - /bin/sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          # Try and ensure etcd is really shut down.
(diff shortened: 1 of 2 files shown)


hexfusion:
continued from https://github.com/openshift/cluster-etcd-operator/pull/383 @retroflexer can you sum briefly how this PR compares to #383

ironcladlou:
Main thing we didn't get around to discussing in enough depth last time: What are the possible failure modes and their impact?

Although I can't think of the problems yet offhand, I'm skeptical there are no downsides here and think we should try and imagine them.

Would it make sense to hold this until master opens for 4.7 and let it soak?

hexfusion:
> Would it make sense to hold this until master opens for 4.7 and let it soak?

Yes this can't land in 4.6 we could consider it as a backport if the demand were strong.

retroflexer:
We can hold this off until 4.7 opens.

Minor differences between 383 and this PR:
1. Defrag is not the first init container (in 383 this was the first init container)
2. Use of `ss` in place of `lsof` for more reliability.

/hold

bindata/etcd/pod.yaml
```diff
@@ -82,6 +82,31 @@ ${COMPUTED_ENV_VARS}
...
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          # Try and ensure etcd is really shut down.
+          timeout 3m /bin/bash -exuo pipefail -c 'while [ -n  "$(ss -Htan '( sport = 2379 or sport = 2380 or sport = 9978 )')" ]; do echo -n "."; sleep 1; done'
+
+          echo "Defragging etcd"
+          time etcdctl defrag --data-dir /var/lib/etcd
```
hexfusion, line 109: etcd would never start you need to make sure that a data-dir exists before you defrag it.

hexfusion: even more you need to make sure it has been intialized

PR #454, opened 2020-09-28
Bug 1883610: Use more reliable ss in place of lsof

retroflexer:
`lsof` doesn't output ports in TIME_WAIT but `ss` does. Therefore, it is more reliable to use `ss` to detect the port condition.

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -142,7 +142,7 @@ ${COMPUTED_ENV_VARS}
         # we cannot use the "normal" port conflict initcontainer because when we upgrade, the existing static pod will never yield,
         # so we do the detection in etcd container itsefl.
         echo -n "Waiting for ports 2379, 2380 and 9978 to be released."
-        while [ -n "$(lsof -ni :2379)$(lsof -ni :2380)$(lsof -ni :9978)" ]; do
+        while [ -n "$(ss -Htan '( sport = 2379 or sport = 2380 or sport = 9978 )')" ]; do
           echo -n "."
           sleep 1
         done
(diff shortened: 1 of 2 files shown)


hexfusion:
Original conversations about `ss` vs `lsof` take place here https://github.com/openshift/cluster-kube-apiserver-operator/pull/864

bindata/etcd/pod.yaml
```diff
@@ -142,7 +142,7 @@ ${COMPUTED_ENV_VARS}
         # we cannot use the "normal" port conflict initcontainer because when we upgrade, the existing static pod will never yield,
         # so we do the detection in etcd container itsefl.
         echo -n "Waiting for ports 2379, 2380 and 9978 to be released."
-        while [ -n "$(lsof -ni :2379)$(lsof -ni :2380)$(lsof -ni :9978)" ]; do
+        while [ -n "$(ss -Htanop \( sport = 2379 \))$(ss -Htanop \( sport = 2380 \))$(ss -Htanop \( sport = 9978 \))" ]; do
```
hexfusion, line 145: `-o` and `-p` seems to additional show meta about the process but I am curious if you felt it added something to the check? Also it seems that the multi-port usage can be a little more readable using an `or` statement.

$(ss -Htan '( sport = 2379 or sport = 2380 or sport = 9978 )') ref [1]

[1] https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/v4.1.0/kube-apiserver/pod.yaml#L27

hexfusion:
/lgtm
@retroflexer this needs a bug.

PR #456, opened 2020-09-29
Bug 1871170: test/e2e: add etcdctl sanity test

hexfusion:
let's not get bugs because we are not testing simple things. this test exposes a few bugs and should keep us honest moving forward.


test/e2e/etcdctl_test.go
```diff
@@ -0,0 +1,125 @@
...
+		{"etcdctl lock", true, "distributed locks should not be done via etcdctl", ""},
+		{"etcdctl make-mirror", true, "", "make-mirror takes one destination argument"},
+
+		// member
+		{"etcdctl member list", false, "", ""},
+		{"etcdctl member add tester", false, "", "member peer urls not provided"},
+		{"etcdctl member promote 8e9e05c52164694d", false, "", "etcdserver: member not found"},
+		{"etcdctl member remove 8e9e05c52164694d", false, "", "etcdserver: member not found"},
+		{"etcdctl member update 8e9e05c52164694d", false, "", "member peer urls not provided"},
+		{"etcdctl member update 8e9e05c52164694d", false, "", "member peer urls not provided"},
```
hexfusion, line 59: dupe

test/e2e/etcdctl_test.go
```diff
@@ -0,0 +1,116 @@
...
+		skip          bool
+		skipReason    string
+		expectedError string
+	}{
+		{"etcdctl alarm disarm", false, "", ""},
+		{"etcdctl alarm list", false, "", ""},
+		{"etcdctl del --prefix /etcdctl-check-perf/", false, "", ""},
+		{"etcdctl check perf", false, "", ""},
+		{"etcdctl compaction 0", false, "", "Error: etcdserver: mvcc: required revision has been compacted"},
+		{"etcdctl defrag", false, "", ""},
```
ironcladlou, line 36: Is this potentially disruptive?

test/e2e/etcdctl_test.go
```diff
@@ -0,0 +1,116 @@
...
+		{"etcdctl member list", false, "", ""},
+		{"etcdctl member add tester", false, "", "member peer urls not provided"},
+		{"etcdctl member promote 8e9e05c52164694d", false, "", "etcdserver: member not found"},
+		{"etcdctl member remove 8e9e05c52164694d", false, "", "etcdserver: member not found"},
+		{"etcdctl member update 8e9e05c52164694d", false, "", "member peer urls not provided"},
+		{"etcdctl move-leader 8e9e05c52164694d", false, "", ""},
+		{"etcdctl put /etcdctl-check-perf/foo bar", false, "", ""},
+		{"etcdctl snapshot save ./backup.db", false, "", ""},
+		{"etcdctl snapshot status ./backup.db", false, "", ""},
+		{"etcdctl snapshot restore ./backup.db", false, "", ""},
```
ironcladlou, line 59: Are these potentially disruptive?

test/e2e/etcdctl_test.go
```diff
@@ -0,0 +1,116 @@
...
+			if tc.skip {
+				t.Skip(tc.skipReason)
+			}
+
+			actual, err := exec.Command("oc", getOcExecArgs(tc.command, podName)...).CombinedOutput()
+			switch {
+			case err == nil && len(tc.expectedError) == 0:
+				//pass
+			case err == nil && len(tc.expectedError) != 0:
+				require.NoError(t, fmt.Errorf("expected error not reported: %q", tc.expectedError))
```
ironcladlou, line 95: Not familiar with this lib and it looks like you're introducing a parallel/competing assertion framework — I expected to see Ginkgo like all the other tests, e.g. `o.Expect(err).NotTo(o.HaveOccurred())`

test/e2e/etcdctl_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/openshift/cluster-etcd-operator/test/e2e/framework"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TestEtcdctlCommands executes all known etcdctl commands inside of the etcdctl container.
+// The test is not intended to be a functional test yet a sanity test that the container
+// ENV is populated correctly and that etcdctl consumes that ENV properly.
+func TestEtcdctlCommands(t *testing.T) {
```
ironcladlou, line 18: Although I personally hate Ginkgo, I expected this to be part of an exiting Ginkgo suite using the usual test framework for consistency and integration with the overall e2e test scaffolding

test/e2e/etcdctl_test.go
hexfusion: it could be now but it should run only against localhost in which case it should not.

test/e2e/etcdctl_test.go
hexfusion: snapshot save could have overhead but it should not be disruptive. the rest would not have disruptive cost.

test/e2e/etcdctl_test.go
hexfusion: I see Ginko used in origin but I don't see it in operator e2e testing feels a little like overkill for these use cases but we can look into it.

test/e2e/etcdctl_test.go
hexfusion: yeah I mean if this were origin I would 100% agree its just not a standard pattern across any of the control-plane operators. In fact the usage across the org is almost nil. the regular testing frameworks is light and saves a pretty large dependency. I am open to considering it.

https://github.com/search?q=org%3Aopenshift+ginko&type=code

test/e2e/etcdctl_test.go
hexfusion: the fact that it runs against all members now is a bug we are looking at now.

test/e2e/etcdctl_test.go
hexfusion: Well if I spelled it right would help... so the usage is a bit more than I expected.

https://github.com/search?q=org%3Aopenshift+ginkgo&type=code

test/e2e/etcdctl_test.go
hexfusion: testify does have heavy usage in the org FTR https://github.com/search?q=org%3Aopenshift+testify&type=code

test/e2e/etcdctl_test.go
ironcladlou: Apologies... I think when I reviewed this before I was totally spaced out and confused — I thought I was looking at a test in the origin repo and didn't realize we were talking about etcd-operator e2e tests. Nevermind everything I said re: Ginkgo, sorry for the confusion.

hexfusion: no worries I know you like to standardize stuff :+1: maybe it should be? I think if operator e2e were to evolve maybe it would need that framework. Let's keep it in mind.

ironcladlou:
/retitle Bug 1871170: test/e2e: add etcdctl sanity test

ironcladlou:
I'm not really persuaded of the value of introducing `testify` in this PR:

```go
switch {
case err != nil && len(tc.expectedError) != 0 && !strings.Contains(string(actual), tc.expectedError):
  require.NoError(t, fmt.Errorf("unexpected error: %q", actual))
}
```

vs

```go
switch {
case err != nil && len(tc.expectedError) != 0 && !strings.Contains(string(actual), tc.expectedError):
  t.Fatalf(fmt.Errorf("unexpected error: %w", actual))
}
```

More indirection, requires learning another library, not sure what it adds.

hexfusion:
> I'm not really persuaded of the value of introducing `testify` in this PR:
> 
> ```go
> switch {
> case err != nil && len(tc.expectedError) != 0 && !strings.Contains(string(actual), tc.expectedError):
>   require.NoError(t, fmt.Errorf("unexpected error: %q", actual))
> }
> ```
> 
> vs
> 
> ```go
> switch {
> case err != nil && len(tc.expectedError) != 0 && !strings.Contains(string(actual), tc.expectedError):
>   t.Fatalf(fmt.Errorf("unexpected error: %w", actual))
> }
> ```
> 
> More indirection, requires learning another library, not sure what it adds.

I was hoping we could move to this as a standard for operator e2e. If the standard is adhered to it would give a clean pattern for  future testing. Currently, I feel like we hand roll a lot of what testify provides. But I am open to other thoughts, libraries for this use. 

cc @marun

marun:
> I was hoping we could move to this as a standard for operator e2e. If the standard is adhered to it would give a clean pattern for future testing. Currently, I feel like we hand roll a lot of what testify provides. But I am open to other thoughts, libraries for this use.
> 
> cc @marun

I'm a big +1 for testify. It's already used in some group b operators like ckao as well as in upstream kube. It provides most of the benefits of ginkgo's helpers without incurring declarative/imperative confusion.

Full benefit, though, often requires a rewrite:

```golang
if len(tc.expectedError) > 0 {
  require.Contains(string(actual), tc.expectedError, "expected error not reported: %q", tc.expectedError)
else {
  require.NoError(t, actual)
}
```

hexfusion:
updated @marun PTAL

PR #457, opened 2020-09-30
Bug 1883772: pkg/operator/clustermembercontroller: resync every minute

hexfusion:
In certain circumstances such as transient client failure followed by the operator process being killed (leader election lost) the operator can get in a situation where we forget to retry scaling members.

In this case we see the caches sync but scaling was not retried.

```
I0929 07:50:39.362608       1 shared_informer.go:223] Waiting for caches to sync for ClusterMemberController
I0929 07:50:40.862752       1 shared_informer.go:230] Caches are synced for ClusterMemberController
```

This PR ensures that we revisit membership in the case of a leaked event or unexpected cache to make sure we don't have any further work to do.

### DIFF
--- a/pkg/operator/clustermembercontroller/clustermembercontroller.go
+++ b/pkg/operator/clustermembercontroller/clustermembercontroller.go
@@ -56,7 +56,7 @@ func NewClusterMemberController(
 		kubeInformers.InformersFor("").Core().V1().Nodes().Informer(),
 		networkInformer.Informer(),
 		operatorClient.Informer(),
-	).WithSync(c.sync).ToController("ClusterMemberController", eventRecorder.WithComponentSuffix("cluster-member-controller"))
+	).WithSync(c.sync).ResyncEvery(time.Minute).ToController("ClusterMemberController", eventRecorder.WithComponentSuffix("cluster-member-controller"))
 }
 
 func (c *ClusterMemberController) sync(ctx context.Context, syncCtx factory.SyncContext) error {


ironcladlou:
Seems harmless to resync periodically but I'm curious about what's going on here. Are you saying the operator remains alive but has missed some edge event that should trigger a sync but doesn't? If the process is killed and restarted I'd expect a resync. But are you saying the process is killed and not restarted, and having missed the event we never catch up with the missed event?

hexfusion:
> Seems harmless to resync periodically but I'm curious about what's going on here. Are you saying the operator remains alive but has missed some edge event that should trigger a sync but doesn't? If the process is killed and restarted I'd expect a resync. But are you saying the process is killed and not restarted, and having missed the event we never catch up with the missed event?

I am saying the operator process is killed and new process is started but it appears we leaked an event and then did not retry.

ironcladlou:
> > Seems harmless to resync periodically but I'm curious about what's going on here. Are you saying the operator remains alive but has missed some edge event that should trigger a sync but doesn't? If the process is killed and restarted I'd expect a resync. But are you saying the process is killed and not restarted, and having missed the event we never catch up with the missed event?
> 
> I am saying the operator process is killed and new process is started but it appears we leaked an event and then did not retry.

Hm, wouldn't the resync happen on startup then? The logic should be level driven, after all

hexfusion:
> Hm, wouldn't the resync happen on startup then? The logic should be level driven, after all
It should yes and in this case it did. 

>  I0929 07:50:39.362608       1 shared_informer.go:223] Waiting for caches to sync for ClusterMemberController
>  I0929 07:50:40.862752       1 shared_informer.go:230] Caches are synced for ClusterMemberController

But we should consider the leak a followup?

ironcladlou:
> > Hm, wouldn't the resync happen on startup then? The logic should be level driven, after all
> 
> It should yes and in this case it did.
> 
> ```
> I0929 07:50:39.362608       1 shared_informer.go:223] Waiting for caches to sync for ClusterMemberController
> I0929 07:50:40.862752       1 shared_informer.go:230] Caches are synced for ClusterMemberController
> ```
> 
> But we should consider the leak a followup?

Are you saying that had process actively resynced every minute, the issue would have resolved before the process ended up killed? Not sure of the timing details here (e.g. how long did it take before the process was nuked at which point the immediate resync fixed stuff)

hexfusion:
> Are you saying that had process actively resynced every minute, the issue would have resolved before the process ended up killed? Not sure of the timing details here (e.g. how long did it take before the process was nuked at which point the immediate resync fixed stuff)

Well I am saying once the event was leaked the controller would not resync. The net result was the cluster failed and never scaled this member.

timing

#### original operator process failure.
>   W0929 07:50:37.860383       1 leaderelection.go:69] leader election lost

#### new operator process few seconds later
```
I0929 07:50:39.362608       1 shared_informer.go:223] Waiting for caches to sync for ClusterMemberController
I0929 07:50:40.862752       1 shared_informer.go:230] Caches are synced for ClusterMemberController
```
### scale other two members
```
425:I0929 07:50:57.371684       1 event.go:278] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"ab4b54e4-a684-4248-96fb-d4cd458f490a", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'MemberAdd' adding new peer https://192.168.1.11:2380
1439:I0929 07:52:53.048252       1 event.go:278] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"ab4b54e4-a684-4248-96fb-d4cd458f490a", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'MemberAdd' adding new peer https://192.168.1.29:2380
```

cache is synced but no event was triggered to attempt scale-up of 3rd. So it seems we leaked one?

PR #459, opened 2020-09-30
*: enable pprof profiler for operator

hexfusion:
enable profiler by default so we can dig deeper into performance improvements with the operator itself without any change to running operator spec or log level. pprof has no perf impact when not used so I don't see any downside to enabling by default.

usage example

```
oc rsh -n openshift-etcd-operator etcd-operator-5d46b88565-lfwdh curl --output - http://127.0.0.1:6060/debug/pprof/heap > heap
```


hexfusion:
need to validate this works has not perf implications (shouldn't) etc/

hexfusion:
I think we want to leave this until 4.7 is cut.

PR #461, opened 2020-10-02
Bug 1882176: Validate certs for the current IP address of the node

retroflexer:
Currently cluster-etcd-operator only checks the existence of the certificates, but doesn't verify the SAN configured in the certs against the current node IP address (hostname).

This PR fixes the problem by parsing the certificate to validate the SAN. If there is a SAN mismatch, it generates a new set of certificates.

### DIFF
--- a/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
+++ b/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
@@ -3,15 +3,18 @@ package etcdcertsigner
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/kubernetes"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/cert"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
@@ -99,10 +102,22 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
 
 	errs := []error{}
 	for _, node := range nodes {
-		if err := c.createSecretForNode(node, recorder); err != nil {
+		nodeInternalIPs, err := dnshelpers.GetInternalIPAddressesForNodeName(node)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		err = c.ensureAndValidateSecrets(node, nodeInternalIPs, recorder)
+		if errors.IsNotFound(err) {
+			// if any of the secrets are not present, create all secrets for the node
+			err = c.createSecretsForNode(node, nodeInternalIPs, recorder)
+		}
+
+		if err != nil {
 			errs = append(errs, err)
 		}
 	}
+
 	if len(errs) > 0 {
 		return utilerrors.NewAggregate(errs)
 	}
@@ -178,24 +193,59 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
 	return utilerrors.NewAggregate(errs)
 }
 
-func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node, recorder events.Recorder) error {
-	etcdPeerClientCertName := tlshelpers.GetPeerClientSecretNameForNode(node.Name)
-	etcdServingCertName := tlshelpers.GetServingSecretNameForNode(node.Name)
-	metricsServingCertName := tlshelpers.GetServingMetricsSecretNameForNode(node.Name)
+// validateCertificateSAN makes sure that a given certificate is valid
+// at least for the SANs defined in the configuration.
+// NOTE: certName argument is used only for forming the error message
+func validateCertificateSAN(certData map[string][]byte, ipAddresses []string, certName string) error {
+	certificates, err := cert.ParseCertsPEM(certData["tls.crt"])
+	if err != nil {
+		return fmt.Errorf("could not parse TLS cert for %s: %w", certName, err)
+	}
+	for _, ipAddress := range ipAddresses {
+		if err := certificates[0].VerifyHostname(ipAddress); err != nil {
+			return fmt.Errorf("SAN for the certificate %s does not include %s: %w", certName, ipAddress, err)
+		}
+	}
+	return nil
+}
 
-	var err error
-	_, err = c.secretLister.Secrets(operatorclient.TargetNamespace).Get(etcdPeerClientCertName)
-	peerClientCertOk := err == nil
-	_, err = c.secretLister.Secrets(operatorclient.TargetNamespace).Get(etcdServingCertName)
-	servingCertOk := err == nil
-	_, err = c.secretLister.Secrets(operatorclient.TargetNamespace).Get(metricsServingCertName)
-	metricsCertOk := err == nil
+// Ensure that all secrets exist for the node and validate that the IP addresses are correct.
+func (c *EtcdCertSignerController) ensureAndValidateSecrets(node *corev1.Node, nodeInternalIPs []string, recorder events.Recorder) error {
+	etcdPeerClientSecretName := tlshelpers.GetPeerClientSecretNameForNode(node.Name)
+	etcdServingSecretName := tlshelpers.GetServingSecretNameForNode(node.Name)
+	metricsServingSecretName := tlshelpers.GetServingMetricsSecretNameForNode(node.Name)
 
-	// if we have all the certs we want, do nothing.
-	if peerClientCertOk && servingCertOk && metricsCertOk {
-		return nil
+	secretNames := []string{etcdPeerClientSecretName, etcdServingSecretName, metricsServingSecretName}
+
+	errs := []error{}
+	for _, secretName := range secretNames {
+		secret, err := c.secretLister.Secrets(operatorclient.TargetNamespace).Get(secretName)
+		if errors.IsNotFound(err) {
+			return err
+		}
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if secret.Data == nil {
+			errs = append(errs, fmt.Errorf("secret data not found in %s", secretName))
+			continue
+		}
+		if err := validateCertificateSAN(secret.Data, nodeInternalIPs, secretName); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) > 0 {
+		return utilerrors.NewAggregate(errs)
 	}
 
+	return nil
+}
+
+func (c *EtcdCertSignerController) createSecretsForNode(node *corev1.Node, nodeInternalIPs []string, recorder events.Recorder) error {
+	etcdPeerClientSecretName := tlshelpers.GetPeerClientSecretNameForNode(node.Name)
+	etcdServingSecretName := tlshelpers.GetServingSecretNameForNode(node.Name)
+	metricsServingSecretName := tlshelpers.GetServingMetricsSecretNameForNode(node.Name)
 	// get the signers
 	etcdCASecret, err := c.secretLister.Secrets(operatorclient.GlobalUserSpecifiedConfigNamespace).Get("etcd-signer")
 	if err != nil {
@@ -206,25 +256,30 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node, record
 		return err
 	}
 
-	nodeInternalIPs, err := dnshelpers.GetInternalIPAddressesForNodeName(node)
+	// create the certificates and update them in the API
+	pCert, pKey, err := tlshelpers.CreatePeerCertKey(etcdCASecret.Data["tls.crt"], etcdCASecret.Data["tls.key"], nodeInternalIPs)
 	if err != nil {
 		return err
 	}
 
-	// create the certificates and update them in the API
-	pCert, pKey, err := tlshelpers.CreatePeerCertKey(etcdCASecret.Data["tls.crt"], etcdCASecret.Data["tls.key"], nodeInternalIPs)
-	err = c.createSecret(etcdPeerClientCertName, pCert, pKey, recorder)
-	if err != nil {
+	if err := c.createSecret(etcdPeerClientSecretName, pCert, pKey, recorder); err != nil {
 		return err
 	}
 	sCert, sKey, err := tlshelpers.CreateServerCertKey(etcdCASecret.Data["tls.crt"], etcdCASecret.Data["tls.key"], nodeInternalIPs)
-	err = c.createSecret(etcdServingCertName, sCert, sKey, recorder)
 	if err != nil {
 		return err
 	}
+
+	if err := c.createSecret(etcdServingSecretName, sCert, sKey, recorder); err != nil {
+		return err
+	}
+
 	metricCert, metricKey, err := tlshelpers.CreateMetricCertKey(etcdMetricCASecret.Data["tls.crt"], etcdMetricCASecret.Data["tls.key"], nodeInternalIPs)
-	err = c.createSecret(metricsServingCertName, metricCert, metricKey, recorder)
 	if err != nil {
+		return err
+	}
+
+	if err := c.createSecret(metricsServingSecretName, metricCert, metricKey, recorder); err != nil {
 		return err
 	}
 
--- a/pkg/operator/etcdcertsigner/etcdcertsignercontroller_test.go
+++ b/pkg/operator/etcdcertsigner/etcdcertsignercontroller_test.go
@@ -1,0 +1,300 @@
+package etcdcertsigner
+
+import (
+	"bytes"
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha1"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
+	"github.com/openshift/library-go/pkg/crypto"
+	"math/big"
+	"testing"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
+	"go.etcd.io/etcd/pkg/mock/mockserver"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	clientgotesting "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
+	u "github.com/openshift/cluster-etcd-operator/pkg/testutils"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+var (
+	rootcaPublicKey  *rsa.PublicKey
+	rootcaPrivateKey *rsa.PrivateKey
+	publicKeyHash    []byte
+	DummyPrivateKey  []byte
+	DummyCertificate []byte
+)
+
+func initialize(t *testing.T) {
+
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+	rootcaPublicKey = &privateKey.PublicKey
+	rootcaPrivateKey = privateKey
+	if err == nil {
+		hash := sha1.New()
+		hash.Write(rootcaPublicKey.N.Bytes())
+		publicKeyHash = hash.Sum(nil)
+	}
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+	ca := &x509.Certificate{
+		SerialNumber: big.NewInt(2020),
+		Subject: pkix.Name{
+			Organization: []string{"Red Hat"},
+			Country:      []string{"US"},
+			Province:     []string{""},
+			Locality:     []string{"Raleigh"},
+		},
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(10, 0, 0),
+		IsCA:                  true,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		BasicConstraintsValid: true,
+	}
+	caBytes, err := x509.CreateCertificate(rand.Reader, ca, ca, rootcaPublicKey, rootcaPrivateKey)
+	if err != nil {
+		t.Fatal(err)
+	}
+	caPEM := new(bytes.Buffer)
+	pem.Encode(caPEM, &pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: caBytes,
+	})
+	caPrivKeyPEM := new(bytes.Buffer)
+	pem.Encode(caPrivKeyPEM, &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(rootcaPrivateKey),
+	})
+	DummyCertificate = caPEM.Bytes()
+	DummyPrivateKey = caPrivKeyPEM.Bytes()
+
+}
+
+func TestValidateSAN(t *testing.T) {
+	initialize(t)
+	mockEtcd, err := mockserver.StartMockServers(3)
+	if err != nil {
+		t.Fatalf("failed to start mock servers: %s", err)
+	}
+	defer mockEtcd.Stop()
+	scenarios := []struct {
+		name            string
+		objects         []runtime.Object
+		staticPodStatus *operatorv1.StaticPodOperatorStatus
+		wantErr         bool
+		validateFunc    func(ts *testing.T, actions []clientgotesting.Action)
+	}{
+		{
+			// All secrets match the node and IP address configuration.
+			name: "NormalSteadyState",
+			objects: []runtime.Object{
+				u.FakeNode("master-0", u.WithMasterLabel(), u.WithNodeInternalIP("10.0.0.1")),
+				u.FakeNode("master-1", u.WithMasterLabel(), u.WithNodeInternalIP("10.0.0.2")),
+				u.FakeNode("master-2", u.WithMasterLabel(), u.WithNodeInternalIP("10.0.0.3")),
+				u.EndpointsConfigMap(
+					u.WithAddress("10.0.0.1"),
+					u.WithAddress("10.0.0.2"),
+					u.WithAddress("10.0.0.3"),
+				),
+				&corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: "openshift-config",
+						Name:      "etcd-signer",
+					},
+					Data: map[string][]byte{
+						"tls.crt": DummyCertificate,
+						"tls.key": DummyPrivateKey,
+					},
+				},
+
+				&corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: "openshift-config",
+						Name:      "etcd-metric-signer",
+					},
+					Data: map[string][]byte{
+						"tls.crt": DummyCertificate,
+						"tls.key": DummyPrivateKey,
+					},
+				},
+				u.FakeSecret("openshift-etcd", "etcd-peer-master-0", makeCerts(t, []string{"localhost", "10.0.0.1"})),
+				u.FakeSecret("openshift-etcd", "etcd-peer-master-1", makeCerts(t, []string{"localhost", "10.0.0.2"})),
+				u.FakeSecret("openshift-etcd", "etcd-peer-master-2", makeCerts(t, []string{"localhost", "10.0.0.3"})),
+
+				u.FakeSecret("openshift-etcd", "etcd-serving-master-0", makeCerts(t, []string{"localhost", "10.0.0.1"})),
+				u.FakeSecret("openshift-etcd", "etcd-serving-master-1", makeCerts(t, []string{"localhost", "10.0.0.2"})),
+				u.FakeSecret("openshift-etcd", "etcd-serving-master-2", makeCerts(t, []string{"localhost", "10.0.0.3"})),
+
+				u.FakeSecret("openshift-etcd", "etcd-serving-metrics-master-0", makeCerts(t, []string{"localhost", "10.0.0.1"})),
+				u.FakeSecret("openshift-etcd", "etcd-serving-metrics-master-1", makeCerts(t, []string{"localhost", "10.0.0.2"})),
+				u.FakeSecret("openshift-etcd", "etcd-serving-metrics-master-2", makeCerts(t, []string{"localhost", "10.0.0.3"})),
+			},
+			staticPodStatus: u.StaticPodOperatorStatus(
+				u.WithLatestRevision(3),
+				u.WithNodeStatusAtCurrentRevision(3),
+				u.WithNodeStatusAtCurrentRevision(3),
+				u.WithNodeStatusAtCurrentRevision(3),
+			),
+			wantErr: false,
+			validateFunc: func(ts *testing.T, actions []clientgotesting.Action) {
+				for _, action := range actions {
+					if action.Matches("update", "secrets") {
+						updateAction := action.(clientgotesting.UpdateAction)
+						actual := updateAction.GetObject().(*corev1.Secret)
+						ts.Errorf("unexpected secret update: %#v", actual)
+					}
+				}
+			},
+		},
+
+		{
+			// IP addresses do not match the SAN configured in the certificates.
+			// master-2 has a new IP addr (10.0.0.33) but certs still have old SAN (10.0.0.3).
+			name: "IPMismatch",
+			objects: []runtime.Object{
+				u.FakeNode("master-0", u.WithMasterLabel(), u.WithNodeInternalIP("10.0.0.1")),
+				u.FakeNode("master-1", u.WithMasterLabel(), u.WithNodeInternalIP("10.0.0.2")),
+				u.FakeNode("master-2", u.WithMasterLabel(), u.WithNodeInternalIP("10.0.0.33")),
+				u.EndpointsConfigMap(
+					u.WithAddress("10.0.0.1"),
+					u.WithAddress("10.0.0.2"),
+					u.WithAddress("10.0.0.33"),
+				),
+				&corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: "openshift-config",
+						Name:      "etcd-signer",
+					},
+					Data: map[string][]byte{
+						"tls.crt": DummyCertificate,
+						"tls.key": DummyPrivateKey,
+					},
+				},
+
+				&corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: "openshift-config",
+						Name:      "etcd-metric-signer",
+					},
+					Data: map[string][]byte{
+						"tls.crt": DummyCertificate,
+						"tls.key": DummyPrivateKey,
+					},
+				},
+				u.FakeSecret("openshift-etcd", "etcd-peer-master-0", makeCerts(t, []string{"localhost", "10.0.0.1"})),
+				u.FakeSecret("openshift-etcd", "etcd-peer-master-1", makeCerts(t, []string{"localhost", "10.0.0.2"})),
+				u.FakeSecret("openshift-etcd", "etcd-peer-master-2", makeCerts(t, []string{"localhost", "10.0.0.3"})),
+
+				u.FakeSecret("openshift-etcd", "etcd-serving-master-0", makeCerts(t, []string{"localhost", "10.0.0.1"})),
+				u.FakeSecret("openshift-etcd", "etcd-serving-master-1", makeCerts(t, []string{"localhost", "10.0.0.2"})),
+				u.FakeSecret("openshift-etcd", "etcd-serving-master-2", makeCerts(t, []string{"localhost", "10.0.0.3"})),
+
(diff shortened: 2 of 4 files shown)


pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,17 +180,56 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+	tlsCrt, ok := secret.Data["tls.crt"]
+	if !ok {
+		return fmt.Errorf("TLS cert data not found in %s", certName)
+	}
+	if len(tlsCrt) == 0 {
+		return fmt.Errorf("%s MISSING tls.crt content\n", certName)
+	}
+	certificates, err := cert.ParseCertsPEM([]byte(tlsCrt))
+	if err != nil {
+		return fmt.Errorf("Could not parse TLS cert for %s: %w", certName, err)
```
hexfusion, line 202: nit: we generally follow the go convention[1], error messages that should not be capitalized.

[1]https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,21 +181,60 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+		return fmt.Errorf("secret data not found in %s", certName)
+	}
+	tlsCrt, ok := secret.Data["tls.crt"]
+	if !ok {
+		return fmt.Errorf("tls cert data not found in %s", certName)
+	}
+	if len(tlsCrt) == 0 {
+		return fmt.Errorf("%s missing tls.crt content\n", certName)
+	}
+	certificates, err := cert.ParseCertsPEM([]byte(tlsCrt))
```
hexfusion, line 201: I think you could simplify this a bit by just passing `secret.Data` directly to `cert.ParseCertsPEM`. The function handles the error checks you manually added above by looking at the code[1].

[1]https://github.com/kubernetes/client-go/blob/master/util/cert/pem.go#L35

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,21 +181,60 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+	if err != nil {
+		return fmt.Errorf("could not parse TLS cert for %s: %w", certName, err)
+	}
+	// Using the first certificate to verify IP address match
+	certificate := certificates[0]
+	for _, ipAddress := range ipAddresses {
+		if err := certificate.VerifyHostname(ipAddress); err != nil {
+			return fmt.Errorf("certificate is invalid for %s, %w", ipAddress, err)
```
hexfusion, line 209: can this error be more clear as to why?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,21 +181,60 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
-	servingCertOk := err == nil
-	_, err = c.secretLister.Secrets(operatorclient.TargetNamespace).Get(metricsServingCertName)
-	metricsCertOk := err == nil
+
+	nodeInternalIPs, err := dnshelpers.GetInternalIPAddressesForNodeName(node)
+	if err != nil {
+		return err
+	}
+
+	certsOkay := true
```
hexfusion, line 227: nit `isCertsValid`

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,21 +181,60 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+	for _, certName := range []string{etcdPeerClientCertName, etcdServingCertName, metricsServingCertName} {
+		if err := c.validateCertificateSAN(certName, nodeInternalIPs); err != nil {
```
hexfusion, line 229: feels like this check should be part of `validateCertificateSAN` methos.

hexfusion: disregard I see it is calling it..

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,21 +181,60 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+			klog.Warningf("validateCertificateSAN: %v", err)
```
hexfusion, line 230: isnt this an error?

retroflexer: We were not printing anything in the past when we regenerated certs :-). But I will make it Errorf.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
retroflexer: Sure. `areCertsValid` perhaps.

retroflexer: Done.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,21 +181,53 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+	}
+	return nil
+}
+
 func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node, recorder events.Recorder) error {
 	etcdPeerClientCertName := tlshelpers.GetPeerClientSecretNameForNode(node.Name)
 	etcdServingCertName := tlshelpers.GetServingSecretNameForNode(node.Name)
 	metricsServingCertName := tlshelpers.GetServingMetricsSecretNameForNode(node.Name)
 
 	var err error
```
hexfusion, line 213: is this useful? why not be explicit about `err` scope. seems like some stuff can get cleaned up below

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,21 +181,53 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
-	servingCertOk := err == nil
-	_, err = c.secretLister.Secrets(operatorclient.TargetNamespace).Get(metricsServingCertName)
-	metricsCertOk := err == nil
+
+	nodeInternalIPs, err := dnshelpers.GetInternalIPAddressesForNodeName(node)
+	if err != nil {
+		return err
+	}
+
+	isCertsOk := true
```
hexfusion, line 220: hmm usually, bool default is false feels weird to set it true as a default. is the failure case defined by the existence of err right

I think you could do.

errs := []error{}
if err != nil
errs = append(errs, err)

if len(errs) > 0 ...

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -206,11 +241,6 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node, record
 		return err
 	}
 
-	nodeInternalIPs, err := dnshelpers.GetInternalIPAddressesForNodeName(node)
-	if err != nil {
-		return err
-	}
-
 	// create the certificates and update them in the API
 	pCert, pKey, err := tlshelpers.CreatePeerCertKey(etcdCASecret.Data["tls.crt"], etcdCASecret.Data["tls.key"], nodeInternalIPs)
```
hexfusion, line 245: I know this is not your code but what happens to this err?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -206,11 +241,6 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node, record
...
 	err = c.createSecret(etcdPeerClientCertName, pCert, pKey, recorder)
```
hexfusion, line 246: why not local scope?

if err := ...

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: if we dont need it or care about it why do we define it?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +181,69 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
 	return utilerrors.NewAggregate(errs)
 }
 
+// validateCertificateSAN makes sure that a given certificate is valid
```
hexfusion, line 184: nit: this is meant for the function below

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +181,69 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+	}
 
 	// if we have all the certs we want, do nothing.
-	if peerClientCertOk && servingCertOk && metricsCertOk {
+	if len(errors) == 0 {
 		return nil
 	}
 
+	_, _, updateErr := v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+		Type:    "EtcdCertSignerControllerDegraded",
```
hexfusion, line 238: The general pattern is to return error to sync resulting in status change or we will have multiple updates on a single sync.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -206,29 +254,41 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node, record
...
+	}
+
+	_, _, updateErr = v1helpers.UpdateStatus(c.operatorClient,
+		v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:   "EtcdCertSignerControllerDegraded",
+			Status: operatorv1.ConditionFalse,
+			Reason: "AsExpected",
+		}))
+
+	return updateErr
```
hexfusion, line 291: same as above status is set in sync based on the result of the sync function syncAllMasters

hexfusion (commented):
@retroflexer thanks for the update, few comments below.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,21 +181,56 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+	if err != nil {
+		return err
+	}
+
+	errors := []error{}
+	for _, certName := range []string{etcdPeerClientCertName, etcdServingCertName, metricsServingCertName} {
+		if err := c.validateCertificateData(certName, nodeInternalIPs); err != nil {
+			klog.Errorf("validateCertificateSAN: %v", err)
+			errors = append(errors, err)
+			break
```
retroflexer, line 228: If we want to aggregate all the SAN mismatches, then, perhaps, this `break` here is not desirable.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -206,29 +244,35 @@ func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node, record
...
 		return err
 	}
 
-	return nil
+	if err := c.createSecret(metricsServingCertName, metricCert, metricKey, recorder); err != nil {
+		return err
+	}
+
+	// Return error to flag the mismatch of SAN to raise a degraded condition.
+	return utilerrors.NewAggregate(errors)
```
hexfusion, line 275: not sure how comfortable I am with rolling new certificates by default in the face of SAN mismatch. Too many corner cases we are not covering.

Scenario 1. If cluster upgrades and on reboot node receives a new IP address.
Scenario 2. Machine is deleted and new machine with same name is created with new IP.

The first we would want to inform the user of the issue so they could resolve it. Would this change solve that problem? If we tolerate the IP change which could have implications on other components. Regardless we fail because membership was not changed to reflect new peerURLs.

Given the first problem, I am not so sure we can reasonably conclude that we should rotate certs with the information that this controller has.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: makes sense

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
retroflexer: Pushed another update. The new code does the follow:

1. If any of the secrets are not found, generate new secrets
2. If all the secrets are found, but do not match SAN, then, raise the error but DO NOT generate new certs.
3. If all the secrets are found, and match the SAN, then, do nothing.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,21 +181,64 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+			secrets = append(secrets, secret)
+		}
+	}
 
-	// if we have all the certs we want, do nothing.
-	if peerClientCertOk && servingCertOk && metricsCertOk {
+	// If we found all the secrets, match the IP SAN.
+	if len(errors) == 0 {
+		sanMismatchErrs := []error{}
+		for index, certName := range certNames {
```
hexfusion, line 231: why do we need two loops?

retroflexer: Because the first loop determines if all the secrets are available. If any of them is not available, then, you immediately jump to regenerating certs.

If all certs are available, then, validate the SAN. If there is a SAN mismatch returning degrading error (no regeneration).

retroflexer: I originally had a single loop for checking the existence of the certs and then validating the SAN. 

The logic for determining to generate certs, I felt, became complicated and less readable. This way, it appears cleaner and more maintainable.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +194,79 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
-		return nil
+	errs := []error{}
+	for _, certName := range certNames {
+		secret, err := c.secretLister.Secrets(operatorclient.TargetNamespace).Get(etcdPeerClientCertName)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if err := verifyCertificateSAN(secret.Data, certName, nodeInternalIPs); err != nil {
+			klog.Errorf("validateCertificateSAN: %v", err)
```
marun, line 256: Out of curiosity, how would I decide whether an error was deserving of appearing in the logs (vs just being recorded in a condition message).

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +194,79 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
-	_, err = c.secretLister.Secrets(operatorclient.TargetNamespace).Get(metricsServingCertName)
-	metricsCertOk := err == nil
+	certNames := []string{etcdPeerClientCertName, etcdServingCertName, metricsServingCertName}
 
-	// if we have all the certs we want, do nothing.
-	if peerClientCertOk && servingCertOk && metricsCertOk {
-		return nil
+	errs := []error{}
+	for _, certName := range certNames {
+		secret, err := c.secretLister.Secrets(operatorclient.TargetNamespace).Get(etcdPeerClientCertName)
```
marun, line 250: Should the secret being retrieved be `certName` rather than `etcdPeerClientCertName`?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +194,79 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
 	metricsServingCertName := tlshelpers.GetServingMetricsSecretNameForNode(node.Name)
 
-	var err error
-	_, err = c.secretLister.Secrets(operatorclient.TargetNamespace).Get(etcdPeerClientCertName)
-	peerClientCertOk := err == nil
-	_, err = c.secretLister.Secrets(operatorclient.TargetNamespace).Get(etcdServingCertName)
-	servingCertOk := err == nil
-	_, err = c.secretLister.Secrets(operatorclient.TargetNamespace).Get(metricsServingCertName)
-	metricsCertOk := err == nil
+	certNames := []string{etcdPeerClientCertName, etcdServingCertName, metricsServingCertName}
```
marun, line 246: nit: Maybe a more appropriate name for this variable would be `secretNames`?

pkg/operator/etcdcertsigner/etcdcertsignercontroller_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"fmt"
+	"github.com/openshift/library-go/pkg/crypto"
+	"math/big"
+	"testing"
+	"time"
+)
+
+func Test_verifyCertificateSAN(t *testing.T) {
```
marun, line 16: (No action required) Given that this is a unit test and you know the implementation, maybe limit the scope of testing to match / doesn't match? `certName` is only used in formulating an error message and the function under test doesn't differentiate between the intended use of the certificate or understand ipv4 vs ipv6 (that's an implementation detail for golang's `VerifyHostname()`).

pkg/operator/etcdcertsigner/etcdcertsignercontroller_test.go
```diff
@@ -0,0 +1,168 @@
...
+			certData := makeCerts(t, tt.args.configHosts)
+			if err := verifyCertificateSAN(certData, tt.args.certName, tt.checkHosts); (err != nil) != tt.wantErr {
+				t.Errorf("verifyCertificateSAN() error = %v, wantErr %v", err, tt.wantErr)
+			}
+		})
+	}
+}
+
+func makeCerts(t *testing.T, hosts []string) map[string][]byte {
+	// below code is copied from vendor/github.com/openshift/library-go/pkg/crypto/crypto.go
```
marun, line 115: I'm assuming the code was copied because library-go didn't expose the functionality for reuse. Could library-go be updated to allow reuse?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +194,79 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+	}
+	return nil
+}
+
+func (c *EtcdCertSignerController) checkSecrets(node *corev1.Node, nodeInternalIPs []string, recorder events.Recorder) error {
+	etcdPeerClientCertName := tlshelpers.GetPeerClientSecretNameForNode(node.Name)
+	etcdServingCertName := tlshelpers.GetServingSecretNameForNode(node.Name)
+	metricsServingCertName := tlshelpers.GetServingMetricsSecretNameForNode(node.Name)
+
+	certNames := []string{etcdPeerClientCertName, etcdServingCertName, metricsServingCertName}
```
marun, line 219: nit: Maybe a more appropriate name for this variable would be `secretNames`?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +194,79 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+
+	errs := []error{}
+	for _, certName := range certNames {
+		secret, err := c.secretLister.Secrets(operatorclient.TargetNamespace).Get(etcdPeerClientCertName)
```
marun, line 223: Should the secret being retrieved be `certName` rather than `etcdPeerClientCertName`?

marun (commented):
First ceo review, my apologies if I'm missing important context. Please educate me!

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -99,7 +102,20 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
 
 	errs := []error{}
 	for _, node := range nodes {
-		if err := c.createSecretForNode(node, recorder); err != nil {
+		nodeInternalIPs, err := dnshelpers.GetInternalIPAddressesForNodeName(node)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if err := c.checkSecrets(node, nodeInternalIPs, recorder); err != nil {
```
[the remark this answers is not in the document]
retroflexer: If you have more validation checks used in other operators, I would be glad to reuse. Thanks.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -99,7 +102,20 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+			// if any of the secrets are not present, create secrets for the node
+			if err := c.createSecretForNode(node, nodeInternalIPs, recorder); err != nil {
+				errs = append(errs, err)
+			}
+		}
+
+		// if all secrets are present, then, validate IP address
```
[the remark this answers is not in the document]
retroflexer: I accidentally deleted a `continue` in the previous section.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
retroflexer: Thanks. copy-paste error. :)

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
retroflexer: I removed klog. We don't use it elsewhere. I was using it for my testing.

pkg/operator/etcdcertsigner/etcdcertsignercontroller_test.go
retroflexer: Yeah. I agree.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -99,10 +101,25 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if err := c.checkSecrets(node, nodeInternalIPs, recorder); err != nil {
+			// if any of the secrets are not present, create secrets for the node
+			if err := c.createSecretForNode(node, nodeInternalIPs, recorder); err != nil {
+				errs = append(errs, err)
+			}
+			continue
```
marun, line 114: nit: Is it your intent that this `continue` exits regardless of whether the secrets are successfully created?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +195,78 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) > 0 {
+		return utilerrors.NewAggregate(errs)
+	}
+	return nil
+}
+
+func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node, nodeInternalIPs []string, recorder events.Recorder) error {
```
marun, line 266: nit: `s/createSecretForNode/createSecretsForNode/`?

marun (commented):
Thank you for responding to my comments! A couple of trivial nits remain.

I am a bit concerned that many of the issues I've commented on were not caught by testing. What's to prevent future maintenance from breaking the proposed changes? Is more test coverage suggested?

Regarding my suggestions to improve secret validation and avoid duplicating client-go cert code, that could be handled in a follow-up PR to avoid delaying this bugfix. Maybe add tracking issues on the repo and assign to yourself?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
retroflexer: yes. If the secrets are not created the `len(errs)` would be greater than zero, and hence exits the function.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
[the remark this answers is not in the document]
retroflexer: Created an issue to address it in a later PR.

https://github.com/openshift/cluster-etcd-operator/issues/487

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +194,79 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+	certNames := []string{etcdPeerClientCertName, etcdServingCertName, metricsServingCertName}
+
+	errs := []error{}
+	for _, certName := range certNames {
+		secret, err := c.secretLister.Secrets(operatorclient.TargetNamespace).Get(etcdPeerClientCertName)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if secret.Data == nil {
```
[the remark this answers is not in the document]
retroflexer: Created an issue to address the validation of secrets in a different PR.

https://github.com/openshift/cluster-etcd-operator/issues/487

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +194,79 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
 
-func (c *EtcdCertSignerController) createSecretForNode(node *corev1.Node, recorder events.Recorder) error {
+// validateCertificateSAN makes sure that a given certificate is valid
+// at least for the SANs defined in the configuration.
+func verifyCertificateSAN(certData map[string][]byte, certName string, ipAddresses []string) error {
+	certificates, err := cert.ParseCertsPEM(certData["tls.crt"])
+	if err != nil {
+		return fmt.Errorf("could not parse TLS cert for %s: %w", certName, err)
+	}
+	// Using the first certificate to verify the IP addresses
```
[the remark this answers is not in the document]
retroflexer: Here are a few examples where they don't even comment why they are using the zero index certificate :-)!
 
1.  https://github.com/openshift/library-go/blob/052c4104706cb9a67fc093cba11baf7980525109/pkg/assets/template.go#L50

2. https://github.com/openshift/service-ca-operator/blob/ac8a280b37d8efc765fa82f94467d30eb5674813/pkg/operator/sync_common.go#L94

3. https://github.com/openshift/service-ca-operator/blob/ac8a280b37d8efc765fa82f94467d30eb5674813/pkg/operator/sync_common.go#L128

4. https://github.com/openshift/service-ca-operator/blob/ac8a280b37d8efc765fa82f94467d30eb5674813/pkg/controller/servingcert/starter/starter.go#L98

5. https://github.com/openshift/service-ca-operator/blob/ac8a280b37d8efc765fa82f94467d30eb5674813/pkg/controller/servingcert/controller/secret_creating_controller.go#L243

6. https://github.com/openshift/cluster-kube-apiserver-operator/blob/c7b1e077d913fbe6e89d4c10312346234e31e4e4/test/e2e/user_certs_test.go#L303

7. https://github.com/openshift/service-ca-operator/blob/09bcf1403670716e32f33b51893a0ddf3cbb3aed/test/e2e/e2e_test.go#L474

:-)

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
marun: Aha, re-reading my question it wasn't clear what I was asking. Maybe add a comment to the `continue` indicating something like 'No need to validate IP addresses - either the secrets remain in an invalid state or they have been created with the correct ip addresses'?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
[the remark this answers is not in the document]
marun: I'm afraid most of the service ca operator examples aren't applicable. What's being managed there is a CA cert (a cert used to generate other certificates rather than secure traffic), and the service ca operator performs seamless automatic rotation by cross-signing. The first cert is the current CA, and any subsequent cert is used only for validating previously issued certificates during a handover period.

For https://github.com/openshift/service-ca-operator/blob/ac8a280b37d8efc765fa82f94467d30eb5674813/pkg/controller/servingcert/starter/starter.go#L95, note the explicit check for only a single certificate.

In the library-go and ckao examples you point to, they are implicitly expecting only a single certificate. 

The comment here says 'Using the first certificate...' which begs the questions 'why the first certificate?' and 'are there other certificates and why are their ip addresses not important to check?'. I'm assuming only a single certificate is expected, which suggests following the example of library-go and ckao and just removing the comment.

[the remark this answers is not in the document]
retroflexer: I will remove the comment. Thanks.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
retroflexer: We want to validate the ip address, only if we already found extant certificates.

If we just created the certificates, it is assumed they are created with the right ip address. I will add a comment.

pkg/operator/etcdcertsigner/etcdcertsignercontroller_test.go
retroflexer: I don't see much point in this unit-test honestly. We are essentially testing VerifyHostname() which is part of the x509 package is already tested by the package (and gazillions users of it)

marun: I agree with your assertion that this test adds little value, but the overarching question from my last review remains: How best to ensure that ip address validation is performed correctly and that the intent will be maintained in the face of future maintenance?

retroflexer: Yeah. We need to add e2e tests or CI tests with new machines created and IP addresses changed.

retroflexer:
> I am a bit concerned that many of the issues I've commented on were not caught by testing. What's to prevent future maintenance from breaking the proposed changes? Is more test coverage suggested?

We definitely need a CI test (or e2e test) involving new machines getting as well as a scenario with changed IP addresses.

pkg/operator/etcdcertsigner/etcdcertsignercontroller_test.go
hexfusion: Can we test kube objects without kube itself? Do we do that already in this repo?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -99,10 +101,26 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
 	for _, node := range nodes {
-		if err := c.createSecretForNode(node, recorder); err != nil {
+		nodeInternalIPs, err := dnshelpers.GetInternalIPAddressesForNodeName(node)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if err := c.checkSecrets(node, nodeInternalIPs, recorder); err != nil {
+			// if any of the secrets are not present, create all secrets for the node
+			if err := c.createSecretsForNode(node, nodeInternalIPs, recorder); err != nil {
```
hexfusion, line 111: I believe you only want to create new secrets if `if errors.IsNotFound(err)` otherwise this appears to try and create a new certificate if there is any err even transient in `checkSecrets` ?  Is that true, what would be the result?

hexfusion:
> raft2020/11/06 15:28:55 INFO: raft.node: dee93288b522d12e elected leader 83b660b1a9693b34 at term 16

Both of the last 2 agnostic tests have etcd in 15+ terms which can be a sign of disruption during scaling. Generally, we expect this to last at 4 - 5. Do we have any theories there?

ref: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/461/pull-ci-openshift-cluster-etcd-operator-master-e2e-agnostic/1324727387086131200

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -99,10 +102,30 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
-		if err := c.createSecretForNode(node, recorder); err != nil {
+		nodeInternalIPs, err := dnshelpers.GetInternalIPAddressesForNodeName(node)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if err := c.checkSecrets(node, nodeInternalIPs, recorder); err != nil {
+			if errors.IsNotFound(err) {
+				// if any of the secrets is not present, create all secrets for the node
+				if err := c.createSecretsForNode(node, nodeInternalIPs, recorder); err != nil {
```
hexfusion, line 113: I think this would make sense to just include in `checkSecrets()`? Maybe then we are `ensureSecrets()`?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -99,10 +102,30 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+				}
+			} else {
+				errs = append(errs, err)
+			}
+			// no need to validate IP addresses, if we just created secrets
+			continue
+		}
+
+		// if all secrets are already extant, then, validate IP addresses
+		if err := c.validateIpAddrs(node, nodeInternalIPs, recorder); err != nil {
```
hexfusion, line 124: if we validate first would that simplify this loop logic?

hexfusion (commented):
@retroflexer thanks for the update, getting closer! We are performing two distinct actions.

1.) verify: is what exists correct?
2.) ensure: do the resources we need exist if not create them.

retroflexer:
/hold

Something wrong with the latest update. The tests are failing. I am looking into it.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -99,10 +102,23 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
 	errs := []error{}
 	for _, node := range nodes {
-		if err := c.createSecretForNode(node, recorder); err != nil {
+		nodeInternalIPs, err := dnshelpers.GetInternalIPAddressesForNodeName(node)
+		if err != nil {
 			errs = append(errs, err)
+			continue
+		}
+		if err := c.ensureAndValidateSecrets(node, nodeInternalIPs, recorder); err != nil {
+			if errors.IsNotFound(err) {
```
marun, line 111: (No action required) When evaluating more than nil, maybe don't use the inline conditional check?

```golang
err := ...
if errors.IsNotFound(err) {
 ...
} else if err != nil {
 ...
}

```

This message brought to you by the society for the minimization of indentation.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +194,59 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+	}
+	if len(errs) > 0 {
+		return utilerrors.NewAggregate(errs)
 	}
 
+	return nil
+}
+
+func (c *EtcdCertSignerController) createSecretsForNode(node *corev1.Node, nodeInternalIPs []string, recorder events.Recorder) error {
+	etcdPeerClientCertName := tlshelpers.GetPeerClientSecretNameForNode(node.Name)
```
marun, line 247: (No action required) Maybe s/CertName/SecretName/ ?

pkg/operator/etcdcertsigner/etcdcertsignercontroller_test.go
marun: @retroflexer As per @hexfusion's comment, please remove this test in favor of one that uses a fake client to simulate API interaction as per [this example](https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/etcdendpointscontroller/etcdendpointscontroller_test.go#L29).

marun (commented):
Code looks fine, but as per an inline comment please add a test that validates the proposed behavior.

retroflexer:
/test e2e-agnostic
--

hexfusion (commented):
@retroflexer looks good a few optional things to look at here. I can tag when you are ready.

/approve

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -99,10 +102,22 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+			errs = append(errs, err)
+			continue
+		}
+		err = c.ensureAndValidateSecrets(node, nodeInternalIPs, recorder)
+		if errors.IsNotFound(err) {
+			// if any of the secrets are not present, create all secrets for the node
+			err = c.createSecretsForNode(node, nodeInternalIPs, recorder)
+		}
+
+		if err != nil {
```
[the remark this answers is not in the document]
retroflexer: Here `err` is reused. The err are appended to errs if the createSecretsForNode also fails. Earlier code was like this:

```
err = c.ensureAndValidateSecrets(node, nodeInternalIPs, recorder)
		if errors.IsNotFound(err) {
			// if any of the secrets are not present, create all secrets for the node
			err := c.createSecretsForNode(node, nodeInternalIPs, recorder)
			if err != nil {
				errs = append(errs, err)
			}
		}

		if err != nil {
			errs = append(errs, err)
		}
```
By re-using `err` we eliminated a few indents, but if you think that makes the code confusing, I can reinstate the earlier version that didn't re-use err.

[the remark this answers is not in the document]
hexfusion: > Earlier code was like this
as noted no change is required, the scope of a variable is what it is. defensive coding with local scope, in general, reduces unexpected leaks. I can see this is covered here.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -178,24 +193,59 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+	errs := []error{}
+	for _, secretName := range secretNames {
+		secret, err := c.secretLister.Secrets(operatorclient.TargetNamespace).Get(secretName)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return err
+			}
+			errs = append(errs, err)
+			continue
+		}
```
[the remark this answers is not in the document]
hexfusion: nesting would be best avoided if possible is the point.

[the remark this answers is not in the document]
retroflexer: Fixed it. Thanks.

PR #474, opened 2020-10-22
Use etcd's /health endpoint for readiness checking

ironcladlou:
Surely it's not this simple...


hexfusion:
looks like it was :)

/lgtm

PR #483, opened 2020-10-29
DEBUG: Safe scale

hexfusion:
based on #449


hexfusion:
cleaning up debug
/close

PR #485, opened 2020-11-02
Bug 1892288: pkg/etcdcli: add IsQuorumFaultTolerant

hexfusion:
In some circumstances, it is not safe for the operator to perform an action that could result in a new static pod revision. One such example would be when the etcd cluster has less than 3 members. In this case any restart to of an etcd process would result in quorum loss. While quorum loss should be transient it can be quite disruptive as we wait for graceful termination of the pervious process. This wait can exceed 1 minute.

The PR addresses this with a new function `IsQuorumFaultTolerant` which allows the cluster to make decisions on when an action is appropriate based on the cluster's ability to tolerate a single fault, which is generally the result of a new static pod revision or upgrade.

TODO

- [x] fix tests https://github.com/openshift/cluster-etcd-operator/blob/3bdb09e493d265257e63aa6c6995f6ed76245c8b/pkg/operator/etcdendpointscontroller/etcdendpointscontroller_test.go


retroflexer:
Unit tests are failing:
```
FAIL: github.com/openshift/cluster-etcd-operator/pkg/operator/etcdendpointscontroller TestBootstrapAnnotationRemoval/NewClusterBootstrapRemoval 0s
FAIL: github.com/openshift/cluster-etcd-operator/pkg/operator/etcdendpointscontroller TestBootstrapAnnotationRemoval 0s 
```

hexfusion:
quota

> ci/prow/e2e-agnostic — Job failed.
> time="2020-11-02T17:51:41Z" level=error msg="Error: Error when reading or editing Instance Group \"ci-op-hytn1fsp-32f07-x4nsg-bootstrap\": googleapi: Error 403: Quota exceeded for quota group 'ReadGroup' and limit 'Read requests per 100 seconds' of service 'compute.googleapis.com' for consumer 'project_number:1053217076791'., rateLimitExceeded"

pkg/etcdcli/health_test.go
```diff
@@ -309,3 +309,106 @@ func TestGetUnhealthyMemberNames(t *testing.T) {
...
 }
+
+func TestIsQuorumFaultTolerant(t *testing.T) {
+	tests := []struct {
+		name         string
+		memberHealth memberHealth
+		want         bool
+	}{
+		{
+			"test all available members",
```
marun, line 320: (No action required) Maybe simplify with helper functions so that the test data is easier to read?

```golang
[]healthCheck{
  healthyMember("etcd-1"),
  unhealthyMember("etcd-2"),
  unstartedMember("etcd-3"),
}
```

hexfusion: Yeah, tests need improved I agree. I will incorporate the comments into the final changes.

hexfusion:
reduced terms from 90ish to 6, getting more testing for validation.

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/485/pull-ci-openshift-cluster-etcd-operator-master-e2e-metal-assisted/1323884374478819328

/test e2e-metal-assisted

pkg/etcdcli/health.go
```diff
@@ -201,6 +204,23 @@ func HasStarted(member *etcdserverpb.Member) bool {
...
 
+// IsQuorumFaultTolerant checks the current etcd cluster and returns true if the cluster can tolerate the
+// loss of a single etcd member. Such loss is common during new static pod revision.
+func IsQuorumFaultTolerant(memberHealth []healthCheck) bool {
+	totalMembers := len(memberHealth)
+	quorum := totalMembers/2 + 1
+	healthyMembers := len(GetHealthyMemberNames(memberHealth))
+	switch {
+	case totalMembers-quorum < 1:
+		klog.Errorf("etcd cluster has quorum of %d which is not fault tolerant", quorum)
```
romfreiman, line 215: won't printing the memberHealth help later on to analyze? Those clauses should not happen to often, right?

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
```diff
@@ -89,13 +92,19 @@ func (c *EtcdEndpointsController) syncConfigMap(recorder events.Recorder) error
 		return fmt.Errorf("couldn't determine bootstrap status: %w", err)
 	}
 
+	etcdMembers, err := c.etcdClient.MemberList()
+	if err != nil {
+		return fmt.Errorf("could not create etcd client: %w", err)
+	}
+	memberHealth := etcdcli.GetMemberHealth(etcdMembers)
```
romfreiman, line 99: this cannot fail?

pkg/etcdcli/health.go
hexfusion: I agree

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
hexfusion: added test for empty []healthCheck but we should not have a failure from nil pointer here. When I added the mock etcd servers to `TestBootstrapAnnotationRemoval` it does test the absolute minimum response configuration else err.

ironcladlou:
/lgtm
/hold

@hexfusion thanks, this is good stuff. Can think about using it with conditions as well. Adding hold which you can remove when you're happy with your tests.

PR #488, opened 2020-11-04
Use 'kubernetes.io/tls' instead of SecretTypeTLS which is not being substituted

jhadvig:
/assign @hexfusion


hexfusion:
/hold

this is breaking etcd and needs a closer look.

PR #489, opened 2020-11-09
Bug 1895509: Make restore-pod yaml node agnostic

retroflexer:
The current restore pod yaml has node specific URLs which makes it not usable across the cluster on any node to restore.

This PR uses template placeholders to replace them with node specific values at the time of restore.

### DIFF
--- a/bindata/etcd/cluster-restore.sh
+++ b/bindata/etcd/cluster-restore.sh
@@ -63,7 +63,18 @@ function extract_and_start_restore_etcd_pod() {
    fi
 
    echo "starting restored etcd-pod.yaml"
-   tar -O -xvf "${BACKUP_FILE}" -C "${MANIFEST_DIR}"/ "${BACKUP_POD_PATH}" > "${MANIFEST_DIR}"/etcd-pod.yaml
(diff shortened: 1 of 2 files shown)


hexfusion (commented):
Does this change break future DR possibilities? Ideally, the restore pod and the regular pod were the same. But since they are not it seems like they should remain as close as possible. Is the problem that we actually should be saving all of the restore pods per revision much as we do with certs? maybe `etcd-all-restore-pod` .

retroflexer:
> Does this change break future DR possibilities? Ideally, the restore pod and the regular pod were the same. But since they are not it seems like they should remain as close as possible. Is the problem that we actually should be saving all of the restore pods per revision much as we do with certs? maybe etcd-all-restore-pod.

I actually think we should have all the restore-pod.yaml files identical on all nodes with values coming from the environment. The same goes for etcd-pod.yaml. I am not sure if they are is no way of generating one version of pod.yaml and be applicable for all nodes just based on the environment.

But given where things are, I will modify to save 3 copies of restore-pod.yaml, so that when restore we can grab the right one for the restoring node.

retroflexer:
Closing as this PR is supersesed by #495. 

/close

PR #491, opened 2020-11-09
Allow rollbackcopier pod to be disabled

ironcladlou:
Allow the `rollbackcopier` container in the etcd pod to to be disabled using a
`openshift.io/disable-rollbackcopier` annotation. The `rollbackcopier` container
is only present in versions 4.4 and 4.5. This patch is implemented in such a way
that the disable function can be introduced in the latest release and
backported. The new code should effectively no-op in releases where
`rollbackcopier` is not present.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 	k8s.io/cri-api v0.18.6
 	k8s.io/klog/v2 v2.3.0
 	k8s.io/utils v0.0.0-20200729134348-d5654de09c73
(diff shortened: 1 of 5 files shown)


ironcladlou:
I need to do some testing in `release-4.5`.

/hold

ironcladlou:
Going to open this against 4.5 instead.

/close

ironcladlou:
Opened https://github.com/openshift/cluster-etcd-operator/pull/492 instead.

PR #492, opened 2020-11-09
WIP: Allow rollbackcopier pod to be disabled

ironcladlou:
Allow the `rollbackcopier` container in the etcd pod to to be disabled using a
`openshift.io/disable-rollbackcopier` annotation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,4 +27,5 @@ require (
 	k8s.io/component-base v0.18.6
 	k8s.io/klog v1.0.0
 	k8s.io/utils v0.0.0-20200324210504-a9aa75ae1b89
+	sigs.k8s.io/yaml v1.2.0
 )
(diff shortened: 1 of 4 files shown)


hexfusion (commented):
I think this makes sense, thanks. General question, since we have the operator to make decisions can we find an observation where we can conclude that we are outside of the need for rollback copier and disable it automatically?

Basically, once we have upgraded to 4.5 it should no longer be needed. Perhaps that can be a follow-up depending on the results seen by customers who disable it. But if we find it to be root cause to perf issues we might need to be more aggressive with disabling. Open to input.

retroflexer:
> I think this makes sense, thanks. General question, since we have the operator to make decisions can we find an observation where we can conclude that we are outside of the need for rollback copier and disable it automatically?
> 
> Basically, once we have upgraded to 4.5 it should no longer be needed. Perhaps that can be a follow-up depending on the results seen by customers who disable it. But if we find it to be root cause to perf issues we might need to be more aggressive with disabling. Open to input.

Yeah. We can get rid of the code entirely.

ironcladlou:
Superseded by https://github.com/openshift/cluster-etcd-operator/pull/494

/close

PR #493, opened 2020-11-09
WIP: Automatically remove the rollbackcopier container when safe

ironcladlou:


ironcladlou:
The thing I don't like about this is the that the baseline assumption is that the container should exist by default and then has to be actively removed, which means if the precondition checks aren't stable over time we might accidentally temporarily restore the container. I can already imagine bugs with what I have here with a subsequent upgrade which would necessitate some modification to the approach (e.g. tracking that we removed it so that we never re-add it no matter what).

I'm not entirely sure inverting the assumptions would help. If we remove the container from the manifest I'm not sure yet what conditions will need to hold true so that we don't accidentally _add_ it back.

Need to think and talk more about it.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -220,17 +238,67 @@ func (c *TargetConfigController) manageRecoveryPod(substitutionReplacer *strings
...
+	if clusteroperatorv1helpers.IsStatusConditionTrue(clusterVersion.Status.Conditions, configv1.OperatorProgressing) {
+		errors = append(errors, fmt.Errorf("clusterversion condition %s=True", configv1.OperatorProgressing))
+	}
+	if clusteroperatorv1helpers.IsStatusConditionTrue(clusterVersion.Status.Conditions, "Failing") {
+		errors = append(errors, fmt.Errorf("clusterversion condition %s=True", "Failing"))
+	}
+	if err := utilerrors.NewAggregate(errors); err != nil {
+		klog.V(3).Infof("won't remove rollback copier because following preconditions failed: %v", err)
+		return etcdPod
+	}
```
ironcladlou, line 291: Pretty sure this is unworkable without additional state because it'll falsely say the copier should exist during a 4.5.z->4.5.z+1 or (e.g.) 4.5->4.6 upgrade

hexfusion (commented):
> The thing I don't like about this is the that the baseline assumption is that the container should exist by default and then has to be actively removed, which means if the precondition checks aren't stable over time we might accidentally temporarily restore the container. I can already imagine bugs with what I have here with a subsequent upgrade which would necessitate some modification to the approach (e.g. tracking that we removed it so that we never re-add it no matter what).
> 
> I'm not entirely sure inverting the assumptions would help. If we remove the container from the manifest I'm not sure yet what conditions will need to hold true so that we don't accidentally _add_ it back.
> 
> Need to think and talk more about it.

To reduce/eliminate flapping can we use your original idea of annotation existence as a condition for removal. Then use logic to ensure the annotation exists based on cluster version observations?

ironcladlou:
Superseded by https://github.com/openshift/cluster-etcd-operator/pull/494

PR #494, opened 2020-11-10
Bug 1894267: Remove the rollbackcopier

ironcladlou:
The rollbackcopier container exists to support a 4.4->4.5 upgrade. Once
4.5 is running, the container is no longer needed provided a 4.4 backup
exists prior to the upgrade, which the rollbackcopier basically assures.

The copier itself incurs a performance penalty that 4.5 users shouldn't
be subject to once the upgrade is complete.

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -233,34 +233,6 @@ ${COMPUTED_ENV_VARS}
         name: cert-dir
       - mountPath: /var/lib/etcd/
         name: data-dir
-  - name: rollbackcopier
-    image: ${OPERATOR_IMAGE}
-    imagePullPolicy: IfNotPresent
-    terminationMessagePolicy: FallbackToLogsOnError
-    command:
-      - /bin/sh
-      - -c
-      - |
-        #!/bin/sh
-        set -euo pipefail
-
-        export ETCD_NAME=${NODE_NODE_ENVVAR_NAME_ETCD_NAME}
-        exec cluster-etcd-operator rollbackcopy
-    resources:
-      requests:
-        memory: 60Mi
-        cpu: 30m
-    securityContext:
-      privileged: true
-    volumeMounts:
-      - mountPath: /etc/kubernetes/rollbackcopy
-        name: rollbackcopy
-      - mountPath: /etc/kubernetes/static-pod-resources
-        name: full-resource-dir
-      - mountPath: /etc/kubernetes/static-pod-certs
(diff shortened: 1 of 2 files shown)


ironcladlou:
4.4 -> 4.5 upgrade test: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1326211719902531584

/retest

hexfusion (approved):
/lgtm

hexfusion:
@sdodson this PR removes rollbackcopier from 4.5. The net gain is that the cluster does not perform snapshots every hour without users consent. We have seen some indications that these snapshots can lead to throttling events on certain providers.

As this is already removed from 4.6 I am going to manually give greens on valid Bugzilla.

ecordell:
Could there not be users that are relying on this behavior? It seems strange to completely remove a component in a z-stream.

hexfusion:
@ecordell this was undocumented thus not a supported feature.

ironcladlou:
>Could there not be users that are relying on this behavior? It seems strange to completely remove a component in a z-stream.

Anything's possible, but the facts are we have substantial evidence the function is causing performance problems for real users today, and we have no evidence of any users relying on it. If we became aware of users relying on it, we would be instructing them to cease relying upon it, and would feel justified in doing so because we never intended or communicated support for such uses in the first place. It was a single purpose tool to mitigate risk in a specific version upgrade, and beyond that it has only deleterious effects on a cluster.

Hope that helps justify the change.

ecordell:
4.4 -> 4.5 upgrades are still supported - is there some reason why this is no longer needed this as a fail-safe when upgrading to newer 4.5.z?

ecordell:
@hexfusion and I spoke about this some, recording the discussion here:

- we don't think users are depending on this, but we don't have information to know for sure
- the worst case is that someone was relying on this for etcd backups in 4.5, updates to a new 4.5.z with this removed, and sometime later has a disaster that requires etcd recovery only to find they don't have them.
- the bug is currently low / medium, but the impact is really quite high/expensive for users that are hitting the io limits in their cloud provider

a couple of options to move forward:

- gather some evidence that this isn't something a customer could reasonably have relied upon, and go ahead and remove
or 
- look into options to disable this for any user that is having problems due to large etcd + small io budgets, with full removal in 4.6

sdodson:
cherry-pick-approved on @ecordell behalf as he's got permissions issues on this repo

PR #498, opened 2020-11-17
Add IBM Cloud managed annotations to CVO manifests

csrwng:
This matches https://github.com/openshift/enhancements/pull/445 and does not change behavior.


ironcladlou:
@csrwng was the dependency bump intended and/or necessary for this change?

csrwng:
> @csrwng was the dependency bump intended and/or necessary for this change?

@ironcladlou yes it was necessary so we could pick up the updated `vendor/github.com/openshift/api/operator/v1/0000_12_etcd-operator_01_config.crd.yaml` file from github.com/openshift/api

PR #501, opened 2020-11-20
Remove unused cruft

ironcladlou:
* Deprecate and remove some unused flags
* Remove unused and confusing openshift-etcd/config configmap
* Remove the endpoints controller which is no longer needed as of https://github.com/openshift/cluster-etcd-operator/pull/354


pkg/cmd/render/render.go
```diff
@@ -124,15 +119,16 @@ func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
 	r.manifest.AddFlags(fs, "etcd")
 	r.generic.AddFlags(fs)
 
+	// TODO: update bootkube.sh in the installer and then remove this
+	var deprecatedClusterEtcdOperatorImage string
```
hexfusion, line 123: We currently have no use but maybe use in the future?

PR #502, opened 2020-11-30
Bug 1844727: Use socket readiness probe to avoid generating zombies

ironcladlou:
Replace the `exec` etcd readiness probe with a `tcpSocket` probe because
the `exec` version generates zombie processes in the container. The new
probe should be functionally equivalent.


hexfusion:
LGTM pending review from node team.

hexfusion:
this needs to go back to 4.4

tnozicka:
btw. it was also in top 3 cpu usage on my master, thx for the fix

PR #505, opened 2020-12-03
Annotate manifests for single-node-developer cluster profile

rkukura:
This partially implements phase 1 of
https://github.com/openshift/enhancements/blob/master/enhancements/single-node-developer-cluster-profile.md
and does not change behavior. Initially, all cluster-etcd-operator
manifests are included in the single-node-developer cluster profile.
Follow-on PRs may exclude any of these that are not needed in the
profile.


manifests/0000_12_etcd-operator_08_etcdquorumguard_deployment.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   namespace: openshift-etcd
   annotations:
     include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
     exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   replicas: 3
```
hexfusion, line 11: I would have figured that you would not want quorum-guard for single node?

hexfusion: I see that you will followup.

rkukura: yes - that one we expect to disable

PR #504, opened 2020-12-02
Bug 1898954: Backup script should get the latest revision of static-pod-resources …

retroflexer:
…in a reliable way.

Currently the backup script relies on the timestamp on the filesystem to determine the latest revision of the static pod resources. Instead, we should use more reliable ways such as inspecting the manifests for the currently running containers to exactly determine the currently active revision.

PR #509, opened 2020-12-09
Bug 1906875: Provide an option force backup even when API is not available to check the health.

retroflexer:
Currently the disruptive tests require taking backup when there is no etcd quorum. The previous PR #504 necessitated the need for the API server to be available to check the reliability of the revision being backed up.

This PR provides an option to force backup even when API is not available to check the health of each of static pods.

### DIFF
--- a/bindata/etcd/cluster-backup.sh
+++ b/bindata/etcd/cluster-backup.sh
@@ -15,9 +15,15 @@ if [[ $EUID -ne 0 ]]; then
 fi
 
 function usage {
-  echo 'Path to backup dir required: ./cluster-backup.sh <path-to-backup-dir>'
+  echo 'Path to backup dir required: ./cluster-backup.sh [--force] <path-to-backup-dir>'
   exit 1
 }
+
+IS_DIRTY=""
+if [ "$1" == "--force" ]; then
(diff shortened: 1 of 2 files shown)


marun (commented):
Please update the output on L47 to include mention of the option to use `--force`. Otherwise I worry that a user might not realize they have the option when they see the error.

hexfusion (commented):
what tells support that user knowingly removed safety checks around correctness to generate the backup. then also is this backup supported?

retroflexer:
> what tells support that user knowingly removed safety checks around correctness to generate the backup. then also is this backup supported?

I added __POSSIBLY_DIRTY__ string to the files generated. 

By the way, do we want to support this flag officially and document it? Otherwise I can remove it from the usage, and it need not be documented.

retroflexer:
> Please update the output on L47 to include mention of the option to use `--force`. Otherwise I worry that a user might not realize they have the option when they see the error.

As I said above (https://github.com/openshift/cluster-etcd-operator/pull/509#issuecomment-742221582), I am not sure if we want to officially support this flag? In any case, I added the message to `Pass the --force flag to skip checks`.

marun:
LGTM

PR #516, opened 2021-01-06
switch to v1beta1 for the p&f APIs

tkashem:
Important Note: We support upgrade and downgrade - 4.7 to 4.6 downgrade would behave very badly (in a skew case). So we should NOT back port to 4.7


tkashem:
master is 4.7, so holding for now. Once master opens to 4.8 we can cancel the hold 

/hold

PR #480, opened 2020-10-26
*: add TLS security policy observer

hexfusion:
Currently, etcd does not provide an explicit cipher suite, the result is etcd is exposed to various exploits on week ciphers such as SWEET32.

This PR adds a TLS security policy observer and then reacts to change to this profile. If the profile is not observed we fallback to the Default as implemented by the library-go crypto package.

The goal is to have a single TLS policy configuration that will be reacted upon by the cluster.

PR #523, opened 2021-01-14
Bug 1851810: *: ensure etcd listens on strong ciphers

hexfusion:
Today etcd will accept any cipher that was part of golang 1.12 unfortunately some of these ciphers are considered weak exposing us to vulnerabilities such as SWEET32. This PR resolved the issue by defining a default using the `TLSProfileIntermediateType`[1]. These ciphers are available in golang 1.12 -> 1.15 so apiserver will have 100% compatibility and are considered strong tls v1.2.

Currently, etcd does not provide an explicit cipher suite, the result is etcd is exposed to various exploits on week ciphers such as SWEET32.

The plans are to supersede this PR by adding a TLS observer, but we have a problem with using the kube-apiservers default for various reasons including lack of default support for tls 1.3.

[1] https://github.com/openshift/api/blob/release-4.6/config/v1/types_tlssecurityprofile.go
[2] https://github.com/openshift/cluster-etcd-operator/pull/480

### DIFF
--- a/pkg/etcdenvvar/etcd_env_test.go
+++ b/pkg/etcdenvvar/etcd_env_test.go
@@ -1,0 +1,40 @@
+package etcdenvvar
+
+import (
+	"reflect"
+	"testing"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+// TestWhitelistEtcdCipherSuites this test is intended to ensure the ciphers we support is explicitly understood overtime.
+// As etcd minor versions increment golang versions will as well, so this test will need to be maintained against changes
+// to the etcd whitelist[1] and or TLSProfileIntermediateType.
+//[1] https://github.com/etcd-io/etcd/blob/release-3.4/pkg/tlsutil/cipher_suites.go
+func TestWhitelistEtcdCipherSuites(t *testing.T) {
+	tests := []struct {
+		name    string
+		ciphers []string
+		want    []string
+	}{
+		{
+			name:    "test TLS v1.2 (current supported runtime)",
+			ciphers: crypto.OpenSSLToIANACipherSuites(configv1.TLSProfiles[configv1.TLSProfileIntermediateType].Ciphers),
+			want: []string{
+				"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256", // https://ciphersuite.info/cs/TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256/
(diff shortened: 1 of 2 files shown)


pkg/etcdenvvar/etcd_env_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+func TestWhitelistEtcdCipherSuites(t *testing.T) {
```
marun, line 11: I'm a bit confused by the intent of this test. What future change would cause it to fail, and what would the consequence of that failure be if uncaught?

hexfusion: yeah, I think it's just exercising the code I added to ensure the logical path was valid with ommissions. Understanding the net ciphers that we honor is the intent. For example, I would expect the results to change if we bumped etcd to 3.5 in which case  tls1.3 test would fail as an example because we would gain that support.

If the supported ciphers change we would want to know is the short answer. I can reduce the test coverage to only test 1.2 and v1.3 `TLSProfileModernType` vs `TLSProfileModernType` v1.2.

marun: Maybe document that intent? Otherwise the test as written might be perceived by the uninitiated to be testing code we don't own.

hexfusion:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/523/pull-ci-openshift-cluster-etcd-operator-master-e2e-agnostic/1349839385847140352
```
2021-01-14 22:34:57.563485 I | pkg/flags: recognized and used environment variable ETCD_CIPHER_SUITES=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
```

pkg/etcdenvvar/etcd_env_test.go
hexfusion: makes sense

hexfusion: I adjusted the test to only test the actual profile we use as default and documented the test's intent.

hexfusion:
This should go back to 4.5 maybe 4.4

PR #527, opened 2021-01-18
Adding single node ocp cluster support

tsorya:
This PR is added to support single node ocp cluster flow:
1. Adding new openshift/api version that adds new infrastructure status field - ControlPlaneTopology  - that describes if cluster is sno or regular one.
2. To be able to manage etcd-quorum-guard deployment dependent on cluster topology it was decided to manage it in etcd controller that will allow not to create deployment in case of single node and create it with expected number of replicas in all other cases.
3. Same regarding etcd-quorum-guard-pdb deployment
4. To be able to observe and to clean any changes made to be etcd-quorum-guard deployment and pdb, we added same mechanism of validation aas in cvo. On every pod, deployment change or every 10 minutes controller will validate that deployment or pdb were the same as expected and if not it will modify them back to expected ones.
5. We added new scalingStrategy "single node", right now it will be the same as isUnsupportedUnsafeEtcd. Was needed to be able to start cluster with single etcd instance.


pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	return nil
+}
+
+// ensureEtcdGuardDeployment if etcd quorum guard deployment doesn't exist - add it, else skip
+func (c *QuorumGuardController) ensureEtcdGuardDeployment(ctx context.Context, replicaCount int32, recorder events.Recorder) error {
+	_, err := c.kubeClient.AppsV1().Deployments(operatorclient.TargetNamespace).Get(ctx,
+		EtcdGuardDeploymentName, metav1.GetOptions{})
+	switch {
```
hexfusion, line 151: optional nit: we really don't need the switch its nice to read `return nil` at bottom of method/function IMO

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+	case err != nil:
+		klog.Warningf("failed to get %s deployment err %v", EtcdGuardDeploymentName, err.Error())
+		return err
+	default:
+		return nil
+	}
+}
+
+// applyDeployment applies new etcd quorum guard deployment
+func (c *QuorumGuardController) applyDeployment(replicaCount int32, recorder events.Recorder) error {
```
hexfusion, line 180: optional nit: if we only need a client can we drop this as a method and just pass the client?

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+		klog.Errorf("Failed to deploy %s, error: %v", deployment.Name, err)
+		return err
+	}
+
+	klog.Infof("New etcd quorum guard deployment is %s", newDeployment.Name)
+	return err
+}
+
+// applyPdbDeployment applies new etcd quorum guard pdb
+func (c *QuorumGuardController) applyPdbDeployment(ctx context.Context, recorder events.Recorder) error {
```
hexfusion, line 196: same here re function vs method or even just inline if only used once.

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+			Selector:       &metav1.LabelSelector{MatchLabels: map[string]string{"k8s-app": EtcdGuardDeploymentName}},
+		},
+	}
+
+	if _, err := c.kubeClient.PolicyV1beta1().PodDisruptionBudgets(operatorclient.TargetNamespace).Create(ctx, pdb, metav1.CreateOptions{}); err != nil {
+		klog.Errorf("Failed to deploy %s, error: %v", pdb.Name, err)
+		return err
+	}
+
+	klog.Infof("New etcd quorum guard pdb deployment is %s", pdb.Name)
```
hexfusion, line 217:
```
New etcd quorum guard PDB is %q
```

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+	recorder.Eventf("CreatedEtcdGuardPDBDeployment", "%s was created", EtcdGuardDeploymentName)
```
hexfusion, line 218: I think I hear an echo ;)

```
recorder.Eventf("CreatedEtcdGuardPDB", "%q", EtcdGuardDeploymentName)
```

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+	eventRecorder events.Recorder,
+	infrastructureLister configv1listers.InfrastructureLister,
+) factory.Controller {
+	c := &QuorumGuardController{
+		operatorClient:       operatorClient,
+		kubeClient:           kubeClient,
+		podLister:            kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Pods().Lister(),
+		nodeLister:           kubeInformers.InformersFor("").Core().V1().Nodes().Lister(),
+		infrastructureLister: infrastructureLister,
+		replicaCount:         0,
```
hexfusion, line 71: do we want to unset this everytime the operator restarts?

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+
+	return nil
+}
+
+// ensureEtcdGuardDeployment if etcd quorum guard deployment doesn't exist - add it, else skip
+func (c *QuorumGuardController) ensureEtcdGuardDeployment(ctx context.Context, replicaCount int32, recorder events.Recorder) error {
+	_, err := c.kubeClient.AppsV1().Deployments(operatorclient.TargetNamespace).Get(ctx,
+		EtcdGuardDeploymentName, metav1.GetOptions{})
+	switch {
+	case apierrors.IsNotFound(err):
```
hexfusion, line 152: we need to not only ensure it exists but that its correct. If user makes a change to the PDB etc we want to stop it 

```
if !reflect.DeepEqual(observed, expected) {
  //stomp user trying to do naughty things and replace proper values for deployment and pdb
}
```

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+) factory.Controller {
+	c := &QuorumGuardController{
+		operatorClient:       operatorClient,
+		kubeClient:           kubeClient,
+		podLister:            kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Pods().Lister(),
+		nodeLister:           kubeInformers.InformersFor("").Core().V1().Nodes().Lister(),
+		infrastructureLister: infrastructureLister,
+		replicaCount:         0,
+	}
+	return factory.New().ResyncEvery(time.Minute).WithInformers(
```
hexfusion, line 73: I think this probably a little aggressive CVO uses 10 minutes I think we can as well.

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+
+// applyPdbDeployment applies new etcd quorum guard pdb
+func (c *QuorumGuardController) applyPdbDeployment(ctx context.Context, recorder events.Recorder) error {
+	klog.Infof("Creating new PDB %q", EtcdGuardDeploymentName)
+	maxUnavailable := intstr.FromInt(pdbDeploymentMaxUnavailableDefault)
+	pdb := &policyv1beta1.PodDisruptionBudget{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      EtcdGuardDeploymentName,
+			Namespace: operatorclient.TargetNamespace,
+			Annotations: map[string]string{"include.release.openshift.io/self-managed-high-availability": "true",
```
hexfusion, line 203: drop annotations they are not authoritative anymore and work to maintain accurately.

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+				"include.release.openshift.io/single-node-developer": "true"},
+		},
+		Spec: policyv1beta1.PodDisruptionBudgetSpec{
+			MaxUnavailable: &maxUnavailable,
+			Selector:       &metav1.LabelSelector{MatchLabels: map[string]string{"k8s-app": EtcdGuardDeploymentName}},
+		},
+	}
+
+	if _, err := c.kubeClient.PolicyV1beta1().PodDisruptionBudgets(operatorclient.TargetNamespace).Create(ctx, pdb, metav1.CreateOptions{}); err != nil {
+		klog.Errorf("Failed to deploy %s, error: %v", pdb.Name, err)
```
hexfusion, line 213: failed to create PDB %q

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+// Get number of expected masters
+func (c *QuorumGuardController) getMastersReplicaCount(ctx context.Context) (int32, error) {
+	if c.replicaCount != 0 {
+		return int32(c.replicaCount), nil
+	}
+
+	klog.Infof("Getting number of expected masters from %s", clusterConfigName)
+	clusterConfig, err := c.kubeClient.CoreV1().ConfigMaps(clusterConfigNamespace).Get(ctx, clusterConfigName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Failed to get ConfigMap %s, err %v", clusterConfigName, err)
```
hexfusion, line 231: can we future proof returned errors by using `%w` vs `%v` please

pkg/operator/ceohelpers/bootstrap_test.go
```diff
@@ -74,25 +85,44 @@ var (
...
 		expectStrategy BootstrapScalingStrategy
+		infraObj       *configv1.Infrastructure
 	}{
 		"default should be HA": {
 			namespace:      defaultNamespace,
 			operatorConfig: defaultOperatorConfig,
 			expectStrategy: HAScalingStrategy,
+			infraObj:       defaultInfra,
 		},
 		"unsupported": {
```
hexfusion, line 109: followup: drop

pkg/operator/starter.go
```diff
@@ -2,9 +2,16 @@ package operator
...
 	"context"
+	"k8s.io/api/policy/v1beta1"
 	"os"
 	"time"
 
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/quorumguardcontroller"
+	"github.com/openshift/library-go/pkg/operator/events"
+	v1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
```
hexfusion, line 13: not: lets not create a new import group please add to existing one.

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,248 @@
...
+		klog.Errorf("Failed to deploy %s, error: %v", pdb.Name, err)
+		return err
+	}
+
+	klog.Infof("New etcd quorum guard pdb deployment is %s", pdb.Name)
+	recorder.Eventf("CreatedEtcdGuardPDBDeployment", "%s was created", EtcdGuardDeploymentName)
+	return nil
+}
+
+// Get number of expected masters
```
hexfusion, line 222: nit:  0:)
```
// getMastersReplicaCount ...
```

hexfusion (commented):
good progress lets start here mostly style and minor nits. But we do need to make sure that no one is messing with our managed resources for quorum-guard. If we observe that it needs to get reverted to the expected state. thanks!

pkg/operator/ceohelpers/bootstrap_test.go
hexfusion: no action required

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
tsorya: what do u mean? when operator restarts we will fill this value from replica count.

hexfusion: yeah disregard we have no way to persist this state unless we have an external observer and that is out of the scope of this PR

pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -68,8 +69,13 @@ func GetBootstrapScalingStrategy(staticPodClient v1helpers.StaticPodOperatorClie
...
 	_, hasDelayedHAAnnotation := etcdNamespace.Annotations[DelayedHABootstrapScalingStrategyAnnotation]
 
+	singleNode, err := IsSingleNodeTopology(infraLister)
+	if err != nil {
+		return strategy, fmt.Errorf("failed to get control plane topology: %w", err)
+	}
+
 	switch {
-	case isUnsupportedUnsafeEtcd:
+	case isUnsupportedUnsafeEtcd || singleNode:
```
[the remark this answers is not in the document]
hexfusion: no action required

pkg/operator/starter.go
tsorya: what does it mean?

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,237 @@
...
+	klog.Infof("Getting number of expected masters from %s", clusterConfigName)
+	clusterConfig, err := c.kubeClient.CoreV1().ConfigMaps(clusterConfigNamespace).Get(ctx, clusterConfigName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Failed to get ConfigMap %s, err %w", clusterConfigName, err)
+		return 0, err
+	}
+
+	rcD := replicaCountDecoder{}
+	if err := yaml.Unmarshal([]byte(clusterConfig.Data[clusterConfigKey]), &rcD); err != nil {
+		err := fmt.Errorf("%s key doesn't exist in ConfigMap %s", clusterConfigKey, clusterConfigName)
```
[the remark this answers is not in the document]
hexfusion: should we consider printing the original error vs clobbering and instead use klog.Errorf?

pkg/operator/quorumguardcontroller/quorumguardcontroller_test.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	fakecore "k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/cache"
+	"testing"
```
hexfusion, line 18: minor style in general I prefer to have core and non broke broken out to seperate blocks, just easier to read.

pkg/operator/starter.go
hexfusion: imports looks like this

```
        "context"
	"k8s.io/api/policy/v1beta1"
	"os"
	"time"

	"github.com/openshift/cluster-etcd-operator/pkg/operator/quorumguardcontroller"
	"github.com/openshift/library-go/pkg/operator/events"
	v1 "k8s.io/api/apps/v1"
	"k8s.io/apimachinery/pkg/util/wait"
	"k8s.io/klog/v2"

	"k8s.io/apimachinery/pkg/api/errors"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/client-go/kubernetes"

	configv1 "github.com/openshift/api/config/v1"
	configv1client "github.com/openshift/client-go/config/clientset/versioned"
	configv1informers "github.com/openshift/client-go/config/informers/externalversions"
	operatorversionedclient "github.com/openshift/client-go/operator/clientset/versioned"
	operatorv1informers "github.com/openshift/client-go/operator/informers/externalversions"
	"github.com/openshift/library-go/pkg/controller/controllercmd"
	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
	"github.com/openshift/library-go/pkg/operator/staleconditions"
	"github.com/openshift/library-go/pkg/operator/staticpod"
	"github.com/openshift/library-go/pkg/operator/staticpod/controller/revision"
	"github.com/openshift/library-go/pkg/operator/staticresourcecontroller"
	"github.com/openshift/library-go/pkg/operator/status"
	"github.com/openshift/library-go/pkg/operator/unsupportedconfigoverridescontroller"
	"github.com/openshift/library-go/pkg/operator/v1helpers"

	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
	"github.com/openshift/cluster-etcd-operator/pkg/etcdenvvar"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/bootstrapteardown"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/clustermembercontroller"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/configobservation/configobservercontroller"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/etcd_assets"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/etcdcertsigner"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/etcdendpointscontroller"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/etcdmemberscontroller"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/metriccontroller"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/resourcesynccontroller"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/scriptcontroller"
	"github.com/openshift/cluster-etcd-operator/pkg/operator/targetconfigcontroller
	```
	
	we already have logical groupings of imports can we choose one instead of making new? By block I mean space seperated.

**Review by hexfusion — COMMENTED** [review]
OK good progress, thank you I think these are mostly minor polishing nits.

Few next steps (not on you but if you can check these off it would be helpful):
- we want to test this manually have you done so? 
- upgrade test via cluster-bot.

File: pkg/operator/ceohelpers/control_plane_topology.go [hunk 18]
```diff
@@ -0,0 +1,34 @@
...
+	infraData, err := infraLister.Get(infrastructureClusterName)
+	if err != nil {
+		klog.Warningf("Failed to get infrastructure resource %s", infrastructureClusterName)
+		return "", err
+	}
+	// Added to make sure that iif ha mode is not set in infrastructure object
+	// we will get the default value configv1.FullHighAvailabilityMode
+	if infraData.Status.ControlPlaneTopology == "" {
+		klog.Infof("HA mode was not set in infrastructure resource setting it to default value %s", configv1.HighlyAvailableTopologyMode)
+		infraData.Status.ControlPlaneTopology = configv1.HighlyAvailableTopologyMode
```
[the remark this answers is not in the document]
tsorya: thought about the upgrade, it means that there is a possibility that we will be degrade for some time. Not sure if it is a big deal or not. The main problem is that we will delete previous etcd guard and new one will be added only after upgrade of infrastructure will do it's job.

pkg/operator/ceohelpers/control_plane_topology.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	return infraData.Status.ControlPlaneTopology, nil
+}
+
+func IsSingleNodeTopology(infraLister configv1listers.InfrastructureLister) (bool, error) {
+	topology, err := GetControlPlaneTopology(infraLister)
+	if err != nil {
+		return false, nil
```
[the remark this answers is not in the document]
tsorya: we return error in case we failed to get the resource

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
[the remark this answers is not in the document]
tsorya: i will wrap it sure, just missed it

hexfusion:
/approve

we need to get 4.8 deps merged and such and verify the upgrade plan.

pkg/operator/starter.go
tsorya: i was aligned with golang, and linter. didn't do it myself. I will go through

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,293 @@
...
+	// if restart occurred, we will apply PDB but if it is the same, nothing will happened
+	actual, modified, err := resourceapply.ApplyPodDisruptionBudgets(ctx, c.kubeClient.PolicyV1beta1(), pdb)
+	if err != nil {
+		klog.Errorf("Failed to verify/apply %s pdb, error %w", EtcdGuardDeploymentName, err)
+		return err
+	}
+
+	// if PDB was modified and is not managed by cvo log and event it
+	// if managed by cvo, delete it
+	if !c.findAndDeleteCVOManagedQuorumGuardPDB(ctx, actual, recorder) && modified {
```
hexfusion, line 230: I think we want to make this a little more level so maybe we check `findAndDeleteCVOManagedQuorumGuardPDB` first? If CVO managed resource exists then delete it event and return nil. This will result in another sync where we will see it is missing and apply. I am not sure if we care if CVO managed resource was altered we just want to delete it quickly. this way one resource action per sync, do you think this is worthwhile?

- sync
- check isCVOManaged()
-  if true delete and return

#### delete creates new sync
- sync
- check isCVOManaged()
- if false apply and event if modified.

hexfusion (commented):
Really close :) minor nits. I do have a question/possible sugestion about delete actions.

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
tsorya: I think that current code allows us to create less api calls and in case of freshly install cluster this part of code is just doing nothing without creating any api call. 
In case of upgrade, the first run will try to update and then delete, i think it is not a big issue. We will not have here loop of retries and race. It will happened only once. As i saw from my observations cluster version is the first one to be update and it stops monitoring etcd in this case. 
In my opinion it is not worth changing

pkg/operator/ceohelpers/control_plane_topology.go
```diff
@@ -0,0 +1,33 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
+	"k8s.io/klog/v2"
+)
+
+const infrastructureClusterName = "cluster"
+
+func GetControlPlaneTopology(infraLister configv1listers.InfrastructureLister) (configv1.TopologyMode, error) {
+	infraData, err := infraLister.Get(infrastructureClusterName)
```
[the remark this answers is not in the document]
tsorya: not really, it must be there. If it is not we will get error and fail but we can't proceed without reading it

bindata/etcd/quorumguard-deployment.yaml
```diff
@@ -3,10 +3,6 @@ kind: Deployment
 metadata:
   name: etcd-quorum-guard
   namespace: openshift-etcd
-  annotations:
-    include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/single-node-developer: "true"
-    exclude.release.openshift.io/internal-openshift-hosted: "true"
```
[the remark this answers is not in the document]
cfergeau: I think this was added for the IBM cloud usecase, @csrwng will most likely know ?

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -0,0 +1,293 @@
...
+		if managedField.Manager == "cluster-version-operator" {
+			deleteDeployment()
+			return true
+		}
+	}
+	return false
+}
+
+// ensureEtcdGuardPdb if etcd quorum guard PDB doesn't exist or ws changed, apply one
+func (c *QuorumGuardController) ensureEtcdGuardPdb(ctx context.Context, recorder events.Recorder) error {
```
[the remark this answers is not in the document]
tsorya: agree, changing

bindata/etcd/quorumguard-deployment.yaml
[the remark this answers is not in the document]
csrwng: Yes, these annotations are in support of [cluster profiles](https://github.com/openshift/enhancements/blob/master/enhancements/update/cluster-profiles.md). Every manifest that is included in the CVO release payload must at least have one `include` annotation.
The exclude annotation predates cluster profiles and is used to keep the manifest from getting applied to IBM ROKS clusters.

If this manifest won't be part of the release payload anymore (I see it's getting moved to bindata), then the annotations are no longer needed.

tsorya:
/test e2e-agnostic
--

yselkowitz:
This causes https://bugzilla.redhat.com/show_bug.cgi?id=1929359

PR #529, opened 2021-01-19
Add summary tables to tls asset docs

marun:


docs/etcd-tls-assets.md
```diff
@@ -22,6 +22,34 @@ See also the [user-facing
...
+
+| CA  (secrets)      | CA bundle  (configmaps)          | CA bundle also appearing in                  |
+| ------------------ | -------------------------------- | -------------------------------------------- |
+| etcd-signer        | etcd-ca-bundle                   | openshift-etcd                               |
+|                    |                                  | openshift-etcd-operator                      |
+|                    |                                  | openshift-etcd/etcd-peer-client-ca           |
+|                    | etcd-serving-ca                  | openshift-etcd                               |
+| etcd-metric-signer | etcd-metric-serving-ca           | openshift-etcd/etcd-metrics-proxy-client-ca  |
+|                    |                                  | openshift-etcd/etcd-metrics-proxy-serving-ca |
+| (bootstrap signer) | initial-etcd-signer              |                                              |
```
hexfusion, line 38: This really isn't a separate CA we just reuse[1] etcd-signer based TLS assets generated by the installer.

https://github.com/openshift/cluster-etcd-operator/blob/release-4.7/pkg/cmd/render/render.go#L129

hexfusion:
small nit otherwise :rocket: yay MOAR docs!

docs/etcd-tls-assets.md
marun: I screwed up the name, it's `initial-etcd-ca` not `signer` (fixed). I'm not sure what you mean by this not being a separate CA - it's a ca bundle (i.e. stored in a configmap) and presumably used to validate certs issued by the installer CA. Any ideas why it still exists after install?

hexfusion: gotcha yeah it was cruft ported over from KAS-O. we removed it in 4.7[1]

[1] https://github.com/openshift/cluster-etcd-operator/pull/501

hexfusion: So we can remove that line then, or mark removed after 4.7?

marun: Done

marun:
@hexfusion Can we just add bugzilla/valid-bug? Having to add a tracking ug for a non-functional doc addition that requires zero qa oversight does not seem reasonable.

hexfusion:
manually adding BZ tag as this is non code change.

PR #531, opened 2021-01-20
[wip] test enable tls1.3 support for etcd

hexfusion:
etcd is compiled with golang 1.12 which does not enable tls 1.3 by default this is a test for enabling it.[1]

[1] https://golang.org/doc/go1.12#tls_1_3


hexfusion:
env should probably be baked into the etcd container but this will test that it works as expected.

hexfusion:
infra

/retest

PR #532, opened 2021-01-20
Bug 1907872: Make dual stack bootstrapping more reliable

ironcladlou:
Due to a parsing bug, the bootstrap rendering logic fails to detect a usable
machine network CIDR when using IPv6 dual stack mode unless the IPv4 CIDR is the
first element in the install config's machineNetwork array, which is brittle.

This patch fixes the parsing mistake so that the IPv4 address can be located
amongst the machineNetwork CIDRs in dual stack mode.

Although ideally etcd would detect and bind to the "preferred" family and that
family would be again located here during bootstrap CIDR detection, such a
change would be more invasive. The scope of this fix is to get dual stack
reliably working even if it means using IPv4 inconsistently during bootstrapping
when IPv6 would be more consistent.

### DIFF
--- a/pkg/cmd/render/render_test.go
+++ b/pkg/cmd/render/render_test.go
@@ -16,6 +16,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ghodss/yaml"
+
 	"github.com/openshift/cluster-etcd-operator/pkg/cmd/render/options"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/cmd/render/render_test.go
```diff
@@ -488,3 +490,105 @@ type fakeBootstrapIPLocator struct {
...
+
+const installConfigDualStack = `
+apiVersion: v1
+metadata:
+  name: my-cluster
+networking:
+  clusterNetwork:
+  - cidr: 10.128.0.0/14
+    hostPrefix: 23
+  machineCIDR: 10.0.0.0/16
```
marun, line 519: What is the relationship between `networking.machineCIDR` and `networking.machineNetwork[x].cidr`?

pkg/cmd/render/render_test.go
```diff
@@ -488,3 +490,105 @@ type fakeBootstrapIPLocator struct {
...
+			installConfig:     installConfigSingleStackIPv6,
+			isSingleStackIPv6: true,
+			expectedCIDR:      "2620:52:0:1302::/64",
+			expectedErr:       nil,
+		},
+	}
+
+	for name, test := range tests {
+		t.Logf("evaluating test %q", name)
+		var installConfig map[string]interface{}
```
marun, line 575: Huh, so rendering forces us to relinquish type safety? I would have thought the struct would be vendored into o/api for reuse.

pkg/cmd/render/render_test.go
ironcladlou: the former is deprecated and used only in a fallback case

pkg/cmd/render/render_test.go
ironcladlou: The install config isn't public API, and vendoring the installer to get the type drags in a huge transitive dependency tree. In other projects I've worked on which consume the install config, I've created a local struct for the subset of installconfig I wanted to deserialize to get some type safety, we could do that here as well

marun: There's no need for a huge transitive dependency tree. That's what go mod is for.

ironcladlou: Can you elaborate? Maybe I'm misunderstanding.

Here's where the `InstallConfig` struct is defined: https://github.com/openshift/installer/blob/master/pkg/types/installconfig.go

Since that package isn't itself a nested module, getting the type in the etcd-operator tree requires (I believe) declaring a dependency on `github.com/openshift/installer` (the module root). By doing that, and running `go mod vendor`, I run into problems immediately as various dependencies I don't want fail to resolve for one reason or another:

```
go: github.com/openshift/installer@v0.9.0-master.0.20210120203138-27fd27be1a03 requires
        github.com/kubevirt/terraform-provider-kubevirt@v0.0.0-00010101000000-000000000000: invalid version: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/*
```

I haven't debugged further to get to a point where `go mod graph` will succeed, but even if go mod is smart enough to drag in only the dependencies rooted from that sub package in this case (not at all obvious to me), it looks like a lot (off the bat you can see that some field of InstallConfig is going to drag in terraform provider code).

Seems like a pretty deep rabbit hole to explore versus re-declaring a small subset of the struct or using `interface{}`.

I wish this were easier.

marun: What's preventing the installer from moving the struct definition to a location that doesn't involve ancillary dependencies?

ironcladlou: I can think of several possible reasons, but a fundamental one is that (I think) the type is considered internal. And now that I've said it, I can imagine we'll be asked a question I actually can't answer yet: can we replace our usage of the internal type with a public API? I'm looking around on a bootstrap node right now to inspect rendered manifests during bootstrapping, and can't yet identify a public OpenShift config API which specifies the machine networks that come in through install config.

hexfusion:
@danwinship PTAL

pkg/cmd/render/render_test.go
ironcladlou: @danwinship do you have any idea if there's some API resource I'm missing we could be using instead of the install config to get the machine network CIDRs?

marun: FWIW this is an overarching concern, shouldn't block merge.

hexfusion:
overall LGTM I would like Networking to sign off this works as expected.

/assign @danwinship

pkg/cmd/render/render_test.go
danwinship: @ironcladlou I don't know; team-sdn doesn't own/use `machineNetwork` (it's the odd man out in the `networking` stanza of install-config).

I'd expect it to get copied into something machine-api related, but I'm not seeing it...

pkg/cmd/render/render_test.go
```diff
@@ -488,3 +490,105 @@ type fakeBootstrapIPLocator struct {
...
+
+func Test_getMachineCIDR(t *testing.T) {
+	tests := map[string]struct {
+		installConfig     string
+		isSingleStackIPv6 bool
+		expectedCIDR      string
+		expectedErr       error
+	}{
+		"ipv4 should ignore the ipv6 cidr": {
+			installConfig:     installConfigSingleStackIPv4,
```
danwinship, line 554: description is bad; there is no ipv6 cidr to be ignored here

pkg/cmd/render/render_test.go
```diff
@@ -488,3 +490,105 @@ type fakeBootstrapIPLocator struct {
...
+		"single stack ipv6 should locate the ipv6 cidr": {
+			installConfig:     installConfigSingleStackIPv6,
+			isSingleStackIPv6: true,
+			expectedCIDR:      "2620:52:0:1302::/64",
+			expectedErr:       nil,
+		},
+	}
+
+	for name, test := range tests {
+		t.Logf("evaluating test %q", name)
```
danwinship, line 574: the cool modern way is to do

    t.Run(name, func(t *testing.T) {
            ...
    })

danwinship (commented):
/lgtm
/hold
non-urgent comments below; cancel either the lgtm or the hold

pkg/cmd/render/render_test.go
danwinship: Do we still have the infra to do staging? The installer could move the install-config API to a staging directory and then let it get mirrored to an "install-api" repo for CEO (and other early-install-time operators?) to consume.

pkg/cmd/render/render_test.go
ironcladlou: d'oh, fixed (and also tried to improve all the descriptions)

pkg/cmd/render/render_test.go
ironcladlou: Nice, thanks for the tip. Refactored

ironcladlou:
Sorry, I used the wrong yaml package and inadvertently changed the module definition. I fixed the import to resolve the module errors.

PR #535, opened 2021-01-27
Bug 1920159: CPU requests overstate actual needs

smarterclayton:
In production environments etcd is over requested for actual use.
The bulk of that over-requesting comes from the sidecars which
overstate their actual needs. Adjust them based on the calculations
from prod clusters to request roughly 100m less CPU than before.

/hold

While we review these changes together


smarterclayton:
/hold cancel

Combined tests looked ok.  Up to sam to decide when he's comfortable (if he needs extra time)

PR #536, opened 2021-02-02
Bug 1924913: A valid etcd endpoint should have the URL scheme prepended.

retroflexer:
A valid etcd endpoint should have a valid URL scheme prefixed. 

The only reason it continues to work is because etcd client was disregarding this setting, and getting a subconn based on the environment variables.


marun:
@retroflexer Is there any testing for this?

retroflexer:
> @retroflexer Is there any testing for this?

This was a one-off feature, and hence we didn't write any tests.

PR #539, opened 2021-02-09
WIP Key etcd certs by uid instead of name

marun:
**NOTE** This PR is not ready for merge until openshift/library-go/pull/991 merges

Identifying etcd member certs by node uid rather than node name ensures that node deletion followed by addition with the same name can be detected as a member change.


marun:
moar disruptive signal!

Issue #487, opened 2020-11-04
Check the validity of certificates


retroflexer:
Currently cert-signer controller only checks for the presence of the secrets but doesn't validate it.

What if:

- the secret is missing either the cert or the key?
- the cert and/or key can't be parsed
- the cert and key are not related?

It may be useful to create the validation of secrets as a utility function and move it to library-go for a wider adoption.

hexfusion:
@marun can this be closed?

PR #540, opened 2021-02-12
Improve cert controller detection and correction of invalid certs

marun:
This refactor enables regenerating certs for nodes that have been removed and added back with the same name, and simplifies the addition of further checks for cert lineage and expiry.

/cc @hexfusion @retroflexer

### DIFF
--- a/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
+++ b/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
@@ -3,7 +3,10 @@ package etcdcertsigner
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -14,7 +17,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
-	"k8s.io/client-go/util/cert"
+	"k8s.io/klog/v2"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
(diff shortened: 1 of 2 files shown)


pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -193,108 +218,151 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+			// etcd cluster membership change may be required. The operator
+			// doesn't currently handle this.
+			return "", utilerrors.NewAggregate(errs)
+		} else {
+			// A different node uid indicates node removal and addition with the
+			// same name. etcd on the node will be a new cluster member and
+			// needs a new certificate.
+			msgs := []string{}
+			for _, err := range errs {
+				msgs = append(msgs, fmt.Sprintf("%v", err))
```
hexfusion, line 356: what does the error look like here it seems that explicitly calling out the fact that node UID is missing OR changed might be useful for debugging? It would be expected that on an upgrade every cert would be rotated. If this is expected we don't want to scare folks with logs. Curious about your thoughts here.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -193,108 +218,151 @@ func (c *EtcdCertSignerController) syncAllMasters(recorder events.Recorder) erro
...
+		return err
+	}
 
-	errs := []error{}
-	for _, secretName := range secretNames {
-		secret, err := c.secretLister.Secrets(operatorclient.TargetNamespace).Get(secretName)
-		if errors.IsNotFound(err) {
+	nodeUIDUnchanged := false
+	if secret != nil {
+		nodeUIDUnchanged = string(node.UID) == secret.Annotations[nodeUIDAnnotation]
```
hexfusion, line 254: we have 2 possible conditions here do we need them handled separately wrt to error message?
- no annotation found (possible upgrade and expected)
- uid does not match

hexfusion (commented):
@marun great progress thank you. I have some questions about upgrade cases (annotation missing) otherwise looks good.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hexfusion: I will say that if we can prove its safe (should be) rotation of these certs in 4.8 would not be a bad thing 0:)

hexfusion:
Closes https://github.com/openshift/cluster-etcd-operator/issues/487

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
marun: Yes, this is definitely in the direction of rotation support.

pkg/operator/etcdcertsigner/etcdcertsignercontroller_test.go
```diff
@@ -1,300 +1,243 @@
...
+		"invalid bytes": {
+			invalidCertPair:  true,
+			expectedRegenMsg: true,
+		},
+		"missing ip address; node uid unchanged": {
+			certIPAddresses: ipAddresses,
+			nodeIPAddresses: differentIpAddresses,
+			expectedErr:     true,
+		},
+		"missing ip address; node uid changed": {
```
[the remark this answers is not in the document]
marun: I was optimistically assuming that upgrade to 4.8 would always be seamless such that the operator would apply the current node uid before any membership changes occurred. No such guarantees in this world, though, so as per your comment I've updated the validation and testing to consider a missing stored uid as if the uid had not changed (resulting in an error if the cert isn't valid for the current node's ip addresses).

hexfusion:
upgrade job shows annotations properly added. LGTM
```
  kind: Secret
  metadata:
    annotations:
      etcd-operator.alpha.openshift.io/cert-secret-node-uid: e42e58f1-bcd8-4f1d-bff9-a253df219ced
```

marun:
Squashed for merge.

PR #543, opened 2021-02-16
Bug 1929359: pkg/operator/quorumguardcontroller: use cli image from release payload

hexfusion:
During the migration of quorum-guard from CVO to controller managed we forgot to populate the image with an explicit image ref from the release payload.

### DIFF
--- a/manifests/0000_12_etcd-operator_06_deployment.yaml
+++ b/manifests/0000_12_etcd-operator_06_deployment.yaml
@@ -57,6 +57,8 @@ spec:
           value: quay.io/openshift/origin-etcd
         - name: OPERATOR_IMAGE
           value: quay.io/openshift/origin-cluster-etcd-operator
+        - name: CLI_IMAGE
+          value: quay.io/openshift/origin-cli
         - name: OPERATOR_IMAGE_VERSION
           value: "0.0.1-snapshot"
         - name: OPERAND_IMAGE_VERSION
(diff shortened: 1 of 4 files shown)


hexfusion:
cc @tsorya

hexfusion:
cc @yselkowitz

manifests/0000_12_etcd-operator_06_deployment.yaml
```diff
@@ -57,6 +57,8 @@ spec:
           value: quay.io/openshift/origin-etcd
         - name: OPERATOR_IMAGE
           value: quay.io/openshift/origin-cluster-etcd-operator
+        - name: CLI_IMAGE
+          value: quay.io/openshift/cli
```
yselkowitz, line 61: value: quay.io/openshift/**origin-**cli:latest

IIUC this needs to match exactly the value in manifests/image-references

yselkowitz (changes requested):

manifests/0000_12_etcd-operator_06_deployment.yaml
hexfusion: aligned image-references with the rest of the images.

yselkowitz: still missing the origin- prefix here

retroflexer:
Adding back the lgtm tag, as it appears @hexfusion closed the PR by mistake, which resulted in removing the lgtm tag.

/lgtm

PR #544, opened 2021-02-18
Manage one aggregated cert secret rather than one per cert type

marun:
Writing a single secret ensures that the result of node addition or cert rotation will trigger at most a single rollout. If more than one aggregated secret needs to be written, there exists the possibility that the static pod controller could start a rollout before all secrets had been written.

/cc @hexfusion @retroflexer

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"context"
 	"os"
+	"strings"
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -23,7 +24,9 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
 
 	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
 	"github.com/openshift/cluster-etcd-operator/pkg/etcdenvvar"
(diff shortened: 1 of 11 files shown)


marun:
Ready for review.

marun:
/retest

Disruptive test is unrelated failure.

pkg/operator/starter.go
```diff
@@ -247,6 +250,35 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+		if err != nil {
+			klog.Warningf("Failed to list secrets when searching for 4.7 secrets to remove: %v", err)
+		}
+		if secrets == nil {
+			return
+		}
+		for _, secret := range secrets.Items {
+			for _, prefix := range secretNamePrefixes {
+				if strings.HasPrefix(secret.Name, prefix) {
+					err := secretClient.Delete(ctx, secret.Name, metav1.DeleteOptions{})
```
hexfusion, line 273: Not sure if we want to do this. Should we mark the revision abandoned or something so pruner just cleans it up? I wonder why keep any assets for a revision if it is not complete.

hexfusion (commented):
looks good just one question to consider otherwise LGTM

pkg/operator/starter.go
marun: These secrets only aggregate certs sourced from node-specific secrets. In the event of a downgrade, these 'all' secrets would just be recreated by the 4.7 operator. I'm not sure I understand the concern, can you be more specific?

hexfusion: Maybe I am just misunderstanding the code here is my concern.

so each revision creates a versioned secret.

```
etcd-all-peer                                                     Opaque                                6      29m
etcd-all-peer-2                                                   Opaque                                6      29m
etcd-all-peer-3                                                   Opaque                                6      21m
etcd-all-serving                                                  Opaque                                6      29m
etcd-all-serving-2                                                Opaque                                6      29m
etcd-all-serving-3                                                Opaque                                6      21m
etcd-all-serving-metrics                                          Opaque                                6      29m
etcd-all-serving-metrics-2                                        Opaque                                6      29m
etcd-all-serving-metrics-3                                        Opaque                                6      21m
```
My concern is going through these old revisions and removing state from them vs having the unused revision completely removed/pruned. Let me reread the code.

marun: How is the pruner going to remove revisions for resources it is no longer tracking?

PR #545, opened 2021-02-23
WIP: manifests/0000_12_etcd-operator_07_clusteroperator: Touch relatedObjects

wking:
I'm pretty sure the CVO will not update this for you, either by adding new entries or removing old entries on updates.  The CVO will create the ClusterOperator for you if it doesn't exist, but after that, managing it is up to the 2nd-level operator.

### DIFF
--- a/manifests/0000_12_etcd-operator_07_clusteroperator.yaml
+++ b/manifests/0000_12_etcd-operator_07_clusteroperator.yaml
@@ -27,5 +27,5 @@ status:
       name: openshift-etcd-operator
       resource: namespaces
     - group: ""
-      name: openshift-etcd
+      name: the-cvo-will-not-add-this-for-you
       resource: namespaces


marun:
afaict the `openshift-etcd` namespace is currently created by the installer. I'm working on moving responsibility for the namespace to ceo so that it will render it on bootstrap. I'm guessing the namespace should also be added as a ceo manifest so the cvo can recreate it on removal? I haven't taken the time to potentially sacrifice a cluster to check whether namespace removal wouldn't remove all etcd pod (I'm a bit fuzzy on how that works for static pods).

wking:
No change on update:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/545/pull-ci-openshift-cluster-etcd-operator-master-e2e-agnostic-upgrade/1364007864661708800/artifacts/e2e-agnostic-upgrade/gather-extra/artifacts/clusteroperators.json
| jq -r '.items[] | select(.metadata.name ==
"etcd").status.relatedObjects[] | tostring'
{"group":"operator.openshift.io","name":"cluster","resource":"etcds"}
{"group":"","name":"openshift-config","resource":"namespaces"}
{"group":"","name":"openshift-config-managed","resource":"namespaces"}
{"group":"","name":"openshift-etcd-operator","resource":"namespaces"}
{"group":"","name":"openshift-etcd","resource":"namespaces"}
```

Demonstrating the etcd-operator's lack of relatedObject reconciliation.

> afaict the openshift-etcd namespace is currently created by the installer.

This demonstration PR is just talking about `relatedObjects` values on the ClusterOperator.  Who (re)creates the referenced namespaces is orthogonal.

/close

PR #546, opened 2021-02-24
Migrate tls bootstrap from the installer

marun:
Migrate tls bootstrap from the installer to ensure that changes to bootstrap, recovery or rotation of tls assets can be validated directly in ceo CI without requiring changes to the installer.

/cc @hexfusion @retroflexer

### DIFF
--- a/bindata/bootkube/manifests/openshift-etcd-svc.yaml
+++ b/bindata/bootkube/manifests/openshift-etcd-svc.yaml
@@ -1,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: etcd
+  namespace: openshift-etcd
+  labels:
+    # this label is used to indicate that it should be scraped by prometheus
+    k8s-app: etcd
+spec:
+  selector:
+    k8s-app: etcd
+  ports:
+  - name: etcd
+    port: 2379
+    protocol: TCP
+  - name: etcd-metrics
+    port: 9979
+    protocol: TCP
--- a/pkg/cmd/render/certs.go
+++ b/pkg/cmd/render/certs.go
@@ -1,0 +1,69 @@
+package render
+
+import (
+	"crypto/x509/pkix"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/openshift/library-go/pkg/crypto"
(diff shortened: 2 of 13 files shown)


marun:
The required installer changes have merged, so it should be possible for CI to pass.

marun:
/retry

Given that this PR only impacts install, and an upgrade doesn't install with this code (it installs a previous release), I can't see how this upgrade failure is due to this PR.

pkg/cmd/render/certs.go
```diff
@@ -0,0 +1,69 @@
...
+	ca := &crypto.CA{
+		Config:          signerCAConfig,
+		SerialGenerator: &crypto.RandomSerialGenerator{},
+	}
+	clientUser := &user.DefaultInfo{
+		Name:   clientName,
+		UID:    uuid.New().String(),
+		Groups: []string{clientName},
+	}
+	clientCertDuration := time.Hour * 24 * 365 * 10 // 10 years
```
hexfusion, line 52: should 10 years be a constant or at least use the same definition?

https://github.com/openshift/cluster-etcd-operator/pull/546/files#diff-28893b89756831caa4aa8af970a87b34173bdfe9a6edb86daaf948b7a6bc03d3R26

hexfusion (commented):
overall LGTM few questions

bindata/bootkube/manifests/openshift-etcd-svc.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
```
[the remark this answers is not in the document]
marun: I don't know if this is even required let alone in this form, but we have no way of knowing until this PR merges, https://github.com/openshift/installer/pull/4691 merges to remove etcd manifest generation, and we can try removing this entirely.

pkg/cmd/render/certs.go
marun: re: reuse I'm assuming you mean 10 years in days? I'm usually pretty DRY but I figured inlining a single-use definition would be easier to read.

Note that this long duration is unlikely to persist given our intent to support rotation.

PR #547, opened 2021-02-24
Bug 1931658: render: refactor scaling strategies

hexfusion:
Currently, the bootstrap in place used with SNO adds an annotation `alpha.installer.openshift.io/etcd-bootstrap` which is intended to be used by the cluster to communicate with the bootstrap etcd instance. But bootstrap in place means that the bootstrap node will never be visible. The result is the cluster never pivots from the bootstrap node endpoint which is not expected. This change fixes that problem by adding the `BootstrapInPlaceStrategy` which omits that annotation.

- [x]  refactor scaling strategies and add BootstrapInPlaceStrategy
- [x] add unit tests

### DIFF
--- a/bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
+++ b/bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
@@ -3,5 +3,10 @@ kind: ConfigMap
 metadata:
   name: etcd-endpoints
   namespace: openshift-etcd
+{{- if ne .BootstrapScalingStrategy "BootstrapInPlaceStrategy" }}
   annotations:
     alpha.installer.openshift.io/etcd-bootstrap: {{ .BootstrapIP }}
+{{- else }}
+data:
+  {{ .EtcdEndpointConfigmapData }}
+{{- end }}
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -1,6 +1,7 @@
 package render
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -166,6 +167,10 @@ type TemplateData struct {
 	// BootstrapIP is address of the bootstrap node.
 	BootstrapIP string
 
+	// BootstrapScalingStrategy describes the invariants which will be enforced when
+	// scaling the etcd cluster.
+	ceohelpers.BootstrapScalingStrategy
+
 	// Platform is the underlying provider the cluster is run on.
 	Platform string
 
@@ -174,6 +179,9 @@ type TemplateData struct {
 
 	// NamespaceAnnotations are addition annotations to apply to the etcd namespace.
 	NamespaceAnnotations map[string]string
+
+	// EtcdEndpointConfigmapData is an optional data field used by etcd-endpoints configmap.
+	EtcdEndpointConfigmapData string
 }
 
 type StaticFile struct {
@@ -245,17 +253,26 @@ func newTemplateData(opts *renderOpts) (*TemplateData, error) {
 	// assume that this is >4.2
 	templateData.Platform = string(infra.Status.PlatformStatus.Type)
 
+	if err := templateData.setBootstrapStrategy(installConfig, opts.delayedHABootstrapScalingStrategyMarker); err != nil {
+		return nil, err
+	}
+
 	if err := templateData.setComputedEnvVars(templateData.Platform); err != nil {
 		return nil, err
 	}
 
-	// Use a marker file to configure the bootstrap scaling strategy.
-	if _, err := os.Stat(opts.delayedHABootstrapScalingStrategyMarker); err == nil {
-		if templateData.NamespaceAnnotations == nil {
-			templateData.NamespaceAnnotations = map[string]string{}
+	// If bootstrap scaling strategy is delayed HA set annotation signal
+	if templateData.BootstrapScalingStrategy == ceohelpers.DelayedHAScalingStrategy {
+		templateData.NamespaceAnnotations = map[string]string{
+			ceohelpers.DelayedHABootstrapScalingStrategyAnnotation: "",
 		}
-		templateData.NamespaceAnnotations[ceohelpers.DelayedHABootstrapScalingStrategyAnnotation] = ""
-		klog.Infof("using delayed HA bootstrap scaling strategy due to presence of marker file %s", opts.delayedHABootstrapScalingStrategyMarker)
+	}
+
+	// If bootstrap scaling strategy is in place set endpoint data
+	if templateData.BootstrapScalingStrategy == ceohelpers.BootstrapInPlaceStrategy {
+		// base64 encode the ip address and use it as unique data key to emulate endpoint controller
+		templateData.EtcdEndpointConfigmapData = fmt.Sprintf("%s: %s",
+			base64.StdEncoding.WithPadding(base64.NoPadding).EncodeToString([]byte(templateData.BootstrapIP)), templateData.BootstrapIP)
 	}
 
 	// Signer key material
@@ -552,6 +569,16 @@ func (t *TemplateData) setComputedEnvVars(platform string) error {
 	return nil
 }
 
+func (t *TemplateData) setBootstrapStrategy(installConfig map[string]interface{}, delayedHAMarkerFile string) error {
+	bootstrapStrategy, err := getBootstrapScalingStrategy(installConfig, delayedHAMarkerFile)
+	if err != nil {
+		return err
+	}
+	klog.Infof("Bootstrapping etcd using: %q", bootstrapStrategy)
+	t.BootstrapScalingStrategy = bootstrapStrategy
+	return nil
+}
+
 func writeManifests(outputDir, templateDir string, templateData interface{}) error {
 	assetPaths := map[string]string{
 		// The etc-kubernetes path will be copied recursively by the installer to /etc/kubernetes/
@@ -642,4 +669,29 @@ func getInfrastructure(file string) (*configv1.Infrastructure, error) {
 		return nil, err
 	}
 	return config, nil
+}
+
+func getBootstrapScalingStrategy(installConfig map[string]interface{}, delayedHAMarkerFile string) (ceohelpers.BootstrapScalingStrategy, error) {
+	// Delayed HA strategy is set if marker file exists on disk.
+	if _, err := os.Stat(delayedHAMarkerFile); err == nil {
+		return ceohelpers.DelayedHAScalingStrategy, nil
+	}
+
+	controlPlane, found := installConfig["controlPlane"].(map[string]interface{})
+	if !found {
+		return "", fmt.Errorf("unrecognized data structure in controlPlane field")
+	}
+	replicaCount, found := controlPlane["replicas"].(float64)
+	if !found {
+		return "", fmt.Errorf("unrecognized data structure in controlPlane replica field")
+	}
+
+	// Bootstrap in place strategy when bootstrapInPlace root key exists in the install-config
+	// and controlPlane replicas is 1.
+	if _, found := installConfig["bootstrapInPlace"]; found && int(replicaCount) == 1 {
+		return ceohelpers.BootstrapInPlaceStrategy, nil
+	}
+
+	// HA "default".
+	return ceohelpers.HAScalingStrategy, nil
 }
--- a/pkg/cmd/render/render_test.go
+++ b/pkg/cmd/render/render_test.go
@@ -7,9 +7,11 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"reflect"
 	"testing"
 
 	"github.com/ghodss/yaml"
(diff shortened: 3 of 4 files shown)


marun:
Why is it desirable to switch to defining the manifests in code? Is there a limitation of template substitution that suggests this change?

hexfusion:
> Why is it desirable to switch to defining the manifests in code? Is there a limitation of template substitution that suggests this change?

In the case of conditional changes against the resource using the raw object seems more natural vs rendering template variables with conditional logic. We could use a base object and append then rewrite to disk but it feels strange.

hexfusion:
> Multi-AZ Clusters should spread the pods of a replication controller across zones

agnostic fail looks like https://bugzilla.redhat.com/show_bug.cgi?id=1929389

hexfusion:
> e2e-aws-single-node

This test is not setting the `bootstrapInPlace`[1] in the install-config.

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/547/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-single-node/1365064469167738880/artifacts/e2e-aws-single-node/gather-extra/artifacts/configmaps.json

cc @romfreiman @tsorya

pkg/cmd/render/render_test.go
```diff
@@ -593,3 +671,61 @@ func Test_getMachineCIDR(t *testing.T) {
...
+  - foo: bar
+    cidr: 2620:52:0:1302::/64
+  networkType: OpenShiftSDN
+  serviceNetwork:
+  - 172.30.0.0/16
+bootstrapInPlace:
+      installationDisk: /dev/sda
+`
+
+func Test_getBootstrapScalingStrategy(t *testing.T) {
```
marun, line 693: (No action required) Given the simplicity of the function under test - 2 conditionals and a return statement constituting 5 lines of code - I'm not sure this level of testing is suggested.

pkg/cmd/render/render_test.go
```diff
@@ -214,6 +264,34 @@ func TestRenderIpv4(t *testing.T) {
 	testRender(config)
 }
 
+func TestRenderScalingStrategyBootstrapInPlace(t *testing.T) {
```
marun, line 267: (No action required) Maybe factor out the test configuration common to this new test and TestRenderIpv4?

pkg/cmd/render/render_test.go
```diff
@@ -173,6 +174,55 @@ data:
         projectID: openshift
         region: us-east1
     publish: External
+`
```
marun, line 177: (No action required) Maybe append the bootstrap-in-place entry to `clusterConfigMap`?

bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
```diff
@@ -2,6 +2,10 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etcd-endpoints
-  namespace: openshift-etcd
+  namespace: openshift-etcd {{ if not .BootstrapScalingStrategy.InPlace }}
   annotations:
     alpha.installer.openshift.io/etcd-bootstrap: {{ .BootstrapIP }}
+{{ end }} {{ if .BootstrapScalingStrategy.InPlace }}
```
marun, line 8: (No action required) Maybe start the new substitution block on the next line?

bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
```diff
@@ -2,6 +2,10 @@ apiVersion: v1
...
+data:
+  {{ .EtcdEndpointConfigmapData }}
```
marun, line 10: (No action required) Note that it's possible to do base64 encoding in the template:


```yaml
data:
  # Use base64-encoded ip address as the key to emulate endpoint controller
  {{ .BootstrapIP | base64 }}: {{ .BootstrapIP }}
```

marun (commented):
/lgtm

Squash for merge?

romfreiman:
> > e2e-aws-single-node
> 
> This test is not setting the `bootstrapInPlace`[1] in the install-config.
> 
> [1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/547/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-single-node/1365064469167738880/artifacts/e2e-aws-single-node/gather-extra/artifacts/configmaps.json
> 
> cc @romfreiman @tsorya

Right. It uses bootstrap vm on aws. 
Bootstrap in place is used only with the e2e-metal-live-iso test

bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
hexfusion: it seemed to inject a newline but I want to play with that so its cleaner I agree.

bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
```diff
@@ -2,6 +2,10 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etcd-endpoints
-  namespace: openshift-etcd
+  namespace: openshift-etcd {{ if not .BootstrapScalingStrategy.InPlace }}
```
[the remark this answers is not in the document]
hexfusion: yeah, that created newline whitespace let me see if I can workaround it.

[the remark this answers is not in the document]
hexfusion: TIL newline suppression...

```
{{- if not .BootstrapScalingStrategy.InPlace }}
```

bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
hexfusion: fixed

bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
hexfusion: nice trick, sold

hexfusion: I see KAS-O does this with the load command but inline it fails. Since it works I am inclined to leave it as is and research for further optimization.

```
 <base64>: wrong type for value; expected []uint8;
```

hexfusion: maybe when you gut the stuff we don't need in render we can add sprig which has some nice helpers.

https://github.com/Masterminds/sprig

hexfusion: adding TODO

pkg/cmd/render/render.go
```diff
@@ -362,6 +367,19 @@ func writeCertKeyFiles(outputDir, allSecretName, nodeSecretName string, certData
 	return nil
 }
 
+func getBootstrapScalingStrategy(installConfig map[string]interface{}, delayedHAMarkerFile string) (ceohelpers.BootstrapScalingStrategy, error) {
+	// Delayed HA strategy is set if marker file exists on disk.
+	if _, err := os.Stat(delayedHAMarkerFile); err == nil {
+		return ceohelpers.DelayedHAScalingStrategy, nil
+	}
+	// Bootstrap in place when `bootstrapInPlace` root key exists in the install-config.
+	if _, found := installConfig["bootstrapInPlace"]; found == true {
```
tsorya, line 376: didn't you wanted to validate control plane replica == 1 ?

hexfusion: is this necessary? it probably is better to be safe will update.

hexfusion:
infra
/retest

bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
```diff
@@ -3,5 +3,11 @@ kind: ConfigMap
 metadata:
   name: etcd-endpoints
   namespace: openshift-etcd
+{{- if ne .BootstrapScalingStrategy "BootstrapInPlaceStrategy" }}
   annotations:
     alpha.installer.openshift.io/etcd-bootstrap: {{ .BootstrapIP }}
+{{- end }}
```
marun, line 9: Maybe simplify to `{{- else }}` ?

pkg/cmd/render/render.go
```diff
@@ -245,17 +253,26 @@ func newTemplateData(opts *renderOpts) (*TemplateData, error) {
...
 	if err := templateData.setComputedEnvVars(templateData.Platform); err != nil {
 		return nil, err
 	}
 
-	// Use a marker file to configure the bootstrap scaling strategy.
-	if _, err := os.Stat(opts.delayedHABootstrapScalingStrategyMarker); err == nil {
-		if templateData.NamespaceAnnotations == nil {
-			templateData.NamespaceAnnotations = map[string]string{}
+	// If bootstrap scaling strategy is delayed HA append annotation signal
+	if templateData.BootstrapScalingStrategy == ceohelpers.BootstrapInPlaceStrategy {
```
marun, line 265: Maybe maybe merge the conditional blocks for setting the annotation and the config map data since they seem to key on the same check?

pkg/cmd/render/render.go
```diff
@@ -552,6 +569,16 @@ func (t *TemplateData) setComputedEnvVars(platform string) error {
 	return nil
 }
 
+func (t *TemplateData) setBootstrapStrategy(installConfig map[string]interface{}, delayedHAMarkerFile string) error {
+	boostrapStrategy, err := getBootstrapScalingStrategy(installConfig, delayedHAMarkerFile)
```
marun, line 573: nit: s/boostrapStrategy/bootstrapStrategy/

pkg/cmd/render/render.go
```diff
@@ -643,3 +670,24 @@ func getInfrastructure(file string) (*configv1.Infrastructure, error) {
 	}
 	return config, nil
 }
+
+func getBootstrapScalingStrategy(installConfig map[string]interface{}, delayedHAMarkerFile string) (ceohelpers.BootstrapScalingStrategy, error) {
+	// Delayed HA strategy is set if marker file exists on disk.
+	if _, err := os.Stat(delayedHAMarkerFile); err == nil {
+		return ceohelpers.DelayedHAScalingStrategy, nil
+	}
+	if _, found := installConfig["controlPlane"]; !found {
```
marun, line 679: Maybe unmarshall installConfig to its golang struct to ensure type safety?

pkg/cmd/render/render_test.go
```diff
@@ -274,6 +315,23 @@ func TestTemplateDataIpv4(t *testing.T) {
 	testTemplateData(config)
 }
 
+func TestRenderScalingStrategyBootstrapInPlace(t *testing.T) {
+	want := TemplateData{
+		BootstrapScalingStrategy:  ceohelpers.BootstrapInPlaceStrategy,
+		EtcdEndpointConfigmapData: "MTAuMC4wLjE: 10.0.0.1",
```
marun, line 321: (No action required) What's the logic behind these obfuscated keys?

hexfusion: it was an implementation detail with the endpoint controller functionally the idea is to generate a unique key by base64 encoding the IP.

pkg/cmd/render/render.go
hexfusion: yeah, the entire render should get reworked.

pkg/cmd/render/render.go
hexfusion: ugh boy :man_facepalming:

hexfusion: Thas actually a typo should be `DelayedHAScalingStrategy`

pkg/cmd/render/render.go
hexfusion: actually looking at this I will clean it up a bit but we would need to import the installer because it is not an actual openshift/api. This feels sort of dirty to me, we can talk about it as a followup.

hexfusion:
added fixup

hexfusion:
rebased ready for review

marun (commented):
/lgtm

pkg/cmd/render/render.go
```diff
@@ -245,17 +253,26 @@ func newTemplateData(opts *renderOpts) (*TemplateData, error) {
...
+
 	if err := templateData.setComputedEnvVars(templateData.Platform); err != nil {
 		return nil, err
 	}
 
-	// Use a marker file to configure the bootstrap scaling strategy.
-	if _, err := os.Stat(opts.delayedHABootstrapScalingStrategyMarker); err == nil {
-		if templateData.NamespaceAnnotations == nil {
-			templateData.NamespaceAnnotations = map[string]string{}
+	// If bootstrap scaling strategy is delayed HA append set annotation signal
```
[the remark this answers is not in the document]
hexfusion: since the upgrade test is down I will clean this up now

PR #549, opened 2021-03-01
Remove unused render command options and configuration

marun:
Cleanup options and configuration from the operator's render command that were carried over from apiserver render and are unused for ceo.

/cc @hexfusion @retroflexer


marun:
@hexfusion I'm afraid this can't pass CI until https://github.com/openshift/installer/pull/4691 merges - it removes usage of arguments that this PR depends on.

marun:
/hold cancel 

single node is passing

bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   containers:
   - name: etcd
-    image: {{ .Images.Etcd }}
+    image: {{ .Image }}
```
hexfusion, line 11: would we ever need to include the operator image or some other image? I suppose we can change that in the future if needed.

pkg/cmd/render/render.go
```diff
@@ -342,12 +303,14 @@ func (r *renderOpts) Run() error {
 		return err
 	}
 
-	if err := r.manifest.ApplyTo(&templateData.ManifestConfig); err != nil {
-		return err
+	// Set the template dir only if not already set to support
+	// overriding the path in unit tests.
+	if len(r.templateDir) == 0 {
```
hexfusion, line 308: I might miss it and can be a follow-up but do we print the write path during render?

hexfusion (commented):
This makes sense mostly cruft from the initial port from kas-o. Thanks!

/lgtm
/approve

bindata/bootkube/bootstrap-manifests/etcd-member-pod.yaml
marun: Yeah, I'm partial to YAGNI.

pkg/cmd/render/render.go
marun: This is the input path, and it is not written. There's no way to set it other than in the code, though. The full output path is logged e.g.

```
Writing asset: /tmp/assets-316323197/manifests/etcd-metric-signer-secret.yaml
```

PR #554, opened 2021-03-12
[POC] defrag controller

hexfusion:
High level ideas for how this should work, will test and harden from here.

### DIFF
--- a/pkg/operator/defragcontroller/defragcontroller.go
+++ b/pkg/operator/defragcontroller/defragcontroller.go
@@ -1,0 +1,113 @@
+package defragcontroller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
+)
+
+const (
+	// TODO these need validated
+	minDefragBytes      int64   = 1073741824 * 2 // 2GiB
+	maxDefragPercentage float64 = 50.00          // 50%
+)
+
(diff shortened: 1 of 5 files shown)


pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -0,0 +1,115 @@
...
+			// defrag
+			klog.Infof("starting defrag on member: %q dbSize: %d , dbInUse: %d", member.Name, status.DbSize, status.DbSizeInUse)
+			_, err := etcdcli.DefragMember(member)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			recorder.Eventf("DefragControllerSuccess", "member: %q has been defragmented", member.Name)
+			// give cluster time to recover before we move to the next.
+			time.Sleep(5 * time.Second)
```
alvaroaleman, line 103: Maybe check in the beginning if all members have the same revision rather than blocking here?

hexfusion: makes sense to validate health before continue vs blind block +1

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -0,0 +1,115 @@
+package defragcontroller
+
+import (
+	"context"
+	"fmt"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
```
hexfusion, line 6: golands gofmt is terrible at understanding import blocks..

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	if len(errs) > 0 {
+		return utilerrors.NewAggregate(errs)
+	}
+	return nil
+}
+
+func getFragmentationPercentage(ondisk, inuse int64) float64 {
+	diff := float64(ondisk - inuse)
+	return (diff / float64(inuse)) * 100
```
retroflexer, line 112: nit: I would have expected that the percentage is on the `ondisk` value (instead of `inuse`).

hexfusion: yeah typo thanks, this all needs tests and such will followup with something more reviewable soonish.

PR #559, opened 2021-03-31
[wip] *: add disruption controller

hexfusion:
This PR adds the disruption controller. The point of this controller is to observe and track the duration of etcd downtime. The controller will be useful in identifying areas of improvement and triage rollout correctness. I performed a performance analysis on the controller and it appears to be very light on memory. This will also be nice to gauging improvements with the socket option change which will merge shortly.

Example of events:

> 7m9s        Warning   ConnectivityRestored                   deployment/etcd-operator                         Connectivity restored after 1m3.00925722s: ip-10-0-190-238.us-west-1.compute.internal: state: READY

While this only observes from s single node I think the data will still be very useful.

TODO:
- [x] add tests


pkg/operator/disruptioncontroller/disruptioncontroller.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+		wg.Add(1)
+		go checkClientConn(ctx, conn, member, eventRecorder, &wg)
+	}
+	wg.Wait()
+	return nil
+}
+
+// checkClientConn periodically checks the connection state of etcd members
+func checkClientConn(ctx context.Context, conn *grpc.ClientConn, member *etcdserverpb.Member, eventRecorder events.Recorder, wg *sync.WaitGroup) error {
```
hexfusion, line 92: pass member name instead of member object.

.gitignore
```diff
@@ -14,3 +14,6 @@ cluster-etcd-operator
 
 # Log output from telepresence
 telepresence.log
+
+# ide
+.idea
```
marun, line 19: Maybe leave this to each developer to globally ignore?

pkg/operator/disruptioncontroller/disruptioncontroller.go
```diff
@@ -0,0 +1,125 @@
...
+		etcdClient:     etcdClient,
+	}
+	return factory.New().ResyncEvery(syncDuration).WithInformers(
+		operatorClient.Informer(),
+	).WithSync(c.sync).ToController("DisruptionController", eventRecorder.WithComponentSuffix("disruption-controller"))
+}
+
+func (c *DisruptionController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	err := c.checkDisruption(ctx, syncCtx.Recorder())
+	if err != nil && !errors.Is(err, context.DeadlineExceeded) {
```
marun, line 48: Is DeadlineExceeded not an error condition?

pkg/operator/disruptioncontroller/disruptioncontroller.go
marun: Why does this function return an error if nothing checks the result and only nil is ever returned?

pkg/operator/disruptioncontroller/disruptioncontroller.go
hexfusion: in this context, it is expected[1] but more of a hack for not having proper error handling of the timeout itself.

[1] https://github.com/openshift/cluster-etcd-operator/pull/559/files#diff-3b8f033e644d1031d01f1b25c8c19640c7b8b84d3382508dd83f69c6a567a3a8R72

hexfusion: to your point, other client timeouts can exist.

pkg/operator/disruptioncontroller/disruptioncontroller.go
hexfusion: because I don't have propper error handling :)

pkg/operator/disruptioncontroller/disruptioncontroller.go
```diff
@@ -0,0 +1,125 @@
...
+func checkClientConn(ctx context.Context, conn *grpc.ClientConn, member *etcdserverpb.Member, eventRecorder events.Recorder, wg *sync.WaitGroup) error {
+	defer wg.Done()
+	for {
+	reset:
+		select {
+		case <-ctx.Done():
+			klog.Info("Restarting connectivity checker to validate membership")
+			conn.Close()
+			return nil
+		case <-time.After(retryDuration):
```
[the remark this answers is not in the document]
hexfusion: not sure I understand what you mean can you clarify?

https://play.golang.org/p/HjzTCjIQpaX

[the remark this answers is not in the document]
marun: I'm highlighting the difference between wait.Poll and wait.PollImmediate. Is it your intent that the first attempt to check connectivity is delayed by the retry interval, vs checking immediately and only waiting for that interval if the initial check fails?

pkg/operator/disruptioncontroller/disruptioncontroller.go
marun: If you see here timeout, though, is the intention that the disruption controller not be indicated as degraded? The current condition is binary (degraded/not degraded) and I wonder if it shouldn't be degraded/not degraded/unknown.

hexfusion:
I understand this is a large addition but I think having the etcd integration suite will allow us to do a lot more interesting tests.

pkg/operator/disruptioncontroller/disruptioncontroller.go
hexfusion: if we have a client timeout[1] that is a degraded condition if the context timeout[2] which is intended to resync fires then that is expected. 

[1] https://github.com/openshift/cluster-etcd-operator/pull/559/files#diff-3b8f033e644d1031d01f1b25c8c19640c7b8b84d3382508dd83f69c6a567a3a8R76

[2] https://github.com/openshift/cluster-etcd-operator/pull/559/files#diff-3b8f033e644d1031d01f1b25c8c19640c7b8b84d3382508dd83f69c6a567a3a8R71

pkg/operator/disruptioncontroller/disruptioncontroller.go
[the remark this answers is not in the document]
hexfusion: ok yeah I see what you mean, my intention is to always provide a `retryDuration` but as you noted reset would actually be 2* retryDuration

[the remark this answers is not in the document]
hexfusion: going to update this I know the code is verbose for now.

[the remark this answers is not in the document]
retroflexer: If you want to have timeout case executed first time too, then, I have done something similar with `rollbackcopier`. I posted an empty structure to the channel, so that the code gets executed the first time too:

https://github.com/openshift/cluster-etcd-operator/blob/release-4.5/pkg/cmd/rollbackcopy/rollbackcopy.go#L64

Actually in that example, the lines L77-L84 are identical with L86-L93. I meant to use `case <-initial, <-leaderCheckTicker.C`, but didn't. :).

hexfusion:
Reworking this to make more sense in the context of the operator.

PR #561, opened 2021-04-09
Bug 1948553: bindata, pkg: Propagate operator log level to etcd itself

lilic:
As per @hexfusion suggestion, logLevel was not used anywhere, its intention is for users to set log-level for etcd operand itself. This PR propagates the logLevel from the etcd CR to set the `--log-level` flag.

Opening PR mainly to get feedback on the following:

I didn't change the original Custom Resource definition of logLevel, but I feel like having that be a 1:1 match to the etcd log level would be a better UX. Should I go ahead and change the `operatorSpec.LogLevel` definition? If yes, I could use some pointers, as while I am familiar with library-go, I am not familiar with the way it does the CRDs. 

cc @marun @retroflexer

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -148,18 +148,12 @@ func createTargetConfig(c TargetConfigController, recorder events.Recorder, oper
 	return false, nil
 }
 
-func loglevelToKlog(logLevel operatorv1.LogLevel) string {
+func loglevelToZap(logLevel operatorv1.LogLevel) string {
 	switch logLevel {
-	case operatorv1.Normal:
-		return "2"
-	case operatorv1.Debug:
-		return "4"
-	case operatorv1.Trace:
-		return "6"
-	case operatorv1.TraceAll:
-		return "8"
+	case operatorv1.Debug, operatorv1.Trace, operatorv1.TraceAll:
+		return "debug"
 	default:
-		return "2"
+		return "info"
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -148,18 +148,18 @@ func createTargetConfig(c TargetConfigController, recorder events.Recorder, oper
 	return false, nil
 }
 
-func loglevelToKlog(logLevel operatorv1.LogLevel) string {
+func loglevelToZap(logLevel operatorv1.LogLevel) string {
 	switch logLevel {
 	case operatorv1.Normal:
```
lilic, line 153: As mentioned in the description, having these be 1:1 match spec to the etcd log level would make it easier I believe.

hexfusion: I agree but `operatorv1.LogLevel` is an enum[1] for v1 API I think the map makes sense here probably. Most of these changes will be from `info` to `debug`.

[1] https://github.com/openshift/api/blob/5935a5beec4bb8e1e81dd0fe9ebc2af36b9a09ae/operator/v1/types.go#L89

lilic: Ack! So the log-level conversion done here looks good to you?

hexfusion: well logically this can be simplified a bit right but otherwise lgtm.

```
case operatorv1.Debug, operatorv1.Trace, operatorv1.TraceAll:
 return "debug"
default:
  return "info"
```

lilic: yes 💯  of course, left it  this way to easily demonstrate the question I had. :)

hexfusion:
let's talk about this, a bit more I have a plan I will share on Monday.

hexfusion:
@lilic we still need rebase I guess :man_shrugging:

PR #565, opened 2021-04-14
[wip] Bug 1946607:*: add healthz endpoint

hexfusion:
The current unauthenticated tcpSocket readinessProbe spams the etcd logs and does not properly reflect ready status. In order for etcd to be ready to accept new requests, a quorate health check should return true. Static Pod Installer is smart enough to not install a new static pod until the existing pods are Ready. So we open up a race that could result in temporary quorum loss by not waiting until etcd is actually Ready.

This PR reuses the healthz term common on the control-plane with a proxy container that performs simple http health check against /health endpoint. This allows httpGet with HTTPS endpoint and only exposes health.

kubelet unauthenticated tcpProbe spam example every 5s

```
2021-03-12 21:35:28.803656 I | embed: rejected connection from "10.0.146.252:36846" (error "EOF", ServerName "")
```

Note: This ability for unauthenticated health checks should be exposed directly by etcd long term.

TODO 

- [ ] add tests?
- [ ] choose/register port with openshift authorities

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -104,6 +104,34 @@ ${COMPUTED_ENV_VARS}
         name: data-dir
     env:
 ${COMPUTED_ENV_VARS}
+  - name: etcd-healthz
+    image: ${OPERATOR_IMAGE}
+    imagePullPolicy: IfNotPresent
+    terminationMessagePolicy: FallbackToLogsOnError
+    command:
+      - /bin/sh
+      - -c
+      - |
+        #!/bin/sh
+        set -euo pipefail
+        cluster-etcd-operator healthz-proxy \
+          --listen-port=10100 \
+          --listen-cert=/etc/kubernetes/static-pod-certs/secrets/etcd-all-certs/etcd-serving-metrics-NODE_NAME.crt \
+          --listen-key=/etc/kubernetes/static-pod-certs/secrets/etcd-all-certs/etcd-serving-metrics-NODE_NAME.key \
+          --client-cert=${ETCDCTL_CERT} \
+          --client-key=${ETCDCTL_KEY} \
+          --client-cacert=${ETCDCTL_CACERT}
+    ports:
+      - containerPort: 10100
+    resources:
+      requests:
+        memory: 50Mi
+        cpu: 5m
+    volumeMounts:
+      - mountPath: /etc/kubernetes/static-pod-certs
+        name: cert-dir
+    env:
+${COMPUTED_ENV_VARS}
   - name: etcd
     image: ${IMAGE}
     imagePullPolicy: IfNotPresent
@@ -165,8 +193,10 @@ ${COMPUTED_ENV_VARS}
         memory: 600Mi
         cpu: 300m
     readinessProbe:
-      tcpSocket:
-        port: 2380
+      httpGet:
+        scheme: HTTPS
+        port: 10100
+        path: healthz
       failureThreshold: 3
       initialDelaySeconds: 3
       periodSeconds: 5
--- a/cmd/cluster-etcd-operator/main.go
+++ b/cmd/cluster-etcd-operator/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	goflag "flag"
 	"fmt"
+	"github.com/openshift/cluster-etcd-operator/pkg/cmd/healthzproxy"
 	"io/ioutil"
 	"math/rand"
 	"os"
@@ -62,6 +63,7 @@ func NewSSCSCommand() *cobra.Command {
 	cmd.AddCommand(prune.NewPrune())
 	cmd.AddCommand(certsyncpod.NewCertSyncControllerCommand(operator.CertConfigMaps, operator.CertSecrets))
 	cmd.AddCommand(waitforceo.NewWaitForCeoCommand(os.Stderr))
+	cmd.AddCommand(healthzproxy.NewHealthzProxyCommand())
 
 	return cmd
 }
--- a/pkg/cmd/healthzproxy/healthzproxy.go
+++ b/pkg/cmd/healthzproxy/healthzproxy.go
@@ -1,0 +1,148 @@
+package healthzproxy
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"k8s.io/klog/v2"
+)
+
+// healthzOpts holds values to drive the healthz proxy.
+type healthzOpts struct {
+	etcdEndpoint string
+
+	// TLS auth for proxy client
+	clientCert   string
+	clientKey    string
+	clientCACert string
+
+	// TLS for /healthz handler
+	listenPort uint16
+	listenCert string
+	listenKey  string
+}
+
+// NewHealthzProxyCommand creates a healthz-proxy command.
+func NewHealthzProxyCommand() *cobra.Command {
+	opts := healthzOpts{
+		listenPort:   10100, // TODO random for testing need to register with proper authorities
+		etcdEndpoint: "https://localhost:2379",
+	}
+	cmd := &cobra.Command{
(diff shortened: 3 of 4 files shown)


bindata/etcd/pod.yaml
```diff
@@ -103,6 +103,34 @@ ${COMPUTED_ENV_VARS}
...
+    terminationMessagePolicy: FallbackToLogsOnError
+    command:
+      - /bin/sh
+      - -c
+      - |
+        #!/bin/sh
+        set -euo pipefail
+        cluster-etcd-operator healthz-proxy \
+          --listen-port=10100 \
+          --listen-cert=/etc/kubernetes/static-pod-certs/secrets/etcd-all-certs/etcd-serving-metrics-NODE_NAME.crt \
```
hexfusion, line 119: server certs probably make more sense here.

pkg/cmd/healthzproxy/healthzproxy.go
```diff
@@ -0,0 +1,148 @@
...
+			klog.Warningf("Failed to read the response body: %v", err)
+			return
+		}
+		json.Unmarshal(body, &check)
+
+		if check.Health == "true" {
+			w.WriteHeader(http.StatusOK)
+		} else {
+			klog.Warningf("health check failed: %s", string(body))
+			w.WriteHeader(http.StatusBadRequest)
```
hexfusion, line 138: should this just be 500?

cmd/cluster-etcd-operator/main.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	goflag "flag"
 	"fmt"
+	"github.com/openshift/cluster-etcd-operator/pkg/cmd/healthzproxy"
```
hexfusion, line 6: goland :roll_eyes:

hexfusion:
this should probably get considered for backport

hexfusion:
Logs look very clean

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/565/pull-ci-openshift-cluster-etcd-operator-master-e2e-agnostic/1382518303179673600/artifacts/e2e-agnostic/gather-extra/artifacts/pods/openshift-etcd_etcd-ip-10-0-141-231.ec2.internal_etcd.log

bindata/etcd/pod.yaml
```diff
@@ -165,8 +193,10 @@ ${COMPUTED_ENV_VARS}
         memory: 600Mi
         cpu: 300m
     readinessProbe:
-      tcpSocket:
-        port: 2380
+      httpGet:
```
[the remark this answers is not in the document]
hexfusion: Yeah I think that is a good question my concern is that socket options added in 4.8 have made the generic tcpSocket probe not effective. `Ready` should mean ready to accept client traffic my experimentations have show tcpSocket showing ready many seconds before quorum-guard checks will pass for the member.

bindata/etcd/pod.yaml
```diff
@@ -103,6 +103,34 @@ ${COMPUTED_ENV_VARS}
...
+        set -euo pipefail
+        cluster-etcd-operator healthz-proxy \
+          --listen-port=10100 \
+          --listen-cert=/etc/kubernetes/static-pod-certs/secrets/etcd-all-certs/etcd-serving-metrics-NODE_NAME.crt \
+          --listen-key=/etc/kubernetes/static-pod-certs/secrets/etcd-all-certs/etcd-serving-metrics-NODE_NAME.key \
+          --client-cert=${ETCDCTL_CERT} \
+          --client-key=${ETCDCTL_KEY} \
+          --client-cacert=${ETCDCTL_CACERT}
+    ports:
+      - containerPort: 10100
```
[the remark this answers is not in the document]
hexfusion: yeah the port was just chosen out of the air so that does need to be changed.

https://github.com/openshift/cluster-etcd-operator/pull/565/files#diff-fad938af40231086decc9dad2ad6a274c6ba5d4a4cdc8696d246d84b5a1770beR34

bindata/etcd/pod.yaml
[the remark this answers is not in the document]
hexfusion: But if we can auth a httpGet on /health from kubelet that would work. I dug on that and didn't find anything obvious but will touch base with Ryan.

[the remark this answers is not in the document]
hexfusion: Also, a Get against /health would require full mTLS so we would need kubelet to not only call with valid ca-bundle but also client cert, key. That's sort of why I feel the proxy is nessisary. Also we are technically `HTTPS`? :)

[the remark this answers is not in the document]
hexfusion: this is exactly how /healthz for kas responds from my testing no auth is required.

```
curl https://127.0.0.1:${heathz_port}/healthz -k
ok
```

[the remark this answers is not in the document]
marun: Would it make sense to fix the tcpSocket probe to work with the new socket options?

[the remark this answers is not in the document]
hexfusion: I am open to the approach but I afraid there is not a direct relationship between listening port and member health so I think we should consider this in the check. As soon as the pod is Ready the static pod controller will install a new pod on the next node. So this timing is very critical.

pkg/cmd/healthzproxy/healthzproxy.go
```diff
@@ -0,0 +1,148 @@
...
+		Certificates: []tls.Certificate{keyPair},
+		RootCAs:      caPool,
+	}
+	client := &http.Client{Transport: &http.Transport{
+		TLSClientConfig: tlsConfig,
+	}}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/healthz", func(w http.ResponseWriter, req *http.Request) {
+		resp, err := client.Get(h.etcdEndpoint + "/health")
```
deads2k, line 118: does etcd require authentication for accessing healthz?  That makes it difficult to add to a pod

pkg/cmd/healthzproxy/healthzproxy.go
```diff
@@ -0,0 +1,148 @@
...
+		var check Health
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			http.Error(w, "failed to read response from etcd health check", http.StatusInternalServerError)
+			klog.Warningf("Failed to read the response body: %v", err)
+			return
+		}
+		json.Unmarshal(body, &check)
+
+		if check.Health == "true" {
```
deads2k, line 134: do we ever get health=true and http status not ok?

Do we ever get health=false and http status ok?

pkg/cmd/healthzproxy/healthzproxy.go
```diff
@@ -0,0 +1,148 @@
...
+		} else {
+			klog.Warningf("health check failed: %s", string(body))
+			w.WriteHeader(http.StatusBadRequest)
+		}
+
+		w.Header().Set("Content-Type", resp.Header.Get("Content-Type"))
+		w.Write(body)
+	})
+
+	addr := fmt.Sprintf("0.0.0.0:%d", h.listenPort)
```
deads2k, line 145: is hostnetwork set at a pod level or a container level? If you're running this on hostnetwork (and I think you need to), then you should bind this down to localhost since we don't want to expose it everywhere.

pkg/cmd/healthzproxy/healthzproxy.go
hexfusion: > do we ever get health=true and http status not ok?
I don't believe that would happen but I will ensure corners are covered.

> Do we ever get health=false and http status ok?
yes but I think in this case we want to return non 200 to the get as a failure?

pkg/cmd/healthzproxy/healthzproxy.go
hexfusion: the client Get requires auth to etcd /health but the endpoint /healthz does not. I get what you're saying though and ideally, etcd would expose this directly we have a few options that are in the works upstream to allow this.

hexfusion:
/hold

reviewing approach

PR #569, opened 2021-04-20
Rename workload annotations

cynepco3hahue:
As per openshift/enhancements#739, the workload annotations names are changing.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

/hold
Hold until after openshift/kubernetes#632 is merged


hexfusion:
/lgtm
/approve

@cynepco3hahue needs a bug

cynepco3hahue:
@hexfusion Thanks
@browsell Do we have a bug for it, or it should be ok without a bug because we got an exception?

PR #580, opened 2021-04-29
Bug 1946607: bindata/etcd: improve readinessProbe checks by dialing unix socket target

hexfusion:
The current unauthenticated tcpSocket readinessProbe spams the etcd logs and does not properly reflect ready status. In order for etcd to be ready to accept new requests, a quorate health check should return true. Static Pod Installer is smart enough to not install a new static pod until the existing pods are Ready. So we open up a race that could result in temporary quorum loss by not waiting until etcd is actually Ready.

```
2021-03-12 21:35:28.803656 I | embed: rejected connection from "10.0.146.252:36846" (error "EOF", ServerName "")
```

socket path includes `NODE_NODE_ENVVAR_NAME_IP` so that mTLS passes auth. etcd today requires the following format for socket path  `<name>:<port>`. By setting 2s timeouts I believe we should remove the window of opportunity for zombie process as a result of exec probe. tl;dr fail fast

I used port 0 which is a restricted port as per IANA for TCP, but since this is unix socket I figured it would possibly eliminate any assumption that it is TCP. This is open to debate.

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -156,8 +156,8 @@ ${COMPUTED_ENV_VARS}
           --peer-key-file=/etc/kubernetes/static-pod-certs/secrets/etcd-all-certs/etcd-peer-NODE_NAME.key \
           --peer-trusted-ca-file=/etc/kubernetes/static-pod-certs/configmaps/etcd-peer-client-ca/ca-bundle.crt \
           --peer-client-cert-auth=true \
-          --advertise-client-urls=https://${NODE_NODE_ENVVAR_NAME_IP}:2379 \
-          --listen-client-urls=https://${LISTEN_ON_ALL_IPS}:2379 \
+          --advertise-client-urls=https://${NODE_NODE_ENVVAR_NAME_IP}:2379,unixs://${NODE_NODE_ENVVAR_NAME_IP}:0 \
+          --listen-client-urls=https://${LISTEN_ON_ALL_IPS}:2379,unixs://${NODE_NODE_ENVVAR_NAME_IP}:0 \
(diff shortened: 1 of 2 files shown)


bindata/etcd/pod.yaml
```diff
@@ -167,8 +167,18 @@ ${COMPUTED_ENV_VARS}
...
     readinessProbe:
-      tcpSocket:
-        port: 2380
+      exec:
+        command:
+        - /bin/bash
+        - -c
+        - |
+          set -xe
+          unset ETCDCTL_ENDPOINTS
```
hexfusion, line 176: going to add a verbose commend here explaining how this works

bindata/etcd/pod.yaml
```diff
@@ -156,8 +156,8 @@ ${COMPUTED_ENV_VARS}
           --peer-key-file=/etc/kubernetes/static-pod-certs/secrets/etcd-all-certs/etcd-peer-NODE_NAME.key \
           --peer-trusted-ca-file=/etc/kubernetes/static-pod-certs/configmaps/etcd-peer-client-ca/ca-bundle.crt \
           --peer-client-cert-auth=true \
-          --advertise-client-urls=https://${NODE_NODE_ENVVAR_NAME_IP}:2379 \
-          --listen-client-urls=https://${LISTEN_ON_ALL_IPS}:2379 \
+          --advertise-client-urls=https://${NODE_NODE_ENVVAR_NAME_IP}:2379,unixs://${NODE_NODE_ENVVAR_NAME_IP}:0\
```
hexfusion, line 159: add space

hexfusion:
I want to add some more extensive tests and will but I think this is important to soak and see if it help upgrade stability at all.

/hold cancel

PR #573, opened 2021-04-22
Bug 1952684: pkg/operator/metriccontroller: dont panic on invalid response from prom client

hexfusion:
caught in CI[1],[2] thanks @trozet. The solution is to check success when casting variable. 

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1042/pull-ci-openshift-cluster-network-operator-master-e2e-gcp-ovn-upgrade/1382687306791522304/artifacts/e2e-gcp-ovn-upgrade/gather-extra/artifacts/pods/openshift-etcd-operator_etcd-operator-7897fff8c9-njwzl_etcd-operator.log
[2] https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1042/pull-ci-openshift-cluster-network-operator-master-e2e-gcp-ovn-upgrade/1382687306791522304

Signed-off-by: Sam Batschelet <sbatsche@redhat.com>

PR #584, opened 2021-05-03
Bug 1952684: pkg/operator/metriccontroller: check vector length

hexfusion:
Follow up to https://github.com/openshift/cluster-etcd-operator/pull/573


lilic (commented):
Small suggestion, non blocking maybe for refactor in 4.9.

PR #593, opened 2021-05-11
[wip][testing] pkg/etcdenvvar: enable log rotation for etcd

hexfusion:
This PR enables log rotation for etcd logs. This functionality ensures that etcd logs are always persisted without the need of loki. The defaults are based on current audit logs config[1].

[1] https://github.com/openshift/cluster-kube-apiserver-operator/blob/b1010cd0da2702678df7c2a8e06dc17ad6e48853/pkg/operator/v410_00_assets/bindata.go#L298

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -12,6 +12,27 @@ metadata:
     revision: "REVISION"
 spec:
   initContainers:
+    - name: setup
+      image: ${IMAGE}
+      imagePullPolicy: IfNotPresent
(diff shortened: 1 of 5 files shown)


bindata/etcd/pod.yaml
```diff
@@ -204,6 +225,8 @@ ${COMPUTED_ENV_VARS}
         name: cert-dir
       - mountPath: /var/lib/etcd/
         name: data-dir
+      - mountPath: /var/log/etcd/
+        name: log-dir
   - name: etcd-metrics
```
hexfusion, line 230: drop unused

PR #594, opened 2021-05-13
[release-4.7] Bug 1958416: pkg/dnshelpers: fallback to spec if status is not populated for serviceNetwork

retroflexer:
Manual cherry-pick of #587 and #591.


derekwaynecarr:
[patch-manager] :rocket: Approved for z-stream by score: 0.20

approved

PR #596, opened 2021-05-14
Bug 1958405: Improve etcd service observability and health checks

hexfusion:
**NOTE**: **the majority of the files that change here are the result of bringing in etcd integration tests. While this is a lot of change having the ability to test etcd outside of OpenShift has great value.**

This PR adds support for an etcd health monitor sidecar for the etcd static pod. This command will perform health checks for etcd on any interval. This solution aligns with etcd operand and uses zap logger to provide a machine consumable rotatable log that can be persisted to the host path. The health monitor can schedule both single and multi target health checks.

Today the cluster performs a few different "health" checks to ensure that etcd service is available.

### Existing checks
**quorum-guard**: Quorum guard performs a health check against the /health endpoint of etcd every 5s. The failure is logged by the kubelet. While this seems to be a logic place for this health check quorum-guard today is not deployed in for example single node. (interval 5s)

**cluster-member-controller**: This controller performs health checks of etcd and has informers on pods and nodes. So during upgrades or more obvious disruptions, this controller will report member health and go degraded. The problem with this approach is that a short disruption could potentially go unoticed. Also the operator needs to be running which creates small windows of unobservable time.
**etcd logs**: etcd will reflect failed health checks with its peers in the etcd logs. This information can help assist in the triage of availability. 

**etcd readinessProbe**: etcd will reflect failed health checks with its peers in the etcd logs. This information can help assist in the triage of availability. (interval 5s) This design considered a 1s interval on health checks. But decided against this for additional control of logging and more complex health checks which would not be dependant on operand support.

While these are somewhat effective they lack a clear durable centralized signal.

**Goals**
- add 1-second granularity health checks of the etcd service on a peer level for both single target and multi target tests. The idea is to ensure that the reporting simulates that of the kube-apiserver. 
- provide the cluster a complete record of etcd service disruption that survives upgrade.
- no direct dependency on monitoring or apiserver.
- check health of etcd whenever it can be available as a service to the cluster.
- provide logs in a machine consumable format that can generate reports and be used for test criteria.
- improved observability on critical timings around scaling during install.

**Non Goals**
- provide static apiserver auth with kubeconfig and cert rotation.
- generate reports or events

**Possible Future Work**
- Add support for cert rotation controller and mint a kubeconfig allowing events to be generated after observation of disruption. This would replace the current[ cluster-member-controller](https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/clustermembercontroller/clustermembercontroller.go).
- Align this logic with library-go [health prober](https://github.com/openshift/library-go/tree/master/pkg/monitor/health).
- Add the host logs to the CI artifacts.
- Add the ability to capture the logs as part of must-gather (similar to audit logs).
- Provide simple tooling to generate availability report based on logs.
- Add support for collecting metrics. example etcd startup time and client failure reporting.
- Add additional testing
- Add detailed docs to ceo repo

[1] https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/etcd/quorumguard-deployment.yaml#L81

TODO
- [x] add tests
- [x] add check-health command to static pod as a container
- [x] add docs for etcd health checking
- [x] performance review

Example output showing an etcd rollout during install.

```
"level":"info","ts":"2021-06-09T17:00:35.915Z","caller":"health/log.go:57","msg":"service restored","addresses":["https://172.55.21.42:2379"],"check":"SerializedReadSingleTarget","health":true,"start":"2021-06-09T16:43:46.836Z","end":"2021-06-09T17:00:35.915Z","duration":"16m49.078852554s"}
{"level":"info","ts":"2021-06-09T17:00:35.916Z","caller":"health/log.go:57","msg":"service restored","addresses":["https://172.55.21.42:2379"],"check":"QuorumReadSingleTarget","health":true,"start":"2021-06-09T16:43:46.836Z","end":"2021-06-09T17:00:35.916Z","duration":"16m49.080058737s"}
{"level":"info","ts":"2021-06-09T17:02:01.646Z","caller":"health/log.go:41","msg":"possible service disruption detected","addresses":["https://172.55.21.42:2379"],"check":"SerializedReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.145Z","error":"slow request"}
{"level":"info","ts":"2021-06-09T17:02:01.704Z","caller":"health/log.go:41","msg":"possible service disruption detected","addresses":["https://172.55.21.42:2379"],"check":"GRPCReadySingleTarget","health":false,"start":"2021-06-09T17:02:01.203Z","error":"slow request"}
{"level":"info","ts":"2021-06-09T17:02:01.814Z","caller":"health/log.go:41","msg":"possible service disruption detected","addresses":["https://172.55.21.42:2379"],"check":"QuorumReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.313Z","error":"slow request"}
{"level":"info","ts":"2021-06-09T17:02:02.146Z","caller":"health/log.go:15","msg":"service disruption detected","addresses":["https://172.55.21.42:2379"],"check":"SerializedReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.145Z","error":"context deadline exceeded"}
{"level":"info","ts":"2021-06-09T17:02:02.314Z","caller":"health/log.go:15","msg":"service disruption detected","addresses":["https://172.55.21.42:2379"],"check":"QuorumReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.313Z","error":"context deadline exceeded"}
{"level":"info","ts":"2021-06-09T17:02:04.147Z","caller":"health/log.go:29","msg":"health check","addresses":["https://172.55.21.42:2379"],"check":"SerializedReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.145Z","error":"client timeout of exceeded: 1s"}
{"level":"info","ts":"2021-06-09T17:02:04.316Z","caller":"health/log.go:29","msg":"health check","addresses":["https://172.55.21.42:2379"],"check":"QuorumReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.313Z","error":"client timeout of exceeded: 1s"}
{"level":"info","ts":"2021-06-09T17:02:06.149Z","caller":"health/log.go:29","msg":"health check","addresses":["https://172.55.21.42:2379"],"check":"SerializedReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.145Z","error":"client timeout of exceeded: 1s"}
{"level":"info","ts":"2021-06-09T17:02:06.317Z","caller":"health/log.go:29","msg":"health check","addresses":["https://172.55.21.42:2379"],"check":"QuorumReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.313Z","error":"client timeout of exceeded: 1s"}
{"level":"info","ts":"2021-06-09T17:02:08.151Z","caller":"health/log.go:29","msg":"health check","addresses":["https://172.55.21.42:2379"],"check":"SerializedReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.145Z","error":"client timeout of exceeded: 1s"}
{"level":"info","ts":"2021-06-09T17:02:08.319Z","caller":"health/log.go:29","msg":"health check","addresses":["https://172.55.21.42:2379"],"check":"QuorumReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.313Z","error":"client timeout of exceeded: 1s"}
{"level":"info","ts":"2021-06-09T17:02:10.153Z","caller":"health/log.go:29","msg":"health check","addresses":["https://172.55.21.42:2379"],"check":"SerializedReadSingleTarget","health":false,"start":"2021-06-09T17:02:01.145Z","error":"client timeout of exceeded: 1s"}
```
steady state perf review.
steady-state


bindata/etcd/quorumguard-deployment.yaml
```diff
@@ -72,13 +72,17 @@ spec:
...
-                  curl --max-time 2 --silent --cert "${cert//:/\:}" --key "$key" --cacert "$cacert" "$health_endpoint" |grep '{ *"health" *: *"true" *}'
+                  /usr/bin/etcdctl \
+                    --key=/var/run/secrets/etcd-client/tls.key
+                    --cert=/var/run/secrets/etcd-client/tls.crt
+                    --cacert=/var/run/configmaps/etcd-ca/ca-bundle.crt
+                    --command-timeout=2s \
+                    --dial-timeout=2s \
+                    --endpoints=https://loclahost:2379 \
+                    endpoint health -w json | grep \"health\":true
+
```
retroflexer, line 85: If we are running `etcdctl` should not the image on L52 be `etcdImage` instead of `cliImage`?

https://github.com/openshift/cluster-etcd-operator/blob/3fde37aace23d4d0978b7b104cfabab4140bb77d/bindata/etcd/quorumguard-deployment.yaml#L52

hexfusion: yeah thanks this is still a bit raw I will ping you guys when it's ready for review.

pkg/cmd/checkhealth/checkhealth.go
```diff
@@ -0,0 +1,294 @@
...
+			EncodeCaller:   zapcore.ShortCallerEncoder,
+		},
+		OutputPaths:      []string{"stderr"},
+		ErrorOutputPaths: []string{"stderr"},
+	}
+	lg, err := logConfig.Build()
+	if err != nil {
+		panic(err)
+	}
+
```
hexfusion, line 165: break out logging setup into a separate function.

pkg/cmd/checkhealth/checkhealth.go
```diff
@@ -0,0 +1,294 @@
...
+		TrustedCAFile: "/var/run/configmaps/etcd-ca/ca-bundle.crt",
+	}
+	tr, err := transport.NewTransport(tls, h.dialTimeout)
+	if err != nil {
+		panic(err)
+	}
+
+	client := http.Client{
+		Transport: tr,
+	}
```
hexfusion, line 131: new client logic can be a function

pkg/cmd/checkhealth/checkhealth.go
```diff
@@ -0,0 +1,294 @@
...
+			disruption:
+				for {
+					select {
+					case <-ticker:
+						_, err := checkHealth()
+						if err == nil {
+							lg.Warn(
+								"etcd service restored",
+								zap.Bool("available", true),
+								zap.Duration("disruption", time.Since(disruptionStart)),
```
hexfusion, line 206: print the error?

bindata/etcd/quorumguard-deployment.yaml
```diff
@@ -72,13 +72,17 @@ spec:
                 - /bin/sh
                 - -c
                 - |
-                  declare -r health_endpoint="https://localhost:2379/health"
-                  declare -r cert="/var/run/secrets/etcd-client/tls.crt"
-                  declare -r key="/var/run/secrets/etcd-client/tls.key"
-                  declare -r cacert="/var/run/configmaps/etcd-ca/ca-bundle.crt"
-                  export NSS_SDB_USE_CACHE=no
+                  set -xe
                   [[ -z $cert || -z $key ]] && exit 1
```
hexfusion, line 76: drop

bindata/etcd/pod.yaml
```diff
@@ -244,6 +264,36 @@ ${COMPUTED_ENV_VARS}
...
+      - name: POD_NAME
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.name
+    volumeMounts:
+      - mountPath: /var/log/etcd/
+        name: log-dir
+    resources:
+      requests:
+        memory: 600Mi
```
hexfusion, line 295: needs profiled and tuned.

pkg/cmd/healthprobe/healthprobe.go
```diff
@@ -0,0 +1,304 @@
...
+	fs.StringVar(&h.clientCertFile, "cert-file", "", "Health probe TLS client certificate file. (required)")
+	fs.StringVar(&h.clientKeyFile, "key-file", "", "Health probe TLS client key file. (required)")
+	fs.StringVar(&h.clientCACertFile, "cacert-file", "", "Health probe TLS client CA certificate file. (required)")
+}
+
+var shutdownSignals = []os.Signal{os.Interrupt, syscall.SIGTERM}
+
+func (h *healthProbeOpts) Validate(_ context.Context) error {
+	if len(h.clientKeyFile) == 0 {
+		return errors.New("missing required flag: --key")
```
hexfusion, line 123: s/--key/--key-file

pkg/cmd/healthprobe/healthprobe.go
```diff
@@ -0,0 +1,304 @@
...
+	if err != nil {
+		return err
+	}
+
+	healthCheckProbe := healthCheck.Probe()
+	if err := wait.PollImmediateUntil(h.interval, func() (bool, error) { return healthCheckProbe() }, ctx.Done()); err != nil {
+		// probe is shutting down report last status of existing disruption
+		if err == wait.ErrWaitTimeout && healthCheck.Disruption {
+			healthCheck.lg.Warn("health probe terminated",
+				zap.String("pod", healthCheck.podName),
```
hexfusion, line 164: add static pod version to logs

bindata/etcd/pod.yaml
```diff
@@ -106,7 +126,7 @@ ${COMPUTED_ENV_VARS}
         name: data-dir
     env:
 ${COMPUTED_ENV_VARS}
-      - name: "ETCD_STATIC_POD_REV"
+      - name: "ETCD_STATIC_POD_VERSION"
```
deads2k, line 129: no particular objection, but why the change?

bindata/etcd/pod.yaml
deads2k: > needs profiled and tuned.

yeah... that should be a lot smaller :)

pkg/cmd/healthprobe/healthprobe.go
```diff
@@ -0,0 +1,316 @@
...
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+)
+
+const (
+	DefaultDialTimeout         = 2 * time.Second
+	DefaultHealthCheckInterval = 5 * time.Second
+	DefaultEndpoint            = "https://localhost:2379"
```
deads2k, line 30: our `etcdctl` container has env vars that I thought enabled it to check all the members.

pkg/cmd/healthprobe/healthprobe.go
```diff
@@ -0,0 +1,316 @@
...
+	if err != nil {
+		return nil, err
+	}
+	client := http.Client{
+		Transport: tr,
+	}
+
+	healthCheck := &healthCheck{
+		client:            client,
+		endpoint:          h.Endpoint,
```
deads2k, line 210: I think I see the following possible checks.

1. use the client as constructed like the kube-apiserver to issue a quorum read of some key?  This should closely approximate what the kube-apiserver will observe at any time, regardless of which individual members are up or down.  Is the client constructed for the `etcdl` container the same as the kube-apiserver?
2. find some way to find all the current members and issue distinct quorum reads to each one.  This allows reporting of failures for individual members, but probably makes for a more complicated implementation as you mux and demux.
3. using localhost is straight-forward for configuration and predictability, but requires the local etcd server instance to be running an available.  This can result in false positives in cases when the local instance is down and the remote instances are up.


Are there other option?  How difficult is it to achieve option 1 in this spot?

pkg/cmd/healthprobe/healthprobe.go
```diff
@@ -0,0 +1,316 @@
...
+		interval:          h.interval,
+		podName:           h.podName,
+		staticPodVersion:  h.staticPodVersion,
+		enableLogRotation: h.enableLogRotation,
+	}
+	return healthCheck, nil
+}
+
+func (h *healthCheck) Probe() HealthCheckFunc {
+	h.lg.Warn("starting health probe",
```
deads2k, line 221: this seems like info

pkg/cmd/healthprobe/healthprobe.go
```diff
@@ -0,0 +1,316 @@
...
+			)
+			h.Disruption = false
+			h.DisruptionStart = time.Time{}
+		}
+		return false, nil
+	}
+}
+
+func checkHealth(client http.Client, endpoint string, podName string, lg *zap.Logger) (bool, error) {
+	resp, err := client.Get(endpoint + "/health")
```
deads2k, line 256: Can we make this a quorum read of some key like the kube-apiserver would do?  namespaces/default perhaps?

The individual health seems like a separate (and valuable) check.

pkg/cmd/healthprobe/healthprobe.go
```diff
@@ -0,0 +1,316 @@
...
+	h.lg.Warn("starting health probe",
+		zap.String("pod", h.podName),
+	)
+	return func() (bool, error) {
+		isHealthy, _ := checkHealth(h.client, h.endpoint, h.podName, h.lg)
+		// track new service disruption
+		if !isHealthy && !h.Disruption {
+			h.Disruption = true
+			h.DisruptionStart = time.Now()
+			h.lg.Warn(
```
deads2k, line 230: minor, can I get this structured as an `h.logDisruptionStart` and `h.logDisruptionEnd`

pkg/cmd/healthprobe/healthprobe.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	return healthCheck, nil
+}
+
+func (h *healthCheck) Probe() HealthCheckFunc {
+	h.lg.Warn("starting health probe",
+		zap.String("pod", h.podName),
+	)
+	return func() (bool, error) {
+		isHealthy, _ := checkHealth(h.client, h.endpoint, h.podName, h.lg)
```
deads2k, line 225: Why structure like this instead of simply

`func (h *healthCheck) CheckQuorumRead()` and have the call site do

```go
func() (bool, error){
  h.CheckQuorumRead(ctx)
  return false, nil
}
```

pkg/cmd/healthprobe/healthprobe.go
```diff
@@ -0,0 +1,316 @@
...
+				zap.Duration("duration", time.Since(h.DisruptionStart)),
+			)
+			h.Disruption = false
+			h.DisruptionStart = time.Time{}
+		}
+		return false, nil
+	}
+}
+
+func checkHealth(client http.Client, endpoint string, podName string, lg *zap.Logger) (bool, error) {
```
deads2k, line 255: any reason this isn't just a function attached to `(h *healthCheck)`?

bindata/etcd/pod.yaml
hexfusion: to standardize against for example KAS `STATIC_POD_VERSION`. We add ETCD_ extenstion so we get it printed in the etcd logs as signal. But it would not be dependent on the overall PR. Just nice to have and not yet release ENV (introduced in 4.8).

pkg/cmd/healthprobe/healthprobe.go
hexfusion: So to be clear this is a quorate get. The current test 
- checks etcd is available (localhost)
- checks that etcd has quorum so N+1 etcds are available thus checking host networking is available.

If this is done on each node that is pretty reasonable coverage with 3 RPS. If each node is checking each member the cost will increase 9 RPS. I didn't think that was required but I am open to better understanding why you feel that is required. KAS needs a single subconn in client balancer to be `Ready` and the cluster to quorate.

pkg/cmd/healthprobe/healthprobe.go
hexfusion: trying to reduce the footprint of the method as the function signature seemed reasonablish was my only reason.

pkg/cmd/healthprobe/healthprobe.go
hexfusion: I can see now your point with local etcd being down but cluster quorate from KAS point of view.

pkg/cmd/healthprobe/healthprobe.go
hexfusion: yes, makes sense.

pkg/cmd/healthprobe/healthprobe.go
hexfusion: I will implement 1 and followup on the rest shortly

lilic (commented):
Really nice!

In the future we could also set the etcd operator status to reflect the health check, what do you think?

bindata/etcd/pod.yaml
hexfusion: still pending.

pkg/cmd/monitor/health/health.go
```diff
@@ -0,0 +1,145 @@
...
+func WithQuorumReadSingleTarget() CheckFunc {
+	return func(c *Check) {
+		c.name = QuorumReadSingleTarget
+		c.checkKVSRead([]clientv3.OpOption{clientv3.WithKeysOnly()})
+	}
+}
+
+// WithGRPCReadySingleTarget performs a gRPC dial to the endpoint and checks if the service is "Ready". This check verifies that
+// subconn will be considered usable by the client balancer for round robin requests.
+// NOTE: This can be flaky during graceful termination due to lack of a context. So it is possible
```
hexfusion, line 78: drop this is not true the bug was with using pauseServer for non linearlized requests.

hexfusion:
@deads2k PTAL I have updated the health monitor to manage both single and multi target health tests. The etcd client has been aligned to use kube-apiserver defaults.

current health tests which are directly reflective of KAS representation of health.

- SerializedReadSingleTarget: performs a serialized RangeRequest against a single target. Unless the watch uses WithRequireLeader the stream is serialized.

-  WithQuorumReadSingleTarget: performs a quorate RangeRequest against a single target.

- WithGRPCReadySingleTarget: performs a gRPC dial to the endpoint and checks if the service is "Ready". This check verifies that subconn will be considered usable by the client balancer for round-robin requests.

- WithQuorumRead: performs a quorate RangeRequest against all targets.

pkg/cmd/monitor/health/health.go
```diff
@@ -0,0 +1,142 @@
...
+	"google.golang.org/grpc/connectivity"
+)
+
+const (
+	SerializedReadSingleTarget CheckName = "SerializedReadSingleTarget"
+	QuorumReadSingleTarget     CheckName = "QuorumReadSingleTarget"
+	GRPCReadySingleTarget      CheckName = "GRPCReadySingleTarget"
+	QuorumRead                 CheckName = "QuorumRead"
+
+	DefaultSlowRequestDuration = 5 * time.Second
```
hexfusion, line 19: we log if request takes longer than DefaultSlowRequestDuration this could be dropped if it is felt to not be useful.

pkg/cmd/healthprobe/healthprobe.go
hexfusion: startup log was dropped in refactor will return this as info

pkg/cmd/healthprobe/healthprobe.go
hexfusion: done

bindata/etcd/pod.yaml
```diff
@@ -244,6 +264,39 @@ ${COMPUTED_ENV_VARS}
         name: cert-dir
       - mountPath: /var/lib/etcd/
         name: data-dir
+  - name: etcd-health-monitor
+    image: ${OPERATOR_IMAGE}
+    imagePullPolicy: IfNotPresent
+    terminationMessagePolicy: FallbackToLogsOnError
+    command: [ "cluster-etcd-operator", "monitor" ]
+    args:
+      - --endpoints=$(ETCDCTL_ENDPOINTS)
```
hexfusion, line 273: pod spec needs to reflect new flags.

pkg/cmd/monitor/health/log.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (c *Check) LogTermination(err error) {
+	c.client.Close()
+	if c.isDisruption() || c.isSlowRequest() {
+		c.status.logDisruptionEnd = time.Now()
+		c.lg.Warn("health probe terminated",
+			zap.Strings("targets", c.targets),
+			zap.Any("check", c.name),
+			zap.Bool("health", false),
```
hexfusion, line 79: standardize log key order.

pkg/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,279 @@
...
+	DefaultStaticPodVersion    = 0
+
+	// DefaultLogRotationConfig is the default configuration used for log rotation.
+	// Log rotation is disabled by default.
+	// MaxSize    = 100 // MB
+	// MaxAge     = 0 // days (no limit)
+	// MaxBackups = 10
+	// LocalTime  = false // use computers local time, UTC by default
+	// Compress   = true // compress the rotated log in gzip format
+	DefaultLogRotationConfig = `{"maxsize": 100, "maxage": 0, "maxbackups": 10, "localtime": false, "compress": true}`
```
hexfusion, line 43: belongs with health/logger.go

pkg/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,279 @@
...
+
+	tlsInfo := transport.TLSInfo{
+		CertFile:      o.clientCertFile,
+		KeyFile:       o.clientKeyFile,
+		TrustedCAFile: o.clientCACertFile,
+	}
+
+	targets := strings.Split(o.Targets, ",")
+
+	// Create single target checks one check per target
```
hexfusion, line 201: improve logic for len(targets) == 1 to skip multi target tests

hexfusion: for example SNO

pkg/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,279 @@
...
+			// health.WithWatchRequireLeader(),
+		),
+		WithMultiTargetHealthCheck(
+			health.WithQuorumRead(),
+			// TODO
+			// health.WithNodeICMP(),
+		),
+	)
+
+	monitor.wg.Add(1)
```
hexfusion, line 176: add startup log

pkg/cmd/monitor/health/log.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"go.uber.org/zap"
+)
+
+func (c *Check) logHealthFail(err error) {
+	if !c.isDisruption() {
+		c.status.disruption = true
+		c.lg.Info(
+			"service disruption detected",
+			zap.Strings("targets", c.targets),
```
hexfusion, line 14: align with grpc logs s/targets/addresses

pkg/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,279 @@
...
+
+func WithMultiTargetHealthCheck(checks ...health.CheckFunc) []health.CheckFunc {
+	return checks
+}
+
+func WithSingleTargetHealthCheck(checks ...health.CheckFunc) []health.CheckFunc {
+	return checks
+}
+
+func newETCD3Client(tlsInfo transport.TLSInfo, endpoints []string) (*clientv3.Client, error) {
```
hexfusion, line 260: send grpc logs to zap if ETCD_DEBUG

```
grpclog.SetLoggerV2(zapgrpc.NewLogger(lg.Named("etcd-client")))
```

hexfusion:
digging into performance issues currently.

pkg/cmd/monitor/health/health.go
deads2k: > we log if request takes longer than DefaultSlowRequestDuration this could be dropped if it is felt to not be useful.

it's useful.  But I would expect it close to 500ms

pkg/cmd/monitor/health/health.go
```diff
@@ -0,0 +1,142 @@
...
+
+// WithSerializedReadSingleTarget performs a serialized RangeRequest against a single target.
+func WithSerializedReadSingleTarget() CheckFunc {
+	return func(c *Check) {
+		c.name = SerializedReadSingleTarget
+		c.checkKVSRead([]clientv3.OpOption{clientv3.WithSerializable(), clientv3.WithKeysOnly()})
+	}
+}
+
+// WithQuorumReadSingleTarget performs a quorate RangeRequest against a single target.
```
deads2k, line 68: This is what kube uses, correct?

pkg/cmd/monitor/health/health.go
```diff
@@ -0,0 +1,142 @@
...
+// WithGRPCReadySingleTarget performs a gRPC dial to the endpoint and checks if the service is "Ready". This check verifies that
+// subconn will be considered usable by the client balancer for round robin requests.
+func WithGRPCReadySingleTarget() CheckFunc {
+	return func(c *Check) {
+		c.name = GRPCReadySingleTarget
+		c.checkDialStatus()
+	}
+}
+
+// WithQuorumRead performs a quorate RangeRequest against all targets.
```
deads2k, line 85: Can you describe in the comments when `WithQuorumRead` and `WithQuorumReadSingleTarget` will return different results and which more closely aligns to what kube is doing?

pkg/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,279 @@
...
+	// Log rotation is disabled by default.
+	// MaxSize    = 100 // MB
+	// MaxAge     = 0 // days (no limit)
+	// MaxBackups = 10
+	// LocalTime  = false // use computers local time, UTC by default
+	// Compress   = true // compress the rotated log in gzip format
+	DefaultLogRotationConfig = `{"maxsize": 100, "maxage": 0, "maxbackups": 10, "localtime": false, "compress": true}`
+)
+
+// defaults used by kube-apiserver
```
deads2k, line 46: if this starts working nicely, we may want to improve granularity by choosing a smaller value.

pkg/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,279 @@
...
+	targets := strings.Split(o.Targets, ",")
+
+	// Create single target checks one check per target
+	for _, target := range targets {
+		// one client per target to eliminate lock racing
+		client, err := newETCD3Client(tlsInfo, targets)
+		if err != nil {
+			return nil, err
+		}
+		for _, healthCheckFunc := range singleTargetChecks {
```
deads2k, line 208: I expected to see the client created here to avoid re-use.

pkg/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,279 @@
...
+			healthChecks = append(healthChecks, healthCheck)
+		}
+	}
+
+	client, err := newETCD3Client(tlsInfo, targets)
+	if err != nil {
+		return nil, err
+	}
+	for _, healthCheckFunc := range multiTargetChecks {
+		healthCheck := health.NewCheck(ctx, lg, client, targets)
```
deads2k, line 220: don't we want distinct clients for every check?

pkg/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,279 @@
...
+		healthCheck.Probe = healthCheckFunc
+		healthChecks = append(healthChecks, healthCheck)
+	}
+
+	var wg sync.WaitGroup
+	monitor := &Monitor{
+		ctx:          ctx,
+		wg:           &wg,
```
deads2k, line 228: I thought we didn't normal do these by reference.  perhaps I got it backwards, but can you double check.

pkg/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,279 @@
...
+
+	return monitor, nil
+}
+
+func (m *Monitor) Schedule() {
+	for _, healthCheck := range m.healthChecks {
+		m.wg.Add(1)
+		go func(healthCheck *health.Check) {
+			defer m.wg.Done()
+			err := wait.PollImmediateUntil(m.interval,
```
deads2k, line 241: update to `UntilWithContext` when it exists.

deads2k: actually `UntilWithContext` may in 1.21

deads2k, line 241: Because this waits for one iteration to be complete before starting another, I think we need to set all the timeouts to 1s or less.  I'm ok doing it this way for simplicity of reading, but part of the point here is high granularity.

hexfusion: nice +1

pkg/cmd/monitor/monitor.go
hexfusion: I think its ok for multitarget because we don't need to set so I will ensure we skip set in multitarget to keep code path clear.

pkg/cmd/monitor/health/health.go
hexfusion: sure will provide matrix in the go doc

pkg/cmd/monitor/monitor.go
hexfusion: the idea is 1 client per target and then each target has multiple health checks using same target. I wanted to reduce the number of clients but I will check perf both ways.

pkg/cmd/monitor/health/log.go
```diff
@@ -0,0 +1,112 @@
...
+			zap.Any("check", c.name),
+			zap.Bool("health", false),
+			zap.Time("start", c.status.logDisruptionStart),
+			zap.Error(err),
+		)
+		return
+	}
+	//sanitize context error
+	if errors.Is(err, context.DeadlineExceeded) {
+		err = fmt.Errorf("client timeout of exceeded: %s", DefaultContextDeadline)
```
hexfusion, line 27: typo here drop `of`

pkg/cmd/monitor/monitor.go
hexfusion: cleaned this up no need for it to be in the Monitor passing it as pointer is the standard.

pkg/cmd/monitor/health/health.go
```diff
@@ -0,0 +1,142 @@
...
+		ctx:     ctx,
+		client:  client,
+		lg:      lg,
+		targets: targets,
+	}
+}
+
+// Health checks will log requests over DefaultSlowRequestDuration with a timeout at DefaultContextDeadline.
+
+// WithSerializedReadSingleTarget performs a serialized RangeRequest against a single target.
```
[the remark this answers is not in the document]
hexfusion: updated docs to reflect why this test exists.

pkg/cmd/monitor/health/health.go
hexfusion: WithQuorumRead would be directly reflective of apiserver usage

bindata/etcd/pod.yaml
hexfusion: steady-state

bindata/etcd/pod.yaml
```diff
@@ -244,6 +264,41 @@ ${COMPUTED_ENV_VARS}
...
+          fieldRef:
+            fieldPath: metadata.name
+    volumeMounts:
+      - mountPath: /var/log/etcd/
+        name: log-dir
+      - mountPath: /etc/kubernetes/static-pod-certs
+        name: cert-dir
+    resources:
+      requests:
+        memory: 600Mi
```
hexfusion, line 300: changed wrong container

hexfusion:
> The direction here looks good. Some of data will only be useful when correlated with information about the cluster state like whether the pods are starting and stopping.
> 
> If I've understood the checks correctly, the `WithQuorumRead: performs a quorate RangeRequest against all targets.` matches what the kube-apiserver check does. This check should never fail in a running cluster. If I've understood correctly, then the next step after this merges is to find a way to report that happening via the API. This may be as simple as a periodic grep of the file every minute and marking some well known configmap with the X latest blips and trying to emit events for them.

quorum-guard is in this general space so maybe it can handle this for now?

hexfusion:
/test e2e-gcp-five-control-plane-replicas
```
The HAProxy router should set Forwarded headers appropriately [Skipped:Disconnected] [Suite:openshift/conformance/parallel] 
```
clayton is looking at this https://github.com/openshift/origin/pull/26213

unrelated to the change

hexfusion:
> failed to acquire lease for "aws-quota-slice": resources not found

AWS quotas......

/retest

hexfusion:
AWS
level=warning msg=DependencyViolation: The vpc 'vpc-0a60ecfa309b0e5a0' has dependencies and cannot be deleted. arn=arn:aws:ec2:us-west-1:460538899914:vpc/vpc-0a60ecfa309b0e5a0

/test e2e-agnostic

hexfusion:
PDB issue is blocked until [1] merges.
[1] https://github.com/openshift/cluster-kube-controller-manager-operator/pull/534

pkg/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,282 @@
+package monitor
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"k8s.io/apimachinery/pkg/util/wait"
```
hexfusion, line 8: could be a followup..

pkg/cmd/monitor/health/health.go
```diff
@@ -0,0 +1,156 @@
...
+		c.logSlowRequest(fmt.Errorf("slow request"))
+	})
+	defer slowRequestTimer.Stop()
+
+	// Without a context timeout we will retry non mutating request indefinitely.
+	ctx, cancel := context.WithTimeout(ctx, DefaultContextDeadline)
+	defer cancel()
+	// pin targets for single target tests
+	if len(c.targets) == 1 {
+		c.client.SetEndpoints(c.targets...)
```
deads2k, line 135: This call here. You're sharing the one client you create here.  If you dropped this line, I wouldn't worry about.  Do you need this line?

deads2k:
Just the comment about shared clients being mutated per instance in go funcs.

Is there a peer PR adding this to must-gather?

pkg/cmd/monitor/health/health.go
hexfusion: I create one client per target so each slice of checks has the same endpoints. While we set we set the same value. I create each client with 3 endpoints. The reason being we will fail to startup if that single endpoint doesn't respond. So in the case of upgrade we could be in fail loop. 

```
for each []targets {
   create client
   for each []check {
     consume client
```

hexfusion: I can start and set early will fix it.

hexfusion:
/test e2e-agnostic-upgrade



<br class="Apple-interchange-newline">

hexfusion:
> Just the comment about shared clients being mutated per instance in go funcs.
> 
> Is there a peer PR adding this to must-gather?

Yes https://github.com/openshift/must-gather/pull/235

hexfusion:
/skip
disruptive tests are not passing unrelated to this PR

geliu2016 (commented):
The container: etcd-health-monitor log show perform health checks every 2s, is it expected?

{"level":"info","ts":"2021-06-10T12:29:56.908Z","caller":"health/log.go:29","msg":"health check","addresses":["https://10.0.187.41:2379"],"check":"QuorumReadSingleTarget","health":false,"start":"2021-06-10T12:29:11.863Z","error":"client timeout exceeded: 1s"}
{"level":"warn","ts":"2021-06-10T12:29:58.753Z","caller":"clientv3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"endpoint://client-a0ff699a-31bd-4771-8e44-2ca5be0a9d72/10.0.140.160:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp 10.0.187.41:2379: connect: connection refused\""}
{"level":"info","ts":"2021-06-10T12:29:58.753Z","caller":"health/log.go:29","msg":"health check","addresses":["https://10.0.187.41:2379"],"check":"SerializedReadSingleTarget","health":false,"start":"2021-06-10T12:29:11.712Z","error":"client timeout exceeded: 1s"}
{"level":"warn","ts":"2021-06-10T12:29:58.909Z","caller":"clientv3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"endpoint://client-a0ff699a-31bd-4771-8e44-2ca5be0a9d72/10.0.140.160:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp 10.0.187.41:2379: connect: connection refused\""}

hexfusion:
@geliu2016 great question here is how the timing breaks down. The code uses UntilWithContext[1] which runs the function then will wait and interval of 1s. So in the case of a healthy test we fire every second. But we also need a timeout because it is not completly reasonable for healthy to mean instant response. This is certainly true in a production cluster under load. So we have a 1 second context timeout for failure.

healthy check less than 2s
```
- start timer 00:00:00
- healthcheck (sub 1s)
- wait 1s
- retry
```

In the healthy case we are checking health sub 2s max with observability on an error once we enter the loop. Meaning no matter how long the test takes to fire the timer considers disruption as soon as the loop starts.

failure check
``` 
- disruption timer 00:00:00
- health check
- timeout 00:00:01
- wait 1s
- retry
- disruption timer 00:00:02
- health check
- timeout 00:00:01
- wait 1s
- disruption timer 00:00:04
```
As a follow-up we can consider more complex retry logic to for example skip wait on failure. But I think this is a good safe place to start.












So to your point it is perhaps not 100% accurate to say this is 100% 1s granularity but the calculations are pretty accurate.


[1]https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#UntilWithContext

hexfusion:
minor upgrade proof pending https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1403012378026577920

hexfusion:
SNO proof pending https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/596/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-single-node/1402974058710044672

hexfusion:
lets not keep grinding known fail non blocking tests

/override ci/prow/e2e-gcp-disruptive-ovn
/override ci/prow/e2e-aws-disruptive 
/override ci/prow/e2e-aws-single-node
/override ci/prow/e2e-gcp-disruptive-ovn
/override ci/prow/e2e-aws-disruptive-ovn

hexfusion:
/override ci/prow/e2e-gcp-disruptive-five-control-plane-replicas

geliu2016:
Hi Sam, thanks for reply, it make sense, LGTM.

hexfusion:
Moved upgrade to GCP where I saw green in 4.8 rehearsal https://github.com/openshift/release/pull/19168

/override ci/prow/e2e-gcp-disruptive

PR #598, opened 2021-05-14
Bug 1884568: etcd-pod: add liveness probe

hexfusion:
etcd should consider a liveness probe, by setting reasonable timeouts we should eliminate disruptive behavior which could result in flapping the service while under pressure.

This PR adds a liveness probe using a serialized get for a key that would not exist as a cheap condition for liveness.

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -193,6 +193,26 @@ ${COMPUTED_ENV_VARS}
       periodSeconds: 5
       successThreshold: 1
       timeoutSeconds: 5
+    livenessProbe:
+      exec:
+        command:
+        - /bin/bash
+        - -c
+        - |
+          set -xe
+
+          # liveness probe uses a serialized get to an non-existant key (liveness) as a condition.
+          unset ETCDCTL_ENDPOINTS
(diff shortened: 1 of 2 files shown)


hexfusion:
This would be 4.9 then consider backport

bindata/etcd/pod.yaml
```diff
@@ -193,6 +193,26 @@ ${COMPUTED_ENV_VARS}
...
       timeoutSeconds: 5
+    livenessProbe:
+      exec:
+        command:
+        - /bin/bash
+        - -c
+        - |
+          set -xe
+
+          # liveness probe uses a serialized get to an non-existant key (liveness) as a condition.
```
deads2k, line 204: I'm thinking about this.  We want to know we're failing, but I'm I want to be careful about indicating failure by crashlooping the thing we use to track what to run.  

Without a startup probe, this could result in a off-cycle etcd pods crashlooping and never starting at the same time to find quorum together, right?

hexfusion: its the main reason I have avoided this is the fear of race but in some ENV such as single node under disruption we might want to help kubelet.

PR #601, opened 2021-05-24
Bug 1966813: Update the flowchart to accommodate modified logic

retroflexer:
The logic is changed using https://github.com/openshift/etcd/pull/81, and the diagram needs to be updated to reflect the modified logic.

If the data-dir exists but it is still not a member, it is reasonable that we give the container a bit longer to be observable.


geliu2016:
LGTM

PR #602, opened 2021-05-26
Bug 1963871: bump(*): Update the go version to 1.16

retroflexer:
Bumping the go version to 1.16.

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: golang-1.15
+  tag: rhel-8-release-golang-1.16-openshift-4.8
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/openshift/cluster-etcd-operator
 
-go 1.15
+go 1.16
 
 require (
 	github.com/davecgh/go-spew v1.1.1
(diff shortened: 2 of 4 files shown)


lilic (commented):
/lgtm

/retest

lilic:
I believe we want to merge https://github.com/openshift/release/pull/18621 first, before bumping to go 1.16.

.ci-operator.yaml
```diff
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-8-release-golang-1.15-openshift-4.8
+  tag: rhel-8-release-golang-1.16-openshift-4.8
```
lilic, line 4: See https://github.com/openshift/cluster-etcd-operator/pull/607

We need 
```suggestion
  tag: golang-1.16
```

lilic (commented):
Based on new information from testplatform team.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/cluster-etcd-operator
 
-go 1.15
+go 1.16
```
sttts, line 3: this needs a `go mod tidy` call. go.sum is expected to change.

.ci-operator.yaml
sttts: actually this one might work. ckao uses the same.

go.mod
lilic: Surprised tests passed, maybe we should add in next release a go mod test as well.

.ci-operator.yaml
retroflexer: I left it as is to match with ckao and pushed the change with go.sum.

go.mod
retroflexer: Pushed the new go.sum.

PR #603, opened 2021-05-26
Bug 1965024: Validate the status of the etcd snapshot during backup and restore

retroflexer:
Backup and restore should perform consistency checks on etcd snapshots.

### DIFF
--- a/bindata/etcd/cluster-backup.sh
+++ b/bindata/etcd/cluster-backup.sh
@@ -35,9 +35,8 @@ if [ ! -d "$1" ]; then
 fi
 
 function check_if_operator_is_progressing {
-   operator="$1"
+   local operator="$1"
 
-   export KUBECONFIG="/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/node-kubeconfigs/localhost.kubeconfig"
    if [ ! -f "${KUBECONFIG}" ]; then
       echo "Valid kubeconfig is not found in kube-apiserver-certs. Exiting!"
       exit 1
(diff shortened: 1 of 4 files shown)


bindata/etcd/cluster-restore.sh
```diff
@@ -66,6 +66,18 @@ function wait_for_containers_to_stop() {
     echo "complete"
   done
 }
+function check_snapshot_status() {
```
hexfusion, line 69: should the snapshot file and backup-dir be inputs to the function?

bindata/etcd/cluster-restore.sh
```diff
@@ -66,6 +66,18 @@ function wait_for_containers_to_stop() {
...
+  dl_etcdctl
+  etcdctl snapshot status "${SNAPSHOT_FILE}" -w json > "${ETCD_DATA_DIR_BACKUP}"/etcd-backup-status-from-snapshot.json
+  tar -C "${ETCD_DATA_DIR_BACKUP}" -xzf "${BACKUP_FILE}" etcd-backup-status.json || true
+  # With the older backups, the backup status file may not exist.
+  if [ -f "${ETCD_DATA_DIR_BACKUP}/etcd-backup-status.json" ]; then
+    if ! diff -q "${ETCD_DATA_DIR_BACKUP}/etcd-backup-status.json" "${ETCD_DATA_DIR_BACKUP}/etcd-backup-status-from-snapshot.json" > /dev/null; then
```
hexfusion, line 75: instead of writing the second file to verify could we ask etcdctl and pipe output as stdin to diff?

```
etcdctl snapshot status "${SNAPSHOT_FILE}" -w json | diff "${ETCD_DATA_DIR_BACKUP}/etcd-backup-status.json" - 
```

bindata/etcd/cluster-restore.sh
```diff
@@ -66,6 +66,18 @@ function wait_for_containers_to_stop() {
...
+       echo "The snapshot file status hash does not match with the hash at the time of the backup. Exiting!"
```
hexfusion, line 76: can we make the error more clear the user may not understand what snapshot means or which hash as the context of the command is backup/restore. If you are goping to store all of the state from status the error should perhaps reflect which key failed. So maybe we only care about the hash as it precludes the others have changed?

```
backup integrity verification failed SHA256SUM mismatch
```

hexfusion:
After a conversation in slack we have decided to only perform bbolt consistency checks which are included for free with `etcdctl status $path` command.

bindata/etcd/cluster-restore.sh
```diff
@@ -67,6 +67,15 @@ function wait_for_containers_to_stop() {
   done
 }
 
+function check_snapshot_status() {
```
hexfusion, line 70: would it make sense for this to be a shared function?

bindata/etcd/cluster-restore.sh
```diff
@@ -67,6 +67,15 @@ function wait_for_containers_to_stop() {
...
+  SNAPFILE="$1"
```
hexfusion, line 71: we have standardized on`local varname="$1"`

https://github.com/openshift/cluster-etcd-operator/blob/c2cab02a3861e5a68d6f4fdc82903273c906d4da/bindata/etcd/cluster-backup.sh#L87

bindata/etcd/cluster-restore.sh
```diff
@@ -67,6 +67,15 @@ function wait_for_containers_to_stop() {
...
+  dl_etcdctl
+  if ! etcdctl snapshot status "${SNAPFILE}" -w json; then
+    echo "Backup integrity verfiication failed. Backup appears corrupted. Aborting cluster-restore!"
```
karmab, line 74: verfiication - verification

pkg/operator/etcd_assets/bindata.go
```diff
@@ -267,6 +275,15 @@ function wait_for_containers_to_stop() {
   done
 }
 
+function check_snapshot_status() {
+  SNAPFILE="$1"
+  dl_etcdctl
+  if ! etcdctl snapshot status "${SNAPFILE}" -w json; then
+    echo "Backup integrity verfiication failed. Backup appears corrupted. Aborting cluster-restore!"
```
karmab, line 282: verfiication typo

bindata/etcd/cluster-restore.sh
retroflexer: In future, we may be extending the function to include generation of the hash. The function provides a logical unit for a set of instructions and make the flow easy to follow and maintain.

retroflexer: Thanks. Well, earlier we were doing different things in `backup` and `restore` and hence didn't think about making it a shared function. Now that the functionality is identical, I made it a common function and moved it to common-tools.

bindata/etcd/cluster-restore.sh
hexfusion: Sorry I was not more clear but that standard we have been trying to adopt is the use of lowercase local variables inside of functions.

Ref: https://tldp.org/LDP/abs/html/localvar.html

retroflexer: I had thought about it too, but since other functions already had uppercase local variables, I went with the current flow.

I can change all of them to lowercase too.

bindata/etcd/cluster-backup.sh
```diff
@@ -110,6 +112,10 @@ if [ ! -f "$ETCDCTL_CACERT" ] && [ ! -d "${CONFIG_FILE_DIR}/static-pod-certs" ];
 fi
 
 dl_etcdctl
-backup_latest_kube_static_resources "${BACKUP_RESOURCE_LIST[@]}"
+backup_latest_kube_static_resources "${BACKUP_TAR_FILE}" "${BACKUP_RESOURCE_LIST[@]}"
```
hexfusion, line 115: if the list is only used to pass to the function why not just use the list in the function?

https://github.com/openshift/cluster-etcd-operator/pull/603/files#diff-b15139535382251f9a3e1ca944f23d804ad410e7a78107940e5f390b64487113R102

retroflexer: To maintain the symmetry with the function restore_static_pods() in `cluster-restore.sh`.

But I can change it.

retroflexer: Pushed the change.

retroflexer:
Manually installed a cluster with this PR, and verified that the backup and restore process works, as expected.

Retesting the disruptive tests here.

/test e2e-aws-disruptive
/test e2e-aws-disruptive-ovn

bindata/etcd/cluster-backup.sh
```diff
@@ -55,49 +54,50 @@ function check_if_operator_is_progressing {
 
 # backup latest static pod resources
 function backup_latest_kube_static_resources {
-  RESOURCES=("$@")
+  local backup_tar_file="$1"
+  local backup_resource_list=("kube-apiserver" "kube-controller-manager" "kube-scheduler" "etcd")
```
hexfusion, line 58: no change required: we could just loop the list vs crate var.

hexfusion:
based on manual testing per conversation with @retroflexer 
/lgtm
/approve

PR #606, opened 2021-05-28
Rotate serving certs when duration less than minimum percent

marun:
This will ensure against expired serving certs breaking the cluster.

/cc @hexfusion

### DIFF
--- a/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
+++ b/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
@@ -30,9 +30,15 @@ import (
 	"github.com/openshift/cluster-etcd-operator/pkg/tlshelpers"
 )
 
-// Annotation key used to associate a cert secret with a node uid. This allows
-// cert regeneration if a node was deleted and added with the same name.
-const nodeUIDAnnotation = "etcd-operator.alpha.openshift.io/cert-secret-node-uid"
+const (
+	// Annotation key used to associate a cert secret with a node uid. This allows
+	// cert regeneration if a node was deleted and added with the same name.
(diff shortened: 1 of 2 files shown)


pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -348,10 +354,22 @@ func checkCertValidity(certBytes, keyBytes []byte, ipAddresses []string, nodeUID
...
-	// TODO(marun) Check that the certificate is not expired or due for replacement
+	// TODO(marun) Check that the certificate was issued by the CA
 
 	// Cert is valid
 	return "", nil
 }
+
+// lessThanMinimumDuration indicates whether the provided cert has less
+// than the provided minimum percentage of its duration remaining.
+func lessThanMinimumDuration(cert *x509.Certificate, minDurationPercent float64) bool {
```
lilic, line 369: One suggestion: if you pass directly the expiry and duration, this could be nicely unit tested with various test cases.

for example:

```go
func lessThanMinimumDuration(expiry, duration time.Time, minDurationPercent float64)
```

and call above with:
```go
lessThanMinimumDuration(leafCert.NotAfter, leafCert.NotBefore, defaultMinDurationPercent)
```

lilic (commented):
One small suggestion from my side.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
marun: Done [#lawofdemeter](https://en.wikipedia.org/wiki/Law_of_Demeter)

hexfusion:
/approve

temping to merge in 4.8, feelings @deads2k?
/hold

marun:
/retest

aws quota issues preventing successful cluster deployment

PR #613, opened 2021-06-18
Migrate alerting mixin from cluster-monitoring-operator

lilic:
As upstream etcd uses jsonnet mixin, as discussed I used jsonnet and jb to generate the manifests. We also override openshift specific things. In the next steps I will be adding fixes for alerts but this just brings in what we already today have in OpenShift via cluster-monitoring-operator. https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/control-plane.libsonnet#L8 
Original PromRule manifest file is located https://github.com/openshift/cluster-monitoring-operator/blob/7f4925a7203622d70b3007fbddfb6bc5cce6c1d9/assets/control-plane/etcd-prometheus-rule.yaml.

Reqs:
- `jsonnet` -> https://jsonnet.org/
- `jb` -> https://github.com/jsonnet-bundler/jsonnet-bundler 
/hold

Mainly opening so we discuss this approach, I can cleanup any generation scripts (its hacky).


hack/generate.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -eux
+
+# Generate jsonnet mixin prometheusrule manifest.
+
+cd jsonnet && jb update && jsonnet -J vendor main.jsonnet  | gojsontoyaml > ../manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```
hexfusion, line 6: question: could we look into adding this to github.com/openshift/build-machinery-go so the rest of the control-plane can emulate this process if they would like. Then also it will become part of our Makefile workflow? `make update`

lilic: Yes that sound good, if you don't mind I would do that in a follow-up. 

Note that this would only be used if we fix things upstream or once a new release is brought in.

hexfusion: should we add a linter test as part of the process? Is jsonnet-lint useful

hexfusion:
/test e2e-agnostic-upgrade



<br class="Apple-interchange-newline">

hack/generate.sh
lilic: I use jsonnetfmt in my vim so it formats everything correctly, but haven't tried jsonnet-lint. Can look into it.

lilic:
/hold 

Until CMO PR is approved, so it's both merged on the same day into the same nightly.

jsonnet/main.jsonnet
```diff
@@ -0,0 +1,35 @@
...
+
+{
+  apiVersion: 'monitoring.coreos.com/v1',
+  kind: 'PrometheusRule',
+  metadata: {
+    name: 'etcd-prometheus-rules',
+    namespace: 'openshift-etcd-operator',
+    annotations:
+      {
+        'include.release.openshift.io/ibm-cloud-managed': 'true',
```
marun, line 27: I think there is a non-zero chance that developers external to the team may need to add more of these annotations in the future. Maybe inject a comment into the manifest after conversion to yaml (since json doesn't support comments) and also add a verify check that prevents a manual change from being committed (since it would be overwritten by a subsequent generate invocation)?

I'm guessing this is what @hexfusion meant by updating make, though, so it can be done in a follow-up.

lilic: As discussed will do in a follow up, and sync with other folks on this.

lilic:
/hold

seems like we have etcdHighNumberOfLeaderChanges fire but only for `e2e-gcp-five-control-plane-replicas`, the alert was not changed so not sure what is up with that, will have a look.

lilic:
Adjusted etcdHighNumberOfLeaderChanges to the value of 5, it was 4 before and when disruption happened it could be the case that leaders change more frequently, but only briefly just hit the mark of 4. The cluster was operational so we know that limit is okay to be hit in OpenShift.

I also brought it in the alert I fixed upstream.

lilic:
Seems like etcdHighNumberOfFailedGRPCRequests is failing on `e2e-gcp-five-control-plane-replicas `, will bring it in another time so we can merge the existing alerts.

lilic:
/retest

unrelated: 
> : [sig-storage] Multi-AZ Cluster Volumes should schedule pods in the same zones as statically provisioned PVs [Suite:openshift/conformance/parallel] [Suite:k8s]

lilic:
/retest

unrelated

lilic:
Seems like the https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/613/pull-ci-openshift-cluster-etcd-operator-master-e2e-gcp-five-control-plane-replicas/1408051931028393984 is known to the storage team.

hexfusion:
upstream test needs fixed, failure not related to code.

/override ci/prow/e2e-gcp-five-control-plane-replicas

PR #619, opened 2021-06-30
Bug 1969633: pkg/operator/targetconfigcontroller: wait for kcm-o to generate certs before rollout

hexfusion:
openshift-managed-config configmap csr-controller-ca is a critical resource for the cluster and is generated by openshift-kube-apiserver-operator render command using ca minted on the installer. When KCM-O starts it deletes and replaces this resource and then manages cert rotation moving forward.

The PR fixes a race condition in SNO where if etcd and or kas rollout and generate disruption before the ca pivots manager we can brick the cluster if the operator loses leadership before the resource is generated. This can lead to node Not Ready and the operator not able to start again. Now we wait for the signal of pivot then continue rollout.

related to: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1169


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -180,6 +182,28 @@ func (c *TargetConfigController) getSubstitutionReplacer(operatorSpec *operatorv
...
+	// key resources have rotated.
+	csrControllerCAConfigMap, err := c.configMapLister.ConfigMaps(operatorclient.GlobalMachineSpecifiedConfigNamespace).Get("csr-controller-ca")
+	if err != nil {
+		return err
+	}
+
+	csrControllerCAConfigMapCopy := csrControllerCAConfigMap.DeepCopy()
+	var isCSRControllerCARotated bool
+	for _, managedField := range csrControllerCAConfigMapCopy.ManagedFields {
+		if managedField.Manager == "cluster-kube-controller-manager-operator" {
```
marun, line 197: I'm curious, is there a precedent for depending on ManagedField metadata to infer whether another agent in the cluster has performed some kind of operation on a resource? And is the presence of ckcmo as a field manager necessarily a guarantee of successful rotation? 

It's possible that this comment is just a reflection of my unease with the prevalence of the many 'API's' in OpenShift for which no explicit contract is published or agreed upon.

/cc @sttts @deads2k

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -180,6 +182,28 @@ func (c *TargetConfigController) getSubstitutionReplacer(operatorSpec *operatorv
...
+func (c *TargetConfigController) checkDependencies() error {
+	// Ensure kube-controller-manager has rotated csr-controller-ca before rolling out.
+	// Waiting here is very important in single node as we want to avoid disruption until
+	// key resources have rotated.
+	csrControllerCAConfigMap, err := c.configMapLister.ConfigMaps(operatorclient.GlobalMachineSpecifiedConfigNamespace).Get("csr-controller-ca")
+	if err != nil {
+		return err
+	}
+
+	csrControllerCAConfigMapCopy := csrControllerCAConfigMap.DeepCopy()
```
marun, line 194: Maybe skip the copy if no modification is performed?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
hexfusion: As the original resource was created by kube-apiserver-operator I believe KCM creating the resource would be indication of the signal required.

hexfusion: I can review the key state to have a more concrete understanding of the signer so I will take that approach and see if the signal is more clear.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
hexfusion: was just trying to be extra cautious with the object but point taken.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -246,3 +264,38 @@ func (c *TargetConfigController) namespaceEventHandler() cache.ResourceEventHand
...
+	block, _ := pem.Decode([]byte(caBundle))
+	leafCert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return err
+	}
+	if len(leafCert.Subject.CommonName) == 0 {
+		return fmt.Errorf("configmap openshift-config-managed/csr-controller-ca is not valid, waiting on kube-controller-manager-operator to generate")
+	}
+
+	isCASubjectExpected := strings.HasPrefix(leafCert.Subject.CommonName, "kube-csr-signer")
```
hexfusion, line 285: this might be too brittle but I added it as another validation signal as manager might not be enough? we could consider additional annotations or labels on the configmap as well.

soltysh: Hmm... I wonder if we couldn't just check that CM's `.metadata.managedFields.manager` instead, if last to touch it is `cluster-kube-controller-manager-operator` it means we got the reaction we cared about. Before the manager will be `cluster-kube-apiserver-operator` iirc, right?

soltysh (approved):
/lgtm

hexfusion:
/override ci/prow/e2e-gcp-five-control-plane-replicas

failure not related to this PR

```
 [sig-storage] Multi-AZ Cluster Volumes should schedule pods in the same zones as statically provisioned PVs [Suite:openshift/conformance/parallel] [Suite:k8s]
 ```

PR #625, opened 2021-07-12
ETCD-58: add support for defrag controller

hexfusion:
enhancement: https://github.com/openshift/enhancements/pull/836


marun:
@hexfusion Is there a design doc or enhancement that I could review to learn more about the rational and intent for this work?

hexfusion:
> @hexfusion Is there a design doc or enhancement that I could review to learn more about the rational and intent for this work?

I am working on a PR to enhancements.

pkg/etcdcli/health.go
```diff
@@ -72,12 +69,7 @@ func GetMemberHealth(etcdMembers []*etcdserverpb.Member) memberHealth {
 				}
 				hc.Healthy = true
 			} else {
-				if err == rpctypes.ErrPermissionDenied {
```
hexfusion, line 75: this was cruft copied from etcdctl `ErrPermissionDenied` is only returned when etcd RBAC is enabled whihc we will never use.

pkg/etcdcli/etcdcli.go
```diff
@@ -258,6 +275,14 @@ func (g *etcdClientGetter) MemberList() ([]*etcdserverpb.Member, error) {
 	return membersResp.Members, nil
 }
 
+func (g *etcdClientGetter) Status(ctx context.Context, member *etcdserverpb.Member) (*clientv3.StatusResponse, error) {
+	cli, err := g.getEtcdClient()
+	if err != nil {
+		return nil, err
+	}
+	return cli.Status(ctx, member.Name)
```
hexfusion, line 283: dial ClientURL not Name!

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (c *DefragController) checkDefrag(ctx context.Context, recorder events.Recorder) error {
+	controlPlaneTopology, err := ceohelpers.GetControlPlaneTopology(c.infrastructureLister)
+	if err != nil {
+		return err
+	}
+	// defrag is blocking and can only be safely performed in HighlyAvailableTopologyMode
+	if controlPlaneTopology != configv1.HighlyAvailableTopologyMode {
+		klog.V(4).Infof("DefragController is disabled for control plane topology: %s", controlPlaneTopology)
+		return nil
```
lilic, line 85: Should we also update the operator status here to reflect this? Otherwise the status would be false degraded and AsExpected.

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -0,0 +1,188 @@
...
+			return err
+		}
+		if leader == nil && status.Leader == member.ID {
+			leader = status
+			continue
+		}
+		endpointStatus = append(endpointStatus, status)
+	}
+
+	// leader last if possible
```
lilic, line 116: I guess this is on best effort basis, for example if by chance there is a re-election by the time we finish these checks, we would not do this correct?

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -0,0 +1,188 @@
...
+				errors = append(errors, err)
+			}
+			postDefragEndpointStatus, err := c.etcdClient.Status(ctx, member)
+			if err != nil {
+				errors = append(errors, err)
+			}
+
+			recorder.Eventf("DefragmentSuccess", "etcd member has been defragmented: %s, dbSize: %d, dbInUse: %d, member ID: %d, leader ID: %d", member.Name, postDefragEndpointStatus.DbSize, postDefragEndpointStatus.DbSizeInUse, member.ID, postDefragEndpointStatus.Leader)
+
+			//give cluster time to recover before we move to the next member.
```
lilic, line 144:
```suggestion
			// Give cluster time to recover before we move to the next member.
```

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -0,0 +1,188 @@
...
+			if err := wait.Poll(
+				waitDuration,
```
lilic, line 146: Is two seconds before we query for member health enough for us to determine this?

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+
+	return v1helpers.NewMultiLineAggregate(errors)
+}
+
+// isEndpointFragmented checks the status of all cluster members to ensure that no members have a fragmented store.
+// This can happen if the operator starts defrag of the cluster but then loses leader status and is rescheduled before
+// the operator can defrag all members.
+func isEndpointFragmented(member *etcdserverpb.Member, endpointStatus *clientv3.StatusResponse) bool {
```
lilic, line 170: Just a suggestion: But if we don't pass the member (which we only use to log, and we can move that one up), we could nicely unit test this function.

lilic (commented):
Nice work! Couple of comments, suggestions and questions.

I wonder if we could add some error metrics or what would be a clear sign for admin that this operation went wrong?

pkg/operator/defragcontroller/defragcontroller.go
hexfusion: good point I added a disabled status

pkg/operator/defragcontroller/defragcontroller.go
hexfusion: right the checks are very fast but its possible but not probable

pkg/operator/defragcontroller/defragcontroller.go
hexfusion: I think so that is 2x election timeout, plus we have a retry until 30s if nessisary.

pkg/operator/defragcontroller/defragcontroller.go
hexfusion: added full unit tests

pkg/operator/defragcontroller/defragcontroller.go
lilic: Could we add a log here then, just to see if this is the case (might be good for troubleshooting), e.g. log level debug "Appending leader last".

hexfusion: sure

hexfusion:
> I wonder if we could add some error metrics or what would be a clear sign for admin that this operation went wrong?

We have the grpc_ metrics on defrag we could consider a warning alert if we dont see a defrag success?

```
grpc_server_handled_total{grpc_method="Defragment", grpc_code="OK"}
```

lilic:
> We have the grpc_ metrics on defrag we could consider a warning alert if we dont see a defrag success?

I think that is a good one, but it should be in general handled by https://github.com/openshift/cluster-etcd-operator/pull/626/files#diff-e1152f2c6995245f9385822461030307fb3662b7039b60a71519bcd8fd604002R62 when we can bring it in. So we can add it to the runbook that defragmentation failure could be an issue, what do you think?

lilic (commented):
Thanks for addressing the comments, this lgtm! 🎉 

/lgtm
/hold

Put a hold in case you want someone else to have a look. 

Let's also adjust the runbooks later on, to not suggest manual defrag anymore.

hexfusion:
well the best way to improve this in my opinion is to exercise it at scale and see if we can flush any corner cases in CI. thanks!

/hold cancel

PR #634, opened 2021-07-29
Bug 1986829: metrics: use client cert auth for metrics scraping

stlaz:
The CMO is now capable of using client cert auth to scrape the metrics endpoints. Use that in our components to reduce API load and allow metrics collection when the API is down - given further authz conditions are met.


lilic (approved):
/lgtm

Thanks! Checked metrics are scrapped correctly, `openshift_etcd_operator_build_info` and `etcd_debugging_raft_terms_total` seems to be the only etcd operator metric we have?

/hold

stlaz:
/retest
I am not quite sure about what metrics the etcd-operator exposes but as long as it's got the serviceMonitor configured, it's going to be generating a TokenReview request to the KAS every 2 minutes or so (I believe we added caching the response a few months back).

marun:
@hexfusion 5 node failure is unrelated, maybe override the job?

PR #635, opened 2021-07-29
Bug 1989335: pkg/operator/targetconfigcontroller: block rollout and report on missing external resource

hexfusion:
The current code reports the invalid resource but does not block a new revision from being created. In the case that configmap openshift-config-managed/csr-controller-ca is not managed by cluster-kube-controller-manager-operator block if SNO. This will ensure we don't rollout and cause disruption before critical resource generation.

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/informers"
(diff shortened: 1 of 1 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -114,8 +116,9 @@ func createTargetConfig(ctx context.Context, c TargetConfigController, recorder
 	if err != nil {
 		return false, err
 	}
-	if err := checkExternalDependencies(ctx, c.configMapLister, recorder); err != nil {
-		errors = append(errors, err)
+	if err := checkExternalDependencies(ctx, c.configMapLister, c.infrastructureLister, recorder); err != nil {
+		recorder.Warning("MissingRequiredResource", err.Error())
```
marun, line 120: Will `checkExternalDependencies` ever return an error that does not indicate a missing resource?

hexfusion: yeah, not in love with that.

hexfusion: updated.

lilic (commented):
/lgtm

Note this still has a hold, CI failure is unrelated.

PR #638, opened 2021-08-09
Bug 1991068: Allow only supported cipher suites

lilic:
/hold

Want to run tests and launch a cluster from this to verify tomorrow. Also want to add some unit tests for this before.


lilic:
Update on failed unit tests:
- kept failing locally but seems like its due to darwin env
- ran in docker and got a proper failure `#9 244.5 E0810 10:00:57.669639   11632 quorumguardcontroller.go:294] failed to convert replica , err %!w(*strconv.NumError=&{Atoi  0xc000046200})` so looking into it now.

hexfusion (approved):
/lgtm

/hold for minor upgrade test via clusterbot

```
test upgrade 4.8.0-0.nightly-2021-08-09-135211 https://github.com/openshift/cluster-etcd-operator/pull/638
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1425105910723776512

hexfusion:
/hold cancel 

upgrade test failed but not as a result of this PR. art still has nightly build failure[1] which I believe is root cause here.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1992673

PR #639, opened 2021-08-10
Bug 1980465: pkg/etcdenvvar: warn on apply duration over 200ms

hexfusion:
Lets not scare folks with logs? Performance issues should be debugged with metrics. Logs are useful but 100ms threshold is quite common in a production cluster and does not necessarily conclude a performance issue. Moving this threshold to 200ms makes the log more of a useful potential signal.


lilic (approved):
/lgtm
/approve

💯 good idea!

geliu2016 (approved):
/lgtm, but I just curious why change to 200ms, but not 300ms or anyother?

hexfusion:
> /lgtm, but I just curious why change to 200ms, but not 300ms or any other?

Its an experiment that we can iterate on. But 200ms is 2x the heartbeat interval and 1/5 of the election timeout.

PR #640, opened 2021-08-16
Bug 1993757: bindata/etcd: remove unix socket from advertised list

hexfusion:
we don't need to advertise and it breaks some documented etcdctl commands using the --cluster flag


lilic:
/retest

unrelated it seems

lilic (approved):
/lgtm
/approve

/retest
Failure looks unrelated to me.

PR #592, opened 2021-05-10
Revert "Bug 1957498: Update policy.v1beta1 to v1 as it is deprecated in v1.21"

dmage:
Reverts openshift/cluster-etcd-operator#589

PR #631, opened 2021-07-22
Bug 1957498: Clean up tech debt

hexfusion:
This PR removes a few areas of code that performed various checks and cleanups for 4.7-> 4.8 and are no longer needed in 4.9.

PR #643, opened 2021-08-17
Bug 1957498: Revert "Revert "Bug 1957498: Update policy.v1beta1 to v1 as it is deprecated in v1.21""

ravisantoshgudimetla:
Reverts openshift/cluster-etcd-operator#592

4.8 by default has v1 enabled so upgrades from 4.8 -> 4.9 should be smooth


ravisantoshgudimetla:
/close

in favor of #631

PR #645, opened 2021-08-19
Bug 1997207: Pass context into clientv3.Config to use same context

ardaguclu:
grpc health check does not use context in here https://github.com/openshift/cluster-etcd-operator/blob/ab70a6487419774dabd514824e20f639e8a555bd/pkg/cmd/monitor/health/health.go#L108. Primary reason of this is context value is nil in here https://github.com/openshift/cluster-etcd-operator/blob/ab70a6487419774dabd514824e20f639e8a555bd/vendor/go.etcd.io/etcd/client/v3/client.go#L293. Thereby, cancellations or any other context based operations have no effect on this function.

This PR passes ctx to stick same context also for grpc health check dials.

### DIFF
--- a/pkg/cmd/monitor/monitor.go
+++ b/pkg/cmd/monitor/monitor.go
@@ -205,7 +205,7 @@ func (o *monitorOpts) newMonitor(ctx context.Context, lg *zap.Logger, singleTarg
 	// Create single target checks one check per target
 	for _, target := range targets {
 		// one client per target to eliminate lock racing
-		client, err := newETCD3Client(tlsInfo, targets)
+		client, err := newETCD3Client(ctx, tlsInfo, targets)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


hexfusion:
I appreciate the PR. In general I think it will be OK to add this context as it can handle client teardown. but to clarify

https://github.com/openshift/cluster-etcd-operator/blob/ab70a6487419774dabd514824e20f639e8a555bd/vendor/go.etcd.io/etcd/client/v3/client.go#L293-L296

ardaguclu:
> I appreciate the PR. In general, I think it will be OK to add this context as it can handle client teardown. but to clarify the dial timeout does set a context for Dial. Although the parent is not explicitly passed from client config. If it were nil it would panic at 296.
> 
> https://github.com/openshift/cluster-etcd-operator/blob/ab70a6487419774dabd514824e20f639e8a555bd/vendor/go.etcd.io/etcd/client/v3/client.go#L293-L296

Thanks for review @hexfusion. Like you said, it is not nil. It is set in here with new context if it not set explicitly https://github.com/openshift/cluster-etcd-operator/blob/d5681e1e1e37038e427d286dc2b44a64a4372f23/vendor/go.etcd.io/etcd/client/v3/client.go#L336 . This PR just sets current context instead new one.

hexfusion:
Ardu this needs a bug if you would not mind?

ardaguclu:
@hexfusion , yes sure I can link a bug after opening in bugzilla.

PR #647, opened 2021-08-24
[wip] pkg/operator: add cluster backup upgrade controller

hexfusion:


bindata/etcd/cluster-backup-pod.yaml
```diff
@@ -0,0 +1,52 @@
...
+        name: cert-dir
+      - mountPath: /etc/kubernetes/manifests
+        name: static-pod-dir
+  priorityClassName: system-node-critical
+  nodeSelector:
+    node-role.kubernetes.io/master: ""
+  restartPolicy: Never
+  hostNetwork: true
+  tolerations:
+  - operator: "Exists"
```
wking, line 39: I'm not all that familiar with tolerations, but from [the docs][1], I'd expect `key` and `effect` here too.  Is a bare `Exists` something that should be documented as a special case in [our conventions][2]?

[1]: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
[2]: https://github.com/openshift/enhancements/blob/39336b3bacc86aa28c01cec5b2a1b9fcd7de81e1/CONVENTIONS.md#taints-and-tolerations

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+		operatorClient:       operatorClient,
+		operatorVersion:      operatorVersion,
+		kubeClient:           kubeClient,
+		podLister:            kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Pods().Lister(),
+		nodeLister:           kubeInformers.InformersFor("").Core().V1().Nodes().Lister(),
+		networkLister:        networkInformer.Lister(),
+		clusterVersionLister: clusterVersionInformer.Lister(),
+		targetImagePullSpec:  targetImagePullSpec,
+	}
+	return factory.New().ResyncEvery(10*time.Minute).WithInformers(
```
wking, line 74: Does this make the latency between the CVO asking for a backup, and etcd claiming the backup's completed 0-to-10 minutes for etcd to notice, and then a few minutes more for the backup itself?  Or do the informers allow the etcd operator to notice with an edge trigger, and this 10 minutes is... something else?

bindata/etcd/cluster-backup-pod.yaml
hexfusion: to be honest it was cargo culted from etcd operand as I threw the pod spec together. I will as a follow-up clean up the spec and ensure tolerations etc are sane.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hexfusion: Good question we have an informer on clusterVersion so any change to that resource would cause sync(). The 10 minute resync would be in addition to any change in informer resources. So assuming steady-state this controller would resync every 10m.

```
clusterVersionInformer.Informer(),
```

hexfusion: The timing here should be very fast. I would expect the normal situation to be a few seconds (20 or less) from the time you update to backup complete.

bindata/etcd/cluster-backup-pod.yaml
```diff
@@ -0,0 +1,52 @@
...
+  containers:
+  - name: cluster-backup
+    imagePullPolicy: IfNotPresent
+    terminationMessagePolicy: FallbackToLogsOnError
+    command:
+      - /bin/sh
+      - -c
+      - |
+        #!/bin/sh
+        /usr/local/bin/cluster-backup.sh /etc/kubernetes/static-pod-resources/cluster-backup
```
hexfusion, line 16: not a huge fan of this path

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+
+		switch pod.Status.Phase {
+		case corev1.PodSucceeded:
+			return operatorv1.ConditionTrue, nil
+		case corev1.PodFailed:
+			if err := delete(pod); err != nil {
+				klog.Errorf("failed to delete pod %s/%s: %v", pod.Namespace, pod.Name, err)
+			}
+			return operatorv1.ConditionFalse, nil
```
hexfusion, line 165: I think we want a way to update status to false giving clear failure signal but then enqueue to ensure fast retry.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+	_, _, err = v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+		Type:   "RecentBackup",
+		Status: backupStatus,
+		Reason: backupVersion,
+	}))
+	return err
+}
+
+func ensureRecentBackup(ctx context.Context, kubeClient kubernetes.Interface, podLister corev1listers.PodLister, nodeLister corev1listers.NodeLister, clusterBackupVersion, targetImagePullSpec string, recorder events.Recorder) (operatorv1.ConditionStatus, error) {
+	delete := func(pod *corev1.Pod) error {
```
hexfusion, line 135: deleteFunc

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+			Type:   "RecentBackup",
+			Status: operatorv1.ConditionUnknown,
+			Reason: backupVersion,
+		}))
+		if err != nil {
+			return err
+		}
+	} else if backupCondition.Status != operatorv1.ConditionUnknown {
+		// Backup was taken or failed, but is not in progress.
+		return nil
```
hexfusion, line 113: I think we need to break this down to ensure timely retry and reduce leaks. 

If its not == true check again quickly.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	backupStatus, err := ensureRecentBackup(ctx, c.kubeClient, c.podLister, c.nodeLister, backupVersion, c.targetImagePullSpec, syncCtx.Recorder())
+	if err != nil {
+		return err
+	}
+	if backupStatus == operatorv1.ConditionUnknown {
+		// Backup is still in progress
+		return nil
```
hexfusion, line 122: again not sure we want to be very patient but also we don't want to abuse resources, needs eval on informer chatter.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+		// Work is done
+		return nil
+	}
+
+	// Find the backup condition
+	_, status, _, err := c.operatorClient.GetOperatorState()
+	if err != nil {
+		return err
+	}
+	backupVersion := c.operatorVersion
```
wking, line 99: How does the version come into this?  If we're currently taking a snapshot (as seen by an `Unknown` status on the current condition with an acceptably recent `lastTransitionTime` or some such), then we can keep waiting that backup out, right?  And if there is no backup running, or the running backup is too old, we start off a new one?  I'm not seeing a use-case for "5s ago I kicked off a snapshot, but now I see a different operator version, so kick off a new one".

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+	backupCondition := v1helpers.FindOperatorCondition(status.Conditions, backupConditionType)
+	newConditionRequired := backupCondition == nil || backupCondition.Reason != backupVersion
+	if newConditionRequired {
+		_, _, err := v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:   "RecentBackup",
+			Status: operatorv1.ConditionUnknown,
+			Reason: backupVersion,
```
wking, line 106: All of these `RecentBackup` conditions should have a chatty `message` too, so users who see ClusterVersion go `Failing=True` can read the message we bubble up from your condition and don't freak out.  Something like: "Taking an etcd snapshot in `$LOCATION` before beginning the update."?

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+		}))
+		if err != nil {
+			return err
+		}
+	} else if backupCondition.Status != operatorv1.ConditionUnknown {
+		// Backup was taken or failed, but is not in progress.
+		return nil
+	}
+
+	backupStatus, err := ensureRecentBackup(ctx, c.kubeClient, c.podLister, c.nodeLister, backupVersion, c.targetImagePullSpec, syncCtx.Recorder())
```
wking, line 116: might want to return a condition pointer and `error` here, instead of just the status, so we can create nice reason/message tuples for the various cases.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
wking: also, I think failing fast with a `return` here is brittle.  Ideally we loop over all the backup pods, aggregate results, and attempt to delete any old/stale/failed pods.  And then if we had a fresh, successful backup, we'd return `True` even if there were some other pods which were old/stale/failed, regardless of how those pods were ordered in `backupPods`.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+	if len(errs) > 0 {
+		return "", utilerrors.NewAggregate(errs)
+	}
+
+	return operatorv1.ConditionUnknown, nil
+}
+
+func manageRecoveryPod(ctx context.Context, nodeName, targetImagePullSpec, clusterBackupVersion string, client coreclientv1.PodsGetter, recorder events.Recorder) (*corev1.Pod, bool, error) {
+	pod := resourceread.ReadPodV1OrDie(etcd_assets.MustAsset("etcd/cluster-backup-pod.yaml"))
+	pod.Name = "cluster-backup-" + nodeName
```
wking, line 190: Do we need `nodeName` here?  In `ensureRecentBackup` you were already failing fast if there were existing backup pods, so I'm having trouble seeing a path to having multiple going at once.  If we do feel like there could be multiple going at once, I don't see any guards to try to keep them on different nodes.  Can we use a random suffix or some such here if we want to support parallel backup pods, or drop the suffix entirely if we don't want to support parallel backup pods?

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+
+	return resourceapply.ApplyPod(ctx, client, recorder, pod)
+}
+
+// isRequireRecentBackup checks the conditions of ClusterVersion to verify if a backup is required.
+func isRequireRecentBackup(config *configv1.ClusterVersion) bool {
+	for _, condition := range config.Status.Conditions {
+		// Check if Failing is true and Message field containers the string RecentBackup.
+		if condition.Type == "Failing" && condition.Status == configv1.ConditionTrue {
+			return strings.Contains(condition.Message, "RecentBackup") == true
```
wking, line 207: nit: no need for `== true` here ;).

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+			continue
+		}
+		node, err := nodeLister.Get(pod.Spec.NodeName)
+		if err != nil {
+			return nil, err
+		}
+
+		// Ensure nodes have adequate storage for backup state.
+		if !isNodeAllocatableStorageAdequate(node) {
+			klog.V(4).Infof("node storage is not adequate for backup: %s, allocatable %s, required: %d", node.Name, node.Status.Allocatable.Storage().String(), etcdenvvar.QuotaBackendBytes*2)
```
wking, line 231: `etcdenvvar.QuotaBackendBytes*2` for `required` is making an assumption about the logic inside `isNodeAllocatableStorageAdequate`.  Can we update `isNodeAllocatableStorageAdequate` to return an `error`, and then just bubble that error into logs here if it's non-`nil`?

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hexfusion: lol

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hexfusion: yeah +1 I think this is the reality I used an initial design that was more robust but simplicity is the key and I will update to reflect.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hexfusion: makes sense

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hexfusion: right version control is probably not necessary in this design as we are not really managing backups for all versions but instead only 4.9 -> 4.9.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
wking: if it's sub-minute, I don't think there's any need for metrics coverage.  Might be interesting to emit events around this so we can spot-check a few in CI.  I dunno if there's an event emitter close by in the code that we could pipe in or not.

wking:
/retitle Bug 1997347: ETCD-223: pkg/operator: add cluster backup upgrade controller

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	pod.Spec.Containers[0].Image = targetImagePullSpec
+
+	return resourceapply.ApplyPod(ctx, client, recorder, pod)
+}
+
+// isRequireRecentBackup checks the conditions of ClusterVersion to verify if a backup is required.
+func isRequireRecentBackup(config *configv1.ClusterVersion) bool {
+	for _, condition := range config.Status.Conditions {
+		// Check if Failing is true and Message field containers the string RecentBackup.
```
lilic, line 205: Too much time in containers ;) 

Unless it's a joke, in which case I am all for leaving it! :D 

```suggestion
		// Check if Failing is true and Message field contains the string RecentBackup.
```

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+		if condition.Type == "Failing" && condition.Status == configv1.ConditionTrue {
+			return strings.Contains(condition.Message, "RecentBackup") == true
+		}
+	}
+	return false
+}
+
+// getNodeWithAdequateAllocatableStorage returns a control-plane node that meet the minimum storage requirements for the backup state.
+func getNodeWithAdequateAllocatableStorage(_ context.Context, podLister corev1listers.PodLister, nodeLister corev1listers.NodeLister) (*corev1.Node, error) {
```
lilic, line 214: I guess the plan is to use ctx in the future? I don't see it being needed right now by any of the existing functions?

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
lilic: +1

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,258 @@
...
+	if node == nil {
+		return false
+	}
+	nodeAllocatableStorageBytes, ok := node.Status.Allocatable.Storage().AsInt64()
+	if !ok {
+		klog.V(4).Infof("error converting allocatable storage to int64 %v", node.Status.Allocatable.Storage())
+		return false
+	}
+	// Require 2x worst case backup size as allocatable storage
+	// TODO is this too extreme? We can ask end point status for the actual size of the snapshot at the expense of a 2 RPC calls.
```
lilic, line 252: I would agree to do the calls here instead. We only do this operation on upgrades and if certain conditions are met so we won't be doing that many calls?

lilic (commented):
Couple of comments. As mentioned before overall idea lgtm, I would just like to see this in 4.8 first but think it was discussed on slack already.

One other thing that would be great here is:
- adding metric for failed backups
- creating alert if this backup fails

Are we blocking upgrades if this fails or we do not have the backup?

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hexfusion: yeah wasn't sure here if I was going to ask etcd about backend status.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hexfusion: OK more time but I want this right vs done.

bindata/etcd/cluster-backup-pod.yaml
```diff
@@ -0,0 +1,46 @@
...
+metadata:
+  name: cluster-backup
+  namespace: openshift-etcd
+  labels:
+    app: cluster-backup
+spec:
+  containers:
+  - name: cluster-backup
+    imagePullPolicy: IfNotPresent
+    t diff
```
wking, line 12: typo?

hexfusion: yeah, I am reworking all of this for Monday.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,288 @@
...
+			Type:    backupConditionType,
+			Reason:  "UpgradeBackupInProgress",
+			Message: fmt.Sprintf("Upgrade backup pod created on node %s", backupNodeName),
+		}, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	// Check the existing Pod Status and update condition respectively.
```
lilic, line 178: This logic does a bit more, and this msg is a bit misleading.

```suggestion
	// Check the existing Pod Status, reconcile pod and update condition respectively.
```

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -0,0 +1,288 @@
...
+
+	if len(errs) > 0 {
+		return "", utilerrors.NewAggregate(errs)
+	}
+
+	return "", ceohelpers.ErrNodeNotFound
+}
+
+func createBackupPod(ctx context.Context, nodeName, recentBackupName, operatorImagePullSpec, targetImagePullSpec string, client coreclientv1.PodsGetter, recorder events.Recorder) error {
+	pod := resourceread.ReadPodV1OrDie(etcd_assets.MustAsset("etcd/cluster-backup-pod.yaml"))
```
lilic, line 250: Do we really want to die here, I would prefer handling this gracefully as backup is not as critical as having a functioning cluster.

I could be misunderstanding what Die does.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller_test.go
```diff
@@ -0,0 +1,146 @@
...
+				Message: fmt.Sprintf("Need RecentBackup"),
+			},
+			wantBackupStatus: operatorv1.ConditionUnknown,
+		},
+		{
+			name: "RecentBackup not required invalid type",
+			clusterversionCondition: configv1.ClusterOperatorStatusCondition{
+				Type:    "NotFailing",
+				Status:  configv1.ConditionTrue,
+				Message: fmt.Sprintf("Need RecentBackup"),
```
lilic, line 57: if this is invalid type should this not fail?

lilic (commented):
Great stuff!

Couple of comments.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hexfusion: it will die if the pod yaml is invalid or does not exist.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller_test.go
hexfusion: well the idea here is to not process the operator unless a valid combination of Type and Message is observed. In this case we have the valid message but not the valid type. So we return nil as it is a no-opp (skip)

bindata/etcd/cluster-backup-pod.yaml
```diff
@@ -0,0 +1,76 @@
...
+    - hostPath:
+        path: /etc/kubernetes/manifests
+      name: static-pod-dir
+    - hostPath:
+        path: /etc/kubernetes/static-pod-resources
+      name: resources-dir
+    - hostPath:
+        path: /etc/kubernetes/static-pod-resources/etcd-certs
+      name: cert-dir
+
```
lilic, line 76: Nit: new line not needed?

pkg/cmd/verify/verify.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+type verifyStorageOpts struct {
+	errOut     io.Writer
+	kubeconfig string
+}
+
+// NewVerifyCommand performs a various checks.
```
lilic, line 14:
```suggestion
// NewVerifyCommand performs various checks.
```

pkg/cmd/verify/backupstorage.go
```diff
@@ -0,0 +1,227 @@
...
+			return v.verifyBackupStorage(ctx)
+		})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (v *verifyBackupStorage) verifyBackupStorage(ctx context.Context) (bool, error) {
```
lilic, line 110: Nit suggestion, fine to not apply:

```suggestion
func (v *verifyBackupStorage) isStorageAdequate(ctx context.Context) (bool, error) {
```

pkg/cmd/verify/backupstorage.go
```diff
@@ -0,0 +1,227 @@
...
+
+	dbSizeBytes, err := getBackupSizeBytes(ctx, cli, members.Members)
+	if err != nil {
+		klog.Warningf("failed checking backup size: %v", err)
+		return false, nil
+	}
+
+	fsAvailableBytes, err := getPathAvailableSpaceBytes(v.backupPath)
+	if err != nil {
+		return true, err
```
lilic, line 137: Should this not be false as we ran into error?

pkg/cmd/verify/backupstorage.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	requiredBytes := 2 * dbSizeBytes
+	if requiredBytes > fsAvailableBytes {
+		return true, fmt.Errorf("available storage is not adequate for path: %q, required bytes: %d, available bytes %d", v.backupPath, requiredBytes, fsAvailableBytes)
```
lilic, line 142: Should this not be false as it's not adequate?

lilic (commented):
Couple of questions, must have mixed it on initial review.

pkg/cmd/verify/backupstorage.go
hexfusion: yeah error should supercede but reads funny

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hexfusion: this was resolved.

lilic (approved):
/lgtm
/approve

🎉

PR #653, opened 2021-09-01
Bug 1997347: pkg/operator/upgradebackupcontroller: update cluster operator status

hexfusion:
This PR fixes an issue with the backup controller where CVO was looking for status reported to ClusterOperator config.openshift.io/v1 but instead, we were publishing to operator.openshift.io/v1.


wking:
/retitle Bug 1997347: pkg/operator/upgradebackupcontroller: update cluster operator status

hexfusion:
Self tagging after manual local testing.

lilic (approved):
/lgtm

PR #654, opened 2021-09-01
Bug 1990489: Reintroduce etcdHighNumberOfFailedGRPCRequests alert for non metal ipi clusters

lilic:
Initially, this alert was reverted as it kept firing on metal ipi clusters in CI. https://bugzilla.redhat.com/show_bug.cgi?id=1990489 For now until we get to the bottom of what is up with etcd on those clusters, and since the alert was not firing on other CI or other types we can reintroduce it back.

/hold
Still need to get a hold of metal ipi cluster to test this, or run the failing CI job.

### DIFF
--- a/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
+++ b/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
@@ -117,6 +117,31 @@ spec:
         severity: warning
   - name: openshift-etcd.rules
     rules:
+    - alert: etcdHighNumberOfFailedGRPCRequests
+      annotations:
+        description: 'etcd cluster "{{ $labels.job }}": {{ $value }}% of requests for {{ $labels.grpc_method }} failed on etcd instance {{ $labels.instance }}.'
+        summary: etcd cluster has high number of failed grpc requests.
+      expr: |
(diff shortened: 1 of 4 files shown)


lilic:
@smarterclayton can you have a look at the query maybe there is a better way to get the metal ipi type but this is what I came up with. Thanks!

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -117,6 +117,31 @@ spec:
         severity: warning
   - name: openshift-etcd.rules
     rules:
+    - alert: etcdHighNumberOfFailedGRPCRequests
+      annotations:
+        description: 'etcd cluster "{{ $labels.job }}": {{ $value }}% of requests for {{ $labels.grpc_method }} failed on etcd instance {{ $labels.instance }}.'
```
smarterclayton, line 122: Should this one also have a runbook?

lilic: No it's warning and only critical alerts, for now, have runbooks as per the monitoring team and SRE folks reqs.

lilic:
/test e2e-gcp-five-control-plane-replicas

with test merged this should be green now. We had one failing test due to this alert firing, so retesting...

ardaguclu:
@lilic metal-ipi job is misconfigured. In favor of this PR and future PRs, to assure not breaking blocking jobs, could you please take a look https://github.com/openshift/release/pull/21608?. Thank you.

lilic:
@hexfusion @hasbro17 this is ready to go now. I tested on baremetal installs and it should not fire. 🎉

PR #656, opened 2021-09-01
Bug 1997347: Fix compile on s390x after #647

yselkowitz:
/cc @hexfusion @lilic﻿


hexfusion:
thanks a lot @yselkowitz I added a few logging improvements since we had the chance. going to carry this work to #657

PR #657, opened 2021-09-01
Bug 1997347: pkg/cmd/verify: bug fixes and improvements

hexfusion:
This PR addresses a few bugs that were missed in the original PR. The fact that logging was muted and the initcontainer was exiting 0 on failure masked the problems which did end up with a successful backup but the disk checks were not properly being performed. This also adds a fix for s390x compile errors carried from #656 thanks again @yselkowitz

### DIFF
--- a/bindata/etcd/cluster-backup-pod.yaml
+++ b/bindata/etcd/cluster-backup-pod.yaml
@@ -20,6 +20,10 @@ spec:
       volumeMounts:
         - mountPath: /etc/kubernetes/cluster-backup
           name: etc-kubernetes-cluster-backup
+        - mountPath: /var/run/secrets/etcd-client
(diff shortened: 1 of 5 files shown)


yselkowitz:
Thanks.  IMHO it would still be appropriate to retitle the second commit to reflect that it is only about logging, but this will compile now.
/hold cancel

wking (approved):
/lgtm

wking (approved):
/lgtm

hexfusion:
This PR was locally tested and manually verified. Self-tagging for the sake of code freeze. I take responsibility for any negative results.

hexfusion:
infra flake 
/test e2e-agnostic

PR #672, opened 2021-09-24
pkg/operator: remove stale condition controller

hexfusion:
Stale condition controller syncs every second and is no longer needed as multiple versions have existed with it enabled so that any cluster that has properly upgraded can't possibly have this condition.

cc @hasbro17


hasbro17 (approved):
/lgtm

hexfusion:
infra
/retest

PR #675, opened 2021-09-27
pkg/operator: disable upgrade backup controller

hexfusion:
This controller was added specifically for 4.8 -> 4.9 upgrades. This PR diable the controller in 4.10 while we consider the fate of the code moving forward.

cc @lilic @hasbro17

### DIFF
--- a/pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
+++ b/pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
@@ -31,6 +31,8 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// NOTE: This controller DISABLED in 4.10+ and can be removed at any time.
+
 // UpgradeBackupController responds to an upgrade request to 4.9 by attempting
 // to ensure the cluster is backed up. The CVO is expected to wait on a ceo
 // condition indicating successful backup before responding to the upgrade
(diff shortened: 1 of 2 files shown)


hexfusion:
Ironically we need to worry about stale status which I just removed the controller for.

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -31,6 +31,8 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// NOTE: This controller DISABLED in 4.10+ and can be removed at any time.
```
lilic, line 34: We should probably not run the tests anymore for this package and skip them instead?

https://github.com/openshift/cluster-etcd-operator/blob/480a387c0f5b3f03478bce1d44a2b83d7416a320/pkg/operator/upgradebackupcontroller/upgradebackupcontroller_test.go

lilic (approved):
One comment, otherwise lgtm

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hexfusion: yeah makes sense

lilic (approved):
Just noticed this still running on a cluster so I remembered this PR, I am fine with test removal in a follow up as well just to be clear.

/lgtm
/hold

PR #676, opened 2021-09-27
Bug 2004451: pkg/operator/upgradebackupcontroller: fix backup dir name in status condition

hasbro17:
The backup directory name was incorrectly being regenerated when writing to the status condition.
It should now be read from the pod spec's env var which reflects the actual backup dir name.

TODO: Backport to 4.9 and 4.8 to fix https://bugzilla.redhat.com/show_bug.cgi?id=2004451

### DIFF
--- a/pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
+++ b/pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
@@ -44,6 +44,7 @@ const (
 	clusterBackupPodName     = "cluster-backup"
 	recentBackupPath         = "/etc/kubernetes/cluster-backup"
 	failedPodBackoffDuration = 30 * time.Second
+	backupDirEnvName         = "CLUSTER_BACKUP_PATH"
 )
 
 type UpgradeBackupController struct {
@@ -161,8 +162,6 @@ func (c *UpgradeBackupController) ensureRecentBackup(ctx context.Context, cluste
 		return nil, nil
 	}
 
-	recentBackupName := fmt.Sprintf("upgrade-backup-%s", time.Now().Format("2006-01-02_150405"))
-
 	backupPod, err := c.podLister.Pods(operatorclient.TargetNamespace).Get(clusterBackupPodName)
 	// No backup found, attempt to create one.
 	if err != nil && apierrors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -195,11 +196,21 @@ func (c *UpgradeBackupController) ensureRecentBackup(ctx context.Context, cluste
 	case backupPod.Status.Phase == corev1.PodSucceeded:
 		backupCondition := configv1helpers.FindStatusCondition(clusterOperatorStatus.Conditions, backupConditionType)
 		if backupCondition.Reason != backupSuccess {
```
lilic, line 198: Should this not be be true for when backup is successful as well:

```suggestion
		if backupCondition.Reason == backupSuccess {
```

otherwise, we will never update the backupDirName once backup is successful first time?

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -195,11 +196,21 @@ func (c *UpgradeBackupController) ensureRecentBackup(ctx context.Context, cluste
...
+
+			// Read the backup directory name from the backup pod's env
+			// TODO(hasbro17): backup pod has already succeeded but should we error if env unset or empty? unlikely it could succeed without the env.
```
lilic, line 201: Agreed it should never probably happen but not sure if we should error out. Status is best-effort basis, rather have something in the status at least the type and reason, incomplete msg is better than erroring out here. WDYT?

lilic (commented):
Nice!

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
hasbro17: IIUC we might not need to update the backupDirName ever again in the status if the backup pod has already succeeded, and we've already once set the backup condition reason to `UpgradeBackupSuccessful`.

We'd only need to rewrite the backupDirName for a new backup, that results in a new backupdir.

And if ([for whatever reason](https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go#L160)), a new backup is initiated , then the backupCondtion would be reset to `UpgradeBackupInProgress`. So here it would just update it again.
https://github.com/openshift/cluster-etcd-operator/blob/2ef42e68e7acdfd7626e481f196d0f1705556c34/pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go#L175-L184

lilic: Thank you! Makes sense to me!

lilic (commented):
/lgtm
/approve

/hold
Letting @hexfusion have a final look.

PR #680, opened 2021-09-30
pkg: Pass context through to functions

lilic:
As part of the previous PR comment, this passes the context and adds it to the function signatures. 

@hasbro17 @hexfusion please take a look, thanks!

### DIFF
--- a/pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
+++ b/pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
@@ -33,7 +33,7 @@ func NewEtcdMembersController(operatorClient v1helpers.OperatorClient,
 }
 
 func (c *EtcdMembersController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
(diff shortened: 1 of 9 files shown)


pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
```diff
@@ -59,10 +59,7 @@ func (c *EtcdMembersController) sync(ctx context.Context, syncCtx factory.SyncCo
 	return nil
 }
 
-func (c *EtcdMembersController) reportEtcdMembers(recorder events.Recorder) error {
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
```
hasbro17, line 64: Spent way too long to figure out if the parent context was already cancellable until I found it 😅 

https://github.com/openshift/cluster-etcd-operator/blob/17978bc6552e27738e3b75b4d2c5c18101fcdd05/vendor/github.com/openshift/library-go/pkg/controller/controllercmd/cmd.go#L84-L92

hasbro17 (approved):
/lgtm

PR #681, opened 2021-10-01
podsecurity: enforce privileged for openshift-etcd namespace

s-urbaniak:
Starting with OpenShift 4.10 we are introducing PodSecurity admission (https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement).

Currently, all pods are marked as privileged, however, over time we want to enforce at least baseline, admirably restricted as default. In order not to break control plane workloads this allows workloads in `openshift-etcd` namespace to run privileged pods.

See https://github.com/openshift/enhancements/pull/899 for more details (and excuse the eventual consistency of updates).

/cc @stlaz

### DIFF
--- a/bindata/etcd/ns.yaml
+++ b/bindata/etcd/ns.yaml
@@ -7,3 +7,6 @@ metadata:
   name: openshift-etcd
   labels:
     openshift.io/run-level: "0"
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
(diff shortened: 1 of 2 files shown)


lilic (commented):
/lgtm
/hold

/retest

lilic:
@hasbro17 @hexfusion This lgtm, but please take a look as well.

hasbro17:
SGTM
@lilic @hexfusion Should we require the same policy for the `openshift-etcd-operator` namespace/pods as well?
https://github.com/openshift/cluster-etcd-operator/blob/master/manifests/0000_12_etcd-operator_00_namespace.yaml

lilic:
No, I asked the same question and @s-urbaniak said the following:
> the operator namespace does (fortunately) not run any privileged pods, so no need to annotate.

PR #682, opened 2021-10-01
[POC] pkg/operator/clustermembercontroller: scale up using raft learners

hexfusion:
This PR is a first pass POC that provides scale up via raft learners. Using learners during scale-up eliminates quorum-loss as we scale from 1->2. This process also allows us to eliminate the current `discover-etcd-initial-cluster` code which is now our largest carry against upstream etcd. Scaling with learners is an important step to providing https://issues.redhat.com/browse/OCPPLAN-5712.

This PR requires https://github.com/openshift/etcd/pull/96 to test.

cluster bot
```
launch openshift/etcd#96,openshift/cluster-etcd-operator#682
```
enhancement: https://github.com/openshift/enhancements/pull/920


pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -83,93 +76,93 @@ func (c *ClusterMemberController) sync(ctx context.Context, syncCtx factory.Sync
...
+		// Promote learner to voting member if log is in sync with leader.
+		if member.IsLearner {
+			err := c.etcdClient.MemberPromote(ctx, member.ID)
+			// TODO: this is not working as expected
+			if err != nil && errors.Is(err, etcdserver.ErrLearnerNotReady) {
+				recorder.Warningf("ScaleUpCluster", "promotion failed etcd learner is not yet ready %s", member.PeerURLs[0])
+				continue
+			}
+			if err != nil {
+				//errs = append(errs, err)
```
lilic, line 96: I think this makes sense as we return the aggregated errors below.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -83,93 +76,93 @@ func (c *ClusterMemberController) sync(ctx context.Context, syncCtx factory.Sync
...
+	members, err := c.etcdClient.MemberList(ctx)
+	if err != nil {
+		return fmt.Errorf("could not get etcd member list: %v", err)
+	}
+
+	var errs []error
+	for _, member := range members {
+		// Promote learner to voting member if log is in sync with leader.
+		if member.IsLearner {
+			err := c.etcdClient.MemberPromote(ctx, member.ID)
```
lilic, line 89: Could we add a debug log level here something like:
` 		klog.V(4).Infof("attempting to promote learner with ID %d", member.ID)`

lilic (commented):
Looking good!

pkg/operator/clustermembercontroller/clustermembercontroller.go
hexfusion: agreed I just have an error handling issue above here that needs to get worked out.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -83,93 +87,132 @@ func (c *ClusterMemberController) sync(ctx context.Context, syncCtx factory.Sync
...
+	members, err := c.etcdClient.MemberList(ctx)
 	if err != nil {
 		return "", err
 	}
-	node, err := c.nodeLister.Get(podToAdd.Spec.NodeName)
+
+	currentMemberCount := len(nodes)
+	maxMemberCount := desiredMemberCount + 1 // Surge up
+	if currentMemberCount == maxMemberCount {
+		klog.Warningf("Skipping scale up etcd membership at maximum size: %d", desiredMemberCount+1)
```
hasbro17, line 186: @hexfusion During a manual test run I see that when I add a new node, `currentMemberCount=4` and `desiredMemberCount=3` since that's sourced from the `cluster-config-v1` configmap's `data.install-config.controlPlane.replicas` which is 3.

So the clustermember controller here will keep spinning on `currentMemberCount == maxMemberCount` even though we're not really at the max membership size yet.

Also at this point seems like there is no etcd static pod on the new node as well:
```yaml
kind: Etcd
metadata:
  name: cluster
...
status:
  conditions:
  - lastTransitionTime: "2021-11-09T20:26:46Z"
    message: 3 nodes are active; 1 nodes are at revision 0; 3 nodes are at revision
      6
    status: "True"
    type: StaticPodsAvailable
  - lastTransitionTime: "2021-11-10T05:22:11Z"
    message: 1 nodes are at revision 0; 3 nodes are at revision 6
    status: "True"
    type: NodeInstallerProgressing
    ...
    nodeStatuses:
  - currentRevision: 6
    nodeName: ip-10-0-136-64.us-west-1.compute.internal
  - currentRevision: 6
    nodeName: ip-10-0-227-172.us-west-1.compute.internal
  - currentRevision: 6
    nodeName: ip-10-0-186-101.us-west-1.compute.internal
  - currentRevision: 0
    nodeName: ip-10-0-143-124.us-west-1.compute.internal
```

It's only when I update the configmap `controlPlane.replicas=4` (and kick the etcd-operator to restart since it's not noticing the change for some reason), does it actually progress past this point, and the cluster settles to 4 available members with all the nodes at revision 7.

But I don't think we want admins to update that configmap since the desired controlplane size should be 3 at the end.


So (other than understanding why the static pod doesn't initially show up on the new node), I think `currentMemberCount` here should be the current number of voting members and not the number of nodes.
E.g we could add 3 new nodes at which point our  `currentMemberCount=6 > maxMemberCount=4`
```suggestion
	currentMemberCount := len(members)
	maxMemberCount := desiredMemberCount + 1 // Surge up
	if currentMemberCount == maxMemberCount {
		klog.Warningf("Skipping scale up etcd membership at maximum size: %d", desiredMemberCount+1)
```

hasbro17: Although that will only allow us to scale up and down serially, 1 at a time, `3->4->3->4->3->4->3` rather than all at once.

hexfusion: That is the goal post bootstrap while we can add new learners in parallel promotion in this circumstance would be serial.  Yes and in general we are moving away from nodes to members so the change makes sense.

PR #684, opened 2021-10-04
pkg/operator/metriccontroller: Fix query

lilic:
Still want to test some things.

/hold


hexfusion:
This looks good, one thing we should try to improve is our event message. Basically, we want admin to understand this is possibly networking but almost certainly storage :)

https://github.com/openshift/cluster-etcd-operator/pull/684/files#diff-1ffe8d5a41289fd1acde2f0fcd5d53c842caffe41d70a478e82b018977b6ac14L106

lilic:
Good idea, done @hexfusion please take a look again, thanks!

hexfusion (commented):
/lgtm

thank you :pray:

PR #691, opened 2021-10-12
Bug 2013243: jssonet: ignore Alarm method in etcdGRPCRequestsSlow alerts

hasbro17:
The cluster-etcd-operator readiness probe uses the `etcdctl endpoint health` cmd to check the health of the cluster
https://github.com/openshift/cluster-etcd-operator/blob/ea4c530cceefe91539a655e8406abbf9c0560da7/bindata/etcd/pod.yaml#L210-L211

As part of the endpoint health command, etcdctl checks the active alarms with grpc calls to the `Alarm` method:
https://github.com/etcd-io/etcd/blob/f2bc5eee916de63ac525ce5a843e849f73921415/etcdctl/ctlv3/command/ep_command.go#L139

To eliminate noise from the `etcdGRPCRequestsSlow` alerts we should discount the effect of `Alarm` grpc calls which could end up being slow due to the repeated calls, and indicating this to the user is not useful.
This is only done downstream through our `custom.libsonnet` rules file for now.

**Side note:**
Depending on when you installed `gojsontoyaml` it may or may not wrap long lines in its output.
The way forward since `v1.0.0` however seems to be non-wrapped lines so we should probably stick to that. Note you may want to run `go get go get github.com/brancz/gojsontoyaml` again if `make generate` results in wrapped lines for you.
See:
- https://github.com/brancz/gojsontoyaml/issues/7
- https://github.com/brancz/gojsontoyaml/commit/d65956103a55ccf82df2cd33cfcf8cdf96296762
- https://github.com/go-yaml/yaml/commit/7649d4548cb53a614db133b2a8ac1f31859dda8c

/cc @lilic @hexfusion

### DIFF
--- a/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
+++ b/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
@@ -13,8 +13,7 @@ spec:
     rules:
     - alert: etcdMembersDown
       annotations:
-        description: 'etcd cluster "{{ $labels.job }}": members are down ({{ $value
-          }}).'
+        description: 'etcd cluster "{{ $labels.job }}": members are down ({{ $value }}).'
         runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/etcdMembersDown.md
         summary: etcd cluster members are down.
       expr: |
@@ -31,8 +30,7 @@ spec:
         severity: critical
     - alert: etcdNoLeader
       annotations:
-        description: 'etcd cluster "{{ $labels.job }}": member {{ $labels.instance
-          }} has no leader.'
+        description: 'etcd cluster "{{ $labels.job }}": member {{ $labels.instance }} has no leader.'
         runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/etcdNoLeader.md
         summary: etcd cluster has no leader.
       expr: |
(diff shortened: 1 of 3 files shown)


manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -120,16 +98,25 @@ spec:
         severity: critical
     - alert: etcdExcessiveDatabaseGrowth
       annotations:
-        description: 'etcd cluster "{{ $labels.job }}": Observed surge in etcd writes
-          leading to 50% increase in database size over the past four hours on etcd
-          instance {{ $labels.instance }}, please check as it might be disruptive.'
+        description: 'etcd cluster "{{ $labels.job }}": Observed surge in etcd writes leading to 50% increase in database size over the past four hours on etcd instance {{ $labels.instance }}, please check as it might be disruptive.'
```
lilic, line 101: > the way forward since v1.0.0 however seems to be non-wrapped lines so we should probably stick to that. 

Yes, I think this formatting changed was due by @hexfusion maybe having an outdated gojsontoyaml install during recent changes? Maybe we can add another check for this here https://github.com/openshift/cluster-etcd-operator/blob/master/hack/generate.sh#L5?

lilic (commented):
/lgtm
/approve

/hold 

in case we want to add check for gojsontoyaml check in this PR. Feel free to remove hold if we want to do that in a separate one.

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
hexfusion: yeah I will keep that in mind but we should bake it into the script.

PR #701, opened 2021-11-11
pkg/operator/etcdendpointscontroller: use etcd membership to populate endpoints

hexfusion:
This PR is a step towards the goal of providing more flexibility in way we communicate etcd membership to the apiserver on a revision level. First of all, we are going to now version the etcd-endpoints configmap. Membership change precludes a new revision and we want to use the etcd-endpoints data to be able to understand this during the revision and as a historical reference.

The second part moves the population of endpoints to actual etcd membership `MemberList` vs blindly using all control-plane nodes. This will prove future flexibility as OpenShift begins to overcome future scaling challenges.

The third part of this change adjusts the key used in the configmap. Today we base64 encode the IP address, while this does generate a unique key per endpoint the data is not useful. We now use the etcd member id as the key. This is an important designation because an IP address (node) can over time have many members. An example of this would be disaster recovery.

example of versioned configmap taken from CI
```
        {
            "apiVersion": "v1",
            "data": {
                "5a4b27de7cb53cad": "10.0.190.46",
                "77564ad437dae32e": "10.0.253.233",
                "ac137595322cc148": "10.0.138.56"
            },
            "kind": "ConfigMap",
            "metadata": {
                "annotations": {
                    "alpha.installer.openshift.io/etcd-bootstrap": "10.0.33.149"
                },
                "creationTimestamp": "2021-11-12T15:30:19Z",
                "name": "etcd-endpoints-5",
                "namespace": "openshift-etcd",
                "ownerReferences": [
                    {
                        "apiVersion": "v1",
                        "kind": "ConfigMap",
                        "name": "revision-status-5",
                        "uid": "118b6a3c-c67c-4f90-ba73-73c38f9a1ed6"
                    }
                ],
                "resourceVersion": "11277",
                "uid": "23d091c2-954c-44eb-bdf2-2a0b776a24da"
            }
        },
 ```

 A noticeable change is that etcd goes from 3 revisions to 6 during install. But because these happen fairly quickly the net revisions per node is still max(3)

### DIFF
--- a/pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
+++ b/pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
@@ -2,11 +2,11 @@ package etcdendpointscontroller
 
 import (
 	"context"
-	"encoding/base64"
 	"fmt"
 	"time"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-etcd-operator/pkg/dnshelpers"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
@@ -15,7 +15,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by hexfusion** [comment]
Nice, so the upgrade failure is a good case that should be covered.

```
event happened 25 times, something is wrong: ns/openshift-etcd-operator deployment/etcd-operator - reason/RequiredInstallerResourcesMissing configmaps: etcd-endpoints-3 | event happened 25 times, something is wrong: ns/openshift-etcd-operator deployment/etcd-operator - reason/RequiredInstallerResourcesMissing configmaps: etcd-endpoints-3
```
Basically during upgrade the new operator starts and expects the existing revision to have all of the resources but because `etcd-endpoints-3` did not exist in the previous version we complain. Now this is expected and transient as the next revision (4) is properly created.

**Comment by hexfusion** [comment]
Thought about this a bit and I believe we should make this exempt for 4.10. The failure is expected and resolving it would require populating the cm with invalid or empty data which doesn't feel right.

File: pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go [hunk 1]
```diff
@@ -114,28 +113,27 @@ func (c *EtcdEndpointsController) syncConfigMap(ctx context.Context, recorder ev
...
 		}
-		if len(nodeInternalIP) == 0 {
-			return fmt.Errorf("unable to determine internal ip address for node %s", node.Name)
+		// Use of PeerURL is expected here because it is a mandatory field, and it will mirror ClientURL.
+		ip, err := dnshelpers.GetIPFromAddress(member.PeerURLs[0])
+		if err != nil {
+			return err
 		}
-		endpointAddresses[base64.StdEncoding.WithPadding(base64.NoPadding).EncodeToString([]byte(nodeInternalIP))] = nodeInternalIP
+		endpointAddresses[fmt.Sprintf("%016x", member.ID)] = ip
```
hasbro17, line 132: Nit: Just curious why this format is being chosen. This is an etcd member ID convention right, not some limitation of the configmap key length or format.

hasbro17 (approved):
/lgtm

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
hexfusion: member id is uint64 this conversion allows you to see it the same way the etcdctl printer does. But this is actually a bug and should be `16` vs `016`

hexfusion:
```
$ printf '%16x\n' 919655480020446617
cc345c95a5c2d99
```
```
+------------------+---------+--------------------------------------------+---------------------------+---------------------------+------------+
|        ID        | STATUS  |                    NAME                    |        PEER ADDRS         |       CLIENT ADDRS        | IS LEARNER |
+------------------+---------+--------------------------------------------+---------------------------+---------------------------+------------+
|  bf238e4c09e3045 | started |  ip-10-0-208-53.us-west-1.compute.internal |  https://10.0.208.53:2380 |  https://10.0.208.53:2379 |      false |
| 766f9b55c68c424f | started | ip-10-0-184-143.us-west-1.compute.internal | https://10.0.184.143:2380 | https://10.0.184.143:2379 |      false |
| d9af47ef64346aaa | started |  ip-10-0-146-57.us-west-1.compute.internal |  https://10.0.146.57:2380 |  https://10.0.146.57:2379 |      false |
+------------------+---------+--------------------------------------------+---------------------------+---------------------------+------------+
```

hexfusion: fixed

hexfusion: ref: https://github.com/etcd-io/etcd/blob/main/etcdctl/ctlv3/command/printer.go#L176

hasbro17 (approved):
/lgtm

PR #706, opened 2021-11-15
pkg/operator/quorumguardcontroller: add pod affinity

hexfusion:
This PR add pod affinity for the quorum-guard pod matching k8s-app etcd. This gives us more flexibility in the future to ensure quorum-guard is scheduled to nodes where etcd static-pod has been installed.

This also will reduce the amount of time during install where quorum-guard reports health check failures while etcd is rolling out during install.


hasbro17 (approved):
/lgtm

PR #710, opened 2021-11-29
pkg/cmd/monitor: handle error on failed monitor

hexfusion:
If the etcd client fails or `newMonitor` for any other reason has an error it is not handled at all which can result in panic. This PR resolves the error handling and adds a test case.


hexfusion:
/hold
adding a test case

hasbro17 (approved):
/lgtm

PR #722, opened 2021-12-17
pkg/cmd/monitor: consume endpoints from configmap

hexfusion:
Since etcd-endpoints are now revisioned we no longer need to rely on the static pod spec to provide this data.


pkg/operator/etcd_assets/bindata.go
```diff
@@ -5,6 +5,8 @@
 // bindata/etcd/cluster-restore.sh
 // bindata/etcd/cm.yaml
 // bindata/etcd/etcd-common-tools
+// bindata/etcd/localhost-recovery-token.yaml
```
hexfusion, line 8: need to drop

PR #724, opened 2021-12-20
operator: remove etcd-endpoints from revisioned configmaps

hexfusion:
While this had the best of intentions the inevitable result is creating 2 revisions for each change to the resource as we also must update the pod.


hasbro17 (approved):
/lgtm

PR #726, opened 2022-01-05
Adding cluster-etcd-operator sub command: ensure-env

dusk125:
This new sub command replaces the [bash implementation found here](https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/etcd/pod.yaml#L35).

Since this subcommand is configurable with command line switches, we can more easily control what validations are done, and how/if they fail. Also, the error reporting is improved with the subcommand; the bash implementation echos the entire bash script so it's very difficult to see where something went wrong (see example below).

For example, in the case where the (Downward API value for NODE_IP is empty)[https://bugzilla.redhat.com/show_bug.cgi?id=1930422#c6], we can change the invocation of ensure-env to continue and try to validate the node IP address against the network interfaces instead of failing.


pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,161 @@
...
+			return
+		}
+
+		for _, iface := range interfaces {
+			var (
+				addrs []net.Addr
+			)
+
+			if addrs, err = iface.Addrs(); err != nil {
+				return
```
hexfusion, line 144: nit: handle return var explicitly vs naked.

```
return err
```

pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,161 @@
...
+		//   interfaces, to see if we find a match.
+		// If we find a match, we'll consider ourselves valid, if we can't
+		//   we'll fail.
+		if interfaces, err = net.Interfaces(); err != nil {
+			return
+		}
+
+		for _, iface := range interfaces {
+			var (
+				addrs []net.Addr
```
hexfusion, line 140: nit: style wise a single var should be initialized like
```
var name type
```
just drop ().

pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,161 @@
...
+			)
+
+			if addrs, err = iface.Addrs(); err != nil {
+				return
+			}
+
+			for _, addr := range addrs {
+				if strings.Contains(addr.String(), fallbackEV) {
```
hexfusion, line 148: careful of contains here 

```
fmt.Println(strings.Contains("192.168.1.100", "192.168.1.1"))
```

pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,161 @@
...
+					// We found a match, we'll assume now that the value of fallbackEV is good
+					//		and we'll continue allowing etcd to start.
+					return nil
```
hexfusion, line 151: Let's print a log that explicitly describes the situation as NODE_IP was empty but ...

pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,161 @@
...
+				}
+			}
+		}
+
+		// If we drop out of the loop, we haven't found a match and we need to error-out
+		return fmt.Errorf("Failed to find any ip addresses in network interfaces that match the value of %v", e.fallbackIPEV)
```
hexfusion, line 157: style nits:

- error message does not start with capital letter.
- "failed to find ip address on network interfaces: %v"

pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return
```
hexfusion, line 160: nit be explicit on returns 
```
return nil
```

pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,161 @@
...
+		},
+	}
+
+	ensure.AddFlags(cmd.Flags())
+	return cmd
+}
+
+func (e *ensureOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&e.allowInvalidIPEV, "allow-invalid-ip-ev-val", true, "If this is set and the value of ip-ev is invalid (not equal to environment variables given by --ip-ev-equals), continue validation and check the value of --fallback-ip-ev against the network interfaces.")
+	fs.StringVar(&e.ipEV, "ip-ev", "NODE_IP", "The environment variable that should contain the node IP address.")
```
hexfusion, line 77: let's simplify and make the default empty string then validate is checking if empty and error.

hexfusion: nit: have the flag name explicitly describe the value maybe something like

```
fs.StringVar(&e.nodeIP, "node-ip".....  "The internal IP of the node provided by downward API"
```

pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+		},
+	}
+
+	ensure.AddFlags(cmd.Flags())
+	return cmd
+}
+
+func (e *ensureOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&e.allowInvalidIPEV, "allow-invalid-ip-ev-val", true, "If this is set and the value of ip-ev is invalid (not equal to environment variables given by --ip-ev-equals), continue validation and check the value of --fallback-ip-ev against the network interfaces.")
```
hexfusion, line 76: I like the thought but let's boil this down to simplify the UX. The flags should only include inputs necessary to perform the tasks. The logic should be straightforward and simple.

pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,161 @@
...
+	fs.StringVar(&e.ipEV, "ip-ev", "NODE_IP", "The environment variable that should contain the node IP address.")
+	fs.StringVar(&e.fallbackIPEV, "fallback-ip-ev", "NODE_NODE_ENVVAR_NAME_IP", "If the value of the environment variable given by --ip-ev is not set (or is empty) and --allow-invalid-ip-ev-val is true, use this environment variable to check against the network interfaces to find the nodes ip address.")
```
hexfusion, line 78: same as above let the default be empty to simplify validate.

```
current-revision-node-ip
```

hexfusion (commented):
Great start a few notes on style and simplifying. In general, inputs should be explicit and only from flags. Another flag could toggle toleration of empty node-ip and best-effort match on interface IP.

But for now as David described let's prove the failure and fail fast with solid logging around the problem. This can be further justification for pressure to resolve. Great work!

We will also want to update the etcd pod spec[1] to include this.

[1] https://github.com/openshift/cluster-etcd-operator/blob/release-4.10/bindata/etcd/pod.yaml#L35

Elbehery:
@dusk125  can u please run `make verify` on the root folder ? .. the ci is failing the verify step

tjungblu:
great addition!
/lgtm

pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+		},
+	}
+
+	ensure.AddFlags(cmd.Flags())
+	return cmd
+}
+
+func (e *ensureOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&e.allowInvalidNodeIP, "allow-invalid-node-ip", true, "If this is set and the value of --node-ip is invalid (not equal to environment variables given by --ip-ev-equals), continue validation and check the value of --current-revision-node-ip against the network interfaces.")
```
hasbro17, line 63: > not equal to environment variables given by --ip-ev-equals

What's `--ip-ev-equals`?

pkg/cmd/ensureenv/ensureenv.go
```diff
@@ -0,0 +1,137 @@
...
+				return err
+			}
+
+			for _, addr := range addrs {
+				switch ip := addr.(type) {
+				// Casting here to remove the mask before checking for equality.
+				case *net.IPNet:
+					if ip.IP.Equal(currentRevIP) {
+						// We found a match, we'll assume now that the value of fallbackEV is good
+						//		and we'll continue allowing etcd to start.
```
hasbro17, line 124: Nit:
```suggestion
						// and we'll continue allowing etcd to start.
```

hasbro17: But you can do this in a follow up if you want since the PR is mostly good to go.

hasbro17 (commented):
Looks good. Just a couple of nits.

pkg/cmd/ensureenv/ensureenv.go
dusk125: An artifact from a previous commit, I've removed this and fixed the comment.

PR #727, opened 2022-01-11
Bug 2039344: Do not include ipv6 node address in cert

mfojtik:
If dual-stack is enabled in cluster, this will cause the etcd not go
degraded, but instead the IPv6 address of Node will be skipped in the
check.

### DIFF
--- a/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
+++ b/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/utils/net"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -318,6 +320,10 @@ func checkCertValidity(certBytes, keyBytes []byte, ipAddresses []string, nodeUID
 	// Check that the cert is valid for the provided ip addresses
 	errs := []error{}
 	for _, ipAddress := range ipAddresses {
+		if net.IsIPv6String(ipAddress) {
+			klog.Warningf("IPv6 address %q was detected for node %q, it will not be added to the certificate", ipAddress, nodeUID)
```
sttts, line 324: this is bad in a verify func. It's not clear how often this is called, potentially very often in the sync loop. I also don't see why this should be a warning in the first place.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -318,6 +331,10 @@ func checkCertValidity(certBytes, keyBytes []byte, ipAddresses []string, nodeUID
 	// Check that the cert is valid for the provided ip addresses
 	errs := []error{}
 	for _, ipAddress := range ipAddresses {
+		// This will skip checking ipv6 addresses as in dual-stack mode, this will cause etcd operator go degraded.
+		if isDualStackMode(ipAddresses) && net.IsIPv6String(ipAddress) {
```
dusk125, line 335: Can we cache the result of isDualStackModel() so that we're not re-invoking it for each element of the ipAddresses list please?

dusk125: Also, I just found that k8s.io/utils/net has a function [IsDualStackIPStrings] that we could use here instead of the custom function.

[IsDualStackIPStrings]: https://pkg.go.dev/k8s.io/utils/net#IsDualStackIPStrings

dusk125 (changes requested):

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
mfojtik: @dusk125 the list is just two items :-)

good catch @ the isDualStackIPStrings, I haven't noticed it there.

mfojtik:
@dusk125 updated, PTAL

PR #736, opened 2022-01-31
wait for other installers to finish

deads2k:
bringing in https://github.com/openshift/library-go/pull/1300


deads2k:
now with real bump

PR #741, opened 2022-02-04
changes the signature of the MemberRemove method to accept a memberID (uint64) not a Name (string)

p0lyn0mial:
previously the method required contacting with the etcd cluster to get the current list of members just to find an ID of the member that matches the given name.

however, all current callers (i.e. the bootstrapteardown, member removal controllers) of this method already have an ID of the member they want to remove.

in addition to that, it is easier to deal with an ID instead of a Name because the latter might be sometimes empty (i.e. member hasn't been started) which leads to awkward error handling in the operator.

### DIFF
--- a/pkg/etcdcli/etcdcli.go
+++ b/pkg/etcdcli/etcdcli.go
@@ -255,33 +255,16 @@ func (g *etcdClientGetter) MemberUpdatePeerURL(ctx context.Context, id uint64, p
 	return err
 }
 
-func (g *etcdClientGetter) MemberRemove(ctx context.Context, member string) error {
-	g.eventRecorder.Eventf("MemberRemove", "removing member %q", member)
-
+func (g *etcdClientGetter) MemberRemove(ctx context.Context, memberID uint64) error {
 	cli, err := g.getEtcdClient()
 	if err != nil {
 		return err
 	}
-	membersResp, err := cli.MemberList(ctx)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


deads2k:
this lgtm

/approve

/assign @hasbro17

pkg/etcdcli/etcdcli.go
```diff
@@ -255,33 +255,16 @@ func (g *etcdClientGetter) MemberUpdatePeerURL(ctx context.Context, id uint64, p
...
-
-			_, err = cli.MemberRemove(cctx, m.ID)
-			if err != nil {
-				return err
-			}
-			return nil
-		}
+	_, err = cli.MemberRemove(ctx, memberID)
+	if err == nil {
+		g.eventRecorder.Eventf("MemberRemove", "removed member with ID: %v", memberID)
```
hasbro17, line 265: I think at a glance it's more useful to see the member name (along with the ID for unstarted members) in the event, so you can more easily tie it to the node that was removed. It's harder to discern that with just the member ID alone.
This is what I was trying to preserve by passing in the entire member struct.
https://github.com/hasbro17/cluster-etcd-operator/blob/62ce76b24d48c749c16f0503111bd4752149a69b/pkg/etcdcli/etcdcli.go#L344

But I'm okay with doing a follow up for this later.

hasbro17 (approved):
/lgtm

PR #742, opened 2022-02-08
reconciles machine deletion hooks

p0lyn0mial:


pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,236 @@
...
+		return nil
+	}
+	// step 1: add hooks for voting members
+	if err := c.addDeletionHookToMasterMachines(ctx); err != nil {
+		return err
+	}
+	// step 2: remove hooks from machines pending deletion and without running etcd member
+	//         before make sure the etcd cluster is healthy and voting members == controlPlane.replicas + 1
+	//
+	//         TODO: step 2 requires updating the ClusterMemberController to not re-add removed members
```
p0lyn0mial, line 94: @deads2k @hasbro17 how about this? does it also make sense?

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,260 @@
...
+		return err
+	}
+	// step 2: attempt to remove hooks from machines pending deletion and without running etcd member
+	//         before make sure the etcd cluster is healthy and the number of voting members >= controlPlane.replicas
+	//
+	//         TODO: step 2 requires updating the ClusterMemberController to not re-add removed members
+	return nil
+}
+
+func (c *machineDeletionHooksController) attemptToDeleteMachineDeletionHook(ctx context.Context) error {
```
p0lyn0mial, line 106: this is the current plan

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
deads2k: > @deads2k @hasbro17 how about this? does it also make sense?

This would be represented as a check in clustermembercontroller to avoid adding members that correspond to machines that are pending deletion?

deads2k:
Is this still a WIP?

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (c *machineDeletionHooksController) sync(ctx context.Context, _ factory.SyncContext) error {
+	// stop if the machine API is not functional
+	if isFunctional, err := c.machineAPIChecker.IsFunctional(); err != nil {
+		return err
+	} else if !isFunctional {
+		return nil
+	}
+	// step 1: add hooks for voting members
+	if err := c.addDeletionHookToVotingMembers(ctx); err != nil {
```
deads2k, line 86: What is the disadvantage of simply adding the hook to every master machine?  that would effectively eliminate one class of races since the deletion hook would be present long before etcd was present on the machine.

p0lyn0mial: It won't solve the issue. In the case that a Machine is deleted before the member is promoted, the etcd operator is expected to not promote the new member, and remove the deletion hook to allow the Machine to be removed from the cluster.

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,352 @@
...
+		return nil
+	}
+
+	// find a machine that doesn't have a member
+	for memberMachinePendingDeletionIP, memberMachinePendingDeletion := range indexMachinesByNodeInternalIP(memberMachinesPendingDeletion) {
+		if !currentVotingMemberIPListSet.Has(memberMachinePendingDeletionIP) && !currentLearnerIPListSet.Has(memberMachinePendingDeletionIP) {
+			memberMachinePendingDeletionCopy := memberMachinePendingDeletion.DeepCopy()
+			removeMachineDeletionHook(memberMachinePendingDeletionCopy)
+			if _, err := c.machineClient.Update(ctx, memberMachinePendingDeletionCopy, metav1.UpdateOptions{}); err == nil {
+				klog.Infof("successfully removed the deletion hook from machine %v", memberMachinePendingDeletion.Name)
```
p0lyn0mial, line 163: do we want to record an event?

p0lyn0mial:
> Is this still a WIP?

it is ready for review, I am working on adding more unit tests.

p0lyn0mial:
/hold

until we update the member removal controller.

p0lyn0mial:
@hasbro17 @deads2k  a friendly ping to review this pr

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,289 @@
...
+		machineAPIChecker:                     machineAPIChecker,
+		masterMachineSelector:                 masterMachineSelector,
+		masterMachineLister:                   machinelistersv1beta1.NewMachineLister(masterMachineInformer.GetIndexer()),
+		configMapListerForKubeSystemNamespace: kubeInformersForNamespaces.InformersFor("kube-system").Core().V1().ConfigMaps().Lister().ConfigMaps("kube-system"),
+	}
+	return factory.New().
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
+		WithInformers(
+			kubeInformersForNamespaces.InformersFor("kube-system").Core().V1().ConfigMaps().Informer(),
```
hasbro17, line 72: Are we only re-syncing on `kube-system` configmaps?
I would have thought this controller would need to react to changes on Machines for hook changes, and potentially nodes or etcd pods to detect member addition/removal.

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,289 @@
...
+	// find a machine that doesn't have a member
+	for memberMachinePendingDeletionIP, memberMachinePendingDeletion := range indexMachinesByNodeInternalIP(memberMachinesPendingDeletion) {
+		if !currentMemberIPListSet.Has(memberMachinePendingDeletionIP) {
+			memberMachinePendingDeletionCopy := memberMachinePendingDeletion.DeepCopy()
+			removeMachineDeletionHook(memberMachinePendingDeletionCopy)
+			if _, err := c.machineClient.Update(ctx, memberMachinePendingDeletionCopy, metav1.UpdateOptions{}); err == nil {
+				klog.Infof("successfully removed the deletion hook from machine %v", memberMachinePendingDeletion.Name)
+			}
+			return err
+		}
```
hasbro17, line 142: Style nit: You don't have to change it :) but I find these blocks more readable when we can return/continue early to reduce nesting if possible.
```suggestion
		if currentMemberIPListSet.Has(memberMachinePendingDeletionIP) {
			// Ignore machines that still have a member present
			continue
		}
		memberMachinePendingDeletionCopy := memberMachinePendingDeletion.DeepCopy()
		removeMachineDeletionHook(memberMachinePendingDeletionCopy)
		if _, err := c.machineClient.Update(ctx, memberMachinePendingDeletionCopy, metav1.UpdateOptions{}); err == nil {
			klog.Infof("successfully removed the deletion hook from machine %v", memberMachinePendingDeletion.Name)
		}
		return err
	}
```

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
hasbro17: I think yes it's probably worth noting when we unblock a machine.

hasbro17 (commented):
Still going through the unit tests but the implementation looks good to me. Just a question on the informers used to sync this controller.

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
p0lyn0mial: np

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
p0lyn0mial: good point, added `masterMachineInformer` and changed the resync time to 1 minute

p0lyn0mial:
applied the review comments, ptal

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,293 @@
...
+		return err
+	}
+	for _, masterMachine := range masterMachines {
+		if masterMachine.DeletionTimestamp != nil {
+			continue
+		}
+		if hasCompleteMachineDeletionHook(masterMachine) {
+			continue
+		}
+		votingMemberMachineCopy := masterMachine.DeepCopy()
```
deads2k, line 164: this is just `machineMemberCopy`, now right?

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+func (c *machineDeletionHooksController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	// stop if the machine API is not functional
+	if isFunctional, err := c.machineAPIChecker.IsFunctional(); err != nil {
+		return err
+	} else if !isFunctional {
+		return nil
+	}
+	if err := c.addDeletionHookToMasterMachines(ctx); err != nil {
```
deads2k, line 86: an error here will stop you from removing your machine deletion hook.  you sure you want to combine them?

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,293 @@
...
+		return nil
+	}
+
+	// get the members, issue a live call instead of reading from the config map
+	// because we need to take into account learners
+	members, err := c.etcdClient.MemberList(ctx)
+	if err != nil {
+		return err
+	}
+	currentMemberIPListSet := sets.NewString()
```
deads2k, line 125: `currentVotingMemberIPListSet`?

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,293 @@
...
+	for memberMachinePendingDeletionIP, memberMachinePendingDeletion := range indexMachinesByNodeInternalIP(memberMachinesPendingDeletion) {
+		if currentMemberIPListSet.Has(memberMachinePendingDeletionIP) {
+			// Ignore machines that still have a member present
+			continue
+		}
+		memberMachinePendingDeletionCopy := memberMachinePendingDeletion.DeepCopy()
+		removeMachineDeletionHook(memberMachinePendingDeletionCopy)
+		if _, err := c.machineClient.Update(ctx, memberMachinePendingDeletionCopy, metav1.UpdateOptions{}); err == nil {
+			klog.V(2).Infof("successfully removed the deletion hook from machine %v", memberMachinePendingDeletion.Name)
+			recorder.Eventf("MachineDeletionHook", "successfully removed the deletion hook from machine %v as it doesn't have a member", memberMachinePendingDeletion.Name)
```
deads2k, line 144: doesn't have a voting member.

pkg/operator/ceohelpers/common.go
```diff
@@ -31,3 +37,27 @@ func IsSingleNodeTopology(infraLister configv1listers.InfrastructureLister) (boo
...
+	host, _, err := net.SplitHostPort(memberURL.Host)
+	if err != nil {
+		return "", err
+	}
+	return host, nil
+}
+
+func memberToURL(member *etcdserverpb.Member) (string, error) {
+	if len(member.PeerURLs) == 0 {
+		return "", fmt.Errorf("unable to extract member's URL address, it has an empty PeerURLs field, member: %v", spew.Sdump(member))
```
deads2k, line 60: spewdump member in an error looks too long to me.  How about an ID or an IP or something short.

deads2k:
this looks plausible to me.

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
p0lyn0mial: yes, will update

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
p0lyn0mial: nope, the list holds both voting members and learners.
we also need to remove hooks from machines that host "non-promotable" learners.

pkg/operator/ceohelpers/common.go
p0lyn0mial: done

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
p0lyn0mial: updated

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	masterMachines, err := c.masterMachineLister.List(c.masterMachineSelector)
+	if err != nil {
+		return err
+	}
+	// machines with the deletion hooks are the ones that host etcd members
+	memberMachines := ceohelpers.FilterMachinesWithMachineDeletionHook(masterMachines)
+
+	// if the number of master machines is > controlPlane.replicas based on data from the cache
+	if len(memberMachines) <= desiredControlPlaneReplicasCount {
```
hasbro17, line 102: Why does the hooks controller care about whether we're over/under the desired control-plane replica count?
I mentioned this in https://github.com/openshift/cluster-etcd-operator/pull/768#discussion_r836926279 but it's the member removal controller that needs to make this consideration when deciding whether it's safe to remove/scale-down a voting member. Plus the member removal controller will also take care of cleaning up learner members pending deletion.

So for the hooks controller, all that would be required is to remove the deletion hook from any machine that isn't mapped to an etcd member (learner or voting).

hasbro17 (changes requested):

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
hasbro17: Apologies if I missed this before on my first pass but this just stood out to me when reviewing #768

hasbro17 (approved):
/approve

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	// machines with the deletion hooks are the ones that host etcd members
+	memberMachines := ceohelpers.FilterMachinesWithMachineDeletionHook(masterMachines)
+
+	// if at least one member machine is pending deletion
+	memberMachinesPendingDeletion := ceohelpers.FilterMachinesPendingDeletion(memberMachines)
+	if len(memberMachinesPendingDeletion) == 0 {
+		klog.V(4).Infof("currently, we don't have a machine pending deletion")
+		return nil
+	}
```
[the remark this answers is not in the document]
p0lyn0mial: kind of, all master machines are machines that host etcd members but yeah, np, I can rename it.

p0lyn0mial:
we are going to wire the controller in https://github.com/openshift/cluster-etcd-operator/pull/768

deads2k:
where does this use ReadDesiredControlPlaneReplicasCount  from https://github.com/openshift/cluster-etcd-operator/pull/767/files ?  I wouldn't expect that value to be pertinent.

p0lyn0mial:
> where does this use ReadDesiredControlPlaneReplicasCount from https://github.com/openshift/cluster-etcd-operator/pull/767/files ?

it doesn't anymore. I removed the check that required it.

p0lyn0mial:
oh, it looks like https://github.com/openshift/cluster-etcd-operator/pull/768 got first.

768 contained commits introduced in this PR.

PR #687, opened 2021-10-08
pkg: Go degraded if disk metrics are above a threshold  

lilic:
/hold

Still testing thresholds.

PR #743, opened 2022-02-08
Bug 2052270: pkg/operator/metriccontroller/fsync_controller: Fix "treshold" -> "thresholds" typos

wking:
These slipped in when the threshold guard landed in 0d4883ed77 (#687).


hasbro17 (approved):
/approve
/lgtm

PR #745, opened 2022-02-10
Bug 2038093: fix etcd endpoints configmap lost

Elbehery:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2038093

### DIFF
--- a/pkg/etcdcli/etcdcli.go
+++ b/pkg/etcdcli/etcdcli.go
@@ -13,17 +13,21 @@ import (
 	"time"
 
 	grpcprom "github.com/grpc-ecosystem/go-grpc-prometheus"
+
 	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/client/pkg/v3/logutil"
 	"go.etcd.io/etcd/client/pkg/v3/transport"
 	clientv3 "go.etcd.io/etcd/client/v3"
+
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/grpclog"
+
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -105,15 +109,16 @@ func (g *etcdClientGetter) getEtcdClient() (*clientv3.Client, error) {
 	}
 
 	configmap, err := g.configmapsLister.ConfigMaps(operatorclient.TargetNamespace).Get("etcd-endpoints")
-	if err != nil {
-		return nil, fmt.Errorf("failed to list endpoints: %w", err)
-	}
-	if bootstrapIP, ok := configmap.Annotations[BootstrapIPAnnotationKey]; ok && bootstrapIP != "" {
-		// escape if IPv6
-		if net.ParseIP(bootstrapIP).To4() == nil {
-			bootstrapIP = "[" + bootstrapIP + "]"
+	if err == nil {
+		if bootstrapIP, ok := configmap.Annotations[BootstrapIPAnnotationKey]; ok && bootstrapIP != "" {
+			// escape if IPv6
+			if net.ParseIP(bootstrapIP).To4() == nil {
+				bootstrapIP = "[" + bootstrapIP + "]"
+			}
+			etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", bootstrapIP))
 		}
-		etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", bootstrapIP))
+	} else if err != nil && !apierrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to list endpoints: %w", err)
 	}
 
 	g.clientLock.Lock()
--- a/pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
+++ b/pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
@@ -6,12 +6,12 @@ import (
 	"time"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
-	"github.com/openshift/cluster-etcd-operator/pkg/dnshelpers"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+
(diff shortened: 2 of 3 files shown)


pkg/operator/etcdendpointscontroller/etcdendpointscontroller_test.go
```diff
@@ -50,6 +52,9 @@ func TestBootstrapAnnotationRemoval(t *testing.T) {
 			name: "NewConfigMapAfterDeletion",
 			objects: []runtime.Object{
 				u.BootstrapConfigMap(u.WithBootstrapStatus("complete")),
+				u.FakeNode(fmt.Sprintf("%016x", etcdMembers[0].ID), u.WithMasterLabel(), u.WithNodeInternalIP(dnshelpers.GetIPFromAddressWithPanic(etcdMembers[0].PeerURLs[0]))),
```
Elbehery, line 55: @hasbro17  @dusk125  I am using the same way the `memberID` is encoded on the configMap within the test. Yet it never get the same value and the tests fails !!

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
```diff
@@ -111,6 +112,39 @@ func (c *EtcdEndpointsController) syncConfigMap(ctx context.Context, recorder ev
...
+			for _, nodeAddress := range node.Status.Addresses {
+				if nodeAddress.Type == corev1.NodeInternalIP {
+					nodeInternalIP = nodeAddress.Address
+					break
+				}
+			}
+			if len(nodeInternalIP) == 0 {
+				return fmt.Errorf("unable to determine internal ip address for node %s", node.Name)
+			}
+			endpointAddresses[node.Name] = nodeInternalIP
```
Elbehery, line 133: @hasbro17  @dusk125  I avoided the `base64` encoding to pass the test, do not know if this is correct way !!

dusk125:
I created a cluster with this PR and observed the values of the 'etcd-endpoints' configmap before and after I deleted it. After the dust settled from the deletion, I verified that the configmap has the same values as it did before deletion.

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
```diff
@@ -111,6 +112,40 @@ func (c *EtcdEndpointsController) syncConfigMap(ctx context.Context, recorder ev
...
+				if nodeAddress.Type == corev1.NodeInternalIP {
+					nodeInternalIP = nodeAddress.Address
+					break
+				}
+			}
+			if len(nodeInternalIP) == 0 {
+				return fmt.Errorf("unable to determine internal ip address for node %s", node.Name)
+			}
+			// the key is not important, the value is used to list etcd members with next sync and the key will be overwritten by the memberId on the next sync
+			endpointAddresses[node.Name] = nodeInternalIP
```
hasbro17, line 134: I know what you mean here but the more I think about it, I'm not really in favor of creating the configmap with incorrect keys and then just relying on a resync to fix that up. It's just confusing to see this in tests and the expected revisions of the configmap.
I understand that the revert quick fix works but I'm hoping we can clean this up while we're addressing this bug.

I'm thinking we could potentially introduce method in the etcdcli wrapper that lets us create the client even when the configmap is not present.
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/etcdcli/etcdcli.go#L107-L117
That way we won't need two separate ways (node list vs member list) of populating the configmap data.

hasbro17 (changes requested):
Good start, but I think we can refine this more so it's easier to reason about how the configmap data is populated on re-creation.

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
```diff
@@ -3,6 +3,7 @@ package etcdendpointscontroller
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/labels"
```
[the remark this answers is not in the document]
Elbehery: @hasbro17  fixed, but `goimports` edit the vendor folder, so i skipped it and manually sorted the imports 👍🏽

Elbehery:
@hasbro17  I have adjusted according to the review and our conversation, please correct me if i am wrong

pkg/etcdcli/etcdcli.go
```diff
@@ -268,7 +321,7 @@ func (g *etcdClientGetter) MemberRemove(ctx context.Context, memberID uint64) er
 }
 
 func (g *etcdClientGetter) MemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
-	cli, err := g.getEtcdClient()
+	cli, err := g.getEtcdClientWithoutConfigMap()
```
hasbro17, line 324: This bypasses the `etcd-endpoints` configmap for all MemberList calls which is not quite what I was intending. Only for the very specific case of when the endpoints controller has to recreate the configmap.

Since the configmap is used to add the `etcd-bootstrap` member endpoint to the client it will still be necessary early on during bootstrap.

To revise my earlier comments, I think we can make the configmap retrieval in `getEtcdClient()` a best effort, so that it is only bypassed when the configmap is actually deleted or missing.

```Go
import (
	apierrrors "k8s.io/apimachinery/pkg/api/errors"
)

...
configmap, err := g.configmapsLister.ConfigMaps(operatorclient.TargetNamespace).Get("etcd-endpoints")
if err == nil {
	if bootstrapIP, ok := configmap.Annotations[BootstrapIPAnnotationKey]; ok && bootstrapIP != "" {
		// escape if IPv6
		if net.ParseIP(bootstrapIP).To4() == nil {
			bootstrapIP = "[" + bootstrapIP + "]"
		}
		etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", bootstrapIP))
	}
} else if err != nil && !apierrrors.IsNotFound(err){
	// Ignore errors if the configmap does not exist or was deleted
	return nil, fmt.Errorf("failed to list endpoints: %w", err)
}
```

That way, MemberList() and subsequently `getEtcdClient()` won't fail if the configmap is missing, but still use it otherwise. And we won't have to add a duplicate method just for that case.

hasbro17 (changes requested):

pkg/etcdcli/etcdcli.go
Elbehery: 👍🏽 🙏🏽

hasbro17:
@Elbehery Looks good but we also need to fix this for printing the warning when the error is actually about the missing configmap.

https://github.com/openshift/cluster-etcd-operator/blob/f3d8db9c071c979b064063e97206f24398e9a854/pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go#L112-L114

```go
if existing, err := c.configmapLister.ConfigMaps(operatorclient.TargetNamespace).Get("etcd-endpoints"); err == nil && existing != nil {
 	 ...
} else if !errors.IsNotFound(err) { 
 	 return fmt.Errorf("could not get etcd-endpoints configmap: %w", err)
 }
 
 klog.Warningf("required configmap %s/%s will be created because it was missing: %w", operatorclient.TargetNamespace, "etcd-endpoints", err)
 
 members, err := c.etcdClient.MemberList(ctx)
 ...
```

Elbehery:
retest

geliu2016 (approved):
/lgtm

hasbro17:
/hold

@Elbehery This will impact some changes from https://github.com/openshift/cluster-etcd-operator/pull/758 so after that merges I need to vet this again to make sure we don't ever add learner member endpoints to the configmap.

PR #754, opened 2022-02-25
Bug 2057644: Fix FSyncController degraded latch

dusk125:
When there's a spike in leader elections and disk metrics, etcd can go degraded and never leave it even if the spike recedes and the cluster is ok.

https://bugzilla.redhat.com/show_bug.cgi?id=2057644


hasbro17 (approved):
/lgtm
/approve

PR #758, opened 2022-03-04
WIP: Scale up new members as learners

hasbro17:


pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -93,83 +118,220 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+	// Add a learner member if next peer found
+	peerURL, err := c.getEtcdPeerURLToAdd(ctx)
 	if err != nil {
-		return fmt.Errorf("could not get etcd peer host :%w", err)
+		return fmt.Errorf("could not get etcd peerURL to add :%w", err)
 	}
-	err = c.etcdClient.MemberAdd(ctx, fmt.Sprintf("https://%s:2380", etcdHost))
-	if err != nil {
-		return fmt.Errorf("could not add member :%w", err)
+	if peerURL != "" {
```
deads2k, line 136: project standard is len(peerURL)>0

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -2,54 +2,79 @@ package clustermembercontroller
...
+	etcdClient            etcdcli.EtcdClient
+	kubeClient            kubernetes.Interface
+	podLister             corev1listers.PodLister
+	masterNodeLister      corev1listers.NodeLister
+	masterMachineLister   machinelistersv1beta1.MachineLister
+	configMapLister       corev1listers.ConfigMapLister
+	networkLister         configv1listers.NetworkLister
+	masterNodeSelector    labels.Selector
+	masterMachineSelector labels.Selector
+	mastersReplicaCount   int
```
deads2k, line 52: what is this value?

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -93,83 +118,220 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
-// yet it will return collected errors.
-func (c *ClusterMemberController) getEtcdPeerHostToScale(podToAdd *corev1.Pod) (string, error) {
-	network, err := c.networkLister.Get("cluster")
+func (c *ClusterMemberController) getPeerURLForNode(node *corev1.Node) (string, error) {
+	internalIP, err := c.getInternalIPForNode(node)
 	if err != nil {
 		return "", err
 	}
-	node, err := c.nodeLister.Get(podToAdd.Spec.NodeName)
+	return fmt.Sprintf("https://%s:2380", internalIP), nil
```
deads2k, line 304: use net join.  you don't want to break on ipv6

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -93,83 +118,220 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+func isURLMappedToMember(peerURL string, members []*etcdserverpb.Member) bool {
+	for _, m := range members {
+		if peerURL == m.PeerURLs[0] {
+			return true
+		}
+	}
+	return false
+}
+
+func countVotingMembers(members []*etcdserverpb.Member) int {
```
deads2k, line 329: do we want to count unhealthy voting members in here?  does it vary by call site?

pkg/operator/quorumguardcontroller/quorumguardcontroller.go
```diff
@@ -235,31 +239,3 @@ func (c *QuorumGuardController) ensureEtcdGuardPDB(ctx context.Context, recorder
 
 	return nil
 }
-
-// getMastersReplicaCount get number of expected masters statically defined by the controlPlane replicas in the install-config.
-func (c *QuorumGuardController) getMastersReplicaCount() (int32, error) {
```
deads2k, line 240: This doesn't seem like a value we should use when deciding whether and how to add members.  It seems simpler to add every master node available.

hasbro17: Knowing the "expected" control plane size was a key requirement so we don't promote to more than `N+1` voting members at any given time (where `N` is the desired control-plane size).

Background from prior proposals:
- https://github.com/openshift/enhancements/pull/943/files/8ed85c3b3108633dced47b3a227c835ead3e921b#diff-8f08f89622ea85c6276166feaa7be109adaae9cf5d280b18ef20f4f218fce87cR137-R139
- https://github.com/openshift/enhancements/pull/920/files/22ad8ce2bcba65a0de70e1ed08d401227b030156#r738694187

When adding members we need to check against the max limit of learner members allowed `ETCD_EXPERIMENTAL_MAX_LEARNERS`:
https://github.com/openshift/cluster-etcd-operator/blob/f3d8db9c071c979b064063e97206f24398e9a854/pkg/etcdenvvar/etcd_env.go#L272-L299

Currently this is set to `N`. So if we simply add every new master node that appears into the etcd cluster as a learner member, then for the `N+1th` node the member add would be rejected.

I do have misgivings on using the InstallConfig's replica count as the desired masters replica count (since it doesn't seem like it was meant to be a declarative config) but the etcd-operator has been consuming it that way for a while now.

pkg/operator/ceohelpers/util.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+type replicaCountDecoder struct {
+	ControlPlane struct {
+		Replicas string `yaml:"replicas,omitempty"`
+	} `yaml:"controlPlane,omitempty"`
+}
+
+// GetMastersReplicaCount get number of expected masters statically defined by the controlPlane replicas in the install-config.
+func GetMastersReplicaCount(configmapLister corev1listers.ConfigMapLister) (int, error) {
```
p0lyn0mial, line 26: I also have this function in my PR. Could you open a new PR with just this function so that we can merge it before this PR? That would allow me to reuse it and would keep my PR smaller.

pkg/operator/ceohelpers/util.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+const (
+	clusterConfigName = "cluster-config-v1"
+	clusterConfigKey  = "install-config"
+)
+
+type replicaCountDecoder struct {
+	ControlPlane struct {
+		Replicas string `yaml:"replicas,omitempty"`
```
p0lyn0mial, line 21: could this be defined as a number?

pkg/etcdenvvar/etcd_env.go
```diff
@@ -272,25 +271,9 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
 // getMaxLearners reads the control-plane replicas from the install-config. We tolerate max learners equal to the
 // desired control-plane replicas so that the admin can surge up N x 2 in case of vertical scaling/replacement.
 func getMaxLearners(envVarContext envVarContext) (map[string]string, error) {
-	clusterConfig, err := envVarContext.configmapLister.ConfigMaps(operatorclient.TargetNamespace).Get(clusterConfigName)
+	replicaCount, err := ceohelpers.GetMastersReplicaCount(envVarContext.configmapLister)
```
p0lyn0mial, line 274: remove `clusterConfigName, clusterConfigKey and replicaCountDecoder` from this file.

pkg/etcdenvvar/etcd_env.go
```diff
@@ -272,25 +271,9 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
...
-
-	replicaCount, err := strconv.Atoi(d.ControlPlane.Replicas)
-	if err != nil {
-		errMsg := fmt.Errorf("failed to convert replica %s: %w", d.ControlPlane.Replicas, err)
-		klog.Error(errMsg)
-		return nil, errMsg
+		return nil, fmt.Errorf("failed to get control-plane replica count: %w", err)
 	}
 
 	return map[string]string{
```
p0lyn0mial, line 279: shouldn't we set `ETCD_EXPERIMENTAL_MAX_LEARNERS` to `2 x replicaCount`?

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -93,83 +118,220 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+	}
+
+	isFunctional, err := c.machineAPIChecker.IsFunctional()
+	if err != nil {
+		return "", fmt.Errorf("failed to determine Machine API availability: %v", err)
+	}
+
+	for _, node := range nodes {
+		if node.DeletionTimestamp != nil {
+			klog.V(4).Infof("Ignoring node (%s) for new member addition as it's pending deletion", node.Name)
```
p0lyn0mial, line 178: I think our default is `V(2)`. This info seems important, would it make sense to decrease the log level.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -93,83 +118,220 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
 
-	// go through the list of all pods, pick one peerFQDN to return from unready pods
-	// and collect dns resolution errors on the way.
-	for _, pod := range pods {
-		if !strings.HasPrefix(pod.Name, "etcd-") {
+	// For a desired size of N voting members in a cluster (specified in the install-config)
+	// we can only add a new learner member if we're under 2N total members (voting+learners).
+	// This abides by the current max limit of N learner members.
+	if len(members) >= 2*c.mastersReplicaCount {
+		klog.V(4).Infof("Skipping learner member addition since the current membership size (%v) exceeds the max learners limit of %v ", len(members), 2*c.mastersReplicaCount)
```
p0lyn0mial, line 162: would it make sense to decrease the log level?

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
```diff
@@ -121,6 +121,15 @@ func (c *EtcdEndpointsController) syncConfigMap(ctx context.Context, recorder ev
...
+		// Since a learner member rejects all requests other than serializable reads and member status API
+		// we have to exclude them from the etcd-endpoints configmap to prevent the API server from sending
+		// requests to it which would fail with "rpc not supported for learner" errors
+		// TODO(haseeb): Exclude learners from MemberList() everywhere else where it might recevie an unsupported request
+		// e.g defrag controller.
+		if member.IsLearner {
+			continue
+		}
+
 		if member.Name == "etcd-bootstrap" {
```
p0lyn0mial, line 133: we should also check if a machine for a member is not pending deletion. In that case, the member should not be added. This will prevent reading previously deleted members.

pkg/operator/starter.go
```diff
@@ -243,7 +243,12 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	)
 
 	clusterMemberController := clustermembercontroller.NewClusterMemberController(
+		masterNodeLabelSelector,
+		masterMachineLabelSelector,
+		ceohelpers.NewMachineAPI(masterMachineInformer, machinelistersv1beta1.NewMachineLister(masterMachineInformer.GetIndexer()), masterMachineLabelSelector),
```
p0lyn0mial, line 248: multiple controllers need `Machine API`, I would factor out and pass a singleton.

pkg/operator/ceohelpers/util.go
p0lyn0mial: I have created one https://github.com/openshift/cluster-etcd-operator/pull/767, ptal

deads2k:
where does this use ReadDesiredControlPlaneReplicasCount  from https://github.com/openshift/cluster-etcd-operator/pull/767/files ?  I wouldn't expect that value to be pertinent.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -93,83 +118,220 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+// getEtcdPeerURLToAdd returns a PeerURL of the next etcd instance to be added as a learner to the cluster. If
+// no elegible member is found return an empty string.
+func (c *ClusterMemberController) getEtcdPeerURLToAdd(ctx context.Context) (string, error) {
+	members, err := c.etcdClient.MemberList(ctx)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 
-	// go through the list of all pods, pick one peerFQDN to return from unready pods
```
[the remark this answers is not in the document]
hasbro17: It is now, since its removal was the cause of the bootstrap failure that I was seeing. Added back in with the latest update.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -93,83 +118,220 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+
+		if machine == nil {
+			klog.V(4).Infof("Ignoring node (%s) for scale-up: no Machine found referencing this node's internal IP (%v)", node.Name, internalIP)
+			continue
+		}
+		if machine.DeletionTimestamp != nil {
+			// Ignore this node since its member previously was, or is in the process of being removed
+			continue
+		}
+
```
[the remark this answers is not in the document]
hasbro17: Added with the latest update.

pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go
hasbro17: This check already exists in the clustermember controller. So any machine with a deletion timestamp would not get re-added to the membership, and as a consequence will never end up in the etcd-endpoints configmap.

hasbro17:
/hold

Still sorting out the unit tests since I'm expecting some rebase churn from after https://github.com/openshift/cluster-etcd-operator/pull/790 is merged.

Other than that this should be good for another review. Notable changes since the last update:
- Learner promotion method implemented
- No longer require the use of `ReadDesiredControlPlaneReplicasCount` or maintaining a `desiredMasterReplicaCount` var since that has no bearing on whether to add a new member as a learner.
- Added back the status check to ensure an etcd container is `Running` and not `Ready` before its added as a learner. Without this, bootstrap (or adding multiple new nodes in parallel) would fail because of peerURL validation errors for multiple unstarted members.

@p0lyn0mial @deads2k PTAL

hasbro17:
Added unit tests for learner addition in `clustermembercontroller_test.go/TestGetEtcdPeerURLToAdd`.

hasbro17:
Seeing a bootstrap failure issue again. 
```console
level=error msg=EtcdEndpointsDegraded: no etcd members are present
level=error msg=EtcdMembersDegraded: 1 of 2 members are available, NAME-PENDING-10.0.135.223 has not started
level=error msg=GuardControllerDegraded: [Missing operand on node ip-10-0-237-156.us-east-2.compute.internal, Missing operand on node ip-10-0-135-223.us-east-2.compute.internal]
level=error msg=StaticPodsDegraded: pod/etcd-ip-10-0-184-38.us-east-2.compute.internal container "etcd" is waiting: CrashLoopBackOff: back-off 5m0s restarting failed container=etcd pod=etcd-ip-10-0-184-38.us-east-2.compute.internal_openshift-etcd(a1d4eb4bf3857629fda3749c6564be68)
level=error msg=StaticPodsDegraded: pod/etcd-ip-10-0-184-38.us-east-2.compute.internal container "etcd-health-monitor" is waiting: CrashLoopBackOff: back-off 5m0s restarting failed container=etcd-health-monitor pod=etcd-ip-10-0-184-38.us-east-2.compute.internal_openshift-etcd(a1d4eb4bf3857629fda3749c6564be68)
level=info msg=Cluster operator etcd Progressing is True with EtcdMembers_MembersNotStarted::NodeInstaller: EtcdMembersProgressing: 1 of 2 members are available, NAME-PENDING-10.0.135.223 has not started
level=info msg=NodeInstallerProgressing: 3 nodes are at revision 0; 0 nodes have achieved new revision 2
level=info msg=Cluster operator etcd Available is False with EtcdMembers_No quorum::StaticPods_ZeroNodesActive: EtcdMembersAvailable: 1 of 2 members are available, NAME-PENDING-10.0.135.223 has not started
level=info msg=StaticPodsAvailable: 0 nodes are active; 3 nodes are at revision 0; 0 nodes have achieved new revision 2
```
Thought I had already fixed that with waiting until the etcd pod is running before adding as a member. Going to do a manual bootstrap test run to confirm if it's the same peerURL validation issue again.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +104,283 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+			return "", fmt.Errorf("failed to get internal IP for node: %v", err)
+		}
+		machine, err := c.findMachineByNodeInternalIP(internalIP)
+		if err != nil {
+			return "", fmt.Errorf("failed to get machine for node (%s): %v", node.Name, err)
+		}
+		if machine == nil {
+			klog.V(2).Infof("Ignoring node (%s) for scale-up: no Machine found referencing this node's internal IP (%v)", node.Name, internalIP)
+			continue
+		}
```
hasbro17, line 183: @p0lyn0mial Just a side note: I've had to had to refactor `isMachinePendingDeletionFor()` from #790 into the conditionals here. We're no longer erroring out if we don't find a machine, since we want to move on to attempt with adding other eligible nodes as learners.
Having a log statement is good enough imo, and if the machine appears, we'll retry anyway.

Also updated the associated unit test case for this from #790

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +104,283 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		// If multiple new nodes are added as members in parallel they will fail with
+		// a discovery error ("error validating peerURLs: ... member count is unequal") because
+		// the unstarted members won't include each other in their ETCD_INITIAL_CLUSTER env.
+		podName := fmt.Sprintf("etcd-%v", node.Name)
+		runningNotReady, err := c.isEtcdContainerRunningNotReady(podName)
+		if err != nil {
+			return "", fmt.Errorf("failed to check if pod is running and not ready: %v", err)
+		}
+		if !runningNotReady {
+			return "", nil
```
p0lyn0mial, line 206: we could continue and examine the next `node`

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +104,283 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
-		return false, nil
+	for _, node := range nodes {
+		if node.DeletionTimestamp != nil {
+			klog.V(2).Infof("Ignoring node (%s) for new member addition as it's pending deletion", node.Name)
+			continue
+		}
+
+		peerURL, err := c.getPeerURLForNode(node)
+		if err != nil {
+			return "", fmt.Errorf("failed to get peerURL for node: %v", err)
```
p0lyn0mial, line 154: we could collect all the errors and aggregate at the end. We took this approach in the other controllers mainly to be able to make progress in case one machine/node is broken.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +104,283 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		machine, err := c.findMachineByNodeInternalIP(internalIP)
+		if err != nil {
+			return "", fmt.Errorf("failed to get machine for node (%s): %v", node.Name, err)
+		}
+		if machine == nil {
+			klog.V(2).Infof("Ignoring node (%s) for scale-up: no Machine found referencing this node's internal IP (%v)", node.Name, internalIP)
+			continue
+		}
+		if machine.DeletionTimestamp != nil {
+			// Ignore this node since its member previously was, or is in the process of being removed
```
p0lyn0mial, line 185: we could log it, it would improve visibility during a potential investigation in the future.

pkg/etcdcli/etcdcli.go
```diff
@@ -186,6 +187,13 @@ func getEtcdClientWithClientOpts(endpoints []string, opts ...ClientOption) (*cli
 	if err != nil {
 		return nil, fmt.Errorf("failed to make etcd client for endpoints %v: %w", endpoints, err)
 	}
+
+	// If the endpoint includes a learner member then we skip the test
+	// as learner members don't support member list
+	if skipConnectionTest {
```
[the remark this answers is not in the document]
hasbro17: > decide on their own whether to test or not
Sorry could you please expand on that?

The callers of `getEtcdClientWithClientOpts()` are already deciding whether or not to skip the connection.
```Go
func getEtcdClientWithClientOpts(endpoints []string, skipConnectionTest bool, opts ...ClientOption) (*clientv3.Client, error) {
```
Here's the health-check deciding to skip it when it knows the connection is to a learner.
https://github.com/openshift/cluster-etcd-operator/pull/758/files#diff-3b4c3d6471887c34bc01770511c7fc811a6832d58ac9ebdcf6776c56075f5dd5R55-R62
```Go
skipConnectionTest := false
if member.IsLearner {
	skipConnectionTest = true
}
cli, err := getEtcdClientWithClientOpts([]string{member.ClientURLs[0]}, skipConnectionTest)
if err != nil {
	..
```

Or did you mean something else?

pkg/cmd/readyz/readyz.go
```diff
@@ -153,12 +153,22 @@ func (r *readyzOpts) getReadyzHandlerFunc(ctx context.Context) http.HandlerFunc
...
-		// linearized request to verify health of member
-		// TODO: Once learner members are supported, update this to differentiate
-		// learner members and only issue a serialized Get() for them
-		ctx, cancel := context.WithCancel(ctx)
-		defer cancel()
-		_, err = etcdClient.Get(ctx, "health")
+		// Learner and voting members both support the endpoint status call
+		resp, err := etcdClient.Status(ctx, r.targetEndpoint)
+		if err != nil {
+			klog.V(2).Infof("failed to get member endpoint status: %v", err)
```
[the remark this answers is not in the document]
hasbro17: The endpoint here is always `https://localhost:2379` but I can add that in the log.
https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/etcd/pod.yaml#L297

pkg/etcdcli/etcdcli.go
```diff
@@ -222,10 +228,38 @@ func (g *etcdClientGetter) MemberAdd(ctx context.Context, peerURL string) error
 		}
 	}
 
-	_, err = cli.MemberAdd(ctx, []string{peerURL})
+	defer func() {
+		if err != nil {
+			g.eventRecorder.Warningf("MemberAddAsLearner", "failed to add new member %v", peerURL)
```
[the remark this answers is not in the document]
hasbro17: Yes, good catch. This would be an infrequent/notable event unlike member promotion failing for a short while.
I'll add the error here and in the other events.

pkg/etcdcli/health.go
```diff
@@ -51,23 +52,40 @@ func getMemberHealth(ctx context.Context, etcdMembers []*etcdserverpb.Member) me
...
 			st := time.Now()
 			ctx, cancel := context.WithCancel(ctx)
-			// linearized request to verify health of member
-			resp, err := cli.Get(ctx, "health")
+			var resp *clientv3.GetResponse
+			if member.IsLearner {
+				// Learner members only support serializable (without consensus) read requests
+				// TODO: See why this is failing and being marked as unhealthy??
+				resp, err = cli.Get(ctx, "health", clientv3.WithSerializable())
+				klog.V(2).Infof("health check for learner: memberID(%v) err(%v)", resp.Header.MemberId, err)
```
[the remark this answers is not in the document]
hasbro17: Forgot to remove this after the last debug session.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -29,30 +31,40 @@ import (
...
 	// this controller can be run only on a cluster that exposes a functional Machine API
 	machineAPIChecker ceohelpers.MachineAPIChecker
 
 	masterMachineLister   machinelistersv1beta1.MachineLister
 	masterMachineSelector labels.Selector
+
+	masterNodeLister   corev1listers.NodeLister
+	masterNodeSelector labels.Selector
+
+	isMachineAPIFunctional bool
```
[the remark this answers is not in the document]
hasbro17: Initially this was just so I didn't have to call `c.machineAPIChecker.IsFunctional()` in multiple places but I can see it's probably confusing and unnecessary to store an additional state var for that. Going to remove this.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +104,283 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+	return true, nil
+}
+
+// findMachineByNodeInternalIP finds the machine that matches the given nodeInternalIP
+// is safe because the MAO:
+//  syncs the addresses in the Machine with those assigned to real nodes by the cloud provider,
+//  checks that the Machine and Node lists match before issuing a serving certification for the kubelet
+//  when the host disappears from the cloud side, it stops updating the Machine so the addresses and information
+//  should persist there as a tombstone as the Machine is marked Failed
+func (c *ClusterMemberController) findMachineByNodeInternalIP(nodeInternalIP string) (*machinev1beta1.Machine, error) {
```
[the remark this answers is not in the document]
hasbro17: Moving to ceohelpers

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +104,283 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+	// Do not promote if the learner member machine is pending deletion
+	if machine.DeletionTimestamp != nil {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// currentMemberMachinesWithDeletionHooks returns machines with the deletion hooks from the lister
+func (c *ClusterMemberController) currentMemberMachinesWithDeletionHooks() ([]*machinev1beta1.Machine, error) {
```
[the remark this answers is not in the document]
hasbro17: Moving to ceohelpers

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +104,283 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
-	if isFunctional, err := c.machineAPIChecker.IsFunctional(); err != nil {
-		return false, err
-	} else if !isFunctional {
-		// always return false when the machine API is off
-		// otherwise we won't be able to add any member
-		return false, nil
+// getEtcdPeerURLToAdd returns a PeerURL of the next etcd instance to be added as a learner to the cluster. If
+// no eligible member is found return an empty string.
+func (c *ClusterMemberController) getEtcdPeerURLToAdd(ctx context.Context) (string, error) {
+	members, err := c.etcdClient.MemberList(ctx)
```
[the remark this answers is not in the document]
hasbro17: The `etcd-endpoints` configmap only has the list of voting member endpoints, not learners by design.
So checking the current nodes against that list could result in the controller mistakenly trying to add a node that is already a learner in the cluster membership.
```Go
// Ignore if the node's PeerURL is mapped to a member and already part of the quorum.
if isURLMappedToMember(peerURL, members) {
	continue
}
```
> current nodes (which could be also retrieved from the cache)

The nodes are already coming from the node lister which should be using the cache, unless I'm missing something:
```Go
nodes, err := c.masterNodeLister.List(c.masterNodeSelector)
if err != nil {
	return "", err
}
```

Not sure if member list should really be that expensive, although I agree with using the cache as much as possible, but unlikely in this case it seems.

pkg/operator/clustermembercontroller/clustermembercontroller.go
hasbro17: Agreed, otherwise this could block promotion if we're failing for an unrelated node here.

pkg/operator/clustermembercontroller/clustermembercontroller.go
hasbro17: Yeah that's a good idea. V(4) I think, otherwise it might spam the logs if the deleted node sticks around for a long time.

hasbro17:
@p0lyn0mial updated, PTAL. Will squash commits before merging at the end.

pkg/operator/clustermembercontroller/clustermembercontroller.go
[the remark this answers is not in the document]
p0lyn0mial: > The etcd-endpoints configmap only has the list of voting member endpoints, not learners by design.

yeah, so if all voting members map to all master nodes then we don't have to retrieve the member list.
I think this will be the basic scenario, after all, scaling would be a one-time operation.
does it make sense?

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +100,255 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		}
+		if !isMachineAPIFunctional {
+			return peerURL, nil
+		}
+
+		// If the Machine API is functional we should ignore the machine
+		// if it is pending deletion to avoid re-adding previously removed members
+		internalIP, err := c.getInternalIPForNode(node)
+		if err != nil {
+			return "", fmt.Errorf("failed to get internal IP for node: %v", err)
```
p0lyn0mial, line 170: do we have to stop here? we could collect the errors inside this method (`getEtcdPeerURLToAdd`) and go on to the next node. I think the order doesn't really matter, right?

pkg/operator/ceohelpers/common.go
```diff
@@ -127,3 +129,36 @@ func memberToURL(member *etcdserverpb.Member) (string, error) {
...
 	return member.PeerURLs[0], nil
 }
+
+// FindMachineByNodeInternalIP finds the machine that matches the given nodeInternalIP
+// is safe because the MAO:
+//  syncs the addresses in the Machine with those assigned to real nodes by the cloud provider,
+//  checks that the Machine and Node lists match before issuing a serving certification for the kubelet
+//  when the host disappears from the cloud side, it stops updating the Machine so the addresses and information
+//  should persist there as a tombstone as the Machine is marked Failed
+func FindMachineByNodeInternalIP(nodeInternalIP string, machineSelector labels.Selector, machineLister machinelistersv1beta1.MachineLister) (*machinev1beta1.Machine, error) {
```
p0lyn0mial, line 139: we could create a new PR, I think that would make @deads2k happy and result in smaller diff

p0lyn0mial: ptal, https://github.com/openshift/cluster-etcd-operator/pull/795

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +100,255 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		if err != nil {
+			// Note: Cannot use errors.Is(err, etcdserver.ErrLearnerNotReady) as that is always false in this case.
+			// because etcdserver.ErrLearnerNotReady always returns an errors.New() error type
+			// So we compare the error strings here instead
+			if err.Error() == etcdserver.ErrLearnerNotReady.Error() {
+				// Not being ready for promotion is an expected state until the learner catches up
+				klog.V(2).Infof("Not ready for promotion: etcd learner member (%s) is not yet in sync with leader's log ", member.PeerURLs[0])
+				continue
+			}
+			return fmt.Errorf("failed to promote learner member (%s): %v", member.PeerURLs[0], err)
```
p0lyn0mial, line 241: why do we have to stop here? could we store the err and move on to the next member?

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +100,255 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		// We have a valid new node with a running etcd pod to add as a new member
+		return peerURL, nil
+	}
+	return "", nil
+}
+
+// ensureEtcdLearnerPromotion checks for all learner members in the cluster membership and tries
+// to promote them if they have the required machine deletion hooks present
+func (c *ClusterMemberController) ensureEtcdLearnerPromotion(ctx context.Context, recorder events.Recorder) error {
+	members, err := c.etcdClient.MemberList(ctx)
```
p0lyn0mial, line 215: We could rely on the information stored in the cache to indicate work. That would make this controller cheap in the steady-state.

p0lyn0mial: similar to https://github.com/openshift/cluster-etcd-operator/pull/758#discussion_r855056766, wdyt?

hasbro17:
@p0lyn0mial Added the check for using etcd-endpoints as the preliminary check before scaling. PTAL.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -61,10 +71,12 @@ func NewClusterMemberController(
...
 		podLister:             kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Pods().Lister(),
-		nodeLister:            kubeInformers.InformersFor("").Core().V1().Nodes().Lister(),
 		networkLister:         networkInformer.Lister(),
 		masterMachineLister:   machinelistersv1beta1.NewMachineLister(masterMachineInformer.GetIndexer()),
 		masterMachineSelector: masterMachineSelector,
+		masterNodeLister:      corev1listers.NewNodeLister(masterNodeInformer.GetIndexer()),
+		masterNodeSelector:    masterNodeSelector,
+		configMapLister:       kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Lister().ConfigMaps(operatorclient.TargetNamespace),
 	}
 	return factory.New().ResyncEvery(time.Minute).
```
p0lyn0mial, line 81: shouldn't we also add `kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps()` to the `WithInformer` method?

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +102,310 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+
+// getEtcdPeerURLToAdd returns a PeerURL of the next etcd instance to be added as a learner to the cluster. If
+// no eligible member is found return an empty string.
+func (c *ClusterMemberController) getEtcdPeerURLToAdd(ctx context.Context) (string, error) {
+	nodes, err := c.masterNodeLister.List(c.masterNodeSelector)
+	if err != nil {
+		return "", err
+	}
+
+	nodesMapToVotingMembers, err := c.allNodesMapToVotingMembers(nodes)
```
p0lyn0mial, line 134: it could return a list of nodes that don't match

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +102,310 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		}
+
+		// If the Machine API is unavailable then we can't evaluate the machine's
+		// deletion timestamp or deletion hooks so we scale it up without any considerations
+		// for vertical scaling
+		// This allows UPI clusters without a machine API to scale up
+		// Check machine API availability
+		isMachineAPIFunctional, err := c.machineAPIChecker.IsFunctional()
+		if err != nil {
+			return "", fmt.Errorf("failed to determine Machine API availability: %v", err)
```
p0lyn0mial, line 175: why not to collect the error and continue?

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +102,310 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		if err != nil {
+			return "", fmt.Errorf("failed to get machine for node (%s): %v", node.Name, err)
+		}
+		if machine == nil {
+			klog.V(2).Infof("Ignoring node (%s) for scale-up: no Machine found referencing this node's internal IP (%v)", node.Name, internalIP)
+			continue
+		}
+		if machine.DeletionTimestamp != nil {
+			// Ignore this node since its member previously was, or is in the process of being removed
+			klog.V(4).Infof("Ignoring node (%s) for scale-up since its machine (%s) is pending deletion", node.Name, machine.Name)
```
p0lyn0mial, line 198: we should be consistent, in other places within this method we do use `V(2)`. I think it is ok, this is important stuff, not an everyday operation. We should have full visibility into why a member has been ignored.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +102,310 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+			continue
+		}
+
+		// Wait until the machine has a deletion hook present before adding it as a member
+		if !ceohelpers.HasMachineDeletionHook(machine) {
+			continue
+		}
+
+		// Wait until the etcd pod is running but not ready on the node which indicates
```
p0lyn0mial, line 207: why we don't need this check when the machine API is not functional? (line 178)

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +102,310 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		}
+		if machine.DeletionTimestamp != nil {
+			// Ignore this node since its member previously was, or is in the process of being removed
+			klog.V(4).Infof("Ignoring node (%s) for scale-up since its machine (%s) is pending deletion", node.Name, machine.Name)
+			continue
+		}
+
+		// Wait until the machine has a deletion hook present before adding it as a member
+		if !ceohelpers.HasMachineDeletionHook(machine) {
+			continue
```
p0lyn0mial, line 204: I would add a log statement at `V(2)` level

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +102,310 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		}
+
+		peerURL, err := c.getPeerURLForNode(node)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to get peerURL for node: %v", err))
+			continue
+		}
+
+		// Ignore if the node's PeerURL is mapped to a member and already part of the quorum.
```
p0lyn0mial, line 163: learners will have assigned `PeerURLs`, right?

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +102,310 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		if isURLMappedToMember(peerURL, members) {
```
p0lyn0mial, line 164: in case of ipv6 `peerURL` will contain an escaped address (`[host]:port`), will `isURLMappedToMember` be able to find a matching address?

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +102,310 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+	}
+	machines, err := ceohelpers.CurrentMemberMachinesWithDeletionHooks(c.masterMachineSelector, c.masterMachineLister)
+	if err != nil {
+		return false, fmt.Errorf("failed to get master machines: %v", err)
+	}
+	nodeIPToMachinesWithDeletionHooks := ceohelpers.IndexMachinesByNodeInternalIP(machines)
+
+	machine, hasMachine := nodeIPToMachinesWithDeletionHooks[nodeInternalIP]
+	if !hasMachine {
+		return false, nil
```
p0lyn0mial, line 316: we don't want to know why a member won't be promoted? I'd add a log statement

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +102,310 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+	}
+
+	// Do not promote if the learner member machine is pending deletion
+	if machine.DeletionTimestamp != nil {
 		return false, nil
```
p0lyn0mial, line 321: same here, add a log statement?

pkg/operator/clustermembercontroller/clustermembercontroller.go
hasbro17: So we only iterate over those nodes? Yeah I guess we could do that.

pkg/operator/clustermembercontroller/clustermembercontroller.go
hasbro17: Good catch, we do need this check for non-functional API or UPI clusters as well.

pkg/operator/clustermembercontroller/clustermembercontroller.go
hasbro17: Yes, learners have identical startup configurations to "normal" members except they just get added with a different flag.

pkg/operator/clustermembercontroller/clustermembercontroller.go
hasbro17: Yes, in that case `member.PeerURLs[0]` would also just be that exact escaped ipv6 address so it should match.
We're not constructing the member URL in `isURLMappedToMember` but just comparing directly.

hasbro17:
@p0lyn0mial Updated for logging statements and the missing check. Also updated the e2e test to correctly health check learners vs voting members.

hasbro17:
Just remembered [the comment](https://github.com/openshift/origin/pull/26917#discussion_r857036849) on counting voting members only during the e2e test applies here too. I'll update for that.

test/e2e/vertical_scaling_test.go
```diff
@@ -275,10 +278,25 @@ func ensureHealthyMember(t testing.TB, etcdClientFactory etcdClientCreator, memb
 	require.NoError(t, err)
 	defer closeFn()
 
-	// since we have a direct connection with the member
-	// getting any response is a good sign of healthiness
 	ctx, cancel := context.WithTimeout(context.TODO(), 15*time.Second)
 	defer cancel()
+
+	// // Learner and voting members both support the endpoint status call
+	// resp, err := etcdClient.Status(ctx, "https://127.0.0.1:2379")
```
p0lyn0mial, line 285: we call it after `ensureVotingMembersCount`, so it seems we don't need this distinction, right?

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -90,117 +103,322 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		if err != nil {
+			// Note: Cannot use errors.Is(err, etcdserver.ErrLearnerNotReady) as that is always false in this case.
+			// because etcdserver.ErrLearnerNotReady always returns an errors.New() error type
+			// So we compare the error strings here instead
+			if err.Error() == etcdserver.ErrLearnerNotReady.Error() {
+				// Not being ready for promotion is an expected state until the learner catches up
+				klog.V(2).Infof("Not ready for promotion: etcd learner member (%s) is not yet in sync with leader's log ", member.PeerURLs[0])
+				continue
+			}
+			errs = append(errs, fmt.Errorf("failed to promote learner member (%s): %v", member.PeerURLs[0], err))
```
p0lyn0mial, line 286: we could fire an event, that would allow us to catch potential issues in CI. The event could be throttled every X min

hasbro17: We already do have an events for promotion failure or success in the `MemberPromote()` method itself.
```Go
	defer func() {
		if err != nil {
			// Not being ready for promotion can be a common event until the learner's log
			// catches up with the leader, so we don't emit events for failing for that case
			if err.Error() == etcdserver.ErrLearnerNotReady.Error() {
				return
			}
			g.eventRecorder.Warningf("MemberPromote", "failed to promote learner member %s: %v", member.PeerURLs[0], err)
		} else {
			g.eventRecorder.Eventf("MemberPromote", "successfully promoted learner member %v", member.PeerURLs[0])
		}
	}()
``` 
https://github.com/openshift/cluster-etcd-operator/pull/758/files#diff-79a9561715e7c906edbb7c786fdbab2cc3f0e6cd83447a8a4aaa5c3c87b0aa2dR256

Or did you mean something else?

test/e2e/vertical_scaling_test.go
hasbro17: Yep that's correct. Forgot to remove this block in the last update.

p0lyn0mial:
informing jobs looks good to me

`e2e-metal-ipi-serial-ovn-ipv6` - was green (https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-cluster-etcd-operator-758-nightly-4.11-e2e-metal-ipi-serial-ovn-ipv6/1519427068218052608)

`e2e-vsphere-serial` - was green (https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-cluster-etcd-operator-758-nightly-4.11-e2e-vsphere-serial/1519427466999894016)

`metal-ipi-serial-ovn-ipv6` - was green (https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-cluster-etcd-operator-758-nightly-4.11-e2e-metal-ipi-serial-ovn-ipv6/1519427068218052608)

PR #763, opened 2022-03-17
Bug 2063831: replace quorumguard and add readyz server

hasbro17:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2063831

This PR replaces the use of the etcd-operator's existing quorum guard controller with the staticpod controller's method of force assigning guard pods directly to the nodes. This should address the issue of two quorum guard pods erroneously landing on the same node.


pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,187 @@
...
+		server.Shutdown(shutdownCtx)
+	}()
+
+	c := net.ListenConfig{}
+	c.Control = permitAddressReuse
+	ln, err := c.Listen(shutdownCtx, "tcp", addr)
+	if err != nil {
+		return err
+	}
+	err = server.Serve(ln)
```
hasbro17, line 119: Just realized this ought to be `ServeTLS()` since the guard controller pods are using HTTPS
```YAML
      readinessProbe:
        httpGet:
          path: healthz
          port: 7443
          host: 10.0.153.60
          scheme: HTTPS
```

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,187 @@
...
+	return err
+}
+
+func (r *readyzOpts) getReadyzHandlerFunc(ctx context.Context) http.HandlerFunc {
+	return func(w http.ResponseWriter, _ *http.Request) {
+		etcdClient, err := r.newETCD3Client(ctx, r.targetEndpoint)
+		if err != nil {
+			// TODO: The error message won't show up in the pod description if the probe fails
+			// e.g Warning  ProbeError  27s   kubelet  Readiness probe error: HTTP probe failed with statuscode: 503
+			// Find a way to get the error message to show there.
```
hasbro17, line 133: Seems like the kubelet's prober ignores the error message written to the body:
https://github.com/kubernetes/kubernetes/blob/56062f7f4fc56e038f9ab99042457183896c47bb/pkg/probe/http/http.go#L137-L138

It's available as a v4 log though, and we can alway query the readyz enpdpoint directly to confirm the error when it's failing so going to remove this TODO.

bindata/etcd/pod.yaml
```diff
@@ -192,23 +192,11 @@ ${COMPUTED_ENV_VARS}
...
-          # required as the kernel will allow the reuse of a port while in TIME_WAIT. etcd requires socket
-          # path in this format <name>:<port> so port 0 is used only to meet this requirement.
-          unset ETCDCTL_ENDPOINTS
-          /usr/bin/etcdctl \
-            --command-timeout=2s \
-            --dial-timeout=2s \
-            --endpoints=unixs://${NODE_NODE_ENVVAR_NAME_IP}:0 \
-            endpoint health -w json | grep \"health\":true
+      httpGet: 
+        port: 7443
```
deads2k, line 196: there's a list of known ports somewhere. I bet @dhellmann knows where it lives.

bindata/etcd/pod.yaml
```diff
@@ -300,6 +288,38 @@ ${COMPUTED_ENV_VARS}
...
+    - containerPort: 7443
+    resources:
+      requests:
+        memory: 50Mi
+        cpu: 10m
+    env:
+${COMPUTED_ENV_VARS}
+      - name: "ETCD_STATIC_POD_VERSION"
+        value: "REVISION"
+      - name: "OPENSHIFT_PROFILE"
```
deads2k, line 312: this looks a little odd, what is it for?

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+const (
+	defaultEndpoint        = "https://localhost:2379"
+	defaultHTTPDialTimeout = 2 * time.Second
+	// keepalive defaults used by kube-apiserver
+	// https://github.com/kubernetes/apiserver/blob/de6ba2aa0a752d077719fcee186291d7af73a825/pkg/storage/storagebackend/factory/etcd3.go#L55-L56
+	keepaliveTime     = 30 * time.Second
+	keepaliveTimeout  = 10 * time.Second
+	DefaultListenPort = 7443
```
deads2k, line 29: public exposure here rubs me the wrong way.

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,187 @@
...
+		},
+	}
+
+	opts.AddFlags(cmd.Flags())
+	return cmd
+}
+
+func (r *readyzOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.Uint16Var(&r.listenPort, "listen-port", r.listenPort, "Listen on this port. Default 7443")
+	fs.DurationVar(&r.dialTimeout, "dial-timeout", defaultHTTPDialTimeout, "Dial timeout for the client. Default 2s")
```
deads2k, line 65: use r.dialTimeout as teh default (and in the other spots) and construct your default options as you want them to appear.  a `NewReadyzOpts() *readyzOpts` is how we do this in kubectl and oc.

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,187 @@
...
+	defaultEndpoint        = "https://localhost:2379"
+	defaultHTTPDialTimeout = 2 * time.Second
+	// keepalive defaults used by kube-apiserver
+	// https://github.com/kubernetes/apiserver/blob/de6ba2aa0a752d077719fcee186291d7af73a825/pkg/storage/storagebackend/factory/etcd3.go#L55-L56
+	keepaliveTime     = 30 * time.Second
+	keepaliveTimeout  = 10 * time.Second
+	DefaultListenPort = 7443
+)
+
+type readyzOpts struct {
```
deads2k, line 32: missing serving cert/key

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	go func() {
+		defer cancel()
+		<-shutdownHandler
+		klog.Infof("Received SIGTERM or SIGINT signal, shutting down server.")
+		server.Shutdown(shutdownCtx)
+	}()
+
+	c := net.ListenConfig{}
+	c.Control = permitAddressReuse
```
deads2k, line 114: ask @p0lyn0mial if this is appropriate here.  I don't actually know. I think it's ok since this is effectively a forwarder, but I'm not certain

pkg/cmd/readyz/readyz.go
deads2k: v(2) if you please.

deads2k: > Seems like the kubelet's prober ignores the error message written to the body:
https://github.com/kubernetes/kubernetes/blob/56062f7f4fc56e038f9ab99042457183896c47bb/pkg/probe/http/http.go#L137-L138

our fork of kube includes body message snippets for certain namespaces.

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,187 @@
...
+		if err != nil {
+			// TODO: The error message won't show up in the pod description if the probe fails
+			// e.g Warning  ProbeError  27s   kubelet  Readiness probe error: HTTP probe failed with statuscode: 503
+			// Find a way to get the error message to show there.
+			http.Error(w, fmt.Sprintf("failed to establish etcd client: %v", err), http.StatusServiceUnavailable)
+			return
+		}
+
+		// linearized request to verify health of member
+		// TODO: Once learner members are supported, update this to differentiate
```
deads2k, line 139: You can add this differentiation after initial merge and before adding learners.

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,187 @@
...
+		if err != nil {
+			http.Error(w, fmt.Sprintf("failed to get member health key: %v", err), http.StatusServiceUnavailable)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+	}
+}
+
+// TODO: Copied from cluster-kube-apiserver. Confirm if this is needed?
+func permitAddressReuse(network, addr string, conn syscall.RawConn) error {
```
deads2k, line 153: ask @p0lyn0mial to review please.

deads2k:
missing a control loop to delete the old quorum guard pods and old PDB *after* etcd is at the new version.

bindata/etcd/pod.yaml
dhellmann: You may mean the host port registry? https://github.com/openshift/enhancements/blob/master/dev-guide/host-port-registry.md

deads2k:
And we need a PR against 4.10 that deletes the new PDB if it is present.

hasbro17:
Still have to update this PR with the controller that cleans up the old quorum guard pods.

bindata/etcd/pod.yaml
hasbro17: Updated to `9980` as the next in series for the etcd ports.
https://github.com/openshift/enhancements/pull/1068

bindata/etcd/pod.yaml
hasbro17: Carryover from the `etcd` container template. Removed as we don't really need these.

pkg/cmd/readyz/readyz.go
hasbro17: I was reusing the on disk etcd client cert/key files for the readyz server certs/key to keep the changes to a min:
```Go
err = server.ServeTLS(ln, r.clientCertFile, r.clientKeyFile)
``` 
That works fine but I'm guessing I should probably just issue a separate self-signed cert/key pair here, e.g:
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/cmd/render/certs.go#L21-L23

Just need to re-format it for how the TLS config expects it.

pkg/cmd/readyz/readyz.go
hasbro17: Is the issue here with not wanting to default r.dialTimeout via a flag?
Or more about the function signature of how the opts are initialized.
I can change this to be more like:
https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/get/get.go#L143-L201

pkg/operator/etcd_assets/bindata.go
```diff
@@ -791,23 +790,12 @@ ${COMPUTED_ENV_VARS}
...
-          # path in this format <name>:<port> so port 0 is used only to meet this requirement.
-          unset ETCDCTL_ENDPOINTS
-          /usr/bin/etcdctl \
-            --command-timeout=2s \
-            --dial-timeout=2s \
-            --endpoints=unixs://${NODE_NODE_ENVVAR_NAME_IP}:0 \
-            endpoint health -w json | grep \"health\":true
+      httpGet: 
+        port: 9980
+        path: readyz
```
hasbro17, line 795: Just figured out why the new guard pods were failing their readiness check.
Apparently the static pod controller guard pods are hardcoded to use `/healthz` instead of `/readyz` as the readiness endpoint check.
https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml#L44

I thought the former was deprecated but I think I'll have to change this to `healthz`.
Or probably I can just keep this, and just also serve the `/healthz` endpoint for the guard pod.

hasbro17:
Added the cleanup controller for removing the old etcd quorum guard pods after the new guard pods are all rolled out and ready.

Will be following up with a separate PR on 4.10 for the converse, i.e clean up the new guard pods in case of a downgrade from 4.11->4.10.

hasbro17:
Also need to change the e2e test that was previously checking if the quorum guard deployment correctly blocks node evictions.
https://github.com/openshift/cluster-etcd-operator/blob/master/test/e2e/etcquorumguard_test.go

Although I think that kind of general test might be better suited in library-go, so it can be tested where the guard controller actually lives.
Not sure if the existing tests cover that though.
https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/guard/guard_controller_test.go

hasbro17:
@deads2k Please feel free to do another review pass while I address the above comments.

pkg/cmd/readyz/readyz.go
deads2k: > I can change this to be more like:
https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/get/get.go#L143-L201

Yes more like this.

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,189 @@
...
+	shutdownCtx, cancel := context.WithCancel(context.Background())
+	shutdownHandler := server.SetupSignalHandler()
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/readyz", r.getReadyzHandlerFunc(shutdownCtx))
+	// Handle the /healthz endpoint as well since the static pod controller's guard pods check the /healthz endpoint
+	// https://github.com/openshift/library-go/blob/edab248e63516c65a93467eaa8224c86d69f5de9/pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml#L44
+	mux.HandleFunc("/healthz", r.getReadyzHandlerFunc(shutdownCtx))
+
+	addr := fmt.Sprintf("0.0.0.0:%d", r.listenPort)
```
deads2k, line 101: can we bind to localhost?

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,189 @@
...
+		server.Shutdown(shutdownCtx)
+	}()
+
+	c := net.ListenConfig{}
+	c.Control = permitAddressReuse
+	ln, err := c.Listen(shutdownCtx, "tcp", addr)
+	if err != nil {
+		return err
+	}
+	err = server.ServeTLS(ln, r.clientCertFile, r.clientKeyFile)
```
deads2k, line 122: `ServeTLS` takes a serving cert/key pair, not the client cert/key pair.  I don't mind if you re-use the existing etcd serving cert/key pair

pkg/cmd/readyz/readyz.go
deads2k: client cert are not the same as serving certs.  They have different usages signed by the CA

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,189 @@
...
+			http.Error(w, fmt.Sprintf("failed to establish etcd client: %v", err), http.StatusServiceUnavailable)
+			return
+		}
+
+		// linearized request to verify health of member
+		// TODO: Once learner members are supported, update this to differentiate
+		// learner members and only issue a serialized Get() for them
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+		_, err = etcdClient.Get(ctx, "health")
```
deads2k, line 144: any response code is good enough?  Is that congruent to current behavior?

deads2k: > any response code is good enough? Is that congruent to current behavior?

looks like it previously checked for health: true.  This should do the same

Also, please write the response body back.

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,189 @@
...
+		if err != nil {
+			klog.V(2).Infof("failed to get member health key: %v", err)
+			http.Error(w, fmt.Sprintf("failed to get member health key: %v", err), http.StatusServiceUnavailable)
+			return
```
deads2k, line 148: Explicitly write the failing response back please, along with the detail we got from etcd if we got any.  If we didn't get a body, explain that too.

pkg/operator/quorumguardcleanup/quorumguardcleanupcontroller.go
```diff
@@ -0,0 +1,141 @@
...
+//  and then deletes the "etcd-quorum-guard" deployment
+func (c *QuorumGuardCleanupController) cleanupOldGuardDeployment(ctx context.Context, recorder events.Recorder) error {
+	// Check if the old quorum guard deployment exists
+	_, err := c.deploymentLister.Deployments(operatorclient.TargetNamespace).Get(oldEtcdGuardDeploymentName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// Old guard deployment already removed, no more cleanup needed
+			return nil
+		}
+		return fmt.Errorf("failed to get etcd quorum guard deployment: %v", err)
```
deads2k, line 95: `%w`

pkg/operator/quorumguardcleanup/quorumguardcleanupcontroller.go
```diff
@@ -0,0 +1,141 @@
...
+		deploymentLister: kubeInformers.InformersFor(operatorclient.TargetNamespace).Apps().V1().Deployments().Lister(),
+		masterNodeLister: corev1listers.NewNodeLister(masterNodeInformer.GetIndexer()),
+	}
+	return factory.New().ResyncEvery(5*time.Minute).WithInformers(
+		kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Pods().Informer(),
+		kubeInformers.InformersFor(operatorclient.TargetNamespace).Apps().V1().Deployments().Informer(),
+	).WithSync(c.sync).ToController("QuorumGuardCleanupController", eventRecorder.WithComponentSuffix("quorum-guard-cleanup-controller"))
+}
+
+func (c *QuorumGuardCleanupController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
deads2k, line 61: You need a check to be sure that the current level of this clusteroperator's version matches the desired level.  Otherwise when this operator starts during an upgrade, it will immediately remove quorum guard protection.

deads2k: > You need a check to be sure that the current level of this clusteroperator's version matches the desired level. Otherwise when this operator starts during an upgrade, it will immediately remove quorum guard protection.

oh, I see it down below by looking for the actual pods.  I think that's more complicated, but if you prefer it, ok.

pkg/operator/quorumguardcleanup/quorumguardcleanupcontroller.go
```diff
@@ -0,0 +1,141 @@
...
+		return fmt.Errorf("failed to list master nodes: %v", err)
+	}
+
+	// Wait until all new guard pods are rolled out to all master nodes, and have the ready condition
+	if countReadyPods(newGuardPods) != len(masterNodes) {
+		return nil
+	}
+
+	// Delete the old quorum guard deployment
+	if err := c.kubeClient.AppsV1().Deployments(operatorclient.TargetNamespace).Delete(ctx, oldEtcdGuardDeploymentName, metav1.DeleteOptions{}); err != nil {
```
deads2k, line 121: need to delete the old PDB as well, right?

pkg/cmd/readyz/readyz.go
hasbro17: I tested it out but I'm not sure that works.
The http-get probe on the `etcd` container will pass if I additionally specify `host: 127.0.0.1` since we are using `hostNetwork: true`.
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#http-probes

But on the actual guard pod the probe uses the etcd pod's IP as the host which isn't accessible since the readyz server is listening on localhost only. So the guard pod's probe fails to reach the readyz server on the etcd pod.
```YAML
      readinessProbe:
        httpGet:
          path: healthz
          port: 9980
          host: 10.0.223.164
          scheme: HTTPS
```
```console
$ kubectl -n openshift-etcd describe pod etcd-guard-ip-10-0-223-164.us-west-1.compute.internal
Events:
  Type     Reason          Age                    From     Message
  ----     ------          ----                   ----     -------
...
  Normal   Started         4m5s                   kubelet  Started container guard
  Warning  ProbeError      3m23s (x11 over 4m4s)  kubelet  Readiness probe error: Get "https://10.0.223.164:9980/healthz": dial tcp 10.0.223.164:9980: connect: connection refused
body:
  Warning  Unhealthy  3m23s (x11 over 4m4s)  kubelet  Readiness probe failed: Get "https://10.0.223.164:9980/healthz": dial tcp 10.0.223.164:9980: connect: connection refused
```

pkg/cmd/readyz/readyz.go
hasbro17: So long as there isn't an error for that call, it seems that should be enough to confirm the etcd endpoint is healthy and responsive.
That's pretty much the health check that the `etcdctl endpoint health` cmd does underneath:
https://github.com/etcd-io/etcd/blob/dcc226491f4e213fa09f4606a998ea387ea28ed9/etcdctl/ctlv3/command/ep_command.go#L136-L143

And what we currently have other places in the operator:
https://github.com/openshift/cluster-etcd-operator/blob/f3d8db9c071c979b064063e97206f24398e9a854/pkg/etcdcli/health.go#L62-L72

https://github.com/openshift/cluster-etcd-operator/blob/f3d8db9c071c979b064063e97206f24398e9a854/bindata/etcd/pod.yaml#L194-L211

pkg/cmd/readyz/readyz.go
hasbro17: Sorry but could you please expand on what you mean by "explicitly write".
The above `http.Error(w, ...)` is writing the full error response from the etcd client Get() request to the http response writer, which shows up in the response body.
https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/net/http/server.go;l=2093-2098

Just double checked the events for a failing probe and the error does show up in the body, e.g:
```console
$ kubectl -n openshift-etcd describe pod etcd-ip-10-0-223-164.us-west-1.compute.internal
Events:
  Type     Reason      Age   From     Message
  ----     ------      ----  ----     -------
...
  Normal   Created     48s   kubelet  Created container etcd-readyz
  Normal   Started     48s   kubelet  Started container etcd-readyz
  Warning  ProbeError  42s   kubelet  Readiness probe error: HTTP probe failed with statuscode: 503
body: failed to establish etcd client: context deadline exceeded
```

Seems like our kublet doesn't swallow the body.

pkg/operator/quorumguardcleanup/quorumguardcleanupcontroller.go
```diff
@@ -0,0 +1,234 @@
...
+	if err := c.removePDB(ctx, syncCtx); err != nil {
+		return err
+	}
+	return nil
+}
+
+// checkClusterOperatorVersion ensures the operator version in clusteroperator/etcd matches the desired version
+// in clusterversion/version
+// i.e all operands have been rolled out to the new image
+func (c *QuorumGuardCleanupController) checkClusterOperatorVersion() (bool, error) {
```
hasbro17, line 146: >You need a check to be sure that the current level of this clusteroperator's version matches the desired level. Otherwise when this operator starts during an upgrade, it will immediately remove quorum guard protection.

> oh, I see it down below by looking for the actual pods. I think that's more complicated, but if you prefer it, ok.

@deads2k Sorry, something is off with github comments and I didn't see your [second comment](https://github.com/openshift/cluster-etcd-operator/pull/763/files#r834334470) until just now.

I've already added this check for the clusteroperator version but just double checking if this is what you meant by checking the desired level. I'm a bit unfamiliar with how the clusteroperator is updating the "operator" version but I think it waits for the operand versions to roll out.
https://github.com/openshift/cluster-etcd-operator/blob/f3d8db9c071c979b064063e97206f24398e9a854/vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go#L143-L158

And yes I already have the check for making sure the new guard pods are all ready before the old ones are cleaned up as well.

pkg/operator/etcd_assets/bindata.go
hasbro17: They don't. I just meant the new ones were failing when they were querying the older `/healthz` endpoint. Although that's fixed now since I've added the handler for both.

hasbro17:
More updates on using serving certs and cleaning up the PDB. Have a couple of nits to do and will squash up the commits at the end. Keeping them separate for ease of review for now.

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,199 @@
...
+	if err == http.ErrServerClosed {
+		err = nil
+	}
+	<-shutdownCtx.Done()
+	return err
+}
+
+func (r *readyzOpts) getReadyzHandlerFunc(ctx context.Context) http.HandlerFunc {
+	return func(w http.ResponseWriter, _ *http.Request) {
+		etcdClient, err := r.newETCD3Client(ctx, r.targetEndpoint)
```
p0lyn0mial, line 142: I think it will create a new client on every HTTP request. Is this what we want?

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	err = server.ServeTLS(ln, r.servingCertFile, r.servingKeyFile)
+	if err == http.ErrServerClosed {
+		err = nil
+	}
+	<-shutdownCtx.Done()
+	return err
+}
+
+func (r *readyzOpts) getReadyzHandlerFunc(ctx context.Context) http.HandlerFunc {
```
p0lyn0mial, line 140: Should we set a maximum time a request should take? Otherwise, it might linger forever. (?)

bindata/etcd/pod.yaml
```diff
@@ -192,23 +192,12 @@ ${COMPUTED_ENV_VARS}
...
-          # path in this format <name>:<port> so port 0 is used only to meet this requirement.
-          unset ETCDCTL_ENDPOINTS
-          /usr/bin/etcdctl \
-            --command-timeout=2s \
-            --dial-timeout=2s \
-            --endpoints=unixs://${NODE_NODE_ENVVAR_NAME_IP}:0 \
-            endpoint health -w json | grep \"health\":true
+      httpGet:
+        port: 9980
+        path: readyz
```
p0lyn0mial, line 197: are we going to hit this endpoint from the operator in the future?

pkg/cmd/readyz/readyz.go
hasbro17: I think that's what the readinessProbe does today as well so I would think it should be okay load wise.
https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/etcd/pod.yaml#L207-L211

And makes sense to me that we would want to test out a new connection for the health check but open to hearing why this might be a bad idea.

bindata/etcd/pod.yaml
hasbro17: Possibly but this is mostly for the etcd pod's own readiness probe for the `etcd` container and also for the guard pod to check readiness.
https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml#L42-L46

Although the guard pod is using the outdated `/healthz` path.

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,199 @@
...
+	c.Control = permitAddressReuse
+	ln, err := c.Listen(shutdownCtx, "tcp", addr)
+	if err != nil {
+		return err
+	}
+	err = server.ServeTLS(ln, r.servingCertFile, r.servingKeyFile)
+	if err == http.ErrServerClosed {
+		err = nil
+	}
+	<-shutdownCtx.Done()
```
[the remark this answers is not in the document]
hasbro17: Need to dig into the shutdown handler but I think yes. Either way, I should probably handle other errors, so good catch.

hasbro17:
Still fiddling with the e2e test (double checking why it is seemingly able to evict more than one pod) but this should be ready for review otherwise.
@deads2k @p0lyn0mial PTAL

pkg/cmd/readyz/readyz.go
deads2k: Ok, thanks

pkg/cmd/readyz/readyz.go
deads2k: > I think it will create a new client on every HTTP request. Is this what we want?

To start, yes.

pkg/cmd/readyz/readyz.go
deads2k: > Should we set a maximum time a request should take? Otherwise, it might linger forever. (?)

This looks worth fixing as a followup

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,205 @@
...
+		},
+	}
+
+	opts.AddFlags(cmd.Flags())
+	return cmd
+}
+
+func (r *readyzOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.Uint16Var(&r.listenPort, "listen-port", r.listenPort, "Listen on this port. Default 9980")
+	fs.DurationVar(&r.dialTimeout, "dial-timeout", defaultHTTPDialTimeout, "Dial timeout for the client. Default 2s")
```
deads2k, line 73: default should be r.dialTimeout and use `newReadyzOpts` to set it to what you want

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,205 @@
...
+	fs.StringVar(&r.targetEndpoint, "target", defaultEndpoint, "Target endpoint to perform health check against. Default https://localhost:2379")
```
deads2k, line 74: same here

pkg/cmd/readyz/readyz.go
```diff
@@ -0,0 +1,205 @@
...
+	fs.StringVar(&r.servingCertFile, "serving-cert-file", "", "Health probe server TLS client certificate file. (required)")
```
deads2k, line 75: and here, even though its `""` wiring to `r.servingCertFile` will make your future life easier if you need a change.

pkg/cmd/readyz/readyz.go
deads2k: ok, I see

pkg/operator/quorumguardcleanup/quorumguardcleanupcontroller.go
```diff
@@ -0,0 +1,238 @@
...
+	etcdClusterOperator, err := c.clusterOperatorLister.Get("etcd")
+	if err != nil {
+		return false, fmt.Errorf("failed to get clusteroperator/etcd: %w", err)
+	}
+	operatorVersion := ""
```
deads2k, line 155: you want this version https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/starter.go#L187

You should pass it through to your controller constructor.

pkg/operator/quorumguardcleanup/quorumguardcleanupcontroller.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	if len(operatorVersion) == 0 {
+		return false, fmt.Errorf("failed to find operator version in clusteroperator/etcd")
+	}
+
+	clusterVersion, err := c.clusterVersionLister.Get("version")
+	if err != nil {
+		return false, fmt.Errorf("failed to check clusterversion/version: %w", err)
+	}
+	if operatorVersion != clusterVersion.Status.Desired.Version {
```
deads2k, line 169: you want to compare against this version https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/starter.go#L187

You should pass it through to your controller constructor.

deads2k:
remember to be sure that https://github.com/openshift/enhancements/pull/1068 merges

pkg/cmd/readyz/readyz.go
hasbro17: Added a TODO for now but testing with https://pkg.go.dev/net/http#TimeoutHandler to see if that would be sufficient.

hasbro17:
Will squash and unhold once the PR looks good to merge.

pkg/operator/quorumguardcleanup/quorumguardcleanupcontroller.go
```diff
@@ -0,0 +1,253 @@
...
+		masterNodeLister:        corev1listers.NewNodeLister(masterNodeInformer.GetIndexer()),
+		clusterOperatorLister:   clusterOperatorInformer.Lister(),
+		clusterVersionLister:    clusterVersionInformer.Lister(),
+		infrastructureInformer:  infrastructureInformer,
+	}
+	return factory.New().ResyncEvery(1*time.Minute).WithInformers(
+		kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Pods().Informer(),
+		kubeInformers.InformersFor(operatorclient.TargetNamespace).Apps().V1().Deployments().Informer(),
+		clusterOperatorInformer.Informer(),
+	).WithSync(c.sync).ToController("QuorumGuardCleanupController", eventRecorder.WithComponentSuffix("quorum-guard-cleanup-controller"))
```
deads2k, line 77: use `WithSyncDegradedOnError` instead of `WithSync`

pkg/operator/quorumguardcleanup/quorumguardcleanupcontroller.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func (c *QuorumGuardCleanupController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	err := c.cleanupOldQuorumGuard(ctx, syncCtx)
+	if err != nil {
```
deads2k, line 82: once the error return handles this by using WithSyncDegradedOnError, you don't need to write it here.

pkg/operator/quorumguardcleanup/quorumguardcleanupcontroller.go
```diff
@@ -0,0 +1,253 @@
...
+	if isSNO {
+		return nil
+	}
+
+	isTargetVersion, err := c.checkClusterOperatorVersion()
+	if err != nil {
+		return fmt.Errorf("failed to ensure clusteroperator version: %w", err)
+	}
+	if !isTargetVersion {
+		syncCtx.Queue().AddAfter(syncCtx.QueueKey(), 1*time.Minute)
```
deads2k, line 124: use `ResyncEvery(time.Minute)` in the `New` method and remove this.

pkg/operator/quorumguardcleanup/quorumguardcleanupcontroller.go
```diff
@@ -0,0 +1,253 @@
...
+		return false, fmt.Errorf("failed to get clusteroperator/etcd: %w", err)
+	}
+	operatorVersion := ""
+	for _, version := range etcdClusterOperator.Status.Versions {
+		if version.Name == "operator" {
+			operatorVersion = version.Version
+		}
+	}
+	if len(operatorVersion) == 0 {
+		return false, fmt.Errorf("failed to find operator version in clusteroperator/etcd")
```
deads2k, line 181: This will happen on initial installation.  Just `return false`

deads2k:
looking good.  please squash when you update

deads2k:
/hold

upgrades are broken

hasbro17:
@deads2k @p0lyn0mial Do either of you know where the `operator conditions` test is defined?
```
operator conditions etcd:
{Operator progressing (NodeInstaller): NodeInstallerProgressing: 1 nodes are at revision 8; 2 nodes are at revision 9  Operator progressing (NodeInstaller): NodeInstallerProgressing: 1 nodes are at revision 8; 2 nodes are at revision 9}
```

Nothing stands out to me in the must gather that indicates a stalled rollout of the pods. I've already ruled out the new PDB blocking rollout so I'm thinking this might be a timeout issue. I suspect with the readyz server the revision rollout would take longer as you have to wait for the readyz server for the health check (compared to the exec probe).

hasbro17:
@deads2k I guess this is good to go. `ci/prow/e2e-gcp-five-control-plane-replicas` is optional and has been broken for a long time so we can ignore that for now.

geliu2016 (commented):
/lgtm

geliu2016 (approved):
/lgtm

PR #766, opened 2022-03-18
Bug 2027744: Change PodDisruptionBudget policy to minavailable

Elbehery:
Change PodDistruptionBudget policy to MinAvailable instead of MaxUnavailable.

The quorum guard purpose is to keep minimum of two etcd pods running to guarantee quorum in the cluster. 

The assumption is that usually there are 3 master nodes, and therefore we can not allow more than one node down. If the cluster master was more than this, the condition still applies unnecessarily. 

Changing the policy to MinAvailable of two has the same expected result

fixes BZ [2027744](https://bugzilla.redhat.com/show_bug.cgi?id=2027744)

### DIFF
--- a/pkg/operator/quorumguardcontroller/quorumguardcontroller.go
+++ b/pkg/operator/quorumguardcontroller/quorumguardcontroller.go
@@ -50,8 +50,8 @@ var pdb = &policyv1.PodDisruptionBudget{
 		Namespace: operatorclient.TargetNamespace,
 	},
 	Spec: policyv1.PodDisruptionBudgetSpec{
-		MaxUnavailable: &intstr.IntOrString{Type: intstr.Int, IntVal: int32(1)},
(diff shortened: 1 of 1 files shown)


Elbehery:
@hasbro17  shall I close this PR ?

hasbro17:
@Elbehery Yes I think we can close this out. Once we switch over to the static guard pods those should be using minavailable as the default policy.
https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/guard/manifests/pdb.yaml#L7

https://github.com/openshift/cluster-etcd-operator/pull/787 was merged but reverted due to a regression (https://github.com/openshift/cluster-etcd-operator/pull/787) and will go back in again soon.

Elbehery:
closed in favour of https://github.com/openshift/cluster-etcd-operator/pull/763

PR #767, opened 2022-03-19
introduces ReadDesiredControlPlaneReplicasCount

p0lyn0mial:
simply reads the desired Control Plane replica count from the cluster-config-v1 configmap in the kube-system namespace.

it will be used by https://github.com/openshift/cluster-etcd-operator/pull/742 and https://github.com/openshift/cluster-etcd-operator/pull/758

### DIFF
--- a/pkg/operator/ceohelpers/common.go
+++ b/pkg/operator/ceohelpers/common.go
@@ -1,0 +1,49 @@
+package ceohelpers
+
+import (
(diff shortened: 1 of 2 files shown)


hasbro17 (approved):
Thanks for breaking this out 👍  I'll update #758 to use this.

/lgtm
/approve

PR #768, opened 2022-03-23
member removal part two

p0lyn0mial:


pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -77,16 +81,88 @@ func NewClusterMemberRemovalController(
...
+	// machines with the deletion hooks are the ones that host etcd members
+	// if the number of master machines is > controlPlane.replicas based on data from the cache
+	desiredControlPlaneReplicasCount, err := ceohelpers.ReadDesiredControlPlaneReplicasCount(c.configMapListerForKubeSystemNamespace)
+	if err != nil {
+		return err
+	}
+	if len(memberMachines) <= desiredControlPlaneReplicasCount {
+		klog.V(4).Infof("haven't found a replacement machine, the number of desired control plane replicas: %d must be greater than the current number of machines that host an etcd member: %d", desiredControlPlaneReplicasCount, len(memberMachines))
+		return nil
+	}
```
hasbro17, line 128: > the number of desired control plane replicas: %d must be greater than the current number of machines that host an etcd member

It's actually the opposite since we want an excess of voting member etcd machines or greater than the desired control-plane size, in order to select a replacement member to scale-down.
So we can probably make this log statement more accurate e.g:
```suggestion
	if len(memberMachines) <= desiredControlPlaneReplicasCount {
		klog.V(4).Infof("Ignoring scale-down since the number of etcd voting member machines (%d) < desired number of control-plane replicas (%d) ", len(memberMachines), desiredControlPlaneReplicasCount)
		return nil
	}
```

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -77,16 +81,88 @@ func NewClusterMemberRemovalController(
...
+}
+
+// attemptToScaleDown attempts to remove the member only once we have identified that
+// a Machine resource is being deleted and a replacement member has been created
+func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context) error {
+	masterMachines, err := c.masterMachineLister.List(c.masterMachineSelector)
+	if err != nil {
+		return err
+	}
+	memberMachines := ceohelpers.FilterMachinesWithMachineDeletionHook(masterMachines)
```
hasbro17, line 117: When deciding whether or not to scale-down a voting member this count should only be voting member machines.

Otherwise given `desiredControlPlaneReplicasCount = 3` and if we have `3 voting + 1 learner` members, then `len(memberMachines) = 4` since learner member machines also have a deletion hook.
Deleting the voting member would trigger a scale-down since  `len(memberMachines) > desiredControlPlaneReplicasCount` but now we've just replaced a voting member with a learner member (which may not get promoted).

So we also need to filter out the learner members here and change `memberMachines => votingMemberMachines` when considering scaling down a voting machine.

hasbro17: With the above you would still need to handle the cleanup of learner members pending deletion. That does not depend on `len(memberMachines) <= desiredControlPlaneReplicasCount`, but rather we remove the learner member whenever its machine is pending deletion.

I would suggest moving the learner member removal logic to its own function to make it simpler, but your call on addressing that wherever makes sense.

hasbro17 (changes requested):
Some questions/notes about how we detect an excess of voting members when deciding to scale-down.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -149,6 +226,42 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, _ factory.Syn
...
+		return memberMachinesPendingDeletion[i].DeletionTimestamp.Before(memberMachinesPendingDeletion[j].DeletionTimestamp)
+	})
+	memberMachineToDelete := memberMachinesPendingDeletion[0]
+
+	// get the members, issue a live call to get the member ID
+	members, err := c.etcdClient.MemberList(ctx)
+	if err != nil {
+		return err
+	}
+	for _, member := range members {
```
p0lyn0mial, line 246: at this point I don't think we have to check the type of the member (voting vs learner)

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
p0lyn0mial: yes, you are right, I have updated the PR, please see if you like it.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -77,16 +82,88 @@ func NewClusterMemberRemovalController(
...
+	if err != nil {
+		return err
+	}
+	var learningMachines []*machinev1beta1.Machine
+	for memberMachineIP, memberMachine := range ceohelpers.IndexMachinesByNodeInternalIP(memberMachines) {
+		if !currentVotingMemberIPListSet.Has(memberMachineIP) {
+			learningMachines = append(learningMachines, memberMachine)
+		}
+	}
+	return c.removeMemberPendingDeletion(ctx, learningMachines, "learning")
```
hasbro17, line 131: Just a general comment, not a suggestion to change :) 
There is a simpler way to tell if members are learning machines by just doing a member list and then just filtering on the `member.IsLearner` field from the response.
Although the benefit of your approach is that it doesn't require a live call to the etcd server so this is fine too.

hasbro17 (approved):
/approve

Still going through the unit tests but the member removal controller logic makes sense.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
p0lyn0mial: > Although the benefit of your approach is that it doesn't require a live call to the etcd server so this is fine too.

yeah, I wanted to avoid a live call on every interation.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -77,16 +82,88 @@ func NewClusterMemberRemovalController(
...
+	currentVotingMemberIPListSet, err := c.votingMemberIPListSet()
+	if err != nil {
+		return err
+	}
+
+	desiredControlPlaneReplicasCount, err := ceohelpers.ReadDesiredControlPlaneReplicasCount(c.configMapListerForKubeSystemNamespace)
+	if err != nil {
+		return err
+	}
+	if currentVotingMemberIPListSet.Len() <= desiredControlPlaneReplicasCount {
```
deads2k, line 146: I'm guessing you're doing this because deletionTimestamp set on a machine can mean either

1. this member should scale down now, or
2. this member should scale down after something else comes up.

However, there needs to be a better signal of this than a static installConfig that is never changed.  Where is this intent actually present.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -149,6 +226,42 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, _ factory.Syn
 	return nil
 }
 
+func (c *clusterMemberRemovalController) removeMemberPendingDeletion(ctx context.Context, memberMachines []*machinev1beta1.Machine, memberType string) error {
+	memberMachinesPendingDeletion := ceohelpers.FilterMachinesPendingDeletion(memberMachines)
```
deads2k, line 230: this method is missing in this PR, is it DeletionTimestamp != nil?

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -149,6 +226,42 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, _ factory.Syn
...
+	if err != nil {
+		return err
+	}
+	for _, member := range members {
+		memberIP, err := ceohelpers.MemberToNodeInternalIP(member)
+		if err != nil {
+			return err
+		}
+		if hasInternalIP(memberMachineToDelete, memberIP) {
+			memberLocator := fmt.Sprintf("[ url: %v, name: %v, id: %v ]", memberIP, member.Name, member.ID)
```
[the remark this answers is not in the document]
p0lyn0mial: `MemberRemove` fires an event

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
p0lyn0mial: yes, https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/ceohelpers/common.go#L93

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -109,6 +120,19 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
 		return fmt.Errorf("could not get etcd peer host :%w", err)
 	}
+
+	masterMachines, err := c.masterMachineLister.List(c.masterMachineSelector)
+	if err != nil {
+		return err
+	}
+
+	if machineForEtcdHost, hasMachine := ceohelpers.IndexMachinesByNodeInternalIP(masterMachines)[etcdHost]; hasMachine && machineForEtcdHost.DeletionTimestamp != nil {
+		klog.V(4).Infof("won't add member: %v to the cluster because its machine is pending deletion: %v", etcdHost, machineForEtcdHost.Name)
```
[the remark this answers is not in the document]
p0lyn0mial: i am okay with `v(2)`

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -77,16 +82,88 @@ func NewClusterMemberRemovalController(
...
 			kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer(),
+			kubeInformersForNamespaces.InformersFor("kube-system").Core().V1().ConfigMaps().Informer(),
 		).ToController("ClusterMemberRemovalController", eventRecorder.WithComponentSuffix("cluster-member-removal-controller"))
 }
 
 func (c *clusterMemberRemovalController) sync(ctx context.Context, _ factory.SyncContext) error {
-	// stop if the machine API is not functional
+	var errs []error
+
+	if err := c.removeMemberWithoutMachine(ctx); err != nil {
```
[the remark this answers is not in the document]
p0lyn0mial: ah, unless there won't be machine objects when the machine API is not functional

PR #770, opened 2022-03-25
turn on initial corruption check

deads2k:
Something to talk about.  I don't fully understand what https://github.com/etcd-io/etcd/blob/main/server/etcdserver/corrupt.go#L39-L121 does yet.

/hold

holding until I understand the impacts and risks.

PR #774, opened 2022-03-30
Bug 2071114: manifests/0000_90_etcd-operator_03_prometheusrule: Add etcdDivergentRevisions

wking:
Proximal trigger is the etcd 3.5.(<=2) split-brain issue ([upstream announcement][1], etcd-io/etcd#13766, [rhbz#2068601][3]). 
  But having this alert in place to look out for future issues seems useful, even once that gets fixed.  This alert supports 139ca20fc0 (#770), but:

* The initial corruption check keeps the corrupted member from coming up, preventing split-braining.  The alert just lets you know if something bad is happening; it doesn't block anything.
* The current initial corruption [check seems to ignore large divergence][4].  This alert will complain about any divergence, regardless of size.
* Adding a new alert PrometheusRule to existing clusters is an easier mitigating patch than adjusting the etcd-launching configuration.

[1]: https://etcd.io/docs/v3.5/op-guide/data_corruption/
[3]: https://bugzilla.redhat.com/show_bug.cgi?id=2068601
[4]: https://github.com/etcd-io/etcd/issues/13766#issuecomment-1083033017

### DIFF
--- a/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
+++ b/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
@@ -184,3 +184,26 @@ spec:
       for: 3m
       labels:
         severity: critical
+    - alert: etcdDivergentRevisions
+      annotations:
+        description: etcd is reporting divergent member revisions, with a pod
+          {{ $labels.pod }} on instance {{ $labels.instance }} trailing
+          {{ $value }} compacted revisions behind pod {{ $labels.max_pod }} on
+          instance {{ $labels.max_instance }}. Members may be stuck or, in
+          extreme circumstances, split brained.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/etcdDivergentRevisions.md
+        summary: etcd is reporting divergent member revisions.
+      expr: |
+        -etcd_debugging_mvcc_compact_revision{job=~".*etcd.*"}
+          + on (job, namespace) group_left (max_instance, max_pod)
+        label_replace(
+          label_replace(
+            topk by (job,namespace) (1, etcd_debugging_mvcc_compact_revision{job=~".*etcd.*"}),
+            "max_instance", "$1", "instance", "(.*)"
+          ),
+          "max_pod", "$1", "pod", "(.*)"
+        )
+          > 0
+      for: 10m
+      labels:
+        severity: critical


manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -184,3 +184,14 @@ spec:
       for: 3m
       labels:
         severity: critical
+    - alert: etcdDivergentRevisions
+      annotations:
+        description: etcd is reporting divergent member revisions,
+          with a standard deviation of {{ $value }}.  Members may be
+          stuck or, in extreme circumstances, split brained.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/FIXME
```
wking, line 192: If folks are ok with this alert, we should create a runbook for it.

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -184,3 +184,14 @@ spec:
...
+        summary: etcd is reporting divergent member revisions.
+      expr: min_over_time(stddev(max by (pod) (etcd_debugging_mvcc_compact_revision{namespace="openshift-etcd"}))[5m:15s]) > 0
```
wking, line 194: hmm `min_over_time(...) > ...` is probably redundant with `for`.  Maybe this should be:

```yaml
expr: =stddev(max by (pod) (etcd_debugging_mvcc_compact_revision{namespace="openshift-etcd"})) > 0
for: 5m
```

wking: I dropped `min_over_time` (and moved to `for: 10m`) with e209966f -> d2ca4071.

wking:
Big reroll with d2ca4071 -> 3bbdfdfc.  I've replaced `stddev` with some complicated PromQL, because alert messages like "there is `$THIS_MUCH` divergence between `$A` and `$B`" seem more actionable than "hey, there's some divergence in there somewhere".  There are some trade-offs in the PromQL, and the 3bbdfdfc commit message walks through the new `expr` and explains how the pieces fit together and the associated risks.

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -184,3 +184,26 @@ spec:
...
       labels:
         severity: critical
+    - alert: etcdDivergentRevisions
+      annotations:
+        description: etcd is reporting divergent member revisions, with a pod
+          {{ $labels.pod }} on instance {{ $labels.instance }} {{ $value }}
+          compacted revisions behind pod {{ $labels.max_pod }} on instance
+          {{ $labels.max_instance }}. Members may be stuck or, in extreme
+          circumstances, split brained.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/etcdDivergentRevisions.md
```
wking, line 194: I've opened openshift/runbooks#45 with a first pass at this runbook.

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -184,3 +184,26 @@ spec:
...
+        summary: etcd is reporting divergent member revisions.
+      expr: |
+        -etcd_debugging_mvcc_compact_revision
+          + on (job, namespace) group_left (max_instance, max_pod)
```
wking, line 198: Here's an example of the rendered message from a pending alert in a cluster-bot cluster.

image

wking: I've pushed 3bbdfdfc -> bd0a01ec to insert "trailing" between `instance` and `value`.  That should make it easier to find the value in the message; in the screenshot above^, my brain turns off when I hit the `instance` IP:port, and having a word before the `value` number would help it turn back on in time to actually read that number.

sdodson:
/retitle manifests/0000_90_etcd-operator_03_prometheusrule: Add etcdDivergentRevisions
Dropping from the urgent bug until we better understand the signal to noise ratio on these alerts

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
mfojtik: @wking is "split brain" a technical term our support and customers will understand?

wking: [Our KCS Solution][1] is saying currently saying "data inconsistency" and [the upstream etcd annoucement][2] is currently saying "data corruption".  [etcd's upstream FAQ][3] includes

> ... changes must go through quorum to avoid membership configuration split brain.

I think it's important to distinguish "some member sticks, but the quorum guards either wedge etcd (quorum loss) or allow a reduced set of healthy members continue with quorum" from "competing members continue to accept writes and serve reads, even though they don't actually hold quorum anymore".  The former is bad.  The latter is terrible.  I'm not all that picky about which words get used, as long as divergent-quorum-ness is being expressed clearly.

[1]: https://access.redhat.com/solutions/6849521
[2]: https://etcd.io/docs/v3.5/op-guide/data_corruption/
[3]: https://etcd.io/docs/v3.5/faq/#why-doesnt-changing---listen-peer-urls-or---initial-advertise-peer-urls-update-the-advertised-peer-urls-in-etcdctl-member-list

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -184,3 +184,26 @@ spec:
...
+      annotations:
+        description: etcd is reporting divergent member revisions, with a pod
+          {{ $labels.pod }} on instance {{ $labels.instance }} trailing
+          {{ $value }} compacted revisions behind pod {{ $labels.max_pod }} on
+          instance {{ $labels.max_instance }}. Members may be stuck or, in
+          extreme circumstances, split brained.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/etcdDivergentRevisions.md
+        summary: etcd is reporting divergent member revisions.
+      expr: |
+        -etcd_debugging_mvcc_compact_revision{job=~".*etcd.*"}
```
stevekuznetsov, line 197: Why is this compacted revision?

wking: From the commit message:

> * In healthy etcd clusters, the members seem to compact every 5 minutes.  There's some spread in Prometheus between the exact time when each compaction round happens.  Perhaps this is just random Prom-scrape ordering.  After 20 or 30s (approximately one default Prom scrape period), the reported compact revisions realign, and divergence within the etcd cluster drops back to zero.

For `etcd_debugging_mvcc_current_revision`, the Prom-scrape-induced latency means there's always some divergence, and you'd need to pick a threshold that depends on revision-velocity which is not as clean as the compact `> 0` threshold.

PR #775, opened 2022-04-01
clusterMemberRemovalController: run removeMemberWithoutMachine only when the machine API is on

p0lyn0mial:


p0lyn0mial:
addresses comments from https://github.com/openshift/cluster-etcd-operator/pull/768

PR #777, opened 2022-04-01
Adding corruption detection and recovery documentation

dusk125:
This documentation will hopefully aid in the detection and validation of corruption from the [Upstream issue](https://github.com/etcd-io/etcd/issues/13766).

### DIFF
--- a/docs/corruption-detection-and-recovery.md
+++ b/docs/corruption-detection-and-recovery.md
@@ -1,0 +1,106 @@
+# etcd Corruption Detection and Recovery
+
+Currently, in etcd 3.5.0-3.5.2, there is a low probability risk of data corruption across the etcd members [Upstream Issue](https://github.com/etcd-io/etcd/issues/13766), [Openshift Documentation](https://access.redhat.com/solutions/6849521). There is a mitigation coming to the Openshift Cluster etcd Operator that enables the etcd 'experimental-initial-corrupt-check' which should help detect corruption and stop the bad member from joining the cluster.
+
+This document will outline some ways to verify if an etcd member is corrupted so that you can be more confident that the corruption recovery steps are necessary and should be taken.
+
+# Corruption Diagnosis
+Unfortunately, there's no single command that will give a precise 'corrupt/not corrupt' at this time; however, if most or all of the following return that there might be corruption, then it's likely that there is corruption and recovery steps should be taken. Please note, that a single failure in the following commands does not necessarily mean that corruption has occurred.
+
+If all of the following report no issues, then you do not need to follow the recovery steps.
+
+---
+
+## Check Prometheus for etcd revision divergence
+Revision divergence is a natural happening of etcd since it takes time to sync a revision to all members; however, continuing divergence is an indication of corruption. If you have access to run Prometheus queries, you can check the following to graph the current and compact revisions to visualize any divergence.
+
+`max by (pod,job) (etcd_debugging_mvcc_current_revision)`
+
+`max by (pod, job) (etcd_debugging_mvcc_compact_revision)`
+
+The former shows instantaneous revision differences where there can be normal divergence that happens in healthy clusters and could resolve itself quickly. If you're seeing divergence here, check the compact revision as well.
+
+The latter shows a delayed history (0-5 minutes) of revision differences and can show true divergence, where the nodes are not going to resolve quickly and could put the cluster at risk to hit this corruption issue.
+
+---
+
+## Get revision differences
+One indicator that there might be corruption is a large difference in revision number. It is normal for some etcd members to lag behind the revision by a few revisions (fewer than 10); if one gets behind, it should catch up to the leader. If a member's revision is largely different from the leader/other members, then it could be a sign that divergence is starting.
+
+1. List the etcd pods
+
+`oc get pods -n openshift-etcd | grep -v etcd-quorum-guard | grep etcd`
+
+2. Log into one of the nodes (choose a healthy node).
+
+`oc rsh -n openshift-etcd [etcd-pod]`
+
+3. Show revisions, these should be all the same or very similar
+
+`etcdctl endpoint status -w fields --cluster`
+
+---
+
+## Check leader elections
+Frequent and failing leader elections are a symptom of the etcd cluster having issues.
+
+1. List the etcd pods
+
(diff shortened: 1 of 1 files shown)


docs/corruption-detection-and-recovery.md
```diff
@@ -0,0 +1,62 @@
+# etcd Corruption Detection and Recovery
+
+Currently, in etcd 3.5.0-3.5.2, there is a low probability risk of data corruption across the etcd members [Upstream Issue](https://github.com/etcd-io/etcd/issues/13766), [Openshift Documentation](https://access.redhat.com/solutions/6849521). There is a mitigation coming to the Openshift Cluster etcd Operator that enables the etcd 'experimental-initial-corrupt-check' which should help detect corruption and stop the bad member from joining the cluster.
```
deads2k, line 3: minor, newline every 100ish characters. Doesn't have to be exact, but helps with comment association.

docs/corruption-detection-and-recovery.md
```diff
@@ -0,0 +1,62 @@
...
+This document will outline some ways to verify if an etcd member is corrupted so that you can be more confident that the corruption recovery steps are necessary and should be taken.
+
+# Corruption Diagnosis
+Unfortunately, there's no single command that will give a precise 'corrupt/not corrupt' at this time; however, if most or all of the following return that there might be corruption, then it's likely that there is corruption and recovery steps should be taken. Please note, that a single failure in the following commands does not necessarily mean that corruption has occurred.
+
+## Check Prometheus for etcd revision divergence
```
deads2k, line 9: This is necessary, but not sufficient right?  In other words, if this reports ok, there is no need to progress further.

docs/corruption-detection-and-recovery.md
```diff
@@ -0,0 +1,62 @@
...
+Revision divergence is a natural happening of etcd since it takes time to sync a revision to all members; however, continuing divergence is an indication of corruption. If you have access to run Prometheus queries, you can check the following to graph the current revisions to visualize any divergence.
```
deads2k, line 10: What is the threshold of delta for deciding good or bad?  Trevor's alert may provide insight.

docs/corruption-detection-and-recovery.md
```diff
@@ -0,0 +1,62 @@
...
+
+### List the etcd pods
+`oc get pods -n openshift-etcd | grep -v etcd-quorum-guard | grep etcd`
+### Check the logs for any failures of leader elections
+`oc logs -n openshfit-etcd [etcd-pod] | grep "failed to renew lease"`
+
+---
+
+## Check the etcd pods for marshalling errors
+Marshalling errors, `failed to unmarshal mvccpb.KeyValue`, from the following command are an indication of corruption; however, no matches does not mean there isn't corruption.
```
deads2k, line 42: We've had this sort of corruption before, but for the case we're chasing now, the data that is stored is deserializable I think.  This was a different problem from the fall.

docs/corruption-detection-and-recovery.md
```diff
@@ -0,0 +1,62 @@
...
+
+`oc -n openshift-etcd get -o yaml pods | grep mvccpb.KeyValue`
+
+---
+
+## Check the output of the startup corruption check
```
deads2k, line 48: Could a step at this point be to manually modify the static pod manifests on disk to add the check?  If it fails, then you have a problem, if it doesn't fail, then you're likely ok.

docs/corruption-detection-and-recovery.md
wking: In my testing, `etcd_debugging_mvcc_current_revision` often had running divergence and rarely zeroed in active clusters.  I'm using `etcd_debugging_mvcc_compact_revision` in #774 to avoid that, at the price of zero to five minutes of additional latency depending on where the hangup occurs within the 5m compaction period.

docs/corruption-detection-and-recovery.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Check Prometheus for etcd revision divergence
+Revision divergence is a natural happening of etcd since it takes time to sync a revision to all members; however, continuing divergence is an indication of corruption. If you have access to run Prometheus queries, you can check the following to graph the current revisions to visualize any divergence.
+
+`max by (pod,job) (etcd_debugging_mvcc_current_revision)`
+
+---
+
+## Get dbSize and revision differences
+One indicator that there might be corruption is a 'significant' (greater than 10%) difference of etcd database size and revision number.
```
wking, line 17: I don't see how you'd get into this state without accumulating significant revision divergence.  It seems very unlikely to me that two members with such wildly divergent storage sizes happen to coincidentally agree on the revision number for very long.  But doesn't hurt to check this one too; maybe I'm wrong and you can get this without significant divergence.  I wouldn't want folks to delay their diagnosis just based on "the size hasn't gotten different enough yet", since that would allow a good deal of split-brain divergence to accumulate.

docs/corruption-detection-and-recovery.md
dusk125: ack, should be a little better.

docs/corruption-detection-and-recovery.md
dusk125: Added a blurb calling this out

docs/corruption-detection-and-recovery.md
dusk125: Since this isn't related, I removed this section

dusk125:
/retitle Adding corruption detection and recovery documentation

docs/corruption-detection-and-recovery.md
```diff
@@ -0,0 +1,106 @@
...
+## Check leader elections
+Frequent and failing leader elections are a symptom of the etcd cluster having issues.
+
+1. List the etcd pods
+
+`oc get pods -n openshift-etcd | grep -v etcd-quorum-guard | grep etcd`
+
+2. Check the logs for any failures of leader elections
+
+`oc logs -n openshfit-etcd [etcd-pod] | grep "failed to renew lease"`
```
iamkirkbater, line 53:
```suggestion
`oc logs -n openshift-etcd -c etcd [etcd-pod] | grep "failed to renew lease"`
```

PR #779, opened 2022-04-04
member removal part two

p0lyn0mial:


pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -35,27 +38,40 @@ type ClusterMemberController struct {
...
+		masterMachineLister:   machinelistersv1beta1.NewMachineLister(masterMachineInformer.GetIndexer()),
+		masterMachineSelector: masterMachineSelector,
 	}
 	return factory.New().ResyncEvery(time.Minute).WithInformers(
 		kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Pods().Informer(),
 		kubeInformers.InformersFor("").Core().V1().Nodes().Informer(),
 		networkInformer.Informer(),
 		operatorClient.Informer(),
+		masterMachineInformer,
 	).WithSync(c.sync).ResyncEvery(time.Minute).ToController("ClusterMemberController", eventRecorder.WithComponentSuffix("cluster-member-controller"))
```
deads2k, line 75: WithSyncDegradedOnError

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -35,27 +38,40 @@ type ClusterMemberController struct {
...
+		machineAPIChecker:     machineAPIChecker,
+		etcdClient:            etcdClient,
+		podLister:             kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Pods().Lister(),
+		nodeLister:            kubeInformers.InformersFor("").Core().V1().Nodes().Lister(),
+		networkLister:         networkInformer.Lister(),
+		masterMachineLister:   machinelistersv1beta1.NewMachineLister(masterMachineInformer.GetIndexer()),
+		masterMachineSelector: masterMachineSelector,
 	}
 	return factory.New().ResyncEvery(time.Minute).WithInformers(
 		kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Pods().Informer(),
```
deads2k, line 70: you need bareinformers for your networkInformer, right?

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -101,21 +117,51 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
+		return false, nil
+	}
+
+	masterMachines, err := c.masterMachineLister.List(c.masterMachineSelector)
+	if err != nil {
+		return false, err
+	}
+
+	// one we have functional machine API, get the corresponding machine and check DeletionTimestamp
+	if machineForEtcdHost, hasMachine := ceohelpers.IndexMachinesByNodeInternalIP(masterMachines)[etcdHost]; hasMachine && machineForEtcdHost.DeletionTimestamp != nil {
```
deads2k, line 154: from here to the end, please make a switch case so it's easier to see.

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -35,27 +38,40 @@ type ClusterMemberController struct {
...
 	networkLister  configv1listers.NetworkLister
+	// machineAPIChecker determines if the precondition for this controller is met,
+	// this controller can be run only on a cluster that exposes a functional Machine API
+	machineAPIChecker ceohelpers.MachineAPIChecker
+
+	masterMachineLister   machinelistersv1beta1.MachineLister
+	masterMachineSelector labels.Selector
 }
 
 func NewClusterMemberController(
```
deads2k, line 49: changes for this controller appear safe regardless, let's make this PR smaller by splitting these out.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -77,16 +82,83 @@ func NewClusterMemberRemovalController(
...
-	// stop if the machine API is not functional
+	// only attempt to scale down if the machine API is functional
 	if isFunctional, err := c.machineAPIChecker.IsFunctional(); err != nil {
 		return err
 	} else if !isFunctional {
 		return nil
 	}
+
+	var errs []error
+	if err := c.removeMemberWithoutMachine(ctx); err != nil {
```
deads2k, line 98: this refactor could be made early to reduce the diff while we argue over the rest.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -149,6 +221,42 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, _ factory.Syn
...
+	memberMachinesPendingDeletion := ceohelpers.FilterMachinesPendingDeletion(memberMachines)
+	if len(memberMachinesPendingDeletion) == 0 {
+		return nil
+	}
+
+	// sort by deletion time and pick up one machine for removal and record it
+	sort.Slice(memberMachinesPendingDeletion, func(i, j int) bool {
+		return memberMachinesPendingDeletion[i].DeletionTimestamp.Before(memberMachinesPendingDeletion[j].DeletionTimestamp)
+	})
+	memberMachineToDelete := memberMachinesPendingDeletion[0]
```
deads2k, line 234: I expected to iterate over the list of member machines, returning after removing a single member.  otherwise, a single stuck member will prevent any progress.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -77,16 +82,83 @@ func NewClusterMemberRemovalController(
...
+		errs = append(errs, err)
+	}
+	if err := c.attemptToRemoveLearningMember(ctx); err != nil {
+		errs = append(errs, err)
+	}
+	return kerrors.NewAggregate(errs)
+}
+
+// attemptToRemoveLearningMember attempts to remove a learning member pending deletion regardless of whether a replacement member has been found
+func (c *clusterMemberRemovalController) attemptToRemoveLearningMember(ctx context.Context) error {
```
deads2k, line 111: this method and usage can be extracted to a different PR while we argue over the rest.

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,219 @@
...
+		etcdClient:                            etcdClient,
+		machineAPIChecker:                     machineAPIChecker,
+		masterMachineSelector:                 masterMachineSelector,
+		masterMachineLister:                   machinelistersv1beta1.NewMachineLister(masterMachineInformer.GetIndexer()),
+		configMapListerForKubeSystemNamespace: kubeInformersForNamespaces.InformersFor("kube-system").Core().V1().ConfigMaps().Lister().ConfigMaps("kube-system"),
+	}
+	return factory.New().
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
+		WithInformers(
```
deads2k, line 62: bareinformers for the machinelister?

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,219 @@
...
+			continue
+		}
+		masterMachineCopy := masterMachine.DeepCopy()
+		if hasPartialMachineDeletionHook(masterMachineCopy) {
+			repairPartialMachineDeletionHook(masterMachineCopy)
+		} else {
+			addPreDrainHook(masterMachineCopy)
+		}
+		if _, err := c.machineClient.Update(ctx, masterMachineCopy, metav1.UpdateOptions{}); err != nil {
+			return err
```
deads2k, line 156: aggregate the errors from all so that you can make partial progress and still fail.

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,219 @@
...
+	for _, hook := range machine.Spec.LifecycleHooks.PreDrain {
+		if hook.Name == ceohelpers.MachineDeletionHookName {
+			continue
+		}
+		newPreDrainHooks = append(newPreDrainHooks, hook)
+	}
+	machine.Spec.LifecycleHooks.PreDrain = newPreDrainHooks
+}
+
+func indexMachinesByNodeInternalIP(machines []*machinev1beta1.Machine) map[string]*machinev1beta1.Machine {
```
deads2k, line 208: how is this different than IndexMachinesByNodeInternalIP from the helpers?

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,219 @@
...
+		return err
+	} else if !isFunctional {
+		return nil
+	}
+	var errs []error
+	if err := c.addDeletionHookToMasterMachines(ctx); err != nil {
+		errs = append(errs, err)
+	}
+	// attempt to remove hooks from machines pending deletion and without running etcd member
+	if err := c.attemptToDeleteMachineDeletionHook(ctx, syncCtx.Recorder()); err != nil {
```
deads2k, line 82: Why do we have the same controller doing the add and the remove?

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,219 @@
...
+			// Ignore machines that still have a member present
+			continue
+		}
+		machinePendingDeletionCopy := machinePendingDeletion.DeepCopy()
+		removeMachineDeletionHook(machinePendingDeletionCopy)
+		if _, err := c.machineClient.Update(ctx, machinePendingDeletionCopy, metav1.UpdateOptions{}); err == nil {
+			klog.V(2).Infof("successfully removed the deletion hook from machine %v", machinePendingDeletion.Name)
+			recorder.Eventf("MachineDeletionHook", "successfully removed the deletion hook from machine %v as it doesn't have a member", machinePendingDeletion.Name)
+		}
+		return err
```
deads2k, line 131: it's been a while, but I think that the `err` from your `if` block is contained to the `if` block, so this won't have the correct `err` in it.

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -0,0 +1,219 @@
...
+
+	// find a machine that doesn't have a member
+	for machinePendingDeletionIP, machinePendingDeletion := range indexMachinesByNodeInternalIP(machinesPendingDeletion) {
+		if currentMemberIPListSet.Has(machinePendingDeletionIP) {
+			// Ignore machines that still have a member present
+			continue
+		}
+		machinePendingDeletionCopy := machinePendingDeletion.DeepCopy()
+		removeMachineDeletionHook(machinePendingDeletionCopy)
+		if _, err := c.machineClient.Update(ctx, machinePendingDeletionCopy, metav1.UpdateOptions{}); err == nil {
```
deads2k, line 127: let's do the standard `if err !=nil`

deads2k, line 127: Aggregate these errors so you can get as right as you can if something is wonky with the API.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -77,16 +82,83 @@ func NewClusterMemberRemovalController(
...
+
+// attemptToScaleDown attempts to remove a voting member only once we have identified that
+// a Machine resource is being deleted and a replacement member has been created
+func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context) error {
+	currentVotingMemberIPListSet, err := c.votingMemberIPListSet()
+	if err != nil {
+		return err
+	}
+
+	desiredControlPlaneReplicasCount, err := ceohelpers.ReadDesiredControlPlaneReplicasCount(c.configMapListerForKubeSystemNamespace)
```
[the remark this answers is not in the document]
p0lyn0mial: /cc @JoelSpeed I think in the KEP we wanted to get this info from the `cluster-config-v1` configmap in `kube-system` namespace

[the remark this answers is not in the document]
p0lyn0mial: > We need a different signal for this. Install-config is write-once data. It cannot be modified via normal means, lacks a schema, lacks validation, and should not be used to feed post-install operator behavior.

does OCP support changing the number of nodes post installation?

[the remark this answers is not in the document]
JoelSpeed: > We need a different signal for this. Install-config is write-once data. It cannot be modified via normal means, lacks a schema, lacks validation, and should not be used to feed post-install operator behavior.

The etcd operator has been relying on this value to know how many replicas it thinks it should have all the way through OpenShift 4. This is the only place in the cluster today where the value of the control plane size actually resides at present.

> It would be best to get this from something like the control plane machine set resource. But even something written into the etcd.operator.openshift.io schema could work. Add it to spec, then observe it from there.

This was written down somewhere in an enhancement, the plan eventually is that the ControlPlaneMachineSet becomes the source of truth for the replica count, but, this feature is going to be delivered before ControlPlaneMachineSet (which is looking likely to miss 4.11), so the plan was to fall back to the old mechanism until CPMS becomes available

> I think in the KEP we wanted to get this info from the cluster-config-v1 configmap in kube-system namespace

Yep, this was discussed with Sam and Lili before they left, it's the only source of the control plane size that we have right now, in the future that will change but we do need to be able to handle upgrades and clusters that won't have CPMS ever 

> does OCP support changing the number of nodes post installation?

Today, no. All customers are required to have 3 replicas in their control plane _unless_ they are granted a support exception, in which case they are allowed to have 5, but that is from install time only. In the future, we want to look into horizontal scaling of the control plane, but this is an explicit non-goal of the first pass of the etcd protection and CPMS projects

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
JoelSpeed: This is correct, better to reverse the condition if you want to capture that `err` inline in the if

pkg/operator/clustermembercontroller/clustermembercontroller.go
p0lyn0mial: done, https://github.com/openshift/cluster-etcd-operator/pull/780

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
p0lyn0mial: thanks for the input.

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
p0lyn0mial: why? we do want to react to changes in `masterMachineInformer`

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
p0lyn0mial: cruft, will reuse once https://github.com/openshift/cluster-etcd-operator/pull/780 merges

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
p0lyn0mial: yes, thanks

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
p0lyn0mial: because this is a controller that reconciles (add, remove) the machine hooks

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
JoelSpeed: I would expect this operator to react to events from master machines so that if a user deletes the hook, it can re-add it as per the requirements (ie re-add it when the machine is not yet deleted)

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
p0lyn0mial: In that scenario, we might remove multiple members, possibly losing quorum. Stable input prevents that. The number of current voting is calculated from the cache. Does it make sense?

pkg/operator/ceohelpers/common.go
```diff
@@ -21,43 +24,34 @@ const MachineDeletionHookName = "EtcdQuorumOperator"
...
+		operatorSpec.UnsupportedConfigOverrides.Raw,
+	)
+	if err != nil {
+		return 0, err
 	}
 
-	var unstructuredInstallConfig map[string]interface{}
-	if err := yaml.Unmarshal([]byte(rawInstallConfig), &unstructuredInstallConfig); err != nil {
-		return 0, fmt.Errorf("failed to unmarshal key: %s to yaml from cm: %s/kube-system, err: %w", installConfigKeyName, clusterConfigConfigMapName, err)
+	var unstructuredConfig map[string]interface{}
```
p0lyn0mial, line 43: as of today, the etcd config is unstructured, in the future we could make it more structured.

pkg/operator/configobservation/controlplanereplicascount/observe_control_plane_replicas_count.go
```diff
@@ -0,0 +1,90 @@
...
+	// read the observed value
+	observedControlPlaneReplicasCountInt, err := readDesiredControlPlaneReplicasCount(genericListers.(configobservation.Listers).ConfigMapListerForKubeSystemNamespace)
+	if err != nil {
+		// we got an error, so we are unable to determine current replicas count
+		return existingConfig, append(errs, err)
+	}
+	observedControlPlaneReplicasCount := strconv.Itoa(observedControlPlaneReplicasCountInt)
+
+	// read the current value
+	currentControlPlaneReplicasCount, _, err := unstructured.NestedString(existingConfig, ControlPlaneReplicasCountPath...)
```
p0lyn0mial, line 34: I'd like to see if we could store it as a `number`

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -97,10 +101,85 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, _ factory.Syn
...
+	if err != nil {
+		return err
+	}
+
+	desiredControlPlaneReplicasCount, err := ceohelpers.ReadDesiredControlPlaneReplicasCount(c.operatorClient)
+	if err != nil {
+		return err
+	}
+	if desiredControlPlaneReplicasCount == 0 {
+		return fmt.Errorf("desired control plane replicas count cannot be empty")
```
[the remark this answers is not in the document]
deads2k: > it cannot be 0, right?

if it has not yet been observed, why can't it be zero?  Error-ing until it is set doesn't bother me right off.  Maybe if it's a practical problem.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -97,10 +101,85 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, _ factory.Syn
...
+	}
+	if currentVotingMemberIPListSet.Len() <= desiredControlPlaneReplicasCount {
```
deads2k, line 126: how will you handle the case on metal where the cluster first needs to be scaled down and then is scaled back up?

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -97,10 +101,85 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, _ factory.Syn
...
+	for _, member := range members {
+		if member.IsLearner {
+			continue
+		}
+		liveVotingMembers = append(liveVotingMembers, member)
+	}
+
+	// do not trust data in the cache, compare with the current state
+	if len(liveVotingMembers) <= desiredControlPlaneReplicasCount {
+		klog.V(2).Infof("Ignoring scale-down since the number of live etcd voting members (%d) < desired number of control-plane replicas (%d) ", len(liveVotingMembers), desiredControlPlaneReplicasCount)
```
deads2k, line 165: yeah, I think before this effort is done, you'll need to create way to "write this event if it has been more than five minutes since I last wrote it"

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -316,6 +387,28 @@ func (c *clusterMemberRemovalController) currentMemberMachines() ([]*machinev1be
...
+			errs = append(errs, fmt.Errorf("failed to get an IP for member: %v, err: %v", memberLocator, err))
+			continue // ignore unhealthy members
+		}
+		if hasInternalIP(machinePendingDeletion, memberIP) {
+			memberLocator := fmt.Sprintf("[ url: %v, name: %v, id: %v ]", memberIP, member.Name, member.ID)
+			if err := c.etcdClient.MemberRemove(ctx, member.ID); err != nil {
+				errs = append(errs, fmt.Errorf("failed to remove member: %v, err: %v", memberLocator, err))
+				break // stop, we have found a matching member for the provided machine
+			}
+			klog.V(2).Infof("successfully removed member: %v from the cluster", memberLocator)
```
deads2k, line 404: event please.

pkg/operator/configobservation/controlplanereplicascount/observe_control_plane_replicas_count.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/ghodss/yaml"
+
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/configobservation"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+var ControlPlaneReplicasCountPath = []string{"controlPlaneReplicasCount"}
```
deads2k, line 14: naw, nest under something.  You know this won't be right.

pkg/operator/configobservation/controlplanereplicascount/observe_control_plane_replicas_count.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		// we got an error, so we are unable to determine current replicas count
+		return existingConfig, append(errs, err)
+	}
+	// since unmarshalling JSON into an interface value always stores JSON number as a float64 convert the observed value
+	observedControlPlaneReplicasCount := float64(observedControlPlaneReplicasCountInt)
+
+	// read the current value
+	currentControlPlaneReplicasCount, _, err := unstructured.NestedFloat64(existingConfig, ControlPlaneReplicasCountPath...)
+	if err != nil {
```
deads2k, line 34: if this path doesn't exist, we don't get an error?  It's been a while, but I think you'll get an error on traversal errors and that isn't the same as "not a float".

pkg/operator/configobservation/controlplanereplicascount/observe_control_plane_replicas_count.go
```diff
@@ -0,0 +1,89 @@
...
+	}()
+
+	// read the observed value
+	observedControlPlaneReplicasCountInt, err := readDesiredControlPlaneReplicasCount(genericListers.(configobservation.Listers).ConfigMapListerForKubeSystemNamespace)
+	if err != nil {
+		// we got an error, so we are unable to determine current replicas count
+		return existingConfig, append(errs, err)
+	}
+	// since unmarshalling JSON into an interface value always stores JSON number as a float64 convert the observed value
+	observedControlPlaneReplicasCount := float64(observedControlPlaneReplicasCountInt)
```
deads2k, line 30: once we have an observed value, why not just write it and return?

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
p0lyn0mial: `MemberRemove` method produces an event, here https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/etcdcli/etcdcli.go#L265

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
p0lyn0mial: sure, initially it can be `0`. But in general, it is safe to assume that eventually it will be filled with something, right?

pkg/operator/configobservation/controlplanereplicascount/observe_control_plane_replicas_count.go
p0lyn0mial: ah, you are right `NestedFloat64` returns `(float64, bool, error)`

p0lyn0mial: ah, sorry, this is actually correct, the current config is not required to have the key. This can be the very first time we need to set a value.

pkg/operator/configobservation/controlplanereplicascount/observe_control_plane_replicas_count.go
JoelSpeed: Nit: Elsewhere we just use `replicas`, so maybe `controlPlane`: `replicas`? Not sure count adds much does it?

pkg/operator/configobservation/controlplanereplicascount/observe_control_plane_replicas_count.go
p0lyn0mial: right, we could simplify.

p0lyn0mial:
I have updated the PR, PTAL

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
deads2k: > But in general, it is safe to assume that eventually it will be filled with something, right?

I think so

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
deads2k: > In short: When a Machine is marked for deletion, if the hook is removed, the etcd operator will not replace it. This will allow users to override the mechanism and continue with their replacement in these scenarios.

and you expect these external parties to handle an etcd member removal?  I don't think that's reasonable.

pkg/operator/configobservation/controlplanereplicascount/observe_control_plane_replicas_count.go
```diff
@@ -0,0 +1,77 @@
...
+
+	unstructuredControlPlaneConfig, exists, err := unstructured.NestedMap(unstructuredInstallConfig, "controlPlane")
+	if err != nil {
+		return 0, fmt.Errorf("failed to extract field: %s.controlPlane from cm: %s/kube-system, err: %v", installConfigKeyName, clusterConfigConfigMapName, err)
+	}
+	if !exists {
+		return 0, fmt.Errorf("required field: %s.controlPlane doesn't exist in cm: %s/kube-system", installConfigKeyName, clusterConfigConfigMapName)
+	}
+	// unmarshalling JSON into an interface value always stores JSON number as a float64
+	desiredReplicas, exists, err := unstructured.NestedFloat64(unstructuredControlPlaneConfig, "replicas")
```
deads2k, line 68: why not just do NestedFloat64(unstructuredInstallConfig, "controlPlane", "replicas")

deads2k:
couple more comments and squash, then lgtm.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
p0lyn0mial: > the etcd operator should be able to detect that the Machine object has been removed and remove the member in these scenarios

implemented in https://github.com/openshift/cluster-etcd-operator/commit/6b8f7c8185adfe08618c7e0ca86adf80803a7994

JoelSpeed: Thanks for that link @p0lyn0mial , so my memory was correct on the plan then, question is, is @deads2k happy with the UX I've described in my previous comment

pkg/operator/configobservation/controlplanereplicascount/observe_control_plane_replicas_count.go
p0lyn0mial: updated.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
deads2k: I suspect that will end up bumpy on etcd consumers.  I'm willing to merge this to get started, but I suspect we'll have an issue.

deads2k:
/hold

seeing reverts open against this repo

p0lyn0mial:
`periodic-ci-openshift-release-master-nightly-4.11-e2e-metal-ipi-serial-ipv4` [green](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-cluster-etcd-operator-779-nightly-4.11-e2e-metal-ipi-serial-ipv4/1515618294126612480)

`periodic-ci-openshift-release-master-nightly-4.11-e2e-vsphere-serial` [green](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-cluster-etcd-operator-779-nightly-4.11-e2e-vsphere-serial/1515618688324079616)

`e2e-metal-ipi-ovn-ipv6` keeps failing on `when FIPS is disabled the HAProxy router should serve routes when configured with a 1024-bit RSA key` which doesn't matter from the perspective of this repo

/hold cancel

PR #780, opened 2022-04-05
ClusterMemberController adds members whose machines are not pending deletion

p0lyn0mial:


pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -35,28 +38,42 @@ type ClusterMemberController struct {
...
+		WithBareInformers(networkInformer.Informer()).
+		WithInformers(
+			kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().Pods().Informer(),
+			kubeInformers.InformersFor("").Core().V1().Nodes().Informer(),
+			operatorClient.Informer(),
+			masterMachineInformer,
+		).WithSync(c.sync).WithSyncDegradedOnError(operatorClient).ResyncEvery(time.Minute).ToController("ClusterMemberController", eventRecorder.WithComponentSuffix("cluster-member-controller"))
 }
 
 func (c *ClusterMemberController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
deads2k, line 79: don't set the condition in this message, just return the error from reconcileMembers

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -101,21 +118,52 @@ func (c *ClusterMemberController) reconcileMembers(ctx context.Context, recorder
...
-	default:
-		recorder.Eventf("FoundPodToScale", "found pod to add to etcd membership: %v", podToAdd.Name)
 	}
 
 	etcdHost, err := c.getEtcdPeerHostToScale(podToAdd)
 	if err != nil {
 		return fmt.Errorf("could not get etcd peer host :%w", err)
 	}
+
+	if machineForHostPendingDeletion, err := c.isMachinePendingDeletionFor(etcdHost); err != nil || machineForHostPendingDeletion {
```
deads2k, line 128: my personal preference is to have an if/else-if with a return err and a return nil, but I'm willing to defer to @hasbro17

JoelSpeed (commented):
/lgtm

pkg/operator/clustermembercontroller/clustermembercontroller.go
p0lyn0mial: oh, right

p0lyn0mial: I think that we don't have to clean up the old condition because the new ones will have the same name

p0lyn0mial: yeah,
```
// WithSyncDegradedOnError encapsulate the controller sync() function, so when this function return an error, the operator client
// is used to set the degraded condition to (eg. "ControllerFooDegraded"). The degraded condition name is set based on the controller name.
```

PR #783, opened 2022-04-06
no pending on etcdHighNumberOfLeaderChanges 1st 1h

jluhrsen:
during initial cluster roll-out there is enough churn that etcd
leadership will change and these alerts would result. This can
result of normal cluster installation and the alerts could cause
false concern of some problem.

this is happening more frquently in CI with OVN as the CNI,
especially in the GCP cloud (for whatever reason) and results
in extra failures in CI.

This should wait for an hour after "cluster_version" goes
initial and should provide enough time before the alert would
be much more meaningful if it came.

Signed-off-by: Jamo Luhrsen <jluhrsen@gmail.com>

### DIFF
--- a/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
+++ b/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
@@ -163,7 +163,11 @@ spec:
           should be investigated.'
         summary: etcd cluster has high number of leader changes.
       expr: |
-        increase((max without (instance) (etcd_server_leader_changes_seen_total{job=~".*etcd.*"}) or 0*absent(etcd_server_leader_changes_seen_total{job=~".*etcd.*"}))[15m:1m]) >= 5
+         (increase((max without (instance) (etcd_server_leader_changes_seen_total{job=~".*etcd.*"})
+         or
+           0*absent(etcd_server_leader_changes_seen_total{job=~".*etcd.*"}))[15m:1m]) >= 5)
+         and
+           on () time() - cluster_version{type="initial"} > 3600
       for: 5m
       labels:
         severity: warning


wking:
This is sort of halfway to [my rhbz#2010989 suggestion][1]:

> Maybe the static pod controller should grow a new metric for config revision, and the `etcdHighNumberOfLeaderChanges` `expr` could be updated to say "when the leader churn is higher than what I'd expect given the revision churn"?

The initial-1h approach in this PR only accounts for install-time revision churn.  We know there is install-time revision churn, but there could also post-install revision churn that would lead to `etcdHighNumberOfLeaderChanges` false-alarms (it is detecting leader changes, they're just not unexpected leader changes that the admin needs to be notified about).  And guessing at an hour will cover most installs, but it's theoretically possible to have an install stretch beyond that.  But the initial-1h approach is something we can do now, in the absence of a static-pod-revision metric, so :+1: from me on landing this as a useful improvement step on the endless road to perfection ;).

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2010989#c8

wking:
[`No CNI configuration file` event spew from Loki][1] and [`KubeAggregatedAPIDown` firing][2] seem unrelated: 

/retest-required

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/783/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-serial/1513595798070235136
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/783/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-single-node/1513595798133149696

PR #785, opened 2022-04-12
Bug 2074544 : Fix ipv6 regression in clustermembercontroller

hardys:
Since #780 ipv6 deployments are broken with an error like:

ClusterMemberControllerDegraded: unable to find machine for member: [fd2e:6f44:5dd8:c956::16]

Machine/Node IPs don't include the port, therefore are unquoted, so
ceohelpers.IndexMachinesByNodeInternalIP gives a map with keys that
do not include brackets.

It seems getEtcdPeerHostToScale can return the unquoted IP which
should reolve this issue, then we can re-add the brackets for the call to `etcdClient.MemberAdd` - I added a test to verify this.

PR #786, opened 2022-04-12
Bug 2074544 : revert #784 and #780 to fix ipv6

hardys:
I tried to fix the issues in #785 but it seems those changes aren't sufficient.

In ipv6 tests (locally and CI) we're seeing installation perma-fail with errors like:

```
ClusterMemberControllerDegraded: unable to find machine for member: [fd2e:6f44:5dd8:c956::16]
```

It seems there are multiple issues - we need to fix the quoting so we're comparing `ip` with `ip` and not `[ip]` (which was handled via #785) but it seems we also need to handle the case where Machine resources may have multiple InternalIP fields in their status (particularly for baremetal this is a common case) - this second issue is not yet resolved so we should revert pending that rework.


hardys:
Ok the ipv6 job failed some tests, but the cluster did deploy:

```
2022-04-13 08:29:32 level=info msg=Install complete!
2022-04-13 08:29:32 level=info msg=To access the cluster as the system:admin user when using 'oc', run 'export KUBECONFIG=/root/dev-scripts/ocp/ostest/auth/kubeconfig'
2022-04-13 08:29:32 level=info msg=Access the OpenShift web-console here: https://console-openshift-console.apps.ostest.test.metalkube.org
```

hardys:
@stbenjam I was thinking I'd close this PR since it seems https://github.com/openshift/cluster-etcd-operator/pull/785 did get past installation in the latest CI run - there are some test failures though which I'm hoping were flakes unrelated to the etcd changes

hardys:
/retitle Bug 2074544 : revert #784 and #780 to fix ipv6

deads2k:
When something breaks the payload, we revert. We do not quibble about a potential fix and "maybe we don't need to revert".  We **always** need to revert and the fix can unrevert to get back in.

deads2k:
This built, installed, and upgraded.  Merging over the rest of the red.

PR #788, opened 2022-04-13
temporarily disables check for machine for the member controller as it fails on ipv6

p0lyn0mial:
as an alternative for https://github.com/openshift/cluster-etcd-operator/pull/786 so that we don't have to revert multiple PRs


hasbro17:
/skip ci/prow/e2e-gcp-five-control-plane-replicas

Optional and broken right now (need to disable that altogether sometime soon).

deads2k:
When something breaks the payload, we revert. We do not quibble about a potential fix and "maybe we don't need to revert".  We **always** need to revert and the fix can unrevert to get back in.

p0lyn0mial:
/close

the revert got first.

PR #798, opened 2022-04-22
Bug 2079724: manually disable defrag

tjungblu:
adding a flag configmap to stop defragmentation

### DIFF
--- a/pkg/operator/defragcontroller/defragcontroller.go
+++ b/pkg/operator/defragcontroller/defragcontroller.go
@@ -7,6 +7,10 @@ import (
 	"math"
 	"time"
 
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
+
+	k8serror "k8s.io/apimachinery/pkg/api/errors"
+
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
@@ -16,6 +20,7 @@ import (
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"k8s.io/apimachinery/pkg/util/wait"
+	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
@@ -30,7 +35,8 @@ const (
 	pollTimeoutDuration             = 45 * time.Second
 	compactionInterval              = 10 * time.Minute
 
-	defragDisabledCondition = "DefragControllerDisabled"
+	defragDisabledCondition    = "DefragControllerDisabled"
+	defragDisableConfigmapName = "etcd-disable-defrag"
 )
 
 // DefragController observes the etcd state file fragmentation via Status method of Maintenance API. Based on these
@@ -39,6 +45,9 @@ type DefragController struct {
 	operatorClient       v1helpers.OperatorClient
 	etcdClient           etcdcli.EtcdClient
 	infrastructureLister configv1listers.InfrastructureLister
+	configmapLister      corev1listers.ConfigMapLister
+
+	defragWaitDuration time.Duration
 }
 
 func NewDefragController(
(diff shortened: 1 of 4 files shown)


Elbehery (commented):
all looks good, great work @tjungblu  👍🏽

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -82,29 +93,29 @@ func (c *DefragController) sync(ctx context.Context, syncCtx factory.SyncContext
 }
 
 func (c *DefragController) checkDefrag(ctx context.Context, recorder events.Recorder) error {
-	// Check for existing status.
-	_, status, _, err := c.operatorClient.GetOperatorState()
-	if err != nil {
-		return err
+	disableConfigMap, err := c.configmapLister.ConfigMaps(operatorclient.OperatorNamespace).Get(defragDisableConfigmapName)
+	if err != nil && !k8serror.IsNotFound(err) {
```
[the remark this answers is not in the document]
tjungblu: all good, this can maybe be written a bit less confusing. 

When getting a config map that doesn't 
exist we get a `disableConfigMap = nil, err = k8serror.NotFound`. So that only ensures that we return errors that are happening when the configmap is actually there (connection errors for example).

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -4,9 +4,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
+	corev1listers "k8s.io/client-go/listers/core/v1"
```
hasbro17, line 8: Nit:
Move these imports down into the non-stdlib groupings below.

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -197,31 +208,26 @@ func (c *DefragController) checkDefrag(ctx context.Context, recorder events.Reco
...
-		}
-	} else {
+func (c *DefragController) ensureControllerDisabledCondition(ctx context.Context, desiredStatus operatorv1.ConditionStatus, recorder events.Recorder) error {
+	_, currentState, _, err := c.operatorClient.GetOperatorState()
+	if err != nil {
+		return err
+	}
+
+	controllerDisabledCondition := v1helpers.FindOperatorCondition(currentState.Conditions, defragDisabledCondition)
+	if controllerDisabledCondition == nil || controllerDisabledCondition.Status != desiredStatus {
```
hasbro17, line 218: I think `v1helpers.UpdateStatus()` will already check if `current status == desired status` before updating so we could just call that directly without this check.

https://github.com/openshift/cluster-etcd-operator/blob/c7ed7ef9c4252feb588db290f84ae63577626dc7/vendor/github.com/openshift/library-go/pkg/operator/v1helpers/helpers.go#L163-L179

pkg/operator/defragcontroller/defragcontroller_test.go
```diff
@@ -3,9 +3,12 @@ package defragcontroller
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
+	corev1listers "k8s.io/client-go/listers/core/v1"
```
hasbro17, line 7: Same as above for grouping like imports together.

hasbro17 (commented):
SGTM overall as a temporary fix.

pkg/operator/defragcontroller/defragcontroller.go
tjungblu: very nice, thanks for telling me. fixed

tjungblu: until I ran the unit tests :) apparently the mocked version does not have the same check so the unit tests were failing. Shall we keep it for now? I'm too lazy to get this patched in library-go

pkg/operator/defragcontroller/defragcontroller_test.go
```diff
@@ -3,9 +3,11 @@ package defragcontroller
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
```
hasbro17, line 6: Nit: Import grouping fix.

hasbro17 (commented):
/approve

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -4,9 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
```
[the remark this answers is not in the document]
tjungblu: apologies, I need to get my goland fixed. That used to work before the update :/ 
will send you another PR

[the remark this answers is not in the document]
tjungblu: fixed

pkg/operator/defragcontroller/defragcontroller_test.go
tjungblu: fixed

Elbehery:
@tjungblu  we need to override `e2e-gcp-five-control-plane-replicas` job 

cc @hasbro17 @deads2k

PR #800, opened 2022-04-25
Bug 2077833: add more logging

tjungblu:


dgoodwin:
No need for running those expensive payload tests on the PR, once this merges we can find you hits in the multitude of CI runs, shouldn't take long as this one looked pretty common.

Elbehery:
@hasbro17  we need to fix this test `e2e-gcp-five-control-plane-replicas` .. it is blocking everyone, any ideas ? 

/lgtm

hasbro17:
@Elbehery That job is optional so it shouldn't be blocking anything. The tide message tells you what's blocking merge:
```
Not mergeable. Needs approved label.
```

Although we really should disable it since it's a waste to run it until we can try to figure out what's wrong with it.

hasbro17:
/approved

@tjungblu Don't think this will fix https://bugzilla.redhat.com/show_bug.cgi?id=2077833 but hopefully it sheds some light on why the members are unhealthy during those upgrades.

PR #801, opened 2022-04-26
fix races in etcdclient

tjungblu:
there is a bug while caching the client, the order in which the endpoints are returned is not deterministic. That causes the cache client to be constantly created from scratch and invalidating the old (potentially while in use). 

the new client works in a pooling fashion, which handles the mutual exclusiveness of returning an existing or newly created client. HealthCheck+Member updates are handled during the retrieval, so is the close. Once a consumer is done, they simply return the client to the pool. 

added some more additional logging as well.

### DIFF
--- a/pkg/etcdcli/etcdcli.go
+++ b/pkg/etcdcli/etcdcli.go
@@ -7,9 +7,7 @@ import (
 	"net"
 	"net/url"
 	"os"
-	"reflect"
 	"strings"
-	"sync"
 	"time"
 
 	grpcprom "github.com/grpc-ecosystem/go-grpc-prometheus"
@@ -28,7 +26,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
@@ -54,13 +51,11 @@ type etcdClientGetter struct {
 
 	eventRecorder events.Recorder
 
-	clientLock          sync.Mutex
-	lastClientConfigKey []string
-	cachedClient        *clientv3.Client
+	clientPool *EtcdClientPool
 }
 
 func NewEtcdClient(kubeInformers v1helpers.KubeInformersForNamespaces, networkInformer configv1informers.NetworkInformer, eventRecorder events.Recorder) EtcdClient {
-	return &etcdClientGetter{
+	g := &etcdClientGetter{
 		nodeLister:             kubeInformers.InformersFor("").Core().V1().Nodes().Lister(),
 		configmapsLister:       kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Lister(),
 		networkLister:          networkInformer.Lister(),
@@ -69,86 +64,67 @@ func NewEtcdClient(kubeInformers v1helpers.KubeInformersForNamespaces, networkIn
 		networkListerSynced:    networkInformer.Informer().HasSynced,
 		eventRecorder:          eventRecorder.WithComponentSuffix("etcd-client"),
 	}
-}
 
-// getEtcdClientWithClientOpts allows customization of the etcd client using ClientOptions. All clients must be manually
-// closed by the caller with Close().
-func (g *etcdClientGetter) getEtcdClientWithClientOpts(endpoints []string, opts ...ClientOption) (*clientv3.Client, error) {
-	return getEtcdClientWithClientOpts(endpoints, false, opts...)
-}
+	endpointFunc := func() ([]string, error) {
+		if !g.nodeListerSynced() {
+			return nil, fmt.Errorf("node lister not synced")
+		}
+		if !g.configmapsListerSynced() {
+			return nil, fmt.Errorf("configmaps lister not synced")
+		}
+		if !g.networkListerSynced() {
+			return nil, fmt.Errorf("network lister not synced")
+		}
 
-// getEtcdClient may return a cached client.  When a new client is needed, the previous client is closed.
-// The caller should not close the client or future calls may fail.
-func (g *etcdClientGetter) getEtcdClient() (*clientv3.Client, error) {
-	if !g.nodeListerSynced() {
-		return nil, fmt.Errorf("node lister not synced")
-	}
-	if !g.configmapsListerSynced() {
-		return nil, fmt.Errorf("configmaps lister not synced")
-	}
-	if !g.networkListerSynced() {
-		return nil, fmt.Errorf("network lister not synced")
-	}
-
-	network, err := g.networkLister.Get("cluster")
-	if err != nil {
-		return nil, fmt.Errorf("failed to list cluster network: %w", err)
-	}
-
-	etcdEndpoints := []string{}
-	nodes, err := g.nodeLister.List(labels.Set{"node-role.kubernetes.io/master": ""}.AsSelector())
-	for _, node := range nodes {
-		internalIP, err := dnshelpers.GetEscapedPreferredInternalIPAddressForNodeName(network, node)
+		network, err := g.networkLister.Get("cluster")
 		if err != nil {
-			return nil, fmt.Errorf("failed to get internal IP for node: %w", err)
+			return nil, fmt.Errorf("failed to list cluster network: %w", err)
 		}
-		etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", internalIP))
+
+		var etcdEndpoints []string
+		nodes, err := g.nodeLister.List(labels.Set{"node-role.kubernetes.io/master": ""}.AsSelector())
+		for _, node := range nodes {
+			internalIP, err := dnshelpers.GetEscapedPreferredInternalIPAddressForNodeName(network, node)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get internal IP for node: %w", err)
+			}
+			etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", internalIP))
+		}
+
+		configmap, err := g.configmapsLister.ConfigMaps(operatorclient.TargetNamespace).Get("etcd-endpoints")
+		if err != nil {
+			return nil, fmt.Errorf("failed to list endpoints from %s/%s: %w",
+				operatorclient.TargetNamespace, "etcd-endpoints", err)
+		}
+		if bootstrapIP, ok := configmap.Annotations[BootstrapIPAnnotationKey]; ok && bootstrapIP != "" {
+			// escape if IPv6
+			if net.ParseIP(bootstrapIP).To4() == nil {
+				bootstrapIP = "[" + bootstrapIP + "]"
+			}
+			etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", bootstrapIP))
+		}
+		return etcdEndpoints, nil
 	}
 
-	configmap, err := g.configmapsLister.ConfigMaps(operatorclient.TargetNamespace).Get("etcd-endpoints")
-	if err != nil {
-		return nil, fmt.Errorf("failed to list endpoints: %w", err)
-	}
-	if bootstrapIP, ok := configmap.Annotations[BootstrapIPAnnotationKey]; ok && bootstrapIP != "" {
-		// escape if IPv6
-		if net.ParseIP(bootstrapIP).To4() == nil {
-			bootstrapIP = "[" + bootstrapIP + "]"
+	newFunc := func() (*clientv3.Client, error) {
+		endpoints, err := endpointFunc()
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving endpoints for new cached client: %w", err)
 		}
-		etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", bootstrapIP))
+		return newEtcdClientWithClientOpts(endpoints, true)
 	}
 
-	g.clientLock.Lock()
-	defer g.clientLock.Unlock()
-	if reflect.DeepEqual(g.lastClientConfigKey, etcdEndpoints) && g.cachedClient != nil {
-		ctx, cancel := context.WithCancel(context.Background())
-		defer cancel()
-		// Test cached client connection by doing a request.
-		_, err = g.cachedClient.MemberList(ctx)
-		if err == nil {
-			return g.cachedClient, nil
-		}
-	}
-
-	c, err := getEtcdClientWithClientOpts(etcdEndpoints, false)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get etcd client: %w", err)
-	}
-	if g.cachedClient != nil {
-		if err := g.cachedClient.Close(); err != nil {
-			utilruntime.HandleError(fmt.Errorf("failed to close cached client: %w", err))
-		}
-	}
-	g.cachedClient = c
-	g.lastClientConfigKey = etcdEndpoints
-
-	return g.cachedClient, nil
+	g.clientPool = NewDefaultEtcdClientPool(newFunc, endpointFunc)
+	return g
 }
 
-func getEtcdClientWithClientOpts(endpoints []string, skipConnectionTest bool, opts ...ClientOption) (*clientv3.Client, error) {
+// newEtcdClientWithClientOpts allows customization of the etcd client using ClientOptions. All clients must be manually
+// closed by the caller with Close().
+func newEtcdClientWithClientOpts(endpoints []string, skipConnectionTest bool, opts ...ClientOption) (*clientv3.Client, error) {
 	grpclog.SetLoggerV2(grpclog.NewLoggerV2(ioutil.Discard, ioutil.Discard, os.Stderr))
 	clientOpts, err := newClientOpts(opts...)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error during clientOpts: %w", err)
 	}
 
 	dialOptions := []grpc.DialOption{
@@ -164,7 +140,7 @@ func getEtcdClientWithClientOpts(endpoints []string, skipConnectionTest bool, op
 	}
 	tlsConfig, err := tlsInfo.ClientConfig()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error during client TLSConfig: %w", err)
 	}
 
 	// Our logs are noisy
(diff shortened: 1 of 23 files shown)


pkg/etcdcli/etcdcli.go
```diff
@@ -208,6 +211,9 @@ func (g *etcdClientGetter) MemberAdd(ctx context.Context, peerURL string) error
 		return err
 	}
 
+	g.clientLock.RLock()
```
tjungblu, line 214: would that be sufficient? I can still see one scenario where after getting this `cli` here and in between the RLock another goroutine can close the g.client in the meantime. It's less likely than before, but can still happen. 

I think the only way to work around this is to directly depend on the `g.cachedClient` instead of the returned client.

tjungblu:
getting some more data here, sorry for the additional cost :/

/payload 4.11 ci required

geliu2016 (approved):
/lgtm

geliu2016 (approved):
/lgtm

pkg/etcdcli/etcdcli.go
```diff
@@ -116,6 +111,9 @@ func (g *etcdClientGetter) getEtcdClient() (*clientv3.Client, error) {
 		etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", bootstrapIP))
 	}
 
+	// ensure we always compare in the right order, the endpoints can be returned in any
+	sort.Strings(etcdEndpoints)
+
 	g.clientLock.Lock()
 	defer g.clientLock.Unlock()
```
hasbro17, line 118: I might be missing something but now that we have `clientLock.RLock()` protecting the client usage in MemberAdd/Remove/List etc... shouldn't we move `g.clientLock.Lock()` down to where we actually modify(close and recreate) the cached client? i.e L131.

Otherwise if goroutine A is in the midst of a `MemberList()` and has acquired `RLock()`, then goroutine B can't get the cached client here when being blocked by `Lock()`.

pkg/etcdcli/etcdcli.go
hasbro17: Arguably better than what we have now, but yes this is still susceptible to the client being closed out before `Rlock()`.

pkg/etcdcli/etcdcli.go
```diff
@@ -411,16 +442,25 @@ func (g *etcdClientGetter) MemberStatus(ctx context.Context, member *etcdserverp
...
+		}
+		if err := cli.Close(); err != nil {
+			klog.Errorf("error closing etcd client for defrag: %v", err)
+		}
+	}()
 
 	resp, err := cli.Defragment(ctx, member.ClientURLs[0])
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error while running defragment defragment: %w", err)
```
hasbro17, line 463:
```suggestion
		return nil, fmt.Errorf("error while running defragment: %w", err)
```

hasbro17 (commented):
Makes sense, but just a question on the placement of the existing Lock() in `getEtcdClient()`.

pkg/etcdcli/etcdcli.go
tjungblu: true, this should move down for sure.  
I found sync.Pool https://pkg.go.dev/sync#Pool in the stdlib, which we could use to get a client and return it once done. I reckon that should alleviate all those issues. Not sure how we can close it then - I'll play around with this a bit today.

pkg/etcdcli/etcdcli.go
tjungblu: apologies, will push this with the next commit

tjungblu:
@EmilyM1 + @dusk125 if you two find the time, please take a look as well. This is such a central piece to the operator it needs to be reviewed by many eyes.

pkg/etcdcli/etcdcli.go
```diff
@@ -68,82 +63,63 @@ func NewEtcdClient(kubeInformers v1helpers.KubeInformersForNamespaces, networkIn
...
-	defer g.clientLock.Unlock()
-	if reflect.DeepEqual(g.lastClientConfigKey, etcdEndpoints) && g.cachedClient != nil {
-		ctx, cancel := context.WithCancel(context.Background())
-		defer cancel()
-		// Test cached client connection by doing a request.
-		_, err = g.cachedClient.MemberList(ctx)
-		if err == nil {
-			return g.cachedClient, nil
+		if bootstrapIP, ok := configmap.Annotations[BootstrapIPAnnotationKey]; ok && bootstrapIP != "" {
+			// escape if IPv6
```
Elbehery, line 98: @tjungblu  here it is what i meant for above :)

pkg/etcdcli/etcdcli.go
```diff
@@ -68,82 +63,63 @@ func NewEtcdClient(kubeInformers v1helpers.KubeInformersForNamespaces, networkIn
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to get etcd client: %w", err)
-	}
-	if g.cachedClient != nil {
-		if err := g.cachedClient.Close(); err != nil {
-			utilruntime.HandleError(fmt.Errorf("failed to close cached client: %w", err))
+	newFunc := func() (*clientv3.Client, error) {
+		endpoints, err := endpointFunc()
+		if err != nil {
+			return nil, err
```
Elbehery, line 110: shall we add more context to the err ?

pkg/etcdcli/etcdcli.go
```diff
@@ -203,11 +179,13 @@ func getEtcdClientWithClientOpts(endpoints []string, opts ...ClientOption) (*cli
 func (g *etcdClientGetter) MemberAdd(ctx context.Context, peerURL string) error {
 	g.eventRecorder.Eventf("MemberAdd", "adding new peer %v", peerURL)
 
-	cli, err := g.getEtcdClient()
+	cli, err := g.clientPool.Get()
 	if err != nil {
 		return err
```
Elbehery, line 184: shall we add context to err here ?

pkg/etcdcli/etcdcli.go
```diff
@@ -243,11 +221,13 @@ func (g *etcdClientGetter) MemberUpdatePeerURL(ctx context.Context, id uint64, p
 		g.eventRecorder.Eventf("MemberUpdate", "updating member %q with peers %v", memberName, strings.Join(peerURLs, ","))
 	}
 
-	cli, err := g.getEtcdClient()
+	cli, err := g.clientPool.Get()
 	if err != nil {
 		return err
```
Elbehery, line 226: same here ?

pkg/etcdcli/etcdcli.go
```diff
@@ -411,16 +404,25 @@ func (g *etcdClientGetter) MemberStatus(ctx context.Context, member *etcdserverp
...
+	cli, err := newEtcdClientWithClientOpts([]string{member.ClientURLs[0]}, WithDialTimeout(DefragDialTimeout))
 	if err != nil {
 		return nil, fmt.Errorf("failed to get etcd client for defragment: %w", err)
 	}
-	defer cli.Close()
+	defer func() {
+		if cli == nil {
+			return
+		}
+		if err := cli.Close(); err != nil {
```
Elbehery, line 418: do we call close or Return() to give it back to the sync.Pool ?

pkg/etcdcli/etcdcli_pool_test.go
```diff
@@ -0,0 +1,176 @@
...
+import (
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/tests/v3/integration"
+	"testing"
+)
+
+// rather poor men's approach to mocking
```
Elbehery, line 12: :D :D

pkg/etcdcli/etcdcli.go
```diff
@@ -68,82 +63,63 @@ func NewEtcdClient(kubeInformers v1helpers.KubeInformersForNamespaces, networkIn
...
-		if net.ParseIP(bootstrapIP).To4() == nil {
-			bootstrapIP = "[" + bootstrapIP + "]"
+		var etcdEndpoints []string
+		nodes, err := g.nodeLister.List(labels.Set{"node-role.kubernetes.io/master": ""}.AsSelector())
+		for _, node := range nodes {
+			internalIP, err := dnshelpers.GetEscapedPreferredInternalIPAddressForNodeName(network, node)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get internal IP for node: %w", err)
+			}
+			etcdEndpoints = append(etcdEndpoints, fmt.Sprintf("https://%s:2379", internalIP))
```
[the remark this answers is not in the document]
tjungblu: this? no, but that's how it was before. Didn't intend to rewrite the whole client logic either.

pkg/etcdcli/etcdcli.go
```diff
@@ -68,82 +63,63 @@ func NewEtcdClient(kubeInformers v1helpers.KubeInformersForNamespaces, networkIn
...
+		}
+
+		configmap, err := g.configmapsLister.ConfigMaps(operatorclient.TargetNamespace).Get("etcd-endpoints")
+		if err != nil {
+			return nil, fmt.Errorf("failed to list endpoints: %w", err)
```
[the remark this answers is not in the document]
tjungblu: haven't changed that, but good idea.

pkg/etcdcli/etcdcli.go
tjungblu: there is more context in both endpointFunc and in the cacheClient. which information do you need here?

pkg/etcdcli/etcdcli.go
tjungblu: as the comment on the func says: 
>// Defragment creates a new uncached clientv3 to the given member url and calls clientv3.Client.Defragment.

I don't know why it did not use the cache client before, I assume due to single endpoint target nature of that method. But Close() is correct here.

pkg/etcdcli/etcdcli.go
tjungblu: yes

tjungblu:
unit tests are still flaky on CI, fixing this.. hard to repro on my box

pkg/etcdcli/etcdcli.go
Elbehery: ok then its fine 👍🏽

pkg/etcdcli/etcdcli.go
Elbehery: sounds good 👍🏽

pkg/etcdcli/etcdcli_pool.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+// etcdClientPool fulfills these requirements:
+// * cache clients to avoid re-creating them all the time (TLS handshakes are expensive after all)
+// * return an exclusively unused client, no other can acquire the same client at that time
+// * health checking a client before using it (using list), return a new one if unhealthy and closing the old one
+// * update endpoints, to be always up to date with the changes
+// * return a used client to the pool, making it available to consume again
+type etcdClientPool struct {
+	pool          *sync.Pool
```
tjungblu, line 22: some learnings from sync.Pool:
* it's really fast
* it does not actually work like a queue, when you return an item there is no guarantee to get it back during the next invocation (meaning it would create a new client)
* matter of fact, load testing this with the readyz endpoint is actually causing it to create gazillions of clients without actually closing the ones that it removed.


So with all the above, I think sync.Pool will also create a leak, will be a bit faster though. Will checkout another way to implement this tomorrow.

Elbehery: @tjungblu  great comments thanks 👍🏽 🙏🏽 

I actually liked the approach, what are other options ? 

I think RWMutex is not a one, it was implemented like this right ?

tjungblu: yeah I thought the sync.Pool would take all the synchronization pain away as well :/
we could wrap this into a single mutex backed with an array (as a queue). I think go channels should also work for that.

tjungblu:
/hold 

for now as it leaks the client again

geliu2016 (approved):
/label cherry-pick-approved 
/lgtm

tjungblu:
/hold

still not really reliable, one issue on a node coming up:
> Readiness probe status=failure output="HTTP probe failed with statuscode: 503" start-of-body=failed to establish etcd client: too many active cache clients, rejecting to create new one

pkg/etcdcli/etcdcli.go
```diff
@@ -69,86 +64,67 @@ func NewEtcdClient(kubeInformers v1helpers.KubeInformersForNamespaces, networkIn
...
-func getEtcdClientWithClientOpts(endpoints []string, skipConnectionTest bool, opts ...ClientOption) (*clientv3.Client, error) {
+// newEtcdClientWithClientOpts allows customization of the etcd client using ClientOptions. All clients must be manually
+// closed by the caller with Close().
+func newEtcdClientWithClientOpts(endpoints []string, skipConnectionTest bool, opts ...ClientOption) (*clientv3.Client, error) {
 	grpclog.SetLoggerV2(grpclog.NewLoggerV2(ioutil.Discard, ioutil.Discard, os.Stderr))
 	clientOpts, err := newClientOpts(opts...)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error during clientOpts: %w", err)
 	}
```
EmilyM1, line 128: Good addition of clearer error messages throughout. Much appreciated!

tjungblu: thank @Elbehery!

pkg/etcdcli/etcdcli_pool_test.go
```diff
@@ -0,0 +1,430 @@
...
+	assert.Equal(t, maxNumClientTickets, poolRecorder.numHealthCalls)
+	assert.Equal(t, 0, poolRecorder.numCloseCalls)
+
+	// this should block and return an error
+	client, err := poolRecorder.pool.Get()
+	assert.Nil(t, client)
+	assert.Errorf(t, err, "too many active cache clients, rejecting to create new one")
+	// returning one should unlock the get again
+	poolRecorder.pool.Return(clients[0])
+	client, err = poolRecorder.pool.Get()
```
hasbro17, line 161: Missing error check.
```suggestion
	client, err = poolRecorder.pool.Get()
	require.NoError(t, err)
```

pkg/etcdcli/etcdcli_pool_test.go
```diff
@@ -0,0 +1,430 @@
...
+	assert.Equal(t, maxNumClientTickets, poolRecorder.numHealthCalls)
+	assert.Equal(t, 0, poolRecorder.numCloseCalls)
+
+	// return all clients to fill the internal cache and cause five to close
+	for i := 0; i < maxNumClientTickets; i++ {
+		poolRecorder.pool.Return(clients[i])
+	}
+	assert.Equal(t, maxNumCachedClients, poolRecorder.numCloseCalls)
+
+	// now we should be able to get the full amount of tickets again
```
hasbro17, line 194:
```suggestion
	// now we should be able to get the full amount of clients again
```
I think you mean clients and not tickets here as the number of available tickets is just 5 at this point.
This threw me off initially as it made me think the tickets were returned once the client was as well.

So I expected the following unit test to succeed.
```Go
func TestAllTicketsAvailableForFullyCachedPool(t *testing.T) {
	integration.BeforeTestExternal(t)
	testServer := integration.NewClusterV3(t, &integration.ClusterConfig{Size: 3})
	defer testServer.Terminate(t)

	poolRecorder := newTestPool(testServer)
	var clients []*clientv3.Client
	for i := 0; i < maxNumCachedClients; i++ {
		client, err := poolRecorder.pool.Get()
		require.NoError(t, err)
		assert.NotNil(t, client)
		clients = append(clients, client)
	}

	// return all clients to the client pool and check that none are closed
	for _, client := range clients {
		poolRecorder.pool.Return(client)
	}
	assert.Equal(t, maxNumCachedClients, poolRecorder.numNewCalls)
	assert.Equal(t, maxNumCachedClients, poolRecorder.numEndpointCalls)
	assert.Equal(t, maxNumCachedClients, poolRecorder.numHealthCalls)
	assert.Equal(t, 0, poolRecorder.numCloseCalls)

	// The tickets should equal the max available (10) since all clients are ready to be taken from the cached pool
	assert.Equal(t, maxNumClientTickets, len(poolRecorder.pool.availableTickets))

	// Grab a cached client from the pool
	client, err := poolRecorder.pool.Get()
	require.NoError(t, err)
	assert.NotNil(t, client)
	assert.Equal(t, maxNumClientTickets-1, len(poolRecorder.pool.availableTickets))

	// Return it back and all tickets should be available again
	poolRecorder.pool.Return(client)
	assert.Equal(t, maxNumClientTickets, len(poolRecorder.pool.availableTickets))
}
```

But looking through the code it seems tickets used == number of clients currently open. So just wanted to confirm this.

pkg/etcdcli/etcdcli_pool_test.go
```diff
@@ -0,0 +1,430 @@
...
+	for i := 0; i < maxNumClientTickets; i++ {
+		client, err := poolRecorder.pool.Get()
+		require.NoError(t, err)
+		assert.NotNil(t, client)
+	}
+
+	// replenish the maxNumCachedClients that were closed earlier
+	assert.Equal(t, maxNumClientTickets+maxNumCachedClients, poolRecorder.numNewCalls)
```
hasbro17, line 202: Can we add a check for the number of available tickets too just to make it explicit that available tickets are the capacity to create more clients.
```suggestion
	assert.Equal(t, maxNumClientTickets+maxNumCachedClients, poolRecorder.numNewCalls)
	assert.Equal(t, maxNumClientTickets-maxNumCachedClients, len(poolRecorder.pool.availableTickets))
```

hasbro17 (commented):
SGTM, just a couple nits and a question on what the tickets actually represent.

pkg/etcdcli/etcdcli_pool.go
```diff
@@ -0,0 +1,184 @@
...
+		return
+	}
+
+	select {
+	case p.pool <- client:
+	default:
+		returnTicket(p.availableTickets)
+		err := p.closeFunc(client)
+		if err != nil {
+			klog.Errorf("could not close cache client exceeding pool capacity: %v", err)
```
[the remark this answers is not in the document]
tjungblu: great suggestion, let me add this

pkg/etcdcli/etcdcli_pool_test.go
tjungblu: done

pkg/etcdcli/etcdcli_pool_test.go
tjungblu: indeed, there are two limits:
* pool size, how many active clients we are keeping cached at max
* tickets, how many clients we are handing out at a time

The first should ensure we are not creating new clients all the time and keeping them cached, because the TLS handshakes are expensive. Yet bound to `maxNumCachedClients` as too many IDLE connections to etcd are also not helpful.

The second should ensure that on connectivity issues we do not overwhelm etcd additionally by creating more connections. The behavior in a previous version before the ticketing was: 
* I have to get the members, can I get a cached client?
* pool is filled, let me grab one and try to health check
* none of the healthy client are available, let's create a new one

The last bullet was basically creating hundreds of new clients as there was no connectivity for some time period. The ticketing basically rate limits this, imagine going to a full amusement park and you have to wait for somebody to come out of it. 

> But looking through the code it seems tickets used == number of clients currently open. So just wanted to confirm this.

yep, good summary. Shall we rename that ticketing aspect to `numOpenClients`?

pkg/etcdcli/etcdcli_pool_test.go
tjungblu: yep, in this case the tickets are zero as all clients are handed out:
```
	// no tickets are available anymore, as we have handed out all clients
	assert.Equal(t, 0, len(poolRecorder.pool.availableTickets))
```

PR #805, opened 2022-04-27
clustermemberremovalcontroller must examine cluster health before removing a member

p0lyn0mial:
otherwise, we might lose quorum since some members might be unavailable.
I think it doesn't matter for `removeMemberWithoutMachine` and  `attemptToRemoveLearningMember` methods.

### DIFF
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
@@ -2,6 +2,7 @@ package clustermemberremovalcontroller
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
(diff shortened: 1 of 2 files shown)


pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
```diff
@@ -696,9 +739,15 @@ func TestAttemptToScaleDown(t *testing.T) {
 				configMapListerForTargetNamespace: configMapTargetNSLister,
 			}
 			err = target.attemptToScaleDown(context.TODO(), eventRecorder)
-			if err != nil {
+			if err == nil && scenario.expectedError != nil {
+				t.Fatal("expected to get an error from attemptToScaleDown method")
```
hasbro17, line 743: Nit: show the expected error
```suggestion
				t.Fatal("expected to get an error from attemptToScaleDown method, expected = %v", scenario.expectedError)
```

PR #808, opened 2022-04-28
fire an event on successfull member removal

p0lyn0mial:


pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -380,6 +380,7 @@ func (c *clusterMemberRemovalController) attemptToRemoveMemberFor(ctx context.Co
 				errs = append(errs, fmt.Errorf("failed to remove member: %v, err: %v", memberLocator, err))
 				break // stop, we have found a matching member for the provided machine
 			}
+			recorder.Eventf("ScaleDown", "successfully removed member: %v from the cluster", memberLocator)
```
hasbro17, line 383: We already do have an event for `MemberRemove` although that method is missing the member name.
https://github.com/openshift/cluster-etcd-operator/blob/8b8c431b3f7fcb6eb80a1229edf403bce5df7be4/pkg/etcdcli/etcdcli.go#L299

I'm fine with adding this event for now since changing that method will also require an update in the bootstrap teardown controller. We can consolidate that later.

PR #810, opened 2022-05-02
[release-4.9] Bug 2068084: manually disable defrag

tjungblu:


pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -23,30 +29,40 @@ import (
...
+	kubeInformers v1helpers.KubeInformersForNamespaces) factory.Controller {
 	c := &DefragController{
 		operatorClient:       operatorClient,
 		etcdClient:           etcdClient,
 		infrastructureLister: infrastructureLister,
+		configmapLister:      kubeInformers.ConfigMapLister(),
+		defragWaitDuration:   minDefragWaitDuration,
 	}
-	return factory.New().ResyncEvery(9*time.Minute).WithInformers(
+	return factory.New().ResyncEvery(compactionInterval+1*time.Minute).WithInformers( // attempt to sync outside of etcd compaction interval to ensure maximum gain by defragmentation.
```
dusk125, line 65: For readability, consider adding parentheses to `compactionInterval+1*time.MInute` so as not to rely on operator precedence.

tjungblu: yeah good point, I haven't actually touched that one. That originated here: https://github.com/openshift/cluster-etcd-operator/commit/aa7ff879599425b169d3e182b312fa2a7f959dae#diff-dfbff74503d147025fe53162fb01aa0bb184ce736fbe346c6f333eef37ebb432R55

dusk125: ack then we should leave as-is and maybe fix it in later revision (and backport), because I believe (I checked the go spec to be sure) but this will evaluate to `compactionInterval+(1*time.Minute)` which I don't believe is what we want, from the change I would assume we wanted `(compactionInterval+1)*time.Minute`.

dusk125: Disregard this, the code is correct: compactionInterval is already a time.Duration and so the `1*time.Minute` happening first is correct to add a minute onto the compactionInterval.

tjungblu: I think it does what was originally intended, the latter doesn't  compile as it overflows int64:
https://go.dev/play/p/pAWgMRwvqOC

PR #812, opened 2022-05-03
Bug 2073901: revisit defrag controller degradation

tjungblu:
only let the defrag controller degrade after failing several attempts at actual defragmentation work

### DIFF
--- a/pkg/etcdcli/helpers.go
+++ b/pkg/etcdcli/helpers.go
@@ -17,6 +17,11 @@ type fakeEtcdClient struct {
 }
 
 func (f *fakeEtcdClient) Defragment(ctx context.Context, member *etcdserverpb.Member) (*clientv3.DefragmentResponse, error) {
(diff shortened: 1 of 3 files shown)


dusk125:
Thoughts on having the max failures configurable (maybe in the future) so if someone wants to degrade after 1 failure they could?

tjungblu:
@dusk125 I like the idea, what would be a good way to configure this? via env vars?

dusk125:
> @dusk125 I like the idea, what would be a good way to configure this? via env vars?

Either that, or with a configmap like [1], @hasbro17 do we have a preferred method for allowing configuration like this?

[1] https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/defragcontroller/defragcontroller.go#L96

tjungblu:
@dusk125 wdyt of adding this to the [RFE-2799](https://issues.redhat.com/browse/RFE-2799)? I reckon we can add this into a CRD as well.

PR #815, opened 2022-05-04
Bug 2077160: Adding Thomas to reviewers and me to approvers

dusk125:
@tjungblu needs review permissions

~~Adding the rest of the etcd team to approvers under the agreement that, except for very minor/insignificant changes, the person who lgtm will not also approve and vice-versa.~~
We found that adding everyone to approvers essentially makes it so there's only one required reviewer instead of two for each change. Adding myself to approvers to limit bottlenecks for now.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -6,9 +6,10 @@ reviewers:
   - hasbro17
   - EmilyM1
   - Elbehery
+  - tjungblu
 approvers:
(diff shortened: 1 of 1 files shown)


hasbro17:
/hold

Need to test something around how the bot handles lgtms and approves from an approver.

PR #823, opened 2022-05-12
Bug 2087983: remove etcd_perf before restore

claneys:
This commit deletes "etcd_perf" files that could present if someone has ran a fio test under /var/lib/etcd
like specified in the OpenShift docs about the recommended host practices.

As this file is known as non essential, it could be freely deleted by anticipation by the restore pod when
restoring a cluster to a previous state.

### DIFF
--- a/bindata/etcd/restore-pod.yaml
+++ b/bindata/etcd/restore-pod.yaml
@@ -26,6 +26,11 @@ spec:
         env | grep ETCD | grep -v NODE
         export ETCD_NODE_PEER_URL=https://${NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST}:2380
 
+        # checking if there are any fio perf file left behind that could be deleted without problems
+        if [ ! -z $(ls -A "/var/lib/etcd/etcd_perf*") ]; then
+          rm -f /var/lib/etcd/etcd_perf*
+        fi
+
         # checking if data directory is empty, if not etcdctl restore will fail
         if [ ! -z $(ls -A "/var/lib/etcd") ]; then
           echo "please delete the contents of data directory before restoring, running the restore script will do this for you"
--- a/pkg/operator/etcd_assets/bindata.go
+++ b/pkg/operator/etcd_assets/bindata.go
@@ -1007,6 +1007,11 @@ spec:
         env | grep ETCD | grep -v NODE
         export ETCD_NODE_PEER_URL=https://${NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST}:2380
 
+        # checking if there are any fio perf file left behind that could be deleted without problems
+        if [ ! -z $(ls -A "/var/lib/etcd/etcd_perf*") ]; then
+          rm -f /var/lib/etcd/etcd_perf*
+        fi
+
         # checking if data directory is empty, if not etcdctl restore will fail
         if [ ! -z $(ls -A "/var/lib/etcd") ]; then
           echo "please delete the contents of data directory before restoring, running the restore script will do this for you"


Elbehery:
@claneys Hello ✋🏽  .. May I ask for Justification of this PR ? Would you mind a detailed description of the reasons and / or why do you need it ?

claneys:
Sure @Elbehery , I forgot to mention that in the commit message, my bad.

The problem here appears when restoring an Openshift cluster in a previous state following the [documentation](https://docs.openshift.com/container-platform/4.10/backup_and_restore/control_plane_backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.html#dr-scenario-cluster-state-issues_dr-restoring-cluster-state). 

Everything went fine until the restauration and restart of the kubelet (Step 8.) because the pod that is in charge of executing the db snapshot restore step isn't able to restore due to that file left behind by fio perf test. The commit context do not show the 4th line after the modification where the script "exit 1".

This fix that problem, It isn't may be not acceptable to handle every files that could land here, but as this fio's file comes from the documentation of recommended host practices, it could be worth it to handle at least file issued from the documentation procedure.

tjungblu:
@claneys thanks for the contribution, could you kindly update the bindata.go and add it to your PR? a `make update` should do the trick. Thanks!

bindata/etcd/restore-pod.yaml
```diff
@@ -26,6 +26,11 @@ spec:
         env | grep ETCD | grep -v NODE
         export ETCD_NODE_PEER_URL=https://${NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST}:2380
 
+        # checking if there are any fio perf file left behind that could be deleted without problems
+        if [ ! -z $(ls -A "/var/lib/etcd/etc_perf*") ]; then
```
Elbehery, line 30: @claneys Hello :) 

Is it `etc_perf*` or `etcd_perf*` ?

Elbehery (commented):
@claneys just a potential nit, could you kindly verify ?

Elbehery:
@claneys can you run `make verify` on the operator root folder ?

bindata/etcd/restore-pod.yaml
claneys: Oh crap, yes you're right. Thanks. I'll fix it and update the PR.

@tjungblu  @Elbehery, ok, no problem. Thanks for the tip.

tjungblu:
/lgtm

@claneys do you need it in a specific version of openshift? We just passed by feature freeze for 4.11 (so we might need to declare this a bug fix).

claneys:
> /lgtm
> 
> @claneys do you need it in a specific version of openshift? We just passed by feature freeze for 4.11 (so we might need to declare this a bug fix).

If it could be included in the 4.10 version, it would perfect :)

claneys:
> /lgtm
> 
> @claneys do you need it in a specific version of openshift? We just passed by feature freeze for 4.11 (so we might need to declare this a bug fix).

If it could be included in the 4.10 version, it would perfect :)

tjungblu:
/retitle Bug 2087983: remove etcd_perf before restore

tjungblu:
@claneys this should merge automatically after the tests go through (no idea why it tests everything again), thanks for your help. I'll get it merged into the 4.10 branch after that. 

I think today was the cut-off for 4.10.16, so it might get into 4.10.17 or following.

claneys:
> @claneys this should merge automatically after the tests go through (no idea why it tests everything again), thanks for your help. I'll get it merged into the 4.10 branch after that.
> 
> I think today was the cut-off for 4.10.16, so it might get into 4.10.17 or following.

It's perfect to me, thanks for the review.

PR #833, opened 2022-05-17
add command to upload the etcd health monitor data to a configmap

deads2k:
This will upload the last 10k-ish (clean line split) of the etcd health monitor logs into a configmap for analysis from must-gather or openshift-tests.

### DIFF
--- a/pkg/cmd/monitoruploader/monitor_uploader.go
+++ b/pkg/cmd/monitoruploader/monitor_uploader.go
@@ -1,0 +1,257 @@
+package monitoruploader
+
+import (
+	"bufio"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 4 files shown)


tjungblu:
@deads2k can't we pull the node log out the same way we do with the must-gather in https://github.com/openshift/must-gather/blob/3af14335d5b30106220643b09349e389a3e07cc0/collection-scripts/gather_service_logs_util#L7

not sure why you want to tail the health probe log into a config map :octocat:

deads2k:
> @deads2k can't we pull the node log out the same way we do with the must-gather in https://github.com/openshift/must-gather/blob/3af14335d5b30106220643b09349e389a3e07cc0/collection-scripts/gather_service_logs_util#L7
> 
> not sure why you want to tail the health probe log into a config map :octocat:

@tjungblu could you put the data onto the openshift-tests timeline data and provide the outage data in cases when the node is offline?

pkg/cmd/monitoruploader/monitor_uploader.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	contentToWrite := strings.Join(lastLines, "\n")
+
+	fmt.Printf("Uploading content\n")
+	configMapToApply := applycorev1.ConfigMap(o.ConfigMapName, o.ConfigMapNamespace).WithData(
+		map[string]string{
+			o.ConfigMapKey: contentToWrite,
+		},
+	)
+	if _, err := o.ConfigMapClient.ConfigMaps(o.ConfigMapNamespace).Apply(ctx, configMapToApply, metav1.ApplyOptions{FieldManager: o.FieldManager, Force: true}); err != nil {
```
deads2k, line 252: server-side apply example here.

PR #838, opened 2022-05-25
Bug 2085997: increase initial delay seconds to match the discovery timeout

p0lyn0mial:


p0lyn0mial:
/hold

until we agree on https://github.com/openshift/etcd/pull/128

deads2k:
This PR needs to be reverted.

This extended all upgrades by a minimum of 18 minutes and all installs by at least 9 minutes because readyz is required before the PDB will allow the next member to go down and is logically required by the operator as well.

Try using startupProbes instead of initialDelaySeconds so you can be responsive when the startup is fast and slow when the startup is slow.

p0lyn0mial:
using a `startupProbes` is way much better than `initialDelaySeconds`, thanks for point it out.

PR #840, opened 2022-05-26
Bug 2085997: defines a startupProbe for etcd container

p0lyn0mial:
The kubelet uses startup probes to know when a container application has started. Once a probe is configured, it disables liveness and readiness checks until it succeeds.

We give the etcd container a three-minute wait with a 10s startup time minimum (initial delay).

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -197,11 +197,20 @@ ${COMPUTED_ENV_VARS}
         port: 9980
         path: readyz
         scheme: HTTPS
-      initialDelaySeconds: 10
       timeoutSeconds: 10
       failureThreshold: 3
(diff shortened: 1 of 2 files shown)


deads2k:
this looks reasonable to me.  Check the output to be sure that the etcd pods on upgrade go ready in about 10 seconds.

deads2k:
Can you eliminate the initialdelayseconds now?

p0lyn0mial:
> Can you eliminate the initialdelayseconds now?

but only for the readiness probe right?

p0lyn0mial:
oh, and we don't run any upgrade job against this repo :(

p0lyn0mial:
/hold cancel

the output from the `e2e-agnostic-upgrade` lgtm

PR #835, opened 2022-05-18
Bug 2079803: fix consecutive backup on consecutive upgrades

Elbehery:
This PR ensures backup is running before each upgrade, regardless of existing backup

PR #841, opened 2022-05-30
fix partial update cluster version

Elbehery:
This PR fixes a bug from https://github.com/openshift/cluster-etcd-operator/pull/835

When the ClusterVersion does not contain `CompletedUpdate` condition, the UpgradeBackup operator goes degraded.
This PR replaces Error with a warning, to avoid going degraded in CI environment.

### DIFF
--- a/pkg/operator/ceohelpers/common.go
+++ b/pkg/operator/ceohelpers/common.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	configv1 "github.com/openshift/api/config/v1"
+	"k8s.io/klog/v2"
 	"net"
 	"net/url"
 
@@ -188,5 +189,6 @@ func GetCurrentClusterVersion(cv *configv1.ClusterVersion) (string, error) {
 			return c.Version, nil
 		}
 	}
-	return "", fmt.Errorf("unable to retrieve cluster version, no completed update was found in cluster version status history: %v", cv.Status.History)
+	klog.Warningf("unable to retrieve cluster version, no completed update was found in cluster version status history: %v", cv.Status.History)
+	return "", nil
 }


pkg/operator/ceohelpers/common.go
```diff
@@ -188,5 +189,6 @@ func GetCurrentClusterVersion(cv *configv1.ClusterVersion) (string, error) {
 			return c.Version, nil
 		}
 	}
-	return "", fmt.Errorf("unable to retrieve cluster version, no completed update was found in cluster version status history: %v", cv.Status.History)
+	klog.Warningf("unable to retrieve cluster version, no completed update was found in cluster version status history: %v", cv.Status.History)
+	return "", nil
```
tjungblu, line 193: I don't quite understand the method now. When there is no cluster version this is a clear error case no? what do you expect somebody to do with "" as a version string?

tjungblu: and please add some unit tests around this method

Elbehery: so I found in the ci https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/816/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-serial/1531208277348585472 that the cluster has `PartialUpdate` and therefore will trigger error and the operator goes degraded 

[Here](https://github.com/openshift/cluster-etcd-operator/blob/bc2ebdadaa29cbace42ab1e5afc11ef85ac68712/pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go#L358-L360) we do nothing if the upgrade is partial and the version is empty string, but we dont go degraded 

Please correct me if I am missing something

tjungblu: I think that behavior is actually fine, if it's a partial upgrade (or install) the cluster is not functional and the operator should degrade here.

Elbehery: ok then i am closing this as of no need to change

tjungblu: @Elbehery could you add some unit tests for this method though? I find it difficult to have a helper method without established tests.

PR #843, opened 2022-06-01
Bug 2092395: etcdHighNumberOfFailedGRPCRequests alerts with wrong results

tjungblu:


tjungblu:
@wking by popular demand of the etcd team, do you also want to take a look?

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -109,22 +101,18 @@ spec:
         severity: warning
     - alert: etcdBackendQuotaLowSpace
       annotations:
-        summary: etcd cluster database size exceeds the defined quota.
-        description: 'etcd cluster "{{ $labels.job }}": database size exceeds the
-          defined quota on etcd instance {{ $labels.instance }}, please defrag or
-          increase the quota as the writes to etcd will be disabled when it is full.'
+        description: 'etcd cluster "{{ $labels.job }}": database size exceeds the defined quota on etcd instance {{ $labels.instance }}, please defrag or increase the quota as the writes to etcd will be disabled when it is full.'
```
wking, line 104: "... exceeds the defined quota..." is a bit premature.  I'd just say "... is {{ $value }}% of the defined quota..." or some such.

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -109,22 +101,18 @@ spec:
...
         runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/etcdBackendQuotaLowSpace.md
+        summary: etcd cluster database is running full.
```
wking, line 106: "running full" is not very specific about what is full.  Maybe rephrase to "etcd cluster database size is near quota" or something?

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -109,22 +101,18 @@ spec:
...
     - alert: etcdExcessiveDatabaseGrowth
       annotations:
-        summary: etcd cluster database size increased by 50 percent over the past four hours.
-        description: 'etcd cluster "{{ $labels.job }}": Observed surge in etcd writes
-          leading to 50% increase in database size over the past four hours on etcd
-          instance {{ $labels.instance }}, please check as it might be disruptive.'
+        description: 'etcd cluster "{{ $labels.job }}": Observed surge in etcd writes leading to 50% increase in database size over the past four hours on etcd instance {{ $labels.instance }}, please check as it might be disruptive.'
+        summary: etcd cluster database growing very fast.
       expr: |
         increase(((etcd_mvcc_db_total_size_in_bytes/etcd_server_quota_backend_bytes)*100)[240m:1m]) > 50
```
wking, line 117: unrelated nit, but putting `etcd_server_quota_backend_bytes` inside the `increase` could lead to false-positives if the user scales down the quota because they had overprovisioned.  What you care about is consumption vs. the current quota, so maybe:

    increase(etcd_mvcc_db_total_size_in_bytes[4h]) / etcd_server_quota_backend_bytes *100 > 50

with a `description` like:

> etcd cluster `"{{ $labels.job }}"`: etcd database size increased `{{ $value }}%` of the configured quota over the past four hours on etcd instance `{{ $labels.instance }}`.  Please defrag (FIXME: doc link?) or increase the quota (FIXME: doc link?) as the writes to etcd will be disabled when it is full.  Alternatively, investigate consumption and see if you can remove whatever's spewing this data into etcd (FIXME: reword.  Doc link).

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -150,43 +138,28 @@ spec:
         runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/etcdHighNumberOfFailedGRPCRequests.md
         summary: etcd cluster has high number of failed grpc requests.
       expr: |
-        100 * sum(rate(grpc_server_handled_total{job="etcd", grpc_code=~"Unknown|FailedPrecondition|ResourceExhausted|Internal|Unavailable|DataLoss|DeadlineExceeded"}[5m])) without (grpc_type, grpc_code)
+        (sum(rate(grpc_server_handled_total{job="etcd", grpc_code=~"Unknown|FailedPrecondition|ResourceExhausted|Internal|Unavailable|DataLoss|DeadlineExceeded"}[5m])) without (grpc_type, grpc_code)
           /
-        sum(rate(grpc_server_handled_total{job="etcd"}[5m])) without (grpc_type, grpc_code)
-          > 5 and on ()(sum(cluster_infrastructure_provider{type!~"ipi|BareMetal"} == bool 1))
+        (sum(rate(grpc_server_handled_total{job="etcd"}[5m])) without (grpc_type, grpc_code)
+          > 2 and on ()(sum(cluster_infrastructure_provider{type!~"ipi|BareMetal"} == bool 1)))) * 100 > 50
```
wking, line 144: Changing from `5` to `2` and adding `> 50` both seem like semantic changes, but there is no `description` update.  It also seems a fair bit different than [the upstream `expr`][1].  I haven't rooted around in history to figure out what's going on here, but I'm curious ;).

[1]: https://github.com/etcd-io/etcd/blame/08f4c340eb92338d0f266795222d9ea04cb69056/contrib/mixin/mixin.libsonnet#L90-L123

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -150,43 +138,28 @@ spec:
...
-          should be investigated.'
+        description: 'etcd cluster "{{ $labels.job }}": {{ $value }} leader changes within the last 15 minutes. Frequent elections may be a sign of insufficient resources, high network latency, or disruptions by other components and should be investigated.'
         summary: etcd cluster has high number of leader changes.
       expr: |
-         (increase((max without (instance) (etcd_server_leader_changes_seen_total{job=~".*etcd.*"})
-         or
-           0*absent(etcd_server_leader_changes_seen_total{job=~".*etcd.*"}))[15m:1m]) >= 5)
-         and
-           on () time() - cluster_version{type="initial"} > 3600
+        increase((max without (instance) (etcd_server_leader_changes_seen_total{job=~".*etcd.*"}) or 0*absent(etcd_server_leader_changes_seen_total{job=~".*etcd.*"}))[15m:1m]) >= 5
```
wking, line 153: Looks like we're dropping the `time() - cluster_version{type="initial"} > 3600` guard accidentally?

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -150,43 +138,28 @@ spec:
...
-          the Kubernetes and OpenShift APIs will reject read and write requests and
-          operations that preserve the health of workloads cannot be performed. This
-          can occur when multiple control plane nodes are powered off or are unable
-          to connect to each other via the network. Check that all control plane nodes
-          are powered on and that network connections between each machine are functional.
+        description: etcd is reporting fewer instances are available than are needed ({{ $value }}). When etcd does not have a majority of instances available the Kubernetes and OpenShift APIs will reject read and write requests and operations that preserve the health of workloads cannot be performed. This can occur when multiple control plane nodes are powered off or are unable to connect to each other via the network. Check that all control plane nodes are powered on and that network connections between each machine are functional.
         runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/etcdInsufficientMembers.md
         summary: etcd is reporting that a majority of instances are unavailable.
```
wking, line 161: unrelated nit: "a majority" is not strictly true is it?  If you're running four members because you haven't read the docs that explain that that's not helpful, or are temporarily transitioning between three-member groups while you replace a member or whatever, and two of those members get lost, you lose quorum, but you aren't quite up to "a majority" of unavailable members.

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -150,43 +138,28 @@ spec:
...
-      expr: sum(up{job="etcd"} == bool 1 and etcd_server_has_leader{job="etcd"} ==
-        bool 1) without (instance,pod) < ((count(up{job="etcd"}) without (instance,pod)
-        + 1) / 2)
+      expr: sum(up{job="etcd"} == bool 1 and etcd_server_has_leader{job="etcd"} == bool 1) without (instance,pod) < ((count(up{job="etcd"}) without (instance,pod) + 1) / 2)
```
wking, line 162: unrelated nit: feels like there should be some distinction between following and voting members in here.

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
tjungblu: I have to be honest, this alert could've never worked in Prometheus. So I think we're going to contribute this back upstream soon after we unblocked the SDN folks with this. 

> Changing from 5 to 2 and adding > 50 both seem like semantic changes, but there is no description update.

5 and 2 are supposed to be a low-pass filter, filtering out the low-QPS endpoints that can have high failure rates without impacting the stability. As to where those numbers come from? No idea, but it seems two QPS is good enough to filter the least called ones out.

The description as `etcd cluster "{{ $labels.job }}": {{ $value }}% of requests for {{ $labels.grpc_method }} failed on etcd instance {{ $labels.instance }}.` is good enough IMHO, any changes you would like to see here?

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
tjungblu: change of plan, let's add this back ;) but I think we need to remove this as this is not actually fixing  the root cause of this alert firing. Let's discuss further in that BZ then.

jsonnet/custom.libsonnet
```diff
@@ -41,7 +41,7 @@
           {
             alert: 'etcdHighNumberOfLeaderChanges',
             expr: |||
-              increase((max without (instance) (etcd_server_leader_changes_seen_total{job=~".*etcd.*"}) or 0*absent(etcd_server_leader_changes_seen_total{job=~".*etcd.*"}))[15m:1m]) >= 5
+              (increase((max without (instance) (etcd_server_leader_changes_seen_total{job=~".*etcd.*"}) or 0*absent(etcd_server_leader_changes_seen_total{job=~".*etcd.*"}))[15m:1m]) >= 5)  and on () time() - cluster_version{type="initial"} > 3600
```
tjungblu, line 44: that was missing from a previous PR that has changed the manifest directly. I added this change as a separate commit.

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
tjungblu: true, unfortunately that's not easy to find with `up`, so that's a metric that would need to get exposed by etcd itself based on the peers.

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
tjungblu: if you're running four nodes, you can still only tolerate a single failure as the quorum would be three. So that alert is technically correct, I assume you're put off by the summary saying that you would need a "majority"?

tjungblu:
.......

PR #852, opened 2022-06-10
Bug 2092880: add static pod revision metrics

tjungblu:


geliu2016 (approved):
/lgtm
/label cherry-pick-approved

pkg/etcdenvvar/envvarcontroller.go
```diff
@@ -57,6 +60,16 @@ func NewEnvVarController(
 	networkInformer configv1informers.NetworkInformer,
 	eventRecorder events.Recorder,
 ) *EnvVarController {
+
+	revGauge := prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd_operator",
+		Name:      "static_pod_current_revision",
```
wking, line 66: Checking one of [your presubmits here][2] in [PromeCIeus][1]:

```
sort_desc(count by (__name__) ({__name__=~".*revision.*"}))
```

finds:

```
etcd_debugging_auth_revision{}
etcd_debugging_mvcc_compact_revision{}
etcd_debugging_mvcc_current_revision{}
kube_statefulset_status_current_revision{}
kube_statefulset_status_update_revision{}
```

So... I don't see your new metric.  And maybe it should be called `static_pod_status_current_revision`?  Squatting on a generic metric name with etcd-specific code feels brittle...  `etcd_static_pod_status_current_revision`?  Or get the Gauge definition landed in library-go, where all interested static-pod consumers can share it?

[1]: https://promecieus.dptools.openshift.org/
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/852/pull-ci-openshift-cluster-etcd-operator-master-e2e-agnostic/1535170318547554304

pkg/etcdenvvar/envvarcontroller.go
```diff
@@ -137,6 +151,10 @@ func (c *EnvVarController) checkEnvVars() error {
 		return err
 	}
 
+	for _, status := range operatorStatus.NodeStatuses {
+		c.revisionGauge.WithLabelValues(status.NodeName).Set(float64(status.CurrentRevision))
```
wking, line 155: [Doesn't seem to work][1].  I think you're missing the piping-it-into-channel, like [this][2].  Although I'm a bit fuzzy on how the existing etcd-operator metrics are wired up.

[1]: https://github.com/openshift/cluster-etcd-operator/pull/852#discussion_r894935496
[2]: https://github.com/openshift/cluster-etcd-operator/blob/c9977ae3bd788a9e9a595001c5bc24995f9d0175/pkg/etcdcli/health.go#L278-L289

wking: Hrm, yeah, it looks like [that][1] is one way to do things.  And see [here][1] for an example using `Collect`.  Might need to ask the monitoring folks to help figure out why your current code doesn't seem to be working.

[1]: https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#hdr-A_Basic_Example
[2]: https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Collector

wking: Hmm, where are you setting up your metrics listener?  Grepping around in the operator code, I'm not finding it...

PR #857, opened 2022-06-15
Bug 2097431: fix degraded missing cluster version

Elbehery:
Fix going degraded on missing cluster version. 

As a result of https://github.com/openshift/cluster-etcd-operator/pull/835, the `Backupupgrade` controller while detecting the cluster version, during install and upgrade, raises an error and goes degraded. Such noise is not needed.

### DIFF
--- a/pkg/operator/upgradebackupcontroller/upgradebackupcontroller_test.go
+++ b/pkg/operator/upgradebackupcontroller/upgradebackupcontroller_test.go
@@ -123,8 +123,7 @@ func Test_ensureRecentBackup(t *testing.T) {
 				Conditions: []configv1.ClusterOperatorStatusCondition{
 					{Type: backupConditionType, Reason: backupSuccess, Message: "UpgradeBackup pre 4.9 located at path "}},
 			},
-			wantBackupStatus: configv1.ConditionUnknown,
-			wantEventCount:   1, // pod created event
+			wantNilBackupCondition: true,
 		},
 		{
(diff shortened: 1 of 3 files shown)


Elbehery:
cc @wking

pkg/operator/upgradebackupcontroller/upgradebackupcontroller_test.go
```diff
@@ -123,8 +123,8 @@ func Test_ensureRecentBackup(t *testing.T) {
 				Conditions: []configv1.ClusterOperatorStatusCondition{
 					{Type: backupConditionType, Reason: backupSuccess, Message: "UpgradeBackup pre 4.9 located at path "}},
 			},
-			wantBackupStatus: configv1.ConditionUnknown,
-			wantEventCount:   1, // pod created event
+			wantNilBackupCondition: true,
+			wantEventCount:         0, // pod created event
```
wking, line 127: `pod created event` doesn't make much sense with `wantEventCount: 0`, does it?

Elbehery: +1 👍🏽

Elbehery:
`ci/prow/e2e-aws-single-node` is failing due to pkg/operator/ceohelpers/common.go:9:2: imported and not used: "github.com/openshift/api/config/v1" as configv1

However, this PR removes the import :/

PR #858, opened 2022-06-15
Bug 2095703: fix multi-address member removal

tjungblu:
This fixes an issue where a machine can have multiple IPs (dualstack or not doesn't actually matter) and inadvertently removes the deletion hook as not all IPs are returned by the etcd member listing.

The fix is to check if any of the addresses of a machine is still returned from etcd, in such case the machine will remain with the hook. Otherwise we can safely remove the hook for this machine.

### DIFF
--- a/pkg/operator/machinedeletionhooks/machinedeletionhooks.go
+++ b/pkg/operator/machinedeletionhooks/machinedeletionhooks.go
@@ -2,17 +2,19 @@ package machinedeletionhooks
 
 import (
 	"context"
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"time"
 
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	machinev1beta1client "github.com/openshift/client-go/machine/clientset/versioned/typed/machine/v1beta1"
 	machinelistersv1beta1 "github.com/openshift/client-go/machine/listers/machine/v1beta1"
-	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
-	"github.com/openshift/cluster-etcd-operator/pkg/operator/ceohelpers"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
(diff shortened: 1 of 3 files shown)


tjungblu:
/retitle Bug 2095703: fix dualstack member removal
/hold cancel

tjungblu:
/retitle Bug 2095703: fix multi-address member removal

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -115,24 +118,33 @@ func (c *machineDeletionHooksController) attemptToDeleteMachineDeletionHook(ctx
...
 		if _, err := c.machineClient.Update(ctx, machinePendingDeletionCopy, metav1.UpdateOptions{}); err != nil {
-			errs = append(errs, err)
-			continue
+			return err
 		}
 		klog.V(2).Infof("successfully removed the deletion hook from machine %v", machinePendingDeletion.Name)
 		recorder.Eventf("MachineDeletionHook", "successfully removed the deletion hook from machine %v as it doesn't have a member", machinePendingDeletion.Name)
 	}
-	return kerrors.NewAggregate(errs)
+	return nil
```
Elbehery, line 147: @tjungblu Why we removed the error aggregation here ?

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
```diff
@@ -115,24 +118,33 @@ func (c *machineDeletionHooksController) attemptToDeleteMachineDeletionHook(ctx
...
+			klog.V(2).Infof("skip removing the deletion hook from machine %s since its member is still present with any of: %v", machinePendingDeletion.Name, machinePendingDeletion.Status.Addresses)
 			continue
 		}
+
 		machinePendingDeletionCopy := machinePendingDeletion.DeepCopy()
 		removeMachineDeletionHook(machinePendingDeletionCopy)
 		if _, err := c.machineClient.Update(ctx, machinePendingDeletionCopy, metav1.UpdateOptions{}); err != nil {
-			errs = append(errs, err)
-			continue
+			return err
```
Elbehery, line 142: Same as below, I think we should accumulate the errors for aggregation, or ?

Elbehery (approved):
@tjungblu nice work, I like it specially the testing 🎉

pkg/operator/machinedeletionhooks/machinedeletionhooks.go
tjungblu: why not fail directly when updating the machine fails? how many machines would be updated normally when running this code?

Elbehery:
`e2e-gcp-five-control-plane-replicas` is optional

/lgtm

hasbro17:
/override e2e-aws-single-node

CI was already green before and is failing for unrelated reasons.

PR #862, opened 2022-06-23
Bug 2093819: add timeout to health checks

tjungblu:
more details can be found in a sister bug with similar problems:
https://bugzilla.redhat.com/show_bug.cgi?id=2090628#c10

### DIFF
--- a/pkg/etcdcli/health.go
+++ b/pkg/etcdcli/health.go
@@ -20,6 +20,7 @@ func init() {
 
 const raftTermsMetricName = "etcd_debugging_raft_terms_total"
 
+// raftTermsCollector is thread-safe internally
 var raftTerms = &raftTermsCollector{
 	desc: prometheus.NewDesc(
 		raftTermsMetricName,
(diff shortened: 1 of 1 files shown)


pkg/etcdcli/health.go
```diff
@@ -41,66 +42,28 @@ type healthCheck struct {
...
-				hc.Error = fmt.Errorf("health check failed: %w", err)
-			}
-			cancel()
-			hch <- hc
-		}(member)
-	}
 
-	wg.Wait()
-	close(hch)
+		resChan := make(chan healthCheck, 1)
```
dusk125, line 52: resChan is never closed, might not matter since the GC will eat it after scope end

tjungblu: good catch, TIL that it's actually also not necessary in this context:

> In Go, the channel is not like the resources like file descriptor. It does not have to be closed, it is just an object, close (channel) is only used to tell the receivers that there is nothing to write, no other purpose.

https://faun.pub/implmenting-timeout-in-golang-ee2bc4aa6ae4

tjungblu: added the close and put the cancel on top, as I just saw the picker_wrapper.go implementation will cancel based on that:
https://github.com/grpc/grpc-go/blob/v1.40.0/picker_wrapper.go#L103-L117

which is where we were actually stuck somehow... I actually thought the deadline exceeded must've kicked in earlier here.. This was literally hanging for half an hour...

pkg/etcdcli/health.go
```diff
@@ -41,66 +42,34 @@ type healthCheck struct {
...
+		go func() {
+			ctx, cancel := context.WithTimeout(ctx, defaultTimeout)
+			defer cancel()
+
+			resChan <- checkSingleMemberHealth(ctx, member)
+		}()
 
-	wg.Wait()
-	close(hch)
+		select {
```
Elbehery, line 61: @tjungblu this select will be running within the loop, so it will block till the go routine write to the channel, or timeout. IMO it would be more concurrent if you spawn the routines from the loop, and after the loop you gather the result (i.e. the original impl using sync.WG) 

This way, you block within each iteration waiting for the result or timeout .. please correct me if I am wrong

Elbehery (commented):
Nice work @tjungblu  👍🏽 

left my review below

pkg/etcdcli/health.go
tjungblu: yeah that was the actual intention of removing it, I don't really understand what you gain from parallelizing three at max 10ms calls here on the happy path (30ms vs. 10ms?). Timeout sucks as it has a 30s penalty, but that is the same when you're parallelizing it.

Elbehery: gotcha, ok make sense then 👍🏽

tjungblu:
@bertinatto thanks, all green again ;)

tjungblu:
since I assume we're not going to make this one today, I assume we have to backport to 4.11.

/cherry-pick release-4.11

geliu2016 (approved):
/lgtm 
/label cherry-pick-approved

PR #865, opened 2022-06-28
[WIP] Bug 2023756

EmilyM1:
Might also fix Bug 2095667


EmilyM1:
@Elbehery Hi Mustafa, it's trying the yaml side to trigger child process exiting properly before parent processes complete.
If it works out I'll take the WIP off when it's ready, if not I'll convert it into trying go routines

PR #873, opened 2022-07-05
remove etcd-health-monitor sidecar from pod manifest

Elbehery:
This PR removes `etcd-health-monitor` side-car from Etcd Pod manifest. 

During https://github.com/openshift/cluster-etcd-operator/pull/869, we have found out that the `etcd-health-monitor` side car is of no use. Therefore, removing it.

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -271,41 +271,6 @@ ${COMPUTED_ENV_VARS}
         name: cert-dir
       - mountPath: /var/lib/etcd/
         name: data-dir
-  - name: etcd-health-monitor
-    image: ${OPERATOR_IMAGE}
-    imagePullPolicy: IfNotPresent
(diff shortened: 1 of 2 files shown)


tjungblu:
interesting, so makes absolutely no difference to whether or not it's there for a functional cluster and e2e tests. @hasbro17 do you think it's worth to remove it?

Elbehery:
@tjungblu  shall we close this ?

tjungblu:
no, why? I think we should merge this

Elbehery:
@tjungblu  Do we need a BZ ? 
@hasbro17 WDYT ?

tjungblu:
/lgtm

Nope, master does not need BZs for the next couple months.

dusk125:
Since the pod yaml was changed, did you regenerate the bindata?

Elbehery:
> Since the pod yaml was changed, did you regenerate the bindata?

Yes indeed 👍🏽

PR #904, opened 2022-08-02
[RELEASE-4.10] Bug 2095667: Etcd readiness probe leaves many zombie process after heavy load

EmilyM1:
This has the readiness updates that Mustafa had made to 4.10 pulled in and the tcp updates

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -193,15 +193,13 @@ ${COMPUTED_ENV_VARS}
         memory: 600Mi
         cpu: 300m
     readinessProbe:
-      httpGet:
-        port: 9980
-        path: readyz
-        scheme: HTTPS
-      initialDelaySeconds: 10
-      timeoutSeconds: 10
-      failureThreshold: 3
-      periodSeconds: 5
-      successThreshold: 1
+    tcpSocket:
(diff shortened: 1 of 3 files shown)


bindata/etcd/pod.yaml
```diff
@@ -193,15 +193,13 @@ ${COMPUTED_ENV_VARS}
...
-      httpGet:
-        port: 9980
-        path: readyz
-        scheme: HTTPS
-      initialDelaySeconds: 10
-      timeoutSeconds: 10
-      failureThreshold: 3
-      periodSeconds: 5
-      successThreshold: 1
+    tcpSocket:
```
tjungblu, line 196: I posted this on slack already, but two general questions:
* why do we need a TCP probe when we have a working http probe? The zombies came from the forking of the etcdctl process, which doesn't exist anymore
* your indentation actually removes the probe configuration, I reckon that's also why the build is green :)

tjungblu (changes requested):

bindata/etcd/pod.yaml
EmilyM1: I can try it with http. I had read a few places that tcp was best practices. Let me put back the working http, as I see what is going on with the indentation.

PR #912, opened 2022-08-26
OCPBUGS-614: ensure healthy quorum before config update

tjungblu:
that's for 4.9, finally. One more to go!


tjungblu:
was just testing this PR with AWS and it worked like a charm.

@geliu2016 could you kindly test the previous 4.10 version so we I can unblock this bug check?

geliu2016 (approved):
/label cherry-pick-approved

PR #913, opened 2022-08-26
ensure-env: propagate error

rphillips:
https://github.com/openshift/cluster-etcd-operator/pull/726 introduced a new sub-command `ensure-env`. The old code is designed to exit 1 in various cases. The refactor missed propagating the error so the [caller](https://github.com/openshift/cluster-etcd-operator/blob/master/cmd/cluster-etcd-operator/main.go#L48-L49) would exit 1.


rphillips:
@dusk125 closing in favor of Allen's fix, since this is failing unit tests.

PR #922, opened 2022-08-31
OCPBUGS-675: preserve log message on failures

tjungblu:
It's actually silly to take the member id from the response header when we know exactly which client url we're talking to in the first place.

### DIFF
--- a/pkg/etcdcli/health.go
+++ b/pkg/etcdcli/health.go
@@ -132,9 +132,7 @@ func checkSingleMemberHealth(ctx context.Context, member *etcdserverpb.Member) h
 		}
 		hc.Healthy = true
 	} else {
-		if resp.Header != nil {
-			klog.Errorf("health check for memberID (%v) failed: err(%v)", resp.Header.MemberId, err)
-		}
+		klog.Errorf("health check for member (%v) failed: err(%v)", member.Name, err)
 		hc.Error = fmt.Errorf("health check failed: %w", err)
 	}
 


kikisdeliveryservice:
Agree! Small q: wondering why client url instead of [member.Name](https://github.com/openshift/cluster-etcd-operator/blob/33fdaf64825c13f578254107f7e7f0d503c724c9/vendor/go.etcd.io/etcd/client/v2/members.go#L34)?

tjungblu:
I personally find client URL a bit easier to work with, because that follows the node names. The `member.Name` I believe is a random hex hash.

kikisdeliveryservice:
> I personally find client URL a bit easier to work with, because that follows the node names. The `member.Name` I believe is a random hex hash.

It seems that yall are [using member.Name](https://github.com/openshift/cluster-etcd-operator/blob/33fdaf64825c13f578254107f7e7f0d503c724c9/pkg/etcdcli/etcdcli.go#L294) in quite a few places for ex: https://github.com/openshift/cluster-etcd-operator/blob/33fdaf64825c13f578254107f7e7f0d503c724c9/pkg/etcdcli/health.go#L68 https://github.com/openshift/cluster-etcd-operator/blob/7e439f359b48fcc4ef124c015ed131a16018cc9f/pkg/etcdcli/etcdcli.go#L221

member.Name seems to look like: `unhealthy members: ci-op-w6q7nmmc-f92fc-s5258-master-1` (from the event log)

kikisdeliveryservice:
> right, thanks for double checking. I think I mistook this for member.id?
> 
> What if etcd can't be reached at all? How does it figure out the member name? At least the client url is always present without network connectivity.

Snooping around the codebase, on an unstarted member it looks like:
`UnstartedEtcdMember | unstarted members: NAME-PENDING-10.0.166.124`

This file seems to use func that generates this:
https://github.com/openshift/cluster-etcd-operator/blob/33fdaf64825c13f578254107f7e7f0d503c724c9/pkg/etcdcli/etcdcli.go#L294

so that func ends up using hostname if member.name doesn't work:
https://github.com/openshift/cluster-etcd-operator/blob/33fdaf64825c13f578254107f7e7f0d503c724c9/pkg/etcdcli/etcdcli.go#L301

PR #929, opened 2022-09-12
Bug 2066671 

EmilyM1:


manifests/0000_20_etcd-operator_10_flowschema.yaml
```diff
@@ -14,16 +14,16 @@ spec:
...
+      - resourcesValue
       verbs:
-      - '*'
+      - verbValue
     subjects:
     - kind: ServiceAccount
       serviceAccount:
-        name: etcd-operator
-        namespace: openshift-etcd-operator
+        namespace: etcd-operator
```
tjungblu, line 28: there's no such namespace

manifests/0000_20_etcd-operator_04_clusterrolebinding.yaml
```diff
@@ -11,4 +11,4 @@ roleRef:
 subjects:
```
tjungblu, line 11: can't comment on the right line, but I thought the goal was to get rid of the cluster-admin role

PR #935, opened 2022-09-22
certificate code cleanup

stlaz:
This PR removes redundant certificate creation code and comments.


stlaz:
/retest-required
looks like OVN serial tests got broken but not as a consequence of this change

stlaz:
/retest-required
It's weird that CI demands `ci/prow/e2e-aws-ovn-single-node` to pass given it's not really required. Also, it does not pass because of other components :eyes:

PR #937, opened 2022-09-28
OCPBUGS-1000: Allow scale-down of unhealthy member

Elbehery:
Currently the clustermemberremoval controller will automatically remove an etcd member when the pre-conditions for scale down are met, i.e we have a machine pending deletion and N+1 voting members where N is the desired control-plane size.
However, if we have the scaling down conditions met, but the etcd cluster is unhealthy (i.e. unhealthy voting member found), 
clustermemberremoval will not attempt to scale down. 

This PR relaxes the scaling down condition, which allows clustermemberremoval to remove unhealthy voting member, if the current voting members count is N+1, while N is the `desiredControlPlaneReplicaCount`.

### DIFF
--- a/pkg/etcdcli/helpers.go
+++ b/pkg/etcdcli/helpers.go
@@ -77,6 +77,11 @@ func (f *fakeEtcdClient) MemberList(ctx context.Context) ([]*etcdserverpb.Member
 	return f.members, nil
 }
 
+func (f *fakeEtcdClient) VotingMemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	members, _ := f.MemberList(ctx)
+	return filterVotingMembers(members), nil
+}
+
 func (f *fakeEtcdClient) MemberRemove(ctx context.Context, memberID uint64) error {
 	var memberExists bool
 	for _, m := range f.members {
@@ -139,12 +144,22 @@ func (f *fakeEtcdClient) UnhealthyMembers(ctx context.Context) ([]*etcdserverpb.
 	return []*etcdserverpb.Member{}, nil
 }
 
+func (f *fakeEtcdClient) UnhealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	members, _ := f.UnhealthyMembers(ctx)
+	return filterVotingMembers(members), nil
+}
+
 func (f *fakeEtcdClient) HealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
 	if f.opts.healthyMember > 0 {
 		// healthy start from end
 		return f.members[f.opts.unhealthyMember:], nil
 	}
 	return []*etcdserverpb.Member{}, nil
+}
+
+func (f *fakeEtcdClient) HealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	members, _ := f.HealthyMembers(ctx)
+	return filterVotingMembers(members), nil
 }
 
 func (f *fakeEtcdClient) MemberStatus(ctx context.Context, member *etcdserverpb.Member) string {
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
@@ -178,18 +178,41 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 	if err != nil {
 		return fmt.Errorf("could not get a list of unhealthy members: %v", err)
 	}
+
+	// map machines pending deletion to node internal ip
+	votingMachinesPendingDeletionIndex := ceohelpers.IndexMachinesByNodeInternalIP(votingMachinesPendingDeletion)
+	// find unhealthy machine pending deletion to remove first from the cluster, without violating quorum
+	var unhealthyMachinePendingDeletion []*machinev1beta1.Machine
 	if len(unhealthyMembers) > 0 {
 		var unhealthyMembersURLs []string
 		for _, unhealthyMember := range unhealthyMembers {
+			unhealthyMemberNodeInternalIP, err := ceohelpers.MemberToNodeInternalIP(unhealthyMember)
+			if err != nil {
+				return fmt.Errorf("cannot get node internal ip for unhealthy member %v: %w", unhealthyMember, err)
+			}
+			unhealthyMemberMachine, ok := votingMachinesPendingDeletionIndex[unhealthyMemberNodeInternalIP]
+			if ok {
+				unhealthyMachinePendingDeletion = append(unhealthyMachinePendingDeletion, unhealthyMemberMachine)
+			}
 			if len(unhealthyMember.PeerURLs) > 0 {
 				unhealthyMembersURLs = append(unhealthyMembersURLs, unhealthyMember.PeerURLs[0])
 			} else {
 				unhealthyMembersURLs = append(unhealthyMembersURLs, unhealthyMember.Name)
 			}
 		}
-		return fmt.Errorf("cannot proceed wth scaling down, unhealthy members found: %v", unhealthyMembersURLs)
+		if len(unhealthyMachinePendingDeletion) > 0 {
+			klog.V(4).Infof("found unhealthy voting members with machine pending deletion: %v", unhealthyMachinePendingDeletion)
+			klog.V(4).Infof("unhealthy members found: %v", unhealthyMembersURLs)
+		} else {
+			return fmt.Errorf("cannot proceed with scaling down, unhealthy members found: %v", unhealthyMembersURLs)
+		}
 	}
 
+	// remove the unhealthy machine pending deletion first
+	// if no unhealthy machine pending deletion found, then attempt to scale down the healthy machines pending deletion
+	if len(unhealthyMachinePendingDeletion) > 0 {
+		votingMachinesPendingDeletion = append(unhealthyMachinePendingDeletion, votingMachinesPendingDeletion...)
+	}
 	var allErrs []error
 	for _, votingMachinePendingDeletion := range votingMachinesPendingDeletion {
 		removed, errs := c.attemptToRemoveMemberFor(ctx, liveVotingMembers, votingMachinePendingDeletion, recorder)
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
@@ -412,21 +412,19 @@ func (c *clusterMemberRemovalController) attemptToRemoveMemberFor(ctx context.Co
 }
 
 func (c *clusterMemberRemovalController) getHealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
-	healthyMembers, err := c.etcdClient.HealthyMembers(ctx)
+	members, err := c.etcdClient.HealthyVotingMembers(ctx)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to list healthy voting members: %w", err)
 	}
-	healthyLiveVotingMembers := getVotingMembers(healthyMembers)
-	return healthyLiveVotingMembers, nil
+	return members, nil
 }
 
 func (c *clusterMemberRemovalController) getAllVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
-	allMembers, err := c.etcdClient.MemberList(ctx)
+	members, err := c.etcdClient.VotingMemberList(ctx)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to list all voting members: %w", err)
 	}
-	liveVotingMembers := getVotingMembers(allMembers)
-	return liveVotingMembers, nil
+	return members, nil
 }
 
 func hasInternalIP(machine *machinev1beta1.Machine, memberInternalIP string) bool {
@@ -436,15 +434,4 @@ func hasInternalIP(machine *machinev1beta1.Machine, memberInternalIP string) boo
 		}
 	}
 	return false
-}
-
-func getVotingMembers(members []*etcdserverpb.Member) []*etcdserverpb.Member {
-	var votingMembers []*etcdserverpb.Member
-	for _, member := range members {
-		if member.IsLearner {
-			continue
-		}
-		votingMembers = append(votingMembers, member)
-	}
-	return votingMembers
 }
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
@@ -182,7 +182,7 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 	// map machines pending deletion to node internal ip
 	votingMachinesPendingDeletionIndex := ceohelpers.IndexMachinesByNodeInternalIP(votingMachinesPendingDeletion)
 	// find unhealthy machine pending deletion to remove first from the cluster, without violating quorum
-	var unhealthyMachinePendingDeletion []*machinev1beta1.Machine
+	var unhealthyMachinesPendingDeletion []*machinev1beta1.Machine
 	if len(unhealthyMembers) > 0 {
 		var unhealthyMembersURLs []string
 		for _, unhealthyMember := range unhealthyMembers {
@@ -192,7 +192,7 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 			}
 			unhealthyMemberMachine, ok := votingMachinesPendingDeletionIndex[unhealthyMemberNodeInternalIP]
 			if ok {
-				unhealthyMachinePendingDeletion = append(unhealthyMachinePendingDeletion, unhealthyMemberMachine)
+				unhealthyMachinesPendingDeletion = append(unhealthyMachinesPendingDeletion, unhealthyMemberMachine)
 			}
 			if len(unhealthyMember.PeerURLs) > 0 {
 				unhealthyMembersURLs = append(unhealthyMembersURLs, unhealthyMember.PeerURLs[0])
@@ -200,8 +200,8 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 				unhealthyMembersURLs = append(unhealthyMembersURLs, unhealthyMember.Name)
 			}
 		}
-		if len(unhealthyMachinePendingDeletion) > 0 {
-			klog.V(4).Infof("found unhealthy voting members with machine pending deletion: %v", unhealthyMachinePendingDeletion)
+		if len(unhealthyMachinesPendingDeletion) > 0 {
+			klog.V(4).Infof("found unhealthy voting members with machine pending deletion: %v", unhealthyMachinesPendingDeletion)
 			klog.V(4).Infof("unhealthy members found: %v", unhealthyMembersURLs)
 		} else {
 			return fmt.Errorf("cannot proceed with scaling down, unhealthy members found: %v", unhealthyMembersURLs)
@@ -210,8 +210,8 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 
 	// remove the unhealthy machine pending deletion first
 	// if no unhealthy machine pending deletion found, then attempt to scale down the healthy machines pending deletion
-	if len(unhealthyMachinePendingDeletion) > 0 {
-		votingMachinesPendingDeletion = append(unhealthyMachinePendingDeletion, votingMachinesPendingDeletion...)
+	if len(unhealthyMachinesPendingDeletion) > 0 {
+		votingMachinesPendingDeletion = append(unhealthyMachinesPendingDeletion, votingMachinesPendingDeletion...)
 	}
 	var allErrs []error
 	for _, votingMachinePendingDeletion := range votingMachinesPendingDeletion {
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
@@ -130,43 +130,35 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 		return nil
 	}
 
+	// machines with master role and deletion hook
 	memberMachines, err := ceohelpers.CurrentMemberMachinesWithDeletionHooks(c.masterMachineSelector, c.masterMachineLister)
 	if err != nil {
 		return err
 	}
 
-	var votingMachines []*machinev1beta1.Machine
+	var votingMembersMachines []*machinev1beta1.Machine
 	for memberMachineIP, memberMachine := range ceohelpers.IndexMachinesByNodeInternalIP(memberMachines) {
 		if currentVotingMemberIPListSet.Has(memberMachineIP) {
-			votingMachines = append(votingMachines, memberMachine)
+			votingMembersMachines = append(votingMembersMachines, memberMachine)
 		}
 	}
 
-	votingMachinesPendingDeletion := ceohelpers.FilterMachinesPendingDeletion(votingMachines)
-	if len(votingMachinesPendingDeletion) == 0 {
+	votingMembersMachinesPendingDeletion := ceohelpers.FilterMachinesPendingDeletion(votingMembersMachines)
+	if len(votingMembersMachinesPendingDeletion) == 0 {
 		return nil
 	}
 
-	// based on the data in the cache it looks like we have something to do
-	// get the live members, to observe the current state and the member IDs
-	members, err := c.etcdClient.MemberList(ctx)
+	// do not trust data in the cache, compare with the current state
+	healthyLiveVotingMembers, err := c.getHealthyVotingMembers(ctx)
 	if err != nil {
 		return err
 	}
 
-	var liveVotingMembers []*etcdserverpb.Member
-	for _, member := range members {
-		if member.IsLearner {
-			continue
-		}
-		liveVotingMembers = append(liveVotingMembers, member)
-	}
-
-	// do not trust data in the cache, compare with the current state
-	if len(liveVotingMembers) <= desiredControlPlaneReplicasCount {
-		klog.V(2).Infof("Ignoring scale down since the number of live etcd voting members (%d) < desired number of control-plane replicas (%d) ", len(liveVotingMembers), desiredControlPlaneReplicasCount)
+	// scaling down invariant
+	if len(healthyLiveVotingMembers) < desiredControlPlaneReplicasCount {
+		klog.V(2).Infof("Ignoring scale down since the number of healthy live etcd voting members (%d) < desired number of control-plane replicas (%d) ", len(healthyLiveVotingMembers), desiredControlPlaneReplicasCount)
 		if time.Now().After(c.lastTimeScaleDownEventWasSent.Add(5 * time.Minute)) {
-			recorder.Eventf("ScaleDown", "Ignoring scale down since the number of live etcd voting members (%d) < desired number of control-plane replicas (%d) ", len(liveVotingMembers), desiredControlPlaneReplicasCount)
+			recorder.Eventf("ScaleDown", "Ignoring scale down since the number of healthy live etcd voting members (%d) < desired number of control-plane replicas (%d) ", len(healthyLiveVotingMembers), desiredControlPlaneReplicasCount)
 			c.lastTimeScaleDownEventWasSent = time.Now()
 		}
 		return nil
@@ -180,9 +172,9 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 	}
 
 	// map machines pending deletion to node internal ip
-	votingMachinesPendingDeletionIndex := ceohelpers.IndexMachinesByNodeInternalIP(votingMachinesPendingDeletion)
+	votingMemberMachinesPendingDeletionIndex := ceohelpers.IndexMachinesByNodeInternalIP(votingMembersMachinesPendingDeletion)
 	// find unhealthy machine pending deletion to remove first from the cluster, without violating quorum
-	var unhealthyMachinesPendingDeletion []*machinev1beta1.Machine
+	var unhealthyVotingMemberMachinesPendingDeletion []*machinev1beta1.Machine
 	if len(unhealthyMembers) > 0 {
 		var unhealthyMembersURLs []string
 		for _, unhealthyMember := range unhealthyMembers {
@@ -190,9 +182,9 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 			if err != nil {
 				return fmt.Errorf("cannot get node internal ip for unhealthy member %v: %w", unhealthyMember, err)
 			}
-			unhealthyMemberMachine, ok := votingMachinesPendingDeletionIndex[unhealthyMemberNodeInternalIP]
+			unhealthyVotingMemberMachinePendingDeletion, ok := votingMemberMachinesPendingDeletionIndex[unhealthyMemberNodeInternalIP]
 			if ok {
-				unhealthyMachinesPendingDeletion = append(unhealthyMachinesPendingDeletion, unhealthyMemberMachine)
+				unhealthyVotingMemberMachinesPendingDeletion = append(unhealthyVotingMemberMachinesPendingDeletion, unhealthyVotingMemberMachinePendingDeletion)
 			}
 			if len(unhealthyMember.PeerURLs) > 0 {
 				unhealthyMembersURLs = append(unhealthyMembersURLs, unhealthyMember.PeerURLs[0])
@@ -200,22 +192,28 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 				unhealthyMembersURLs = append(unhealthyMembersURLs, unhealthyMember.Name)
 			}
 		}
-		if len(unhealthyMachinesPendingDeletion) > 0 {
-			klog.V(4).Infof("found unhealthy voting members with machine pending deletion: %v", unhealthyMachinesPendingDeletion)
+		if len(unhealthyVotingMemberMachinesPendingDeletion) > 0 {
+			klog.V(4).Infof("found unhealthy voting members with machine pending deletion: %v", unhealthyVotingMemberMachinesPendingDeletion)
 			klog.V(4).Infof("unhealthy members found: %v", unhealthyMembersURLs)
 		} else {
-			return fmt.Errorf("cannot proceed with scaling down, unhealthy members found: %v", unhealthyMembersURLs)
+			return fmt.Errorf("cannot proceed with scaling down, unhealthy voting members found: %v, none are pending deletion", unhealthyMembersURLs)
 		}
 	}
 
 	// remove the unhealthy machine pending deletion first
 	// if no unhealthy machine pending deletion found, then attempt to scale down the healthy machines pending deletion
-	if len(unhealthyMachinesPendingDeletion) > 0 {
-		votingMachinesPendingDeletion = append(unhealthyMachinesPendingDeletion, votingMachinesPendingDeletion...)
+	if len(unhealthyVotingMemberMachinesPendingDeletion) > 0 {
+		votingMembersMachinesPendingDeletion = append(unhealthyVotingMemberMachinesPendingDeletion, votingMembersMachinesPendingDeletion...)
 	}
+
+	liveVotingMembers, err := c.getAllVotingMembers(ctx)
+	if err != nil {
+		return err
+	}
+
 	var allErrs []error
-	for _, votingMachinePendingDeletion := range votingMachinesPendingDeletion {
-		removed, errs := c.attemptToRemoveMemberFor(ctx, liveVotingMembers, votingMachinePendingDeletion, recorder)
+	for _, votingMemberMachinePendingDeletion := range votingMembersMachinesPendingDeletion {
+		removed, errs := c.attemptToRemoveMemberFor(ctx, liveVotingMembers, votingMemberMachinePendingDeletion, recorder)
 		if removed {
 			break // we want to remove only one member at a time
 		}
@@ -413,6 +411,24 @@ func (c *clusterMemberRemovalController) attemptToRemoveMemberFor(ctx context.Co
 	return removed, errs
 }
 
+func (c *clusterMemberRemovalController) getHealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	healthyMembers, err := c.etcdClient.HealthyMembers(ctx)
+	if err != nil {
+		return nil, err
+	}
+	healthyLiveVotingMembers := getVotingMembers(healthyMembers)
+	return healthyLiveVotingMembers, nil
+}
+
+func (c *clusterMemberRemovalController) getAllVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	allMembers, err := c.etcdClient.MemberList(ctx)
+	if err != nil {
+		return nil, err
+	}
+	liveVotingMembers := getVotingMembers(allMembers)
+	return liveVotingMembers, nil
+}
+
 func hasInternalIP(machine *machinev1beta1.Machine, memberInternalIP string) bool {
 	for _, addr := range machine.Status.Addresses {
 		if addr.Type == corev1.NodeInternalIP && addr.Address == memberInternalIP {
@@ -420,4 +436,15 @@ func hasInternalIP(machine *machinev1beta1.Machine, memberInternalIP string) boo
 		}
 	}
 	return false
+}
+
+func getVotingMembers(members []*etcdserverpb.Member) []*etcdserverpb.Member {
+	var votingMembers []*etcdserverpb.Member
+	for _, member := range members {
+		if member.IsLearner {
+			continue
+		}
+		votingMembers = append(votingMembers, member)
+	}
+	return votingMembers
 }
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
@@ -685,7 +685,7 @@ func TestAttemptToScaleDown(t *testing.T) {
 				return []runtime.Object{cm}
 			}(),
 			fakeEtcdClientOptions: etcdcli.WithFakeClusterHealth(&etcdcli.FakeMemberHealth{Healthy: 3, Unhealthy: 1}),
-			expectedError:         fmt.Errorf("cannot proceed wth scaling down, unhealthy members found: [https://10.0.139.78:1234]"),
+			expectedError:         fmt.Errorf("cannot proceed with scaling down, unhealthy members found: [https://10.0.139.78:1234]"),
 			validateFn: func(t *testing.T, fakeEtcdClient etcdcli.EtcdClient) {
 				memberList, err := fakeEtcdClient.MemberList(context.TODO())
 				if err != nil {
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
@@ -426,6 +426,7 @@ func TestAttemptToScaleDown(t *testing.T) {
 				cm.Data["m-4"] = "10.0.139.81"
 				return []runtime.Object{cm}
 			}(),
+			fakeEtcdClientOptions: etcdcli.WithFakeClusterHealth(&etcdcli.FakeMemberHealth{Healthy: 4, Unhealthy: 0}),
 			validateFn: func(t *testing.T, fakeEtcdClient etcdcli.EtcdClient) {
 				memberList, err := fakeEtcdClient.MemberList(context.TODO())
 				if err != nil {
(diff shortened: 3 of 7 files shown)


pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -177,19 +177,50 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 	if err != nil {
 		return fmt.Errorf("could not get a list of unhealthy members: %v", err)
 	}
+
+	// map machines pending deletion to node internal ip
+	votingMachinesPendingDeletionIndex := ceohelpers.IndexMachinesByNodeInternalIP(votingMachinesPendingDeletion)
```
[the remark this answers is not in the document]
tjungblu: here the usage seems safe, but please add a unit test to exercise a node with multiple IPs pending deletion. 

also cc @geliu2016 fyi for verifying on dualstack vsphere

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -177,19 +177,50 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
...
 	if len(unhealthyMembers) > 0 {
 		var unhealthyMembersURLs []string
 		for _, unhealthyMember := range unhealthyMembers {
+			unhealthyMemberNodeInternalIP, err := ceohelpers.MemberToNodeInternalIP(unhealthyMember)
+			if err != nil {
+				return fmt.Errorf("cannot get node internal ip for unhealthy member %v: %w", unhealthyMember, err)
+			}
+			unhealthyMemberMachine, ok := votingMachinesPendingDeletionIndex[unhealthyMemberNodeInternalIP]
+			if ok {
+				unhealthyMachinePendingDeletion = unhealthyMemberMachine
```
tjungblu, line 194: why not append to `votingMachinesPendingDeletion` and break the loop?

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
Elbehery: >  can you ensure this PR also works on dualstack setups?
 sure, I am trying to create a cluster with this PR atop, but I need to use the installer, once I test this i will try with dual stack 👍🏽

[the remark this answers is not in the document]
Elbehery: > but please add a unit test to exercise a node with multiple IPs pending deletion. 
I am working on the test now, will add this 

Do you know if there is a similar test using dual stack somewhere ? just I wanna see how to create the scenario to avoid confusion :)

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
Elbehery: gr8 question 

so I do not use a slice, the last unhealthy member with machine pending deletion found, will be removed first .. 
then next sync we continue, to guarantee we do not break quorum, so we do it one by one, and if the current voting members size is <= the `controlPlaneReplicaCount` then we do nothing, even if we have unhealthy member

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
tjungblu: I added one in the above PR:
https://github.com/openshift/cluster-etcd-operator/pull/858/files#diff-b3ed5284128aa9fb29b8b026649f6401ca9e67922974c15ddff2a795d0225d76R345

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -177,19 +177,50 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
...
-		return fmt.Errorf("cannot proceed wth scaling down, unhealthy members found: %v", unhealthyMembersURLs)
+		if unhealthyMachinePendingDeletion != nil {
+			klog.V(4).Infof("found unhealthy voting member's machine pending deletion: %v, attempt to scale this member down since it does not violate quorum", unhealthyMachinePendingDeletion)
+			klog.V(4).Infof("unhealthy members found: %v", unhealthyMembersURLs)
+		} else {
+			return fmt.Errorf("cannot proceed wth scaling down, unhealthy members found: %v", unhealthyMembersURLs)
+		}
 	}
 
+	// trying to remove the unhealthy machine pending deletion first
```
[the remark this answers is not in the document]
Elbehery: gr8 question, 

so there could be many machines pending deletions, however, we would need to remove the `unhealthy` members first. Otherwise, we could remove a healthy member pending deletion, while there is an `unhealthy` member pending deletion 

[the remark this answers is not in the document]
tjungblu: still thinking about it, you might want to simplify the logic by just putting the unhealthy member at the start of the `votingMachinesPendingDeletion` slice. That way we don't duplicate the deletion logic. It will also stop on the first deletion.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
tjungblu: > then next sync we continue, to guarantee we do not break quorum, so we do it one by one, and if the current voting members size is <= the controlPlaneReplicaCount then we do nothing, even if we have unhealthy member

shouldn't we better fallback to the error when there are too many unhealthy members?
> return fmt.Errorf("cannot proceed wth scaling down, unhealthy members found: %v", unhealthyMembersURLs)

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
Elbehery: I agree with you, I will try to fix it now 👍🏽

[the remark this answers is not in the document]
Elbehery: @tjungblu  fixed :) .. can u have a look now please ?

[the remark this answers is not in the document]
tjungblu: much better! thank you sir

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -177,18 +177,39 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
...
 			} else {
 				unhealthyMembersURLs = append(unhealthyMembersURLs, unhealthyMember.Name)
 			}
 		}
-		return fmt.Errorf("cannot proceed wth scaling down, unhealthy members found: %v", unhealthyMembersURLs)
+		if len(unhealthyMachinePendingDeletion) > 0 {
+			klog.V(4).Infof("found unhealthy voting members with machine pending deletion: %v", unhealthyMachinePendingDeletion)
+			klog.V(4).Infof("unhealthy members found: %v", unhealthyMembersURLs)
+		} else {
+			return fmt.Errorf("cannot proceed wth scaling down, unhealthy members found: %v", unhealthyMembersURLs)
```
tjungblu, line 206: maybe it's already too late for my brain, but wouldn't this never delete a node when there are only healthy members?

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
tjungblu: depends how "safe" you want to play it, I think it's better to be explicit and do not attempt to delete anything when you know there are too many unhealthy nodes. I guess there was also a good reason that check was in there before.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
tjungblu: okay wait, this is confusing:

> if len(unhealthyMembers) > 0 

so there are some unhealthy members, then we join with the pending deletions. If there are no unhealthy members pending deletion we are returning an error?

gotta look at this tomorrow again :) let me know what you think @Elbehery

Elbehery: Gr8 Question .. 

So if this  `if len(unhealthyMembers) > 0` condition is true, then we have unhealthy members. Usually we should **not** allow scaling down in this case, however, if the unhealthy voting member is `pending deletion` && N+1 voting members exist, where N is the `desiredControlPlaneReplicaCount` .. then we should remove the unhealthy voting member whose machine is pending deletion .. 

On the other hand, when the condition `if len(unhealthyMembers) > 0` is false, then we skip this whole block and proceed to delete voting members who machine is pending deletion .. 

cc @hasbro17 @JoelSpeed

JoelSpeed: How would we get to `N+1` voting members? Normally we have `N`, one goes unhealthy, we can't add a new member while one of the `N` is unhealthy, we can only remove. So as far as i can tell, as soon as one voting member goes unhealthy, the only way we can go is down from `N`.

Do we mean pods here rather than voting members?

Elbehery:
I tested on a 4.12 cluster && this PR atop .. 

Scenario : 
- Create a new master machine
- wait till the new etcd pod joins the etcd cluster and the cluster is healthy / stable
- Terminate the ec2 instance backing the master-0 machine
- wait till the machine becomes `failed` 
- the etcd cluster is unhealthy due to unreachable voting member
- run `oc delete machine master-0`
- the etcd voting member within master-0 is being removed from the cluster, even though the cluster is `unhealthy`
- the etcd cluster recovers and becomes healthy again with 3 voting member ( i.e. desiredControlPlaneReplicaCount` )


Logs from etcd operator 

```
I0929 11:40:05.548255       1 machinedeletionhooks.go:150] skip removing the deletion hook from machine ci-ln-jyvkrct-76ef8-7wxlj-master-0 since its member is still present with any of: [{InternalIP 10.0.167.93} {InternalDNS ip-10-0-167-93.us-east-2.compute.internal} {Hostname ip-10-0-167-93.us-east-2.compute.internal}]
I0929 11:40:05.777591       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"8dfd52b2-879d-4a48-85a7-bfc6c2c99fe1", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Warning' reason: 'UnhealthyEtcdMember' unhealthy members: ip-10-0-167-93.us-east-2.compute.internal
I0929 11:40:05.785685       1 clustermemberremovalcontroller.go:407] successfully removed member: [ url: 10.0.167.93, name: ip-10-0-167-93.us-east-2.compute.internal, id: 1403051452037064279 ] from the cluster
I0929 11:40:05.786373       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"8dfd52b2-879d-4a48-85a7-bfc6c2c99fe1", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'MemberRemove' removed member with ID: 1403051452037064279
I0929 11:40:05.786586       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"8dfd52b2-879d-4a48-85a7-bfc6c2c99fe1", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'ScaleDown' successfully removed member: [ url: 10.0.167.93, name: ip-10-0-167-93.us-east-2.compute.internal, id: 1403051452037064279 ] from the cluster
I0929 11:40:05.794345       1 clustermemberremovalcontroller.go:166] Ignoring scale down since the number of live etcd voting members (3) < desired number of control-plane replicas (3) 
I0929 11:40:05.794615       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"8dfd52b2-879d-4a48-85a7-bfc6c2c99fe1", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'ScaleDown' Ignoring scale down since the number of live etcd voting members (3) < desired number of control-plane replicas (3) 
I0929 11:40:08.582614       1 machinedeletionhooks.go:136] current members [ID:4311233373965647731 name:"ip-10-0-159-85.us-east-2.compute.internal" peerURLs:"https://10.0.159.85:2380" clientURLs:"https://10.0.159.85:2379"  ID:6911491470152851383 name:"ip-10-0-235-212.us-east-2.compute.internal" peerURLs:"https://10.0.235.212:2380" clientURLs:"https://10.0.235.212:2379"  ID:12883664184677543177 name:"ip-10-0-171-169.us-east-2.compute.internal" peerURLs:"https://10.0.171.169:2380" clientURLs:"https://10.0.171.169:2379" ] with IPSet: map[10.0.159.85:{} 10.0.171.169:{} 10.0.235.212:{}]
I0929 11:40:08.604048       1 machinedeletionhooks.go:159] successfully removed the deletion hook from machine ci-ln-jyvkrct-76ef8-7wxlj-master-0
I0929 11:40:08.604388       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"8dfd52b2-879d-4a48-85a7-bfc6c2c99fe1", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'MachineDeletionHook' successfully removed the deletion hook from machine ci-ln-jyvkrct-76ef8-7wxlj-master-0 as it doesn't have a member
```

logs from etcd member pod 

```
{"level":"info","ts":"2022-09-29T11:45:54.719Z","caller":"membership/cluster.go:473","msg":"removed member","cluster-id":"d3e9ce9abfaf1868","local-member-id":"b2cbf850bd20ad09","removed-remote-peer-id":"1378a3dca843e657","removed-remote-peer-urls":["https://10.0.167.93:2380"],"removed-remote-peer-is-learner":false}
```

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -177,18 +177,41 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
...
 			if len(unhealthyMember.PeerURLs) > 0 {
 				unhealthyMembersURLs = append(unhealthyMembersURLs, unhealthyMember.PeerURLs[0])
 			} else {
 				unhealthyMembersURLs = append(unhealthyMembersURLs, unhealthyMember.Name)
 			}
 		}
-		return fmt.Errorf("cannot proceed wth scaling down, unhealthy members found: %v", unhealthyMembersURLs)
+		if len(unhealthyMachinePendingDeletion) > 0 {
+			klog.V(4).Infof("found unhealthy voting members with machine pending deletion: %v", unhealthyMachinePendingDeletion)
+			klog.V(4).Infof("unhealthy members found: %v", unhealthyMembersURLs)
```
Elbehery, line 204: these two logs did not show up on the operator logs .. I expected them to be there, needs more investigation

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: > cannot proceed with scaling down, unhealthy members found

The voting members are etcd pods which are members of the etcd cluster and **not** learner

> How would we get to N+1 voting members

Assuming N which is `desiredControlPlaneReplicaCount` is `3` 

We added a new master machine, so now we have `4` master machine (i.e. voting etcd members). Now one goes unhealthy / failed. Running `oc delete machine` on the failed machine, will remove the corresponding etcd voting member from the etcd cluster. This is allowed since we have N+1 members while the required for Quorum safety is only N

> Normally we have N, one goes unhealthy, we can't add a new member while one of the N is unhealthy, we can only remove . So as far as i can tell, as soon as one voting member goes unhealthy, the only way we can go is down from N.

So as discussed with @hasbro17 , currently we do not allow scaling up while the cluster is unhealthy.

JoelSpeed: > We added a new master machine, so now we have 4 master machine (i.e. voting etcd members). Now one goes unhealthy / failed.

The scenario I would like to resolve is what happens when a Machine goes unhealthy *before* you are able to add the 4th master machine. This is likely what will happen in the real world. A machine goes down for some reason, so you want to replace that, you ask for the machine to be deleted, you expect everything to recover, right now it won't, and if we rely on this N+1 voting members instead of just N+1 pods, I still don't believe that will resolve the issue.

When the cluster is unhealthy and there is a new node waiting to be added as a learner, IMO, we should scale down to allow the new learner to be added. I don't think what's described is resolving that issue just yet

Elbehery: > When the cluster is unhealthy and there is a new node waiting to be added as a learner, IMO, we should scale down to allow the new learner to be added. I don't think what's described is resolving that issue just yet 

this most probably violates Quorum safety, and needs a new design

Elbehery: wdyt herr @tjungblu ?

tjungblu: > The scenario I would like to resolve is what happens when a Machine goes unhealthy before you are able to add the 4th master machine. This is likely what will happen in the real world. A machine goes down for some reason, so you want to replace that,

Yeah that's true, it's unsafe however. On n=3, one unhealthy node already brings you to the point where you have to still have quorum but any additional failure will break quorum. Adding a learner in that scenario will make it more likely that you overload the remaining two that also must handle both reads + writes of the usual cluster traffic. Imagine that third member was overloaded and thus became unhealthy - now you're in a cascading failure.

It's still better IMHO to have a human go and evaluate the health of the lost member and ensure that the others are sufficiently able to handle another learner member or restore directly from snapshot. Having said that, this is just being conservative on my end. Maybe @hasbro17 has another take on this.

Since https://github.com/openshift/cluster-etcd-operator/pull/872 we're also stopping to rollout any static revision as this will break quorum. Before the PR adding a new learner would definitely create a new static pod revision and that certainly broke quorum in lots of clusters already rendering them unusable.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
tjungblu: you can check the unit test failures, I believe that will tell you what's wrong.

Elbehery: Yes I am working on it now :) 👍🏽

Elbehery: it was a type in the expected error :D :D

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
JoelSpeed: > It's still better IMHO to have a human go and evaluate the health of the lost member and ensure that the others are sufficiently able to handle another learner member or restore directly from snapshot

I guess my retort to that is that, if the user deleted the Machine, that's their signal this is going away, but maybe you think the second, oh right, I need to check etcd, step is of additional help

Elbehery:
@JoelSpeed  I added two test cases to cover the changes, please review 👍🏽

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
hasbro17: So just to clarify, this bug/PR is specifically meant to address the scenario of removing an unhealthy member when we have >= N healthy voting members present in addition, e.g deleting the machine for the unhealthy member in a cluster with 3/4 healthy members.

@JoelSpeed I agree that the real world scenario is more likely to be wanting to delete the 1/N unhealthy machines and expecting to auto-recover. As you've mentioned, that's not possible today (since that was a non-goal of the quorum protection enhancement) but we can certainly try and address it now.

I've created a ticket for us to track that work separately given it's not going to be a simple change without considering the effects of the quorum checker that we have to prevent revision rollouts that would cause quorum loss. So let's follow up the discussion on there.
https://issues.redhat.com/browse/ETCD-328

Elbehery:
I checked the logs of the failing test, it is mainly alerts and unit test not related to the changes :/

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
geliu2016: > here the usage seems safe, but please add a unit test to exercise a node with multiple IPs pending deletion.
> 
> also cc @geliu2016 fyi for verifying on dualstack (eg vsphere or so)
@tjungblu I tried dualstack on baremetal, and destroy one etcd-pod.yaml to make etcd member in unhealth status, then try to create a new machine to enable N+1 master node, but unhealth etcd member failed to be deleted automatically.

# oc get machine
NAME                              PHASE         TYPE   REGION   ZONE   AGE
geliudual2-5dss6-master-0         Failed                               3h23m
geliudual2-5dss6-master-03        Provisioned                          32m
geliudual2-5dss6-master-1         Running                              3h23m
geliudual2-5dss6-master-2         Running                              3h23m
geliudual2-5dss6-worker-0-5kvmd   Running                              177m
geliudual2-5dss6-worker-0-fv99s   Running                              177m

# oc get machine geliudual2-5dss6-master-0 -o yaml
.................
status:
  conditions:
  - lastTransitionTime: "2022-10-10T05:14:40Z"
    message: 'Drain operation currently blocked by: [{Name:EtcdQuorumOperator Owner:clusteroperator/etcd}]'
    reason: HookPresent
    severity: Warning
    status: "False"
    type: Drainable
  - lastTransitionTime: "2022-10-10T07:37:13Z"
    message: Instance not found on provider
    reason: InstanceMissing
    severity: Warning
    status: "False"
    type: InstanceExists
  - lastTransitionTime: "2022-10-10T05:12:03Z"
    status: "True"
    type: Terminable
  errorMessage: Can't find created instance.

dusk125:
lgtm, will approve when we get another

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -177,18 +177,41 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 	if err != nil {
 		return fmt.Errorf("could not get a list of unhealthy members: %v", err)
 	}
+
+	// map machines pending deletion to node internal ip
+	votingMachinesPendingDeletionIndex := ceohelpers.IndexMachinesByNodeInternalIP(votingMachinesPendingDeletion)
+	// find unhealthy machine pending deletion to remove first from the cluster, without violating quorum
+	var unhealthyMachinePendingDeletion []*machinev1beta1.Machine
```
hasbro17, line 184: Since this set is plural
```suggestion
	var unhealthyMachinesPendingDeletion []*machinev1beta1.Machine
```

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -177,18 +177,41 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
...
+			klog.V(4).Infof("unhealthy members found: %v", unhealthyMembersURLs)
+		} else {
+			return fmt.Errorf("cannot proceed with scaling down, unhealthy members found: %v", unhealthyMembersURLs)
+		}
 	}
 
+	// remove the unhealthy machine pending deletion first
+	// if no unhealthy machine pending deletion found, then attempt to scale down the healthy machines pending deletion
+	if len(unhealthyMachinePendingDeletion) > 0 {
+		votingMachinesPendingDeletion = append(unhealthyMachinePendingDeletion, votingMachinesPendingDeletion...)
```
hasbro17, line 213: Resurrecting the earlier [thread](https://github.com/openshift/cluster-etcd-operator/pull/937) on this here: 
If we have 2/4 unhealthy members, we're appending both of them for deletion.
While I can see that in practice we'll break after only removing one, from a code readability perspective it's confusing to have the intent to remove multiple unhealthy members since that could cause quourm loss(although we already have quorum loss at 2/4 anyway).

Also @Elbehery for your earlier comment:
>shouldn't we better fallback to the error when there are too many unhealthy members?

> well, I discussed this with @hasbro17 yesterday, we concluded that as long as the current number of voting members is greater than the controlPlaneReplicaCount , then we can remove members whose machine is pending deletions

I probably mentioned this but the key not just voting members but `healthy voting members == controlPlaneReplicaCount` which would be 3/4 healthy.
And even if it's safe to remove I would err on the side of caution when we have multiple unhealthy members.

So I think we should err out in that case and only address the single member unhealthy case here.

hasbro17 (changes requested):
Looks good but made a comment on the multiple unhealthy member scenario.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
Elbehery: @geliu2016 Thanks for your follow up 

So this PR tackles only the case when we have N+1 master machines, and then one goes unhealthy .. 

The scenario you have tested is another [issue](https://issues.redhat.com/browse/ETCD-328), i am still working on it. However, the result of your test is as expected

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: Thanks @hasbro17  

> I probably mentioned this but the key not just voting members but healthy voting members == controlPlaneReplicaCount which would be 3/4 healthy.

So [here](https://github.com/openshift/cluster-etcd-operator/blob/17238fb25a3e14d0ef123f26e4070a2673315eb8/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go#L166) we check that the number of `liveVotingMembers <= controlPlaneReplicaCount`. if the invariant does not hold, we skip scaling down, regardless of the member health. 

We can replace `c.etcdClient.MemberList(ctx)` with `c.etcdClient.HealthyMembers(ctx)` and then we sure that we hold the invariant before proceeding with scaling down, wdyt ? 

> And even if it's safe to remove I would err on the side of caution when we have multiple unhealthy members.

So here I am logging the unhealthy voting members && unhealthy voting members pending deletion. I can error on the case of multiple unhealthy members, but then we prevent scaling down at all, please correct me if i am wrong 

I will create a separate commit with the possible fix to continue the discussion.

Elbehery: @hasbro17  I have made the changes in this [commit](https://github.com/openshift/cluster-etcd-operator/pull/937/commits/3968a8c50ba3a049e6310694fdc29aa4a046ad51)

> I probably mentioned this but the key not just voting members but healthy voting members == controlPlaneReplicaCount which would be 3/4 healthy.

so basically I have restricted the invariant of scaling down, considering only `healthy voting members < controlPlaneReplicaCount` breaks the invariant and therefore we skip scaling down regardless

Elbehery: > And even if it's safe to remove I would err on the side of caution when we have multiple unhealthy members.

I added this [test case](https://github.com/openshift/cluster-etcd-operator/blob/2d5cb605ec80aebe86cd95426a801bfa2f2e6d79/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go#L780) to show that we can allow scaling down having multiple unhealthy machines without violating quorum. If you still wanna to error if we have more than one unhealthy machine pending deletion, ok I change it .. This is just a demonstration :)

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: done 👍🏽

Elbehery:
@dusk125 @hasbro17  Can we override `e2e-agnostic-ovn-upgrade` ? 

the failure has nothing to do with my changes :)

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -130,69 +130,89 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
...
 			} else {
 				unhealthyMembersURLs = append(unhealthyMembersURLs, unhealthyMember.Name)
 			}
 		}
-		return fmt.Errorf("cannot proceed wth scaling down, unhealthy members found: %v", unhealthyMembersURLs)
+		if len(unhealthyVotingMemberMachinesPendingDeletion) > 0 {
+			klog.V(4).Infof("found unhealthy voting members with machine pending deletion: %v", unhealthyVotingMemberMachinesPendingDeletion)
+			klog.V(4).Infof("unhealthy members found: %v", unhealthyMembersURLs)
+		} else {
+			return fmt.Errorf("cannot proceed with scaling down, unhealthy voting members found: %v", unhealthyMembersURLs)
```
tjungblu, line 204: maybe add "... members found, none are pending deletion" to make it clear why we're skipping this scale down?

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
```diff
@@ -655,8 +657,11 @@ func TestAttemptToScaleDown(t *testing.T) {
 						m5Found = true
 					}
 				}
-				if m4Found && m5Found {
-					t.Fatal("expected either m4 or m5 to be deleted but not both")
+				if m4Found {
+					t.Fatal("expected m4 to be deleted")
+				}
+				if !m5Found {
+					t.Fatal("expected m5 not to be deleted")
```
tjungblu, line 664: the way you wrote it, you can just error in the loop:

```
for _, member := range memberList {
   if member.ID == 4 {
       t.Errorf("not expected member with id %d", member.ID)
   }
}
```

or so, m5 not being deleted would've been caught by the member list length earlier already.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
```diff
@@ -696,6 +701,148 @@ func TestAttemptToScaleDown(t *testing.T) {
...
+				return []runtime.Object{cm}
+			}(),
+			fakeEtcdClientOptions: etcdcli.WithFakeClusterHealth(&etcdcli.FakeMemberHealth{Healthy: 3, Unhealthy: 1}),
+			expectedError:         nil,
+			validateFn: func(t *testing.T, fakeEtcdClient etcdcli.EtcdClient) {
+				memberList, err := fakeEtcdClient.MemberList(context.TODO())
+				if err != nil {
+					t.Fatal(err)
+				}
+				if len(memberList) != 3 {
```
tjungblu, line 739: Could you add an additional check on which member was deleted? we're relying on the `FakeMemberHealth` to set the last (m4) to be unhealthy machine that is deleted but it can still be in that member list.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
```diff
@@ -696,6 +701,148 @@ func TestAttemptToScaleDown(t *testing.T) {
...
+				for _, member := range memberList {
+					if member.ID == 1 {
+						m1Found = true
+					}
+					if member.ID == 2 {
+						m2Found = true
+					}
+				}
+				if m1Found {
+					t.Fatal("expected m1 to be deleted")
```
tjungblu, line 839: why m1? m-4 and m-5 are unhealthy and deleted

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -390,6 +410,40 @@ func (c *clusterMemberRemovalController) attemptToRemoveMemberFor(ctx context.Co
 	return removed, errs
 }
 
+func (c *clusterMemberRemovalController) getHealthyLiveVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
```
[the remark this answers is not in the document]
tjungblu: > I can add the unit tests, but they do nothing but calling the client :)

yes, but the output matters and should be asserted.

> I actually wanted to do that, but the issue is that I need to change the etcdcli and move some helper methods to ceohelpers .. I just did not want to make many changes for bug fix :)

That's okay, it's not really a bugfix is it? or do you want to backport it to 4.11?

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
Elbehery: the healthFake sets the first `n` members to unhealthy, where `n` is a number of `unhealthy` argument. 

instead of changing the healthFake to have control which members is healthy/unhealthy, I just override it for my test case

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
Elbehery: I think so, or wdyt @hasbro17 , @JoelSpeed ?

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
tjungblu: ah right, and then we just test that m1 get's deleted first and m2 stays for the time being. 

Thanks for the explanation!

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
hasbro17: It is technically a bugfix (since I didn't intend this behavior in the original implementation) but there isn't a pressing need to backport this back to 4.11 since it's not like a customer nor CPMS is exercising this particular scenario. This is mostly just to bring this in line with the original design.
I'm okay with deduplicating the helper functions though.

Elbehery:
I tested on a 4.12 cluster && this PR atop .. 

Scenario : 
- Create two new master machine ( i.e. master-3, master-4 )
- wait till the new etcd pods joins the etcd cluster and the cluster is healthy / stable
- Terminate the ec2 instance backing the master-0 machine
- Terminate the ec2 instance backing the master-1 machine
- wait till the machines becomes `failed` 
- the etcd cluster is unhealthy due to unreachable voting member
- run `oc delete machine master-0`
- run `oc delete machine master-1`
- the etcd voting member within master-0 is being removed from the cluster, even though the cluster is `unhealthy`
- the etcd voting member within master-1 is being removed from the cluster, even though the cluster is `unhealthy`
- the etcd cluster recovers and becomes healthy again with 3 voting member ( i.e. desiredControlPlaneReplicaCount` )


Logs from etcd operator 
*master-0*

```
I1012 21:30:33.874864       1 machinedeletionhooks.go:151] skip removing the deletion hook from machine ci-ln-kscjfbb-76ef8-rn695-master-0 since its member is still present with any of: [{InternalIP 10.0.176.99} {InternalDNS ip-10-0-176-99.ec2.internal} {Hostname ip-10-0-176-99.ec2.internal}]
I1012 21:30:33.874877       1 machinedeletionhooks.go:151] skip removing the deletion hook from machine ci-ln-kscjfbb-76ef8-rn695-master-1 since its member is still present with any of: [{InternalIP 10.0.244.103} {InternalDNS ip-10-0-244-103.ec2.internal} {Hostname ip-10-0-244-103.ec2.internal}]
I1012 21:30:34.430911       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"824c3c25-dfa9-4d6f-9d3a-dc34041343c0", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Warning' reason: 'UnhealthyEtcdMember' unhealthy members: ip-10-0-244-103.ec2.internal,ip-10-0-176-99.ec2.internal
I1012 21:30:34.434903       1 clustermemberremovalcontroller.go:405] successfully removed member: [ url: 10.0.176.99, name: ip-10-0-176-99.ec2.internal, id: 15209380422659960618 ] from the cluster
I1012 21:30:34.435201       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"824c3c25-dfa9-4d6f-9d3a-dc34041343c0", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'MemberRemove' removed member with ID: 15209380422659960618
I1012 21:30:34.435412       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"824c3c25-dfa9-4d6f-9d3a-dc34041343c0", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'ScaleDown' successfully removed member: [ url: 10.0.176.99, name: ip-10-0-176-99.ec2.internal, id: 15209380422659960618 ] from the cluster
I1012 21:30:36.891331       1 machinedeletionhooks.go:137] current members [ID:2141193993532186933 name:"ip-10-0-191-7.ec2.internal" peerURLs:"https://10.0.191.7:2380" clientURLs:"https://10.0.191.7:2379"  ID:2161484939922234315 name:"ip-10-0-244-103.ec2.internal" peerURLs:"https://10.0.244.103:2380" clientURLs:"https://10.0.244.103:2379"  ID:15105148013921830572 name:"ip-10-0-190-249.ec2.internal" peerURLs:"https://10.0.190.249:2380" clientURLs:"https://10.0.190.249:2379"  ID:17431867959992985327 name:"ip-10-0-134-37.ec2.internal" peerURLs:"https://10.0.134.37:2380" clientURLs:"https://10.0.134.37:2379" ] with IPSet: map[10.0.134.37:{} 10.0.190.249:{} 10.0.191.7:{} 10.0.244.103:{}]
I1012 21:30:36.908593       1 machinedeletionhooks.go:160] successfully removed the deletion hook from machine ci-ln-kscjfbb-76ef8-rn695-master-0

```
*master-1*

```
E1012 21:31:01.335646       1 base_controller.go:272] ClusterMemberController reconciliation failed: unhealthy members found during reconciling members
I1012 21:31:04.080278       1 machinedeletionhooks.go:137] current members [ID:2141193993532186933 name:"ip-10-0-191-7.ec2.internal" peerURLs:"https://10.0.191.7:2380" clientURLs:"https://10.0.191.7:2379"  ID:2161484939922234315 name:"ip-10-0-244-103.ec2.internal" peerURLs:"https://10.0.244.103:2380" clientURLs:"https://10.0.244.103:2379"  ID:15105148013921830572 name:"ip-10-0-190-249.ec2.internal" peerURLs:"https://10.0.190.249:2380" clientURLs:"https://10.0.190.249:2379"  ID:17431867959992985327 name:"ip-10-0-134-37.ec2.internal" peerURLs:"https://10.0.134.37:2380" clientURLs:"https://10.0.134.37:2379" ] with IPSet: map[10.0.134.37:{} 10.0.190.249:{} 10.0.191.7:{} 10.0.244.103:{}]
I1012 21:31:04.080331       1 machinedeletionhooks.go:151] skip removing the deletion hook from machine ci-ln-kscjfbb-76ef8-rn695-master-1 since its member is still present with any of: [{InternalIP 10.0.244.103} {InternalDNS ip-10-0-244-103.ec2.internal} {Hostname ip-10-0-244-103.ec2.internal}]
I1012 21:31:04.500904       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"824c3c25-dfa9-4d6f-9d3a-dc34041343c0", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Warning' reason: 'UnhealthyEtcdMember' unhealthy members: ip-10-0-244-103.ec2.internal
I1012 21:31:04.505255       1 clustermemberremovalcontroller.go:405] successfully removed member: [ url: 10.0.244.103, name: ip-10-0-244-103.ec2.internal, id: 2161484939922234315 ] from the cluster
I1012 21:31:04.505302       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"824c3c25-dfa9-4d6f-9d3a-dc34041343c0", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'MemberRemove' removed member with ID: 2161484939922234315
I1012 21:31:04.505801       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"824c3c25-dfa9-4d6f-9d3a-dc34041343c0", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'ScaleDown' successfully removed member: [ url: 10.0.244.103, name: ip-10-0-244-103.ec2.internal, id: 2161484939922234315 ] from the cluster
I1012 21:31:07.114636       1 machinedeletionhooks.go:160] successfully removed the deletion hook from machine ci-ln-kscjfbb-76ef8-rn695-master-1
```

Elbehery:
I tested on a 4.12 cluster && this PR atop .. 

cluster state : 3 running master machines ( master-2, master-3, master-4 ) 
Scenario : 
- Terminate the ec2 instance backing the master-2 machine
- wait till the machine becomes `failed` 
- the etcd cluster is unhealthy due to unreachable voting member
- machine-2 state is `failed`
- run `oc delete machine master-2`
- machine-2 state is `deleting`
- No replacement has been created 

*expected ( a replacement master machine is being created, and master-2 should have been deleted )*
*actual ( master-2 stuck in `deleting` phase and not replacement has been created )*

- run `oc create machine master-5`

*expected ( master-2 should have been deleted and master-5 should become running )*
*actual ( master-2 stuck in `deleting` phase master-5 stuck is running but did not join the etcd cluster )*

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
Elbehery: done

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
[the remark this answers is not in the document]
Elbehery: @tjungblu 

> actually with getAllLiveVotingMembers - do you think we should put both functions into the etcdClient directly? Seems like a common functionality and duplicate code...

I can 

```
func getVotingMembers(members []*etcdserverpb.Member) []*etcdserverpb.Member {
	var votingMembers []*etcdserverpb.Member
	for _, member := range members {
		if member.IsLearner {
			continue
		}
		votingMembers = append(votingMembers, member)
	}
	return votingMembers
}
```

but I am confused between `ceohelpers` and `etcdcli` ?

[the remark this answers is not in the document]
Elbehery: if I keep both helper functions in `clustermemberremovalcontroller` I can refactor and use this helper, and add tests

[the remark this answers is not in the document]
tjungblu: Just another example which we should refactor with that is:

https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/defragcontroller/defragcontroller.go#L124-L135

So how about we add a new method to `etcdcli.MemberLister`  like:

```
type MemberLister interface {
        // MemberList lists all members in a cluster 
	MemberList(ctx context.Context) ([]*etcdserverpb.Member, error)
        // VotingMemberList lists all voting members in a cluster (ie are not learner)
	VotingMemberList(ctx context.Context) ([]*etcdserverpb.Member, error)
}
```
That way it's well contained and you can test it there along with the existing MemberList.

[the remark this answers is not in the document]
tjungblu: I reckon we could also add equivalent methods to the `UnhealthyMemberLister` and the `HealthyMemberLister`, so we don't have to write filter methods for those all the time.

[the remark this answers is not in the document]
Elbehery: sounds like a good plan 👍🏽  .. i am on it

pkg/etcdcli/helpers.go
```diff
@@ -77,6 +77,16 @@ func (f *fakeEtcdClient) MemberList(ctx context.Context) ([]*etcdserverpb.Member
 	return f.members, nil
 }
 
+func (f *fakeEtcdClient) VotingMemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	members, _ := f.MemberList(ctx)
```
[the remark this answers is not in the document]
Elbehery: > why do you swallow the err here? don't we have to return it?

the fake client always return `nil` for error 

```
func (f *fakeEtcdClient) MemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
	return f.members, nil
}
```

pkg/etcdcli/helpers.go
```diff
@@ -242,3 +257,15 @@ func WithFakeDefragErrors(errors []error) FakeClientOption {
 		fo.defragErrors = errors
 	}
 }
+
+// FilterVotingMembers filters out learner members
+func FilterVotingMembers(members []*etcdserverpb.Member) []*etcdserverpb.Member {
```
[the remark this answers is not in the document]
Elbehery: Du hast Recht :)

tjungblu (commented):
/lgtm

PR #942, opened 2022-10-05
timeout for quorum check

tjungblu:
The health check is known to hang sometimes since https://github.com/openshift/cluster-etcd-operator/pull/862 - it might also get stuck on the listing, thus explicitly adding a 10s timeout to the whole operation here instead of `context.Background`.

Not sure whether that's resolving the issue in https://github.com/openshift/installer/pull/6124 yet.


dusk125 (commented):
lgtm, will approve when we get another

tjungblu:
/hold

yeah let's wait for a while, even adding that PR doesn't resolve everything.

PR #949, opened 2022-10-19
OCPBUGS-2577: fix cert rotation on IP changes

tjungblu:
pick and fix of https://github.com/openshift/cluster-etcd-operator/pull/945

plus adding goland to gitignore as separate commit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ cluster-etcd-operator
 
 # Log output from telepresence
 telepresence.log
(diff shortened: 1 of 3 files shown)


tjungblu:
@geliu2016 fyi when you find some time, that's the last of the chain. Thanks for testing this all the way down so far :)

geliu2016 (approved):
/label cherry-pick-approved

PR #950, opened 2022-10-19
library-go vendoring for installer cmd timeout

neisw:
Continuing research via [Trt-589](https://issues.redhat.com//browse/Trt-589) into an issue where installer pods timeout, shutdown and enter a failed state triggering KubePodNotReady alerts. library-go was updated to lower the installer cmd timeout (from 1m to 14s) to allow for more frequent retries. Bumping the version to pull those changes in.


neisw:
/hold cancel

found an issue in library-go from another commit, got the update in.

PR #957, opened 2022-10-27
[4.11] OCPBUGS-2919: update prometheus alerts

tjungblu:


Elbehery:
That's gr88 .. actually i thought it does not work because the version of the etcd, did it :)

tjungblu:
> That's gr88 .. actually i thought it does not work because the version of the etcd, did it :)

@Elbehery aside from openshift 4.8, everything above is on 3.5.5, that's what we recently rebased towards if you remember.

tjungblu:
@geliu2016 I'd also need a CPA here please :pray: :)

geliu2016 (approved):
/label cherry-pick-approved

PR #962, opened 2022-11-07
ETCD-349: add etcdmemberlister controller

Elbehery:
resolves https://issues.redhat.com/browse/ETCD-349

cc @tjungblu @hasbro17  minimal controller to discuss on this PR the right directions

### DIFF
--- a/pkg/etcdcli/cached_members.go
+++ b/pkg/etcdcli/cached_members.go
@@ -1,0 +1,125 @@
+package etcdcli
+
+import (
+	"context"
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	"sync"
+	"time"
+)
+
+const refreshPeriod = time.Minute
+
+type AllMemberLister interface {
+	MemberLister
+	HealthyMemberLister
+	UnhealthyMemberLister
+}
+
+type CachedMembers struct {
+	client  EtcdClient
+	lock    sync.RWMutex
+	members memberHealth
+
+	lastRefresh time.Time
+}
+
+func NewMemberCache(client EtcdClient) AllMemberLister {
+	return &CachedMembers{
+		client:      client,
+		lock:        sync.RWMutex{},
+		lastRefresh: time.UnixMilli(0),
+	}
+}
+
+func (c *CachedMembers) shouldRefresh() bool {
+	return c.lastRefresh.Add(refreshPeriod).Before(time.Now())
+}
+
+func (c *CachedMembers) refresh(ctx context.Context) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if c.shouldRefresh() {
+		health, err := c.client.MemberHealth(ctx)
+		if err != nil {
+			return err
+		}
+
+		c.members = health
+		c.lastRefresh = time.Now()
+	}
+
+	return nil
+}
+
+func (c *CachedMembers) convertToProtoMemberList() []*etcdserverpb.Member {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	var members []*etcdserverpb.Member
+	for _, member := range c.members {
+		members = append(members, member.Member)
+	}
+	return members
+}
+
+func (c *CachedMembers) MemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	return c.convertToProtoMemberList(), nil
+}
+
+func (c *CachedMembers) VotingMemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	list := c.convertToProtoMemberList()
+	return filterVotingMembers(list), nil
+}
+
+func (c *CachedMembers) HealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.members.GetHealthyMembers(), nil
+}
+
+func (c *CachedMembers) HealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return filterVotingMembers(c.members.GetHealthyMembers()), nil
+}
+
+func (c *CachedMembers) UnhealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.members.GetUnhealthyMembers(), nil
+}
+
+func (c *CachedMembers) UnhealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return filterVotingMembers(c.members.GetUnhealthyMembers()), nil
+}


pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -368,9 +368,9 @@ func (c *ClusterMemberController) isEtcdContainerRunningNotReady(node *corev1.No
 	return true, nil
 }
 
-// allNodesMapToVotingMembers returns nodes that don't map to voting members in the etcd cluster membership.
-// The voting members are read from the etcd-endpoints configmap
-func (c *ClusterMemberController) allNodesMapToVotingMembers(nodes []*corev1.Node) ([]*corev1.Node, error) {
+// allNodesMapToNonVotingMembers returns nodes that don't map to voting members (i.e. non learner) in the etcd cluster membership.
+// The voting members are read from the etcd cluster membership.
+func (c *ClusterMemberController) allNodesMapToNonVotingMembers(nodes []*corev1.Node) ([]*corev1.Node, error) {
```
Elbehery, line 373: @hasbro17  I think this name reflects more what the function does, isn't it ? 

cc @tjungblu

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+// EtcdMemberListerController keeps etcd members cached, and updates the cache when the membership changes
+type EtcdMemberListerController struct {
+	operatorClient v1helpers.OperatorClient
+	etcdClient     etcdcli.EtcdClient
+	membersCache   map[string]*etcdserverpb.Member
```
tjungblu, line 22: I think this will need some kind of locking, given this will be consumed by other controllers?

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -0,0 +1,86 @@
...
+		membersCache:   make(map[string]*etcdserverpb.Member),
+	}
+	return factory.New().ResyncEvery(30*time.Second).WithSync(c.sync).ToController("EtcdMemberListerController", eventRecorder.WithComponentSuffix("etcd-member-lister-controller"))
+}
+
+func (c *EtcdMemberListerController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	err := c.syncEtcdMembers(ctx, syncCtx.Recorder())
+	if err != nil {
+		_, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:    "EtcdMemberListerControllerDegraded",
```
tjungblu, line 41: we should only degrade if this fails for several consecutive times, a transient failure every 30s shouldn't be a big issue to degrade the whole operator on

Elbehery:
@tjungblu  thanks a lot for your review .. 

I just want to clarify what this controller is going to do 

- cache All etcd members 
- cache healthy etcd members
- cache unhealthy etcd members 

I have chosen to use t `map[string]*Member`. If this choice is not adequate please let me know your thoughts 

Also please correct me if I am wrong :)

tjungblu:
this :+1: cache All etcd members

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+var (
+	errNilMembersCache = errors.New("member cache is nil")
+	errKeyExist        = errors.New("key already exist")
+	errKeyNotExist     = errors.New("key does not exist")
+)
+
+type membersCache struct {
```
Elbehery, line 26: @tjungblu  would this make sense ? .. I am gonna embed this within the controller

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -0,0 +1,195 @@
...
+
+	mc.mux.Lock()
+	_, ok := mc.cache[key]
+	if !ok {
+		return false, errKeyNotExist
+	}
+	// safe to update
+	mc.cache[key] = *value
+	return true, nil
+}
```
hasbro17, line 72: Do we really need add/update/delete/get methods on the member cache if we're just outright replacing the cache when the membership changes?
If not, then we can remove these.

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -0,0 +1,195 @@
...
+	members, err := c.etcdClient.MemberList(ctx)
+	if err != nil {
+		return err
+	}
+
+	var syncErrs []error
+	// build map from current live members
+	currentMembersMap, errs := c.currentMembersMap(members)
+	if len(errs) > 0 {
+		syncErrs = append(syncErrs, errs...)
```
hasbro17, line 160: Why are we continuing to check the cache here if we have errors from building the current members map?
Shouldn't we return the errors?

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
Elbehery: yes u r right, initially i was using them to update the cache with new entries and remove stale entries .. but we dont need them now, removing ...

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
Elbehery: well I did not know what to do tbh :) 

So I can proceed with building the map while accumulating errors as I do now, or fail-fast once i encounter an error 

wdyt ? 

cc @tjungblu

tjungblu: fail-fast, but only degrade the operator after several consecutive failures.

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	// check membership changes against members cache
+	if !reflect.DeepEqual(currentMembersMap, c.membersCache) {
+		klog.V(2).Infof("detected changes in etcd cluster membership, updating controller cache")
```
tjungblu, line 103: can you add the previous members and the current members? I've done something similar in the client and it's super useful when going through must gathers

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+var (
+	errNilMembersCache = errors.New("member cache is nil")
+	errKeyExist        = errors.New("key already exist")
+	errKeyNotExist     = errors.New("key does not exist")
```
tjungblu, line 27: I think the errors can also be deleted now that you've removed the diff/add/remove logic

Elbehery: done

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
Elbehery: done

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (c *EtcdMemberListerController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	err := c.syncEtcdMembers(ctx, syncCtx.Recorder())
+	if err != nil {
+		// update failure counts
+		c.lastConsecutiveFailures++
+		if c.lastConsecutiveFailures >= 5 {
+			// reset counter before going degraded
+			c.lastConsecutiveFailures = 0
```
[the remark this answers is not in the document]
Elbehery: +1, thanks a lot 👍🏽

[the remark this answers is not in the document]
Elbehery: done

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -52,6 +55,9 @@ type ClusterMemberController struct {
 
 	masterNodeLister   corev1listers.NodeLister
 	masterNodeSelector labels.Selector
+	memberListerCtrl   *etcdmemberlistercontroller.EtcdMemberListerController
+	members            []etcdserverpb.Member
```
Elbehery, line 59: this controller has two usage of `memberlist()`
- one uses to validate if a member is already part of the cluster 
- one iterates over the members and attempt to promote learners

Therefore i added both these fields and i update them on notifications from `etcdmemberlistercontroller` 

please let me know wdyt  cc @tjungblu @hasbro17

pkg/operator/starter.go
```diff
@@ -316,6 +323,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		configInformers.Config().V1().Networks(),
 		etcdClient,
 		controllerContext.EventRecorder,
+		&etcdmemberlistercontroller,
```
Elbehery, line 326: this line does not compile, i do not know how to fix it :/ 

So the `NewClusterMemberController` returns a `factory.Controller` type . I need to have a ref for the `etcdmemberlistercontroller` to update the local cache in each controller upon notifications 

please let me know wdyt

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -254,16 +262,11 @@ func (c *ClusterMemberController) ensureEtcdLearnerPromotion(ctx context.Context
 		return nil
 	}
 
-	members, err := c.etcdClient.MemberList(ctx)
-	if err != nil {
-		return fmt.Errorf("could not get etcd member list: %v", err)
-	}
-
```
Elbehery, line 261: this call will be removed with my changes .. so instead of issuing a request to etcd-server, the controller retrieve them locally .. `etcdmemberlistercontroller` makes sure other controllers do not have stale cached data

pkg/operator/starter.go
tjungblu: I suggest you to create an interface that is returned by `NewEtcdMemberListerController` instead. You're also missing to actually run that controller.

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -143,3 +147,31 @@ func (c *EtcdMemberListerController) getMembersFromMap(membersMap map[string]etc
...
+}
+
+func (c *EtcdMemberListerController) AddListener(listener interfaces.Listener) {
+	for _, l := range c.listeners {
+		if l == listener {
+			return
+		}
+	}
+	c.listeners = append(c.listeners, listener)
+	return
```
tjungblu, line 170: why return?

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -143,3 +147,31 @@ func (c *EtcdMemberListerController) getMembersFromMap(membersMap map[string]etc
 	}
 	return members
 }
+
+func (c *EtcdMemberListerController) GetMembers() []etcdserverpb.Member {
+	members := make([]etcdserverpb.Member, 0, len(c.membersCache.cache))
+	defer c.membersCache.mux.RUnlock()
+
+	c.membersCache.mux.RLock()
```
tjungblu, line 155: it's not wrong what you wrote here, but for better readability I would put the locking to the start of the method:

```
	c.membersCache.mux.RLock()
	defer c.membersCache.mux.RUnlock()

	members := make([]etcdserverpb.Member, 0, len(c.membersCache.cache))
        ...
```

So when people add more lines later, they have an easier time to reason about the scope of the lock.

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -9,6 +9,7 @@ import (
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 
 	errorsutil "k8s.io/apimachinery/pkg/util/errors"
+	interfaces "k8s.io/apiserver/pkg/server/dynamiccertificates"
```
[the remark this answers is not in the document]
Elbehery: i found `Notifier` and `Listener` already defined there so I used them 

I can also create my own interface

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
Elbehery: +1

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
Elbehery: +1

pkg/operator/clustermembercontroller/clustermembercontroller.go
```diff
@@ -414,12 +417,11 @@ func (c *ClusterMemberController) getInternalIPForNode(node *corev1.Node) (strin
 	return internalIP, nil
 }
 
-// isURLMappedToMember returns true if the given peerURL matches any of the members' peerURLs
-func isURLMappedToMember(peerURL string, members []*etcdserverpb.Member) bool {
-	for _, m := range members {
-		if peerURL == m.PeerURLs[0] {
-			return true
-		}
+func (c *ClusterMemberController) Enqueue() {
```
tjungblu, line 420: I think this can create a race condition. Imagine this controller is running every 10s on a different goroutine than the member lister controller.

The updates done in this Enqueue method might not be fully visible during the controller sync. I like this idea with the set, but you maybe just want to implement it in the boring way by listing all members when the controller sync runs and just reuse that return throughout the whole lifecycle of the sync.

OR alternatively add some locking around it, not sure it's really worth it because the controller runs never the less. If you were to tie the controller sync period to only run when the member listing changes, that would be useful but likely not necessary.

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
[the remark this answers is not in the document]
tjungblu: what's the envvar controller using?

pkg/operator/clustermembercontroller/clustermembercontroller.go
Elbehery: I am a bit confused now :D 

So `Enqueue` is invoked by the `etcdmemberlistercontroller` when the member list changes only. it updates it own cache and then sync Listener's caches by invoking `Enqueue()` .. Moreover, Enqueue calls 
`(c *EtcdMemberListerController) GetMembers()`  which uses the lock before reading the cache and returning a copy to the Listener 

Listener's Sync will use the local updated cache to reconcile 

Where is the race condition ?

tjungblu: simply put, you're adjusting the internal structure of `clustermembercontroller` here that runs on a different controller goroutine.

Enqueue is called from the `etcdmemberlistercontroller` controller goroutine, so you need to lock here in this controller as well if you want to do this listener-style updates. 

I'm sure if you were to run both controllers in goroutines in a test with `-race` it would flag that.

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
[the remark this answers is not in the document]
Elbehery: it's own interfaces .. I created mine, I can use others .. maybe move them into commons ?

pkg/operator/clustermembercontroller/clustermembercontroller.go
Elbehery: i see it now, but what is the right way then, without using extra locking ? .. i think we should reduce this, or ?

pkg/operator/starter.go
Elbehery: fixed 👍🏽

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -0,0 +1,195 @@
...
+	eventRecorder events.Recorder,
+) *EtcdMemberListerController {
+	c := &EtcdMemberListerController{
+		operatorClient:          operatorClient,
+		etcdClient:              etcdClient,
+		membersCache:            newMembersCache(),
+		listeners:               []Listener{},
+		lastConsecutiveFailures: 0,
+	}
+	//return factory.New().ResyncEvery(30*time.Second).WithSync(c.sync).ToController("EtcdMemberListerController", eventRecorder.WithComponentSuffix("etcd-member-lister-controller"))
```
Elbehery, line 73: I do not know how to replace this to still get a controller while returning the `*EtcdMemberListerController` :/

pkg/operator/clustermembercontroller/clustermembercontroller.go
tjungblu: in `clustermembercontroller.sync` I would just call `getMembers()` as it was with the client. This is properly locked and you always get consistent data back for the lifetime of the sync call. We can add the notification fanciness later to only run the controller when the member list changes.

Elbehery: yes I agree, I am also using `[]Member` instead of `map[string]Member` since the controller need not lookup by memberIP .. 

I am adding also `GetHealthyMember` and `GetUnhealthyMembers` so that other controllers need not to use the etcdclient directly for any member listing 

WIP .. :D

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
tjungblu: You have two options
* see how the `NewEnvVarController` does this, it doesn't use the factory, it merely implements `Run`
* compose the two interfaces

The latter looks clearer to me, think of something like that:

```
type ListerController interface {
	factory.Controller
	EtcdMemberLister
}
```

Then just return the ListerController from New*

Elbehery:
/retest-required

failing unit test is flake

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
Elbehery: this is a great solution, thx so much .. i was stuck debugging the test cases :D

Elbehery: no it is not working :/ 

```
Running `gofmt -s -l` on 97 file(s).
go vet -mod=vendor ./...
# github.com/openshift/cluster-etcd-operator/pkg/operator/etcdmemberlistercontroller
pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go:79:9: cannot use factory.New().ResyncEvery(30 * time.Second).WithSync(c.sync).ToController("EtcdMemberListerController", eventRecorder.WithComponentSuffix("etcd-member-lister-controller")) (value of type factory.Controller) as type ListerController in return statement:
        factory.Controller does not implement ListerController (missing AddListener method)
# github.com/openshift/cluster-etcd-operator/pkg/operator/etcdmemberlistercontroller
vet: pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go:79:9: cannot use factory.New().ResyncEvery(30 * time.Second).WithSync(c.sync).ToController("EtcdMemberListerController", eventRecorder.WithComponentSuffix("etcd-member-lister-controller")) (value of type factory.Controller) as ListerController value in return statement: factory.Controller does not implement ListerController (missing method AddListener)
make: *** [verify-govet] Error 2

```

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
```diff
@@ -56,23 +57,25 @@ type EtcdMemberListerController struct {
 	lastConsecutiveFailures int
 	operatorClient          v1helpers.OperatorClient
 	etcdClient              etcdcli.EtcdClient
+	controllerFactory       *factory.Factory
```
Elbehery, line 60: @tjungblu  i followed the same pattern from `StatusSyncer` from `library-go` :D

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
tjungblu: why don't you implement what it says then? :roll_eyes: 
> (missing AddListener method)

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
tjungblu: that also works, cool workaround

Elbehery: thx a lot 🙏🙏

so i need to add logs, events .. anything else ?

Elbehery:
@hasbro17  can we override this failing job ?

Elbehery:
i gave up :D

pkg/operator/etcdmemberlistercontroller/etcdmemberlistercontroller.go
tjungblu: I think the struct makes sense, I'm not sure why you wanted to have it as a dedicated controller (or maybe I did).

it's more sensible to me to put this into `etcdClientGetter` struct in the etcdcli, then `MemberList` can always return a cached response or fetch when outdated.

tjungblu: something along those lines:

```
package etcdcli

import (
	"context"
	"go.etcd.io/etcd/api/v3/etcdserverpb"
)

type CachedMembers struct {
	// some caching logic
}

func (c CachedMembers) MemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
	//TODO implement me
	panic("implement me")
}

func (c CachedMembers) VotingMemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
	//TODO implement me
	panic("implement me")
}

func (c CachedMembers) HealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
	//TODO implement me
	panic("implement me")
}

func (c CachedMembers) HealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
	//TODO implement me
	panic("implement me")
}

func (c CachedMembers) UnhealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
	//TODO implement me
	panic("implement me")
}

func (c CachedMembers) UnhealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
	//TODO implement me
	panic("implement me")
}

```

PR #965, opened 2022-11-21
OCPBUGS-3841: update apf configuration to use v1beta2

tkashem:


soltysh:
/hold 
first master only then 4.12

tkashem:
master has merged
/hold cancel

PR #972, opened 2022-12-13
[release-4.8] OCPBUGS-4775: only allow safe TLS1.2 and 1.3 ciphersuites in etcd and CEO

tjungblu:


tjungblu:
/hold

actually I just noticed that this could potentially wait, the 4.8.55 cutoff was last week, so there won't be another z release until the new year anyway.

tjungblu:
@geliu2016 need a CPA here please :) last one, promised!

PR #978, opened 2023-01-05
OCPBUGS-5304: should not scale-down when all members are healthy

Elbehery:
fixes https://issues.redhat.com/browse/OCPBUGS-5304

### DIFF
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"time"
 
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
@@ -207,6 +208,14 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 			klog.V(2).Infof("cannot proceed with scaling down, unhealthy voting etcd members found: %v but none are pending deletion", unhealthyMembersURLs)
 			return fmt.Errorf("cannot proceed with scaling down, unhealthy voting etcd members found: %v but none are pending deletion", unhealthyMembersURLs)
 		}
+	}
+
+	// Don't scale down if all members are healthy and the cluster is already at the desired control-plane size
+	// When at the desired control-plane size, scale down is only allowed if a member is unhealthy so that a new machine can replace it.
+	// For healthy members, deleting a machine should result in a new machine being created and member addition before scale down can occur
+	if reflect.DeepEqual(liveVotingMembers, healthyLiveVotingMembers) && len(healthyLiveVotingMembers) <= desiredControlPlaneReplicasCount {
+		klog.V(2).Infof("skip scale down: all voting members are healthy and membership does not exceed the desired cluster size of %v", desiredControlPlaneReplicasCount)
+		return nil
 	}
 
 	// remove the unhealthy machine pending deletion first
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
@@ -474,18 +474,53 @@ func TestAttemptToScaleDown(t *testing.T) {
 			},
 		},
 		{
-			name:                       "skip scaling down by one unhealthy machine from 3 master nodes", // no machine pending deletion
+			name:                       "scale down by one unhealthy machine pending deletion from 3 master nodes",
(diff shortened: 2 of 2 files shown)


pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -209,6 +210,12 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 		}
 	}
 
+	// fixes https://issues.redhat.com/browse/OCPBUGS-5304
+	if reflect.DeepEqual(liveVotingMembers, healthyLiveVotingMembers) && len(healthyLiveVotingMembers) <= desiredControlPlaneReplicasCount {
+		klog.V(2).Infof("ignoring scaling down, all voting member are healthy, no unhealthy voting member found")
```
hasbro17, line 215: Clarify the log message.
```suggestion
		klog.V(2).Infof("skip scale down: all voting members are healthy and membership does not exceed the desired cluster size of %v", desiredControlPlaneReplicasCount)
```
Also can you please confirm by looking at the operator's logs but I think this is the default case for a normal cluster and this would end up spamming the logs so V(2) may not be appropriate here.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -209,6 +210,12 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 		}
 	}
 
+	// fixes https://issues.redhat.com/browse/OCPBUGS-5304
+	if reflect.DeepEqual(liveVotingMembers, healthyLiveVotingMembers) && len(healthyLiveVotingMembers) <= desiredControlPlaneReplicasCount {
```
hasbro17, line 214: This whole function could really benefit from restructuring to make this more readable but I'm okay with keeping this change small to unblock us on fixing the scaling CI.

We can do a follow up PR to clean up this controller.

hasbro17 (changes requested):
Looks okay but a couple nits needed.

hasbro17:
`e2e-aws-ovn-etcd-scaling` is going to fail since a fix on the CPMSO (https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/112) got merged which causes an issue in the vertical scaling test because it uses a non CPMS workflow.

We have an update for that here https://github.com/openshift/origin/pull/27497 but we can't update that test to use CPMS before we tighten this scale down condition. So I'm going to override the failing test for now.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -209,6 +210,12 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 		}
 	}
 
+	// fixes https://issues.redhat.com/browse/OCPBUGS-5304
```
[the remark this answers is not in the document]
Elbehery: done

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: alright, shall i use `v(4)` ? @hasbro17

Elbehery:
@hasbro17  can you override the failing job, it is flaky due to CPMS changes

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -209,6 +210,14 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 		}
 	}
 
+	// Don't scale down if all members are healthy and the cluster is already at the desired control-plane size
+	// When at the desired control-plane size, scale down is only allowed if a member is unhealthy so that a new machine can replace it.
+	// For healthy members, deleting a machine should result in a new machine being created and member addition before scale down can occur
+	if reflect.DeepEqual(liveVotingMembers, healthyLiveVotingMembers) && len(healthyLiveVotingMembers) <= desiredControlPlaneReplicasCount {
+		klog.V(2).Infof("skip scale down: all voting members are healthy and membership does not exceed the desired cluster size of %v", desiredControlPlaneReplicasCount)
```
hasbro17, line 217: >alright, shall i use v(4)

Actually on second though this won't be the default case since a machine is pending deletion, but we're awaiting a new one to surge up the membership first. So we could keep it v(2).

But can you confirm what the operator's log show while 3 members are healthy with a machine pending deletion. Just want to gauge if this isn't too spammy.

PR #982, opened 2023-01-10
OCPBUGS-5516: avoid lister cache inconsistency in etcd cert signer

tjungblu:
This is to avoid restore inconsistencies where we rely on the lister cache. For the cert generation we should directly get the secret from the API like the apply does.


geliu2016 (approved):
/label cherry-pick-approved

tjungblu:
Let's close this for now, in favor of documenting the restart of the operators.

PR #992, opened 2023-01-19
OCPBUGS-5988: add dedicated success status for bootstrap removal

tjungblu:
/hold

less complicated than  #991

### DIFF
--- a/pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
+++ b/pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog/v2"
 
 	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/ceohelpers"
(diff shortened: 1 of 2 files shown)


pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -139,7 +146,19 @@ func (c *BootstrapTeardownController) removeBootstrap(ctx context.Context, syncC
...
+}
+
+func setSuccessfulBoostrapRemovalStatus(ctx context.Context, client v1helpers.StaticPodOperatorClient) error {
+	_, _, updateErr := v1helpers.UpdateStatus(ctx, client, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+		Type:    "EtcdBoostrapMemberRemoved",
+		Status:  operatorv1.ConditionTrue,
+		Reason:  "BootstrapMemberRemoved",
+		Message: "etcd bootstrap member is removed",
+	}))
+	return updateErr
```
hasbro17, line 161: Is this new condition intended to block and signal to the bootkube process to proceed once the member is truly removed?
https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L481-L493

We would need to update wait-for-ceo in that case. If so, we could just make sure that `EtcdRunningInCluster ` is set only after the bootstrap member is no longer present in the membership.

tjungblu: dunno yet, I did a similar version earlier in https://github.com/openshift/cluster-etcd-operator/pull/991 and it failed terribly in CI.

At first it should signal assisted-installer to get these guys unblocked, then we can talk about the bootkube behavior.

tsorya: Adding it for wait_for_ceo needed in case of DelayedStrategy that assisted-installer is setting.
I don't think anybody else will need it

tjungblu: @tsorya checkout https://github.com/openshift/cluster-etcd-operator/pull/996

PR #995, opened 2023-01-23
adding a skip hash to restore pod yaml, to be able to restore from the db file

vwalek:
Adding this PR to just add `--skip-hash-check=true` to the restore command to be able to restore from the `db` file if the snapshot is not generated.

Signed-off-by: Vladislav Walek <22072258+vwalek@users.noreply.github.com>

### DIFF
--- a/bindata/etcd/restore-pod.yaml
+++ b/bindata/etcd/restore-pod.yaml
@@ -59,7 +59,8 @@ spec:
          --initial-cluster=$ETCD_INITIAL_CLUSTER \
          --initial-cluster-token "openshift-etcd-${UUID}" \
          --initial-advertise-peer-urls $ETCD_NODE_PEER_URL \
-         --data-dir="/var/lib/etcd/restore-${UUID}"
+         --data-dir="/var/lib/etcd/restore-${UUID}" \
+         --skip-hash-check=true
 
         mv /var/lib/etcd/restore-${UUID}/* /var/lib/etcd/
 
--- a/pkg/operator/etcd_assets/bindata.go
+++ b/pkg/operator/etcd_assets/bindata.go
@@ -1049,7 +1049,8 @@ spec:
          --initial-cluster=$ETCD_INITIAL_CLUSTER \
          --initial-cluster-token "openshift-etcd-${UUID}" \
          --initial-advertise-peer-urls $ETCD_NODE_PEER_URL \
-         --data-dir="/var/lib/etcd/restore-${UUID}"
+         --data-dir="/var/lib/etcd/restore-${UUID}" \
+         --skip-hash-check=true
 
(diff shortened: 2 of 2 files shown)


Elbehery:
Thanks @vwalek  for this PR 

There is another generated file [here](https://github.com/openshift/cluster-etcd-operator/blob/4d03335886fe26be247adb1b0209bd114beff966/pkg/operator/etcd_assets/bindata.go)

you need to run `make generate` to update this file from the manifest, then  `make verify` from the cluster-etcd-operator root folder, to make sure everything in place

would you kindly update the PR accordingly ?

Elbehery:
the failing `verify` job will pass if `make verify` passes locally

Elbehery:
@vwalek which go version u have ? please make sure it is go.1.19.x 

also reset to the openshift/master branch, and run `go mod tidy && go mod vendor`, before adding ur changes && using `make generate`

- `make update`
- `make`
- `make generate`
- `make verify`

Elbehery:
> > you need to run `make generate` to update this file from the manifest, then `make verify` from the cluster-etcd-operator root folder, to make sure everything in place
> 
> Where I need to run `make generate` from? From the root directory it shows no rule found.
> 
> ```
> make: *** No rule to make target 'generate'.  Stop.
> ```
> 
> Thx

try `make update` && `make verify`

tjungblu:
@vwalek what's the purpose of this? can you detail out what your scenario is and why it should be the default? I would expect to restore only valid snapshots.

vwalek:
> @vwalek what's the purpose of this? can you detail out what your scenario is and why it should be the default? I would expect to restore only valid snapshots.

@tjungblu Hi, the `--skip-hash-check=true` will give you ability to recover from the db file, instead of the snapshot. If you attempt to recover from db file without that option, the etcd database will not be working properly. We have this option in the our recovery doc for 3.11 and for 4 as well.

There are times where system admin cannot take fresh snapshot of the database or something goes wrong. There is still the DB file, however, it contains data about the etcd members (where snapshot does not). Recover from the db file might be the last option to recover the production clusters.

lance5890:
only set --skip-hash-check=true is ont enough, maybe u should also adjust the restore-pod logic to use the db file

vwalek:
> only set --skip-hash-check=true is ont enough, maybe u should also adjust the restore-pod logic to use the db file

The restore pod will restore it from the snapshot.db file, so if you name your `db` file as `snapshot.db` it should recover from it. However, I see that recovery script actually uses needs it to be `snapshot.db`.

Thanks for the comment, it kept me thinking about the process today :) And yeah, there are some minor adjustments that needs to be done for it to work.

vwalek:
the issue will be handled by jira

/close

Issue #1000, opened 2023-02-07
How can I merge the ceo bug into ocp


lance5890:
I have found two ceo bug when I intall ocp release

1. When install SingleNodeTopology ocp release, the etcd env miss ETCD_URL_HOST , ETCD_NAME , and lead to single mode failed occasionally
2. When install HABootstrapScaling ocp relesase, when I remove one master node ,the ceo will stuck in etcd memer list about 15 minutes

I have resolved this two issue , I would like to merge my code into ceo for you review

@tjungblu @hexfusion

tjungblu:
Sounds great, can you send us a pull request?

lance5890:
these two bugs have all merged into master, i will close this issue

PR #1001, opened 2023-02-08
fix ceo miss key env values for etcd check

lance5890:
ocp single install  missing key env var values about "_ETCD_URL_HOST", "_ETCD_NAME","_IP"; which be checked when etcd etcd-ensure-env-vars init container start

[issue 1000 merge the ceo bug](https://github.com/openshift/cluster-etcd-operator/issues/1000)

Signed-off-by: lan.tian <lance5890@163.com>

### DIFF
--- a/pkg/etcdenvvar/etcd_env.go
+++ b/pkg/etcdenvvar/etcd_env.go
@@ -139,8 +139,12 @@ func getAllEtcdEndpoints(envVarContext envVarContext) (map[string]string, error)
 func getEtcdName(envVarContext envVarContext) (map[string]string, error) {
 	ret := map[string]string{}
 
-	for _, nodeInfo := range envVarContext.status.NodeStatuses {
-		ret[fmt.Sprintf("NODE_%s_ETCD_NAME", envVarSafe(nodeInfo.NodeName))] = nodeInfo.NodeName
+	if nodeStatus := envVarContext.status.NodeStatuses; nodeStatus != nil {
+		for _, nodeInfo := range nodeStatus {
+			ret[fmt.Sprintf("NODE_%s_ETCD_NAME", envVarSafe(nodeInfo.NodeName))] = nodeInfo.NodeName
+		}
+	} else {
+		return nil, fmt.Errorf("nodeStatuses not found in StaticPodOperator")
 	}
 
 	return ret, nil
@@ -153,17 +157,21 @@ func getEscapedIPAddress(envVarContext envVarContext) (map[string]string, error)
 	}
 
 	ret := map[string]string{}
-	for _, nodeInfo := range envVarContext.status.NodeStatuses {
-		node, err := envVarContext.nodeLister.Get(nodeInfo.NodeName)
-		if err != nil {
-			return nil, err
-		}
+	if nodeStatus := envVarContext.status.NodeStatuses; nodeStatus != nil {
+		for _, nodeInfo := range nodeStatus {
+			node, err := envVarContext.nodeLister.Get(nodeInfo.NodeName)
+			if err != nil {
+				return nil, err
+			}
 
-		escapedIPAddress, err := dnshelpers.GetEscapedPreferredInternalIPAddressForNodeName(network, node)
-		if err != nil {
-			return nil, err
+			escapedIPAddress, err := dnshelpers.GetEscapedPreferredInternalIPAddressForNodeName(network, node)
+			if err != nil {
+				return nil, err
+			}
+			ret[fmt.Sprintf("NODE_%s_IP", envVarSafe(nodeInfo.NodeName))] = escapedIPAddress
 		}
-		ret[fmt.Sprintf("NODE_%s_IP", envVarSafe(nodeInfo.NodeName))] = escapedIPAddress
+	} else {
+		return nil, fmt.Errorf("nodeStatuses not found in StaticPodOperator")
 	}
 
 	return ret, nil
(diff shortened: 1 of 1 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -94,6 +94,19 @@ func (c TargetConfigController) sync(ctx context.Context, syncCtx factory.SyncCo
...
+            if strings.Contains(k, "_ETCD_URL_HOST"){
+                checkEnvNum++
+            }else if strings.Contains(k, "_ETCD_NAME"){
+                checkEnvNum++
+            }else if strings.Contains(k, "_IP"){
+                checkEnvNum++
+            }
+        }
+        if checkEnvNum < 3 {
+            return fmt.Errorf("TargetConfigController mismatch checkEnvNum")
```
tjungblu, line 108: besides the formatting issues, I don't really understand what the purpose of this is.
The variables are generated in etcd_env.go directly from the node lister and the static pod machinery:

https://github.com/openshift/cluster-etcd-operator/blob/b561d08d887fc0916ba6008308b8621754e30302/pkg/etcdenvvar/etcd_env.go#L172-L193


if there are none of those we need to fix this up the stack in https://github.com/openshift/library-go and not here. 

Do you know why your static pod operator status does not contain the required node status?

pkg/operator/targetconfigcontroller/targetconfigcontroller_test.go
```diff
@@ -138,6 +138,9 @@ func TestTargetConfigController(t *testing.T) {
 
 			envVar := etcdenvvar.FakeEnvVar{EnvVars: map[string]string{
 				"ALL_ETCD_ENDPOINTS": scenario.etcdMembersEnvVar,
+                                "_ETCD_URL_HOST": "127.0.0.1",
+                                "_ETCD_NAME": "single",
+                                "_IP": "127.0.0.1",
```
tjungblu, line 143: please add a test that exercises the logic you've added, this doesn't test anything.

tjungblu:
@lance5890 you've mentioned in the issue a second problem:

> When install HABootstrapScaling ocp relesase, when I remove one master node ,the ceo will stuck in etcd memer list about 15 minutes

I don't see how this PR will fix this, can you elaborate more?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
tjungblu: not to say that I'm opposed against validation, but maybe we should add this in the place where it occurs, for example here:

https://github.com/openshift/cluster-etcd-operator/blob/b561d08d887fc0916ba6008308b8621754e30302/pkg/etcdenvvar/etcd_env.go#L142

so if the length of `envVarContext.status.NodeStatuses` is zero, this should error out directly.

lance5890: yes ,Under normal circumstances, the EtcdURLHost is generated by the EnvVarController, and the TargetConfigController will use the GetEnvVars to get the env to generate etcd yaml

but in some cases，the GetEnvVars is not zero, bu not containes "_ETCD_URL_HOST", "_ETCD_NAME","_IP"; but TargetConfigController  will still generate etcd yaml, ans this yaml will checked fail by etcd-ensure-env-vars init container like ""#!/bin/sh\nset -euo pipefail\n\n: \"${NODE_master0_ETCD_URL_HOST?not set}\"\n: \"${NODE_master0_ETCD_NAME?not set}\"\n: \"${NODE_master0_IP?not set}\"

lance5890:
> n y

lance5890:
> @lance5890 you've mentioned in the issue a second problem:
> 
> > When install HABootstrapScaling ocp relesase, when I remove one master node ,the ceo will stuck in etcd memer list about 15 minutes
> 
> I don't see how this PR will fix this, can you elaborate more?

this PR only containes the single etcd yaml err

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
lance5890: like this:
image

tjungblu: okay, I think there's a bigger bug we need to go after in the static pod machinery. Which OpenShift version are you running? 

Could also you please add your validation directly into the etcd_env.go on the respective functions? I think that's the better place, all other controllers will benefit from this too.

lance5890: i am using ocp 4.9 and 4.11, both have the problem

lance5890: I have read all the ceo code, and GetEnvVars is only used by the TargetConfigController, and my modification will resolve this problem

lance5890: > 

I have read the ceo code, and the GetEnvVars is only used by the TargetConfigController; as u said"add your validation directly into the etcd_env.go", i think the problem is not in the EnvVarController, but in the timing when should TargetConfigController use the env, TargetConfigController  should wait all the env it needs to be ready, and then it can generate the etcd yaml

tjungblu: I have read the code too, yet I would like to avoid a case where other people add a new controller that requires env vars and read incomplete data. 

> TargetConfigController should wait all the env it needs to be ready, and then it can generate the etcd yaml

If this is a race condition between the controllers we need to add some locking or serialization for them, not add some if conditions on the consumer side of the API. I'll see if I can reproduce this and will come back to you on this.

lance5890: I think we can also add check in following:
https://github.com/openshift/cluster-etcd-operator/blob/f695ca59f559e6b7a23451f42e1d430aa1d4e380/pkg/etcdenvvar/etcd_env.go#L156

we can check the nodeInfo should exist, will also resolve this

lance5890: like this env get :
https://github.com/openshift/cluster-etcd-operator/blob/f695ca59f559e6b7a23451f42e1d430aa1d4e380/pkg/etcdenvvar/etcd_env.go#L204

pkg/operator/targetconfigcontroller/targetconfigcontroller_test.go
lance5890: I have changed the env get check ,please check

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
lance5890: I have changed the etcdenv check func, please review the new code

tjungblu: sounds good, I just added some more tests around this in https://github.com/openshift/cluster-etcd-operator/pull/1002

can you please take a look at that one?

lance5890: ok

lance5890:
this pr has also resolved in https://github.com/openshift/cluster-etcd-operator/pull/1002, which use the same way ,but with more test cases, i will close this pr

PR #1002, opened 2023-02-08
fail early on missing node status envs

tjungblu:
fixing #1000 and similar to #1001 with tests


lance5890:
sounds good  for fixing https://github.com/openshift/cluster-etcd-operator/issues/1000

PR #1003, opened 2023-02-10
OCPBUGS-7374: set default timeouts in etcdcli

lance5890:
The ceo operator connect to etcd without timeout, which may lead to some ceo controllers(like EtcdEndpointsController, EtcdCertSignerController and so on ) may be stuck in etcd member list interface about 15minutes, which finally may be closed by the operating system after 15 minutes

[issue1000: fix ceo connection timeout ](https://github.com/openshift/cluster-etcd-operator/issues/1000)

1. this problem can be reproduced by rm one master node, and the pprof top info will show like this:
image

2. maybe the 60s is not suitable for most conditions，we can discuss about this

Signed-off-by: lan.tian <lance5890@163.com>

### DIFF
--- a/pkg/cmd/verify/backupstorage.go
+++ b/pkg/cmd/verify/backupstorage.go
@@ -132,6 +132,8 @@ func (v *verifyBackupStorage) isStorageAdequate(ctx context.Context) (bool, erro
 	}
 	defer cli.Close()
 
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
 	members, err := cli.MemberList(ctx)
 	if err != nil {
 		klog.Warningf("failed checking member list: %v", err)
--- a/pkg/etcdcli/etcdcli.go
+++ b/pkg/etcdcli/etcdcli.go
@@ -39,6 +39,7 @@ const (
 	BootstrapIPAnnotationKey = "alpha.installer.openshift.io/etcd-bootstrap"
 	DefaultDialTimeout       = 15 * time.Second
 	DefragDialTimeout        = 60 * time.Second
+	DefaultClientTimeout     = 30 * time.Second
 )
 
 type etcdClientGetter struct {
@@ -135,7 +136,7 @@ func newEtcdClientWithClientOpts(endpoints []string, skipConnectionTest bool, op
 	}
 
 	// Test client connection.
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithTimeout(context.Background(), DefaultClientTimeout)
 	defer cancel()
 	_, err = cli.MemberList(ctx)
 	if err != nil {
@@ -156,6 +157,8 @@ func (g *etcdClientGetter) MemberAddAsLearner(ctx context.Context, peerURL strin
 
 	defer g.clientPool.Return(cli)
 
+	ctx, cancel := context.WithTimeout(ctx, DefaultClientTimeout)
+	defer cancel()
 	membersResp, err := cli.MemberList(ctx)
 	if err != nil {
 		return err
@@ -203,6 +206,8 @@ func (g *etcdClientGetter) MemberPromote(ctx context.Context, member *etcdserver
 		}
 	}()
 
+	ctx, cancel := context.WithTimeout(ctx, DefaultClientTimeout)
+	defer cancel()
 	_, err = cli.MemberPromote(ctx, member.ID)
 	return err
 }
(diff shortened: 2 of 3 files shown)


lance5890:
@dusk125

tjungblu:
Thanks. @dusk125 is out for the coming weeks, but I'm happy to assist.
60s sounds high for almost all calls, maybe we should reduce this to 30s? Maybe defrag needs a different timeout, that can take much longer.

I do remember I have changed something similar in the past, we discovered controllers getting stuck in some calls before already. I believe I've added a timeout for the respective calls there too. Let me dig this up again.

lance5890:
> 

ok, we can reduce the timeout time by different situation, this PR only care about the connection timeout with the etcd，and we always need client timeout , i will modify the default timeout to 30s

lance5890:
> 

and the defrag controller use the etcd status interface, we can only change the etcd status interface timeout to suitable value

lance5890:
> Thanks. @dusk125 is out for the coming weeks, but I'm happy to assist. 60s sounds high for almost all calls, maybe we should reduce this to 30s? Maybe defrag needs a different timeout, that can take much longer.
> 
> I do remember I have changed something similar in the past, we discovered controllers getting stuck in some calls before already. I believe I've added a timeout for the respective calls there too. Let me dig this up again.

I have reposted my code again, change default timeout to 30s, change etcd status timeout to DefragDialTimeout(60s), and that would be my honor to merge my contribution within ocp

tjungblu:
found it, that was in a slightly different place with the same issue: 
https://github.com/openshift/cluster-etcd-operator/commit/4816fab709e11e0681b760003be3f1de12c9c103

I think adding this directly to the client makes sense with 30s as a default. Let me quickly create all the bug tickets around this, I'll carry this through the other releases for you.

pkg/cmd/verify/backupstorage.go
```diff
@@ -132,6 +132,8 @@ func (v *verifyBackupStorage) isStorageAdequate(ctx context.Context) (bool, erro
 	}
 	defer cli.Close()
 
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
 	members, err := cli.MemberList(ctx)
```
tjungblu, line 137: why would that listing take 60s?

pkg/etcdcli/etcdcli.go
```diff
@@ -39,6 +39,7 @@ const (
 	BootstrapIPAnnotationKey = "alpha.installer.openshift.io/etcd-bootstrap"
 	DefaultDialTimeout       = 15 * time.Second
 	DefragDialTimeout        = 60 * time.Second
+	DefaultCliTimeout        = 30 * time.Second
```
tjungblu, line 42: feel free to spell out Client, we're not short on characters today

pkg/cmd/verify/backupstorage.go
lance5890: I will reduce this to 30s

pkg/etcdcli/etcdcli.go
lance5890: ok， i will use DefaultClientTimeout

pkg/etcdcli/etcdcli.go
```diff
@@ -282,6 +293,8 @@ func (g *etcdClientGetter) Status(ctx context.Context, clientURL string) (*clien
 	}
 
 	defer g.clientPool.Return(cli)
+	ctx, cancel := context.WithTimeout(ctx, DefragDialTimeout)
```
tjungblu, line 296: I think that's fine and can be 30s, the actual defragmentation call could take longer. Can you change that back to `DefaultCliTimeout` please?

lance5890: ok， i will change it back to DefaultClientTimeout

lance5890:
> found it, that was in a slightly different place with the same issue: [4816fab](https://github.com/openshift/cluster-etcd-operator/commit/4816fab709e11e0681b760003be3f1de12c9c103)
> 
> I think adding this directly to the client makes sense with 30s as a default. Let me quickly create all the bug tickets around this, I'll carry this through the other releases for you.

all done， please review again , thanks!

tjungblu:
/retitle OCPBUGS-7374: set default timeouts in etcdcli

tjungblu:
awesome, thank you. 

/lgtm

lance5890:
> /cherry-pick release-4.12

4.11 has the same problem, could u cherry-pick to 4.11

tjungblu:
yes, I will backport both of your fixes to 4.9.  We have to backport them in order of releases.

tjungblu:
Thanks for your contribution @lance5890, I'll get the backports as soon as possible. It might take some time to end up in an official z-release however.

lance5890:
> Thanks for your contribution @lance5890, I'll get the backports as soon as possible. It might take some time to end up in an official z-release however.

it is my pleasure to cooperate with u, i will keep following with ceo, and i hope to contribute more than these two bugfixs in https://github.com/openshift/cluster-etcd-operator/issues/1000

PR #1006, opened 2023-02-15
OCPBUGS-7510: [release-4.11] fail early on missing node status envs

tjungblu:
cherry-pick from #1004 and manual go mod tidy


geliu2016 (approved):
/lgtm

tjungblu:
@geliu2016 I think that CPA command did not go through :/

Elbehery:
> @geliu2016 I think that CPA command did not go through :/

Yes, @mfojtik or @deads2k can do though 

Also I think @geliu2016 should be added to `patch-manager` list 🤔

PR #1012, opened 2023-02-21
ETCD-385: Prefer ipv6 on bootstrap in v6-primary dual stack

cybertron:
Previously we only supported v4-primary dual stack, so it was correct to always use ipv4 unless we were single stack v6. Now we need to add support for v6-primary dual stack so we need to be more nuanced about ip selection.

This change looks at the first CIDR to determine whether the cluster should be v4 or v6 primary. The other logic stays essentially the same since it was already set up to look for ipv6 addresses when required.

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -158,8 +158,8 @@ type TemplateData struct {
 	// MachineCIDR is the IP range for machine IPs.
 	MachineCIDR string
 
-	// SingleStackIPv6 is true if the stack is IPv6 only.
-	SingleStackIPv6 bool
+	// PreferIPv6 is true if IPv6 is the primary IP version.
+	PreferIPv6 bool
 
 	// Hostname as reported by the kernel.
 	Hostname string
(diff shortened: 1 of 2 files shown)


pkg/cmd/render/render.go
```diff
@@ -621,18 +621,16 @@ func writeManifests(outputDir, templateDir string, templateData interface{}) err
...
 	}
-	for _, cidr := range serviceCIDRs {
-		ip, _, err := net.ParseCIDR(cidr)
-		if err != nil {
-			return false, err
-		}
-		if ip.To4() != nil {
-			return false, nil
-		}
+	ip, _, err := net.ParseCIDR(serviceCIDRs[0])
```
tjungblu, line 628: can you please leave a comment here why we do that? 

maybe with a quick example of a serviceCIDR listing and it's desired behavior in terms of bootstrap IPs

pkg/cmd/render/render.go
```diff
@@ -621,18 +621,16 @@ func writeManifests(outputDir, templateDir string, templateData interface{}) err
 	return nil
 }
 
-func isSingleStackIPv6(serviceCIDRs []string) (bool, error) {
+func preferIPv6(serviceCIDRs []string) (bool, error) {
```
tjungblu, line 624: would you mind adding a quick unit test for this func specifically? just testing a couple of input/output pairs

cybertron: Yep, will do.

pkg/cmd/render/render.go
cybertron: Sure.

cybertron (commented):
Will update this shortly to address the inline comments.

pkg/cmd/render/render.go
tjungblu: perfect, thanks!

pkg/cmd/render/render.go
tjungblu: awesome, thanks again.

tjungblu:
@cybertron I think there's just a formatting issue left:

> verify-gofmt failed - please run `make update-gofmt`


otherwise I'm lgtm :)

cybertron:
Oops, fixed!

Elbehery:
@cybertron do we need backports ?

cybertron:
> @cybertron do we need backports ?

No, this is for a new feature in 4.13.

PR #1013, opened 2023-02-24
document adaptation with ceo

lance5890:
1. The Prune func are general methods for all static pod(like  kube-apiserver,  kube-controller-manager, kube-scheduler, etcd)
2. But in the ceo docs, the description should  be used for cluster-etcd-operator as an example, to avoid misleading


lance5890:
@Elbehery

tjungblu:
/lgtm

thanks @lance5890

PR #1018, opened 2023-03-03
OCPBUGS-1153: Add etcd operator roles and clusterrole

Elbehery:
This PR add replaces usage ClusterRole cluster-admin with ClusterRole and Role for etcd operator.

The ClusterRole and Role added in this PR are reverse engineering using https://github.com/liggitt/audit2rbac.
After inspecting the kube-apiserver audit logs using the tools, I was able to identify the resources and verbs required by the operator.

resolves https://issues.redhat.com/browse/OCPBUGS-1153

### DIFF
--- a/manifests/0000_20_etcd-operator_11_clusterrole.yaml
+++ b/manifests/0000_20_etcd-operator_11_clusterrole.yaml
@@ -1,0 +1,83 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: etcd-operator-cluster-role
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+  rules:
+  - apiGroups:
+    - ""
+    resources:
+    - configmaps
+    verbs:
+    - get
+    - list
+    - patch
+    - update
+    - watch
+  - apiGroups:
+    - ""
+    resources:
+    - namespaces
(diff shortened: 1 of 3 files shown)


tjungblu:
@Elbehery can you help me on Monday to go through all the job logs and look for cases where the operator might not have sufficient permissions? 

I'm 99% sure that we don't have enough test coverage for all functionality in place today.

tjungblu:
/hold 

for the time being

manifests/0000_20_etcd-operator_11_clusterrole.yaml
```diff
@@ -0,0 +1,83 @@
...
+  - apiGroups:
+    - ""
+    resources:
+    - namespaces
+    - nodes
+    - secrets
+    verbs:
+    - get
+    - list
+    - watch
```
tjungblu, line 29: I think we're missing a create here to create the openshift-etcd namespace. @Elbehery can you go through all the resources that bootkube must create? They can be found here: https://github.com/openshift/cluster-etcd-operator/tree/master/bindata/bootkube/manifests

tjungblu: even though I believe this should happen before the operator runs, so I'll dig deeper on Monday, maybe you can figure it out already from the install bundles :)

Elbehery: SGTM .. it will be done before Monday ;)

Elbehery:
closing this in favour of https://github.com/openshift/cluster-etcd-operator/pull/1020

/close

PR #1027, opened 2023-03-16
OCPBUGS-10351: skip machine deletion during boostrap

Elbehery:
resolves https://issues.redhat.com/browse/OCPBUGS-10351 

cc @tjungblu @hasbro17 @wking @damdo

### DIFF
--- a/pkg/operator/ceohelpers/bootstrap.go
+++ b/pkg/operator/ceohelpers/bootstrap.go
@@ -107,7 +107,7 @@ func CheckSafeToScaleCluster(
 	infraLister configv1listers.InfrastructureLister,
 	etcdClient etcdcli.EtcdClient) error {
 
-	bootstrapComplete, err := IsBootstrapComplete(configmapLister, staticPodClient, etcdClient, infraLister)
+	bootstrapComplete, err := IsBootstrapComplete(configmapLister, staticPodClient, etcdClient)
 	if err != nil {
 		return fmt.Errorf("CheckSafeToScaleCluster failed to determine bootstrap status: %w", err)
 	}
@@ -161,7 +161,7 @@ func CheckSafeToScaleCluster(
 }
 
 // IsBootstrapComplete returns true if bootstrap has completed.
-func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPodClient v1helpers.StaticPodOperatorClient, etcdClient etcdcli.EtcdClient, infraLister configv1listers.InfrastructureLister) (bool, error) {
+func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPodClient v1helpers.StaticPodOperatorClient, etcdClient etcdcli.EtcdClient) (bool, error) {
 	// do a cheap check to see if the annotation is already gone.
 	// check to see if bootstrapping is complete
 	bootstrapFinishedConfigMap, err := configMapClient.ConfigMaps("kube-system").Get("bootstrap")
@@ -195,15 +195,6 @@ func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPo
 			klog.V(4).Infof("bootstrap considered incomplete because revision %d is still in progress", status.LatestAvailableRevision)
 			return false, nil
 		}
-	}
-
-	// skip listing etcd membership for SingleReplicaTopologyMode
-	singleNode, err := IsSingleNodeTopology(infraLister)
-	if err != nil {
-		return false, fmt.Errorf("failed to get control plane topology: %w", err)
-	}
-	if singleNode {
-		return true, nil
 	}
 
 	// check if etcd-bootstrap member is still present within the etcd cluster membership
--- a/pkg/operator/ceohelpers/bootstrap.go
+++ b/pkg/operator/ceohelpers/bootstrap.go
@@ -107,7 +107,7 @@ func CheckSafeToScaleCluster(
 	infraLister configv1listers.InfrastructureLister,
 	etcdClient etcdcli.EtcdClient) error {
 
-	bootstrapComplete, err := IsBootstrapComplete(configmapLister, staticPodClient, etcdClient)
+	bootstrapComplete, err := IsBootstrapComplete(configmapLister, staticPodClient, etcdClient, infraLister)
 	if err != nil {
 		return fmt.Errorf("CheckSafeToScaleCluster failed to determine bootstrap status: %w", err)
 	}
@@ -161,7 +161,7 @@ func CheckSafeToScaleCluster(
 }
 
 // IsBootstrapComplete returns true if bootstrap has completed.
-func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPodClient v1helpers.StaticPodOperatorClient, etcdClient etcdcli.EtcdClient) (bool, error) {
+func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPodClient v1helpers.StaticPodOperatorClient, etcdClient etcdcli.EtcdClient, infraLister configv1listers.InfrastructureLister) (bool, error) {
 	// do a cheap check to see if the annotation is already gone.
 	// check to see if bootstrapping is complete
 	bootstrapFinishedConfigMap, err := configMapClient.ConfigMaps("kube-system").Get("bootstrap")
@@ -195,6 +195,15 @@ func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPo
 			klog.V(4).Infof("bootstrap considered incomplete because revision %d is still in progress", status.LatestAvailableRevision)
 			return false, nil
 		}
+	}
+
+	// skip listing etcd membership for SingleReplicaTopologyMode
+	singleNode, err := IsSingleNodeTopology(infraLister)
+	if err != nil {
+		return false, fmt.Errorf("failed to get control plane topology: %w", err)
+	}
+	if singleNode {
+		return true, nil
 	}
 
 	// check if etcd-bootstrap member is still present within the etcd cluster membership
--- a/pkg/operator/ceohelpers/bootstrap.go
+++ b/pkg/operator/ceohelpers/bootstrap.go
@@ -3,6 +3,7 @@ package ceohelpers
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
 
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
@@ -11,7 +12,6 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
-	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
 )
 
 // BootstrapScalingStrategy describes the invariants which will be enforced when
@@ -60,7 +60,7 @@ const (
 	DelayedHABootstrapScalingStrategyAnnotation = "openshift.io/delayed-ha-bootstrap"
 )
 
-// GetBootstrapScalingStrategy determines the scaling strategy to use.
+// GetBootstrapScalingStrategy determines the scaling strategy to use
 func GetBootstrapScalingStrategy(staticPodClient v1helpers.StaticPodOperatorClient, namespaceLister corev1listers.NamespaceLister, infraLister configv1listers.InfrastructureLister) (BootstrapScalingStrategy, error) {
 	var strategy BootstrapScalingStrategy
 
@@ -107,7 +107,7 @@ func CheckSafeToScaleCluster(
 	infraLister configv1listers.InfrastructureLister,
 	etcdClient etcdcli.EtcdClient) error {
 
-	bootstrapComplete, err := IsBootstrapComplete(configmapLister, staticPodClient)
+	bootstrapComplete, err := IsBootstrapComplete(configmapLister, staticPodClient, etcdClient)
 	if err != nil {
 		return fmt.Errorf("CheckSafeToScaleCluster failed to determine bootstrap status: %w", err)
 	}
@@ -161,7 +161,7 @@ func CheckSafeToScaleCluster(
 }
 
 // IsBootstrapComplete returns true if bootstrap has completed.
-func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPodClient v1helpers.StaticPodOperatorClient) (bool, error) {
+func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPodClient v1helpers.StaticPodOperatorClient, etcdClient etcdcli.EtcdClient) (bool, error) {
 	// do a cheap check to see if the annotation is already gone.
 	// check to see if bootstrapping is complete
 	bootstrapFinishedConfigMap, err := configMapClient.ConfigMaps("kube-system").Get("bootstrap")
@@ -196,5 +196,18 @@ func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPo
 			return false, nil
 		}
 	}
+
+	// check if etcd-bootstrap member is still present within the etcd cluster membership
+	membersList, err := etcdClient.MemberList(context.Background())
+	if err != nil {
+		return false, fmt.Errorf("IsBootstrapComplete couldn't list the etcd cluster members: %w", err)
+	}
+	for _, m := range membersList {
+		if m.Name == "etcd-bootstrap" {
+			klog.V(4).Infof("(etcd-bootstrap) member is still present in the etcd cluster membership")
+			return false, nil
+		}
+	}
+
 	return true, nil
 }
--- a/pkg/operator/ceohelpers/bootstrap_test.go
+++ b/pkg/operator/ceohelpers/bootstrap_test.go
@@ -165,36 +165,56 @@ func Test_IsBootstrapComplete(t *testing.T) {
 	tests := map[string]struct {
 		bootstrapConfigMap *corev1.ConfigMap
 		nodes              []operatorv1.NodeStatus
+		etcdMembers        []*etcdserverpb.Member
 		expectComplete     bool
 		expectError        error
 	}{
 		"bootstrap complete, nodes up to date": {
 			bootstrapConfigMap: bootstrapComplete,
 			nodes:              twoNodesAtCurrentRevision,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			expectComplete:     true,
 			expectError:        nil,
 		},
 		"bootstrap progressing, nodes up to date": {
 			bootstrapConfigMap: bootstrapProgressing,
 			nodes:              twoNodesAtCurrentRevision,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			expectComplete:     false,
 			expectError:        nil,
 		},
 		"bootstrap configmap missing": {
 			bootstrapConfigMap: nil,
 			nodes:              twoNodesAtCurrentRevision,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			expectComplete:     false,
 			expectError:        nil,
 		},
 		"bootstrap complete, no recorded revisions": {
 			bootstrapConfigMap: bootstrapComplete,
 			nodes:              zeroNodesAtAnyRevision,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			expectComplete:     true,
 			expectError:        nil,
 		},
 		"bootstrap complete, node progressing": {
 			bootstrapConfigMap: bootstrapComplete,
 			nodes:              twoNodesProgressingTowardsCurrentRevision,
+			etcdMembers:        u.DefaultEtcdMembers(),
+			expectComplete:     false,
+			expectError:        nil,
+		},
+		"bootstrap complete, etcd-bootstrap removed": {
+			bootstrapConfigMap: bootstrapComplete,
+			nodes:              twoNodesAtCurrentRevision,
+			etcdMembers:        u.DefaultEtcdMembers(),
+			expectComplete:     true,
+			expectError:        nil,
+		},
+		"bootstrap complete, etcd-bootstrap exists": {
+			bootstrapConfigMap: bootstrapComplete,
+			nodes:              twoNodesAtCurrentRevision,
+			etcdMembers:        append(u.DefaultEtcdMembers(), u.FakeEtcdBoostrapMember(0)),
 			expectComplete:     false,
 			expectError:        nil,
 		},
@@ -216,25 +236,20 @@ func Test_IsBootstrapComplete(t *testing.T) {
 			}
 			fakeStaticPodClient := v1helpers.NewFakeStaticPodOperatorClient(nil, operatorStatus, nil, nil)
 
-			actualComplete, actualErr := IsBootstrapComplete(fakeConfigMapLister, fakeStaticPodClient)
+			fakeEtcdClient, err := etcdcli.NewFakeEtcdClient(test.etcdMembers)
+			if err != nil {
+				t.Fatal(err)
+			}
 
-			if test.expectComplete != actualComplete {
-				t.Errorf("expected complete=%v, got %v", test.expectComplete, actualComplete)
-			}
-			if test.expectError != actualErr {
-				t.Errorf("expected error=%v, got %v", test.expectError, actualErr)
-			}
+			actualComplete, actualErr := IsBootstrapComplete(fakeConfigMapLister, fakeStaticPodClient, fakeEtcdClient)
+
+			assert.Equal(t, test.expectComplete, actualComplete)
+			assert.Equal(t, test.expectError, actualErr)
 		})
 	}
 }
 
 func Test_CheckSafeToScaleCluster(t *testing.T) {
-	defaultEtcdMembers := []*etcdserverpb.Member{
-		u.FakeEtcdMemberWithoutServer(0),
-		u.FakeEtcdMemberWithoutServer(1),
-		u.FakeEtcdMemberWithoutServer(2),
-	}
-
 	tests := map[string]struct {
 		namespace          *corev1.Namespace
 		bootstrapConfigMap *corev1.ConfigMap
@@ -250,7 +265,7 @@ func Test_CheckSafeToScaleCluster(t *testing.T) {
 			bootstrapConfigMap: bootstrapComplete,
 			operatorConfig:     defaultOperatorConfig,
 			nodes:              threeNodesAtCurrentRevision,
-			etcdMembers:        defaultEtcdMembers,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			infraObj:           defaultInfra,
 			expectError:        nil,
 		},
@@ -259,7 +274,7 @@ func Test_CheckSafeToScaleCluster(t *testing.T) {
 			bootstrapConfigMap: bootstrapComplete,
 			operatorConfig:     defaultOperatorConfig,
 			nodes:              twoNodesAtCurrentRevision,
-			etcdMembers:        defaultEtcdMembers,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			infraObj:           defaultInfra,
 			expectError:        fmt.Errorf("CheckSafeToScaleCluster 3 nodes are required, but only 2 are available"),
 		},
@@ -268,7 +283,7 @@ func Test_CheckSafeToScaleCluster(t *testing.T) {
 			bootstrapConfigMap: bootstrapComplete,
 			operatorConfig:     unsupportedOperatorConfig,
 			nodes:              oneNodeAtCurrentRevision,
-			etcdMembers:        defaultEtcdMembers,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			infraObj:           defaultInfra,
 			expectError:        nil,
 		},
@@ -277,7 +292,7 @@ func Test_CheckSafeToScaleCluster(t *testing.T) {
 			bootstrapConfigMap: bootstrapComplete,
 			operatorConfig:     unsupportedOperatorConfig,
 			nodes:              zeroNodesAtAnyRevision,
-			etcdMembers:        defaultEtcdMembers,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			infraObj:           defaultInfra,
 			expectError:        nil,
 		},
@@ -286,7 +301,7 @@ func Test_CheckSafeToScaleCluster(t *testing.T) {
 			bootstrapConfigMap: bootstrapProgressing,
 			operatorConfig:     defaultOperatorConfig,
 			nodes:              twoNodesAtCurrentRevision,
-			etcdMembers:        defaultEtcdMembers,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			infraObj:           defaultInfra,
 			expectError:        nil,
 		},
@@ -295,7 +310,7 @@ func Test_CheckSafeToScaleCluster(t *testing.T) {
 			bootstrapConfigMap: bootstrapProgressing,
 			operatorConfig:     defaultOperatorConfig,
 			nodes:              oneNodeAtCurrentRevision,
-			etcdMembers:        defaultEtcdMembers,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			infraObj:           defaultInfra,
 			expectError:        nil,
 		},
@@ -304,7 +319,7 @@ func Test_CheckSafeToScaleCluster(t *testing.T) {
 			bootstrapConfigMap: bootstrapComplete,
 			operatorConfig:     defaultOperatorConfig,
 			nodes:              threeNodesAtCurrentRevision,
-			etcdMembers:        defaultEtcdMembers,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			infraObj:           defaultInfra,
 			expectError:        nil,
 		},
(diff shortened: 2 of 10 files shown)


pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -132,6 +132,10 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 		return fmt.Errorf("could not list etcd members: %w", err)
 	}
 
+	// removing etcd-boostrap from live voting members
+	// see https://issues.redhat.com/browse/OCPBUGS-10351
+	liveVotingMembers = removeBootstrapMember(liveVotingMembers)
+
 	if len(liveVotingMembers) < desiredControlPlaneReplicasCount {
 		klog.V(2).Infof("Ignoring scale-down since the number of etcd members (%d) < desired number of control-plane replicas (%d) ", len(liveVotingMembers), desiredControlPlaneReplicasCount)
```
wking, line 140: nit: Now that this might include bootstrap exclusion, did we want to soften "the number of etcd members" to  "the number of production etcd members" or "the number of non-bootstrap etcd members" or something?

Elbehery: iiuc from haseeb comment, we want the number of voting member excluding the bootstrap member, to be more the desiredControlPlane 

cc @hasbro17

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -132,6 +132,10 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
 		return fmt.Errorf("could not list etcd members: %w", err)
 	}
 
+	// removing etcd-boostrap from live voting members
+	// see https://issues.redhat.com/browse/OCPBUGS-10351
+	liveVotingMembers = removeBootstrapMember(liveVotingMembers)
```
wking, line 137: Do we want similar filtering [down here][1] to remove a possible bootstrap member from `healthyLiveVotingMembers`, [the quorum check][2], and [the all-healthy check][3]?

[1]: https://github.com/openshift/cluster-etcd-operator/blob/0584b0d1c8868535baf889d8c199f605aef4a3ae/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go#L158-L162
[2]: https://github.com/openshift/cluster-etcd-operator/blob/0584b0d1c8868535baf889d8c199f605aef4a3ae/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go#L171
[3]: https://github.com/openshift/cluster-etcd-operator/blob/0584b0d1c8868535baf889d8c199f605aef4a3ae/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go#L216

Elbehery: +1, thats why i wanted to skip scaling down as long as `etcd-boostrap` exist, imho it is cleaner fix :) 

wdyt @hasbro17 @tjungblu

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
wking: > we want the number of voting member excluding the bootstrap member, to be more the desiredControlPlane

I agree with "non-bootstrap voters >= desired count" as the criteria.  I'm just quibbling with logs.  For example, say your desired count was 3 and you had one bootstrap and two non-bootstrap voters.  We want to ignore scale-down, because (2 non-bootstrap voters) < (3 desired count).  So far, so good.  But as your pull stands, we'll log:

> ... since the number of etcd members (2) < desired number of control-plane replicas (3)

But there are 3 voting members.  It's just that one of those members is the bootstrap node.  I want us to qualify  the logs to say something like:

> ... since the number of voting, non-bootstrap etcd members (2) < desired number of control-plane replicas (3)

And also, staring at this line for too long, I see the current code has `replicas (%d) "` with trailing whitespace, which should probably be `replicas (%d)"`.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
wking: I agree  on skipping scale-down until we have a full set of non-bootstrap voting members.  But the presence of such doesn't mean the non-bootstrap voting members are all healthy yet.  I was suggesting something like:

```diff
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
@@ -165,6 +165,12 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
                return fmt.Errorf("could not list etcd healthy members: %w", err)
        }
 
+       // We pruned bootstrap from liveVotingMembers above.  Prune
+       // healthyVotingMembers the same way, so we can compare them, and
+       // don't count a healthy bootstrap member when comparing with
+       // minTolerableQuorum or desiredControlPlaneReplicasCount.
+       healthyVotingMembers = removeBootstrapMember(healthyVotingMembers)
+
        // scaling down invariant
        minTolerableQuorum, err := etcdcli.MinimumTolerableQuorum(desiredControlPlaneReplicasCount)
        if err != nil {
```

But I also don't have a solid grasp on this logic, so it's certainly possible I'm missing something, and your pivot is fine as it stands.

romfreiman:
Please verify that it does not break neither assisted installer nor sno. Thanks!

Elbehery:
> Please verify that it does not break neither assisted installer nor sno. Thanks!

Thanks @romfreiman for your review 
/ 
May I ask how to verify this

wking:
[e2e-aws-ovn-single-node passing][1] is a good sign that single-node wasn't broken :)

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/1027/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-ovn-single-node/1636469029705093120

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: done

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: done

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
```diff
@@ -25,6 +27,28 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
+var (
+	bootstrapComplete = &corev1.ConfigMap{
```
tjungblu, line 31: I think we have those in test utils already: https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/testutils/testutils.go#L168-L181

tjungblu: also saw it's duplicated in `bootstrap_teardown_controller_test.go`, if you want to refactor the extra mile :)

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -114,6 +120,11 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, syncCtx facto
 // attemptToScaleDown attempts to remove a voting member only once we have identified that
 // a Machine resource is being deleted and a replacement member has been created
 func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context, recorder events.Recorder) error {
+	scalingStrategy, err := ceohelpers.GetBootstrapScalingStrategy(c.operatorClient, c.namespaceLister, c.infrastructureLister)
```
tjungblu, line 123: I think I have something simpler for you, I would check two things in sync() already:
* is it currently bootstrapping?
* is it safe based on the number of healthy/unhealthy members?

Right next to the method you've moved into common, there's another called `CheckSafeToScaleCluster` which does exactly that. I think you can just call it early in `sync` and just make it a little smarter, for example by updating `IsBootstrapComplete` to also include an etcdclient memberList to see if there's still a bootstrap node in there.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -132,8 +143,13 @@ func (c *clusterMemberRemovalController) attemptToScaleDown(ctx context.Context,
...
 	}
 
+	// removing etcd-boostrap from live voting members count for HAScalingStrategy
+	if scalingStrategy == ceohelpers.HAScalingStrategy {
+		liveVotingMembers = removeBootstrapMember(liveVotingMembers)
+	}
+
 	if len(liveVotingMembers) < desiredControlPlaneReplicasCount {
-		klog.V(2).Infof("Ignoring scale-down since the number of etcd members (%d) < desired number of control-plane replicas (%d) ", len(liveVotingMembers), desiredControlPlaneReplicasCount)
+		klog.V(2).Infof("Ignoring scale-down since the number of voting, non-bootstrap etcd members is (%d) < desired number of control-plane replicas (%d)", len(liveVotingMembers), desiredControlPlaneReplicasCount)
```
wking, line 152: The "non-bootstrap" qualifier [I'd asked for earlier][1] is a bit fiddly now that `removeBootstrapMember` is conditioned on `scalingStrategy == ceohelpers.HAScalingStrategy`.  Maybe use something more generic like "the number of production etcd members"?  Or use a string variable to hold the relevant specific qualifier:

```go
memberQualifier := "voting"
if scalingStrategy == ceohelpers.HAScalingStrategy {
	memberQualifier = "voting, non-bootstrap"
	liveVotingMembers = removeBootstrapMember(liveVotingMembers)
}
```	

and then add that to logs with `... number of %s etcd members...`?

[1]: https://github.com/openshift/cluster-etcd-operator/pull/1027#discussion_r1139351212

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
Elbehery: lets do it 👍🏽

Elbehery: done

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: thanks @wking for the review

May I ask what is the benefit from `memberQualifier = "voting, non-bootstrap` ? 

I check the scaling strategy in the beginning of the control-loop (i.e. sync), and use this assertion before deducting the `etcd-bootstrap` member from the voting-members. 

iiuc, we need not deducing the boot-strap count from the membership in other strategies, please correct me if I am wrong :)

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: done 👍🏽

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -449,3 +479,14 @@ func hasInternalIP(machine *machinev1beta1.Machine, memberInternalIP string) boo
 	}
 	return false
 }
+
+func removeBootstrapMember(members []*etcdserverpb.Member) []*etcdserverpb.Member {
```
tsorya, line 483: Name is actually very confusing, it seems like it expects from this function to actually remove bootstrap and not just filter it from list

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
tsorya: I am not sure why this new logic is not relevant for delayed strategy but message is not right in that case as you will have bootstrap member in case it is not HaScalingStrategy, right?

tsorya: But why this not needed for DelayedHAScalingStrategy?

Elbehery: Thanks @tsorya for your review 

iiuc, we trying to maintain the etcd membership for HAScalingStrategy preserving the quorum. CPMS triggered master machine replacement while the boot-strap member still exist in the etcd cluster membership. This caused [bug](https://issues.redhat.com/browse/OCPBUGS-8038).

for DelayedHAScalingStrategy, I am not sure if similar issue was found 


cc @JoelSpeed @damdo

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: done 👍🏽

Elbehery:
@romfreiman  `ci/prow/e2e-metal-single-node-live-iso — Job succeeded` 🎉

Elbehery:
anyone fancy `lgtm/approve` here ?

tjungblu:
/hold

I'll take a look in the afternoon today. @hasbro17 please also review

Elbehery:
> /test e2e-aws-ovn-etcd-scaling

i was just trying to save some Dollars 💵

tjungblu:
> i was just trying to save some Dollars dollar

maybe we shouldn't save dollars on the only test that actually tests the code that you've changed ;)

Elbehery:
> > i was just trying to save some Dollars dollar
> 
> maybe we shouldn't save dollars on the only test that actually tests the code that you've changed ;)

u r totally right 👍🏽 , my apologies 🙏🏽

pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -196,5 +159,17 @@ func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPo
 			return false, nil
 		}
 	}
+
+	// check if etcd-bootsrap member is still present within the etcd cluster membership
+	memberHealth, err := etcdClient.MemberHealth(context.Background())
```
tjungblu, line 164: why memberHealth and not plain listing? you're not even using the health result

pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -196,5 +159,17 @@ func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPo
...
+	if err != nil {
+		return false, fmt.Errorf("IsBootstrapComplete couldn't determine member health: %w", err)
+	}
+	for _, m := range memberHealth {
+		if m.Member.Name == "etcd-bootstrap" {
+			return false, fmt.Errorf("(%v) member is till present the etcd cluster membership", m.Member.Name)
```
tjungblu, line 170: still*

pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -60,42 +59,6 @@ const (
 	DelayedHABootstrapScalingStrategyAnnotation = "openshift.io/delayed-ha-bootstrap"
 )
 
-// GetBootstrapScalingStrategy determines the scaling strategy to use.
-func GetBootstrapScalingStrategy(staticPodClient v1helpers.StaticPodOperatorClient, namespaceLister corev1listers.NamespaceLister, infraLister configv1listers.InfrastructureLister) (BootstrapScalingStrategy, error) {
```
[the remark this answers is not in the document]
Elbehery: I moved it because it is being used here, and i needed it in `clustermemberremovalcontroller`

pkg/operator/ceohelpers/bootstrap.go
Elbehery: +1, fix on the way

pkg/operator/ceohelpers/bootstrap_test.go
```diff
@@ -165,36 +165,42 @@ func Test_IsBootstrapComplete(t *testing.T) {
...
 			expectError:        nil,
 		},
 		"bootstrap complete, no recorded revisions": {
 			bootstrapConfigMap: bootstrapComplete,
 			nodes:              zeroNodesAtAnyRevision,
+			etcdMembers:        u.DefaultEtcdMembers(),
 			expectComplete:     true,
 			expectError:        nil,
 		},
 		"bootstrap complete, node progressing": {
```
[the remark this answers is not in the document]
Elbehery: sure adding this now  👍🏽 

actually `defaultEtcdMembers` are being used in many tests, sometimes using this and sometimes inlined iirc 

i just did not want to make a lot of changes in the same PR

pkg/operator/ceohelpers/common.go
```diff
@@ -197,3 +200,39 @@ func VotingMemberIPListSet(ctx context.Context, cli etcdcli.EtcdClient) (sets.St
 
 	return currentVotingMemberIPListSet, nil
 }
+
+// GetBootstrapScalingStrategy determines the scaling strategy to use
+func GetBootstrapScalingStrategy(staticPodClient v1helpers.StaticPodOperatorClient, namespaceLister corev1listers.NamespaceLister, infraLister configv1listers.InfrastructureLister) (BootstrapScalingStrategy, error) {
```
tjungblu, line 205: can you please move that one back to its original file? it's hard to see what changed here.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
tjungblu: not quite what I meant, but thanks for adding this one here too.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
tjungblu: I also don't like that `if scalingStrategy == ceohelpers.HAScalingStrategy {` littered all over the place, it's very hard to follow the logic. Can you please add this to `CheckSafeToScaleCluster`?

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -449,3 +479,14 @@ func hasInternalIP(machine *machinev1beta1.Machine, memberInternalIP string) boo
 	}
 	return false
 }
+
+func removeBootstrapMemberCount(members []*etcdserverpb.Member) []*etcdserverpb.Member {
```
tjungblu, line 483: please add a test for that function, I would also think this could be a better fit for our etcdcli.

pkg/operator/ceohelpers/bootstrap.go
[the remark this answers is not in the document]
tjungblu: you can use it from bootstrap.go just fine, why does it need to be in common.go?

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller_test.go
```diff
@@ -985,6 +1020,81 @@ func TestAttemptToScaleDown(t *testing.T) {
 				}
 			},
 		},
+		{
```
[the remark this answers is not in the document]
tjungblu: IIRC the problem occurred when we had 2 masters + one bootstrap that was considered "safe" to delete, so you're missing that case too.

pkg/testutils/testutils.go
```diff
@@ -394,3 +394,11 @@ func FakeNetworkLister(t *testing.T, isIPv6 bool) configv1listers.NetworkLister
 	}
 	return configv1listers.NewNetworkLister(indexer)
 }
+
+func DefaultEtcdMembers() []*etcdserverpb.Member {
```
tjungblu, line 398: great addition!

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: shall i move it to commons ?

pkg/operator/ceohelpers/common.go
Elbehery: :O ok

pkg/operator/ceohelpers/bootstrap.go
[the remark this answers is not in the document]
Elbehery: ok 👍🏽

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: what am i missing ?? .. i still can change it

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: > I also don't like that `if scalingStrategy == ceohelpers.HAScalingStrategy {` littered all over the place, it's very hard to follow the logic. Can you please add this to `CheckSafeToScaleCluster`?

so i skip the sync loop for other strategies 🤔  .. cant get what you mean here

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: moved to common and added test

pkg/operator/ceohelpers/common.go
Elbehery: moved back to bootstrap

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -90,6 +99,12 @@ func NewClusterMemberRemovalController(
 }
 
 func (c *clusterMemberRemovalController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	// skip reconciling this controller, if it is unsafe to scale the etcd cluster membership
+	err := ceohelpers.CheckSafeToScaleCluster(c.configMapLister, c.operatorClient, c.namespaceLister, c.infrastructureLister, c.etcdClient)
+	if err != nil {
+		return err
+	}
```
tjungblu, line 106: that's already your job done, you don't need to add any other code change in this controller. At this point, there is only a safe cluster without a bootstrap node and etcd member.

tjungblu: apologies, I got the conditions backwards. At this point there may be a bootstrap member still present. So instead of this `CheckSafeToScaleCluster` you can use `bootstrapComplete, err := IsBootstrapComplete(configmapLister, staticPodClient, etcdClient)`.

Only continue the sync loop if bootstrap is really completed.

Elbehery: done 👍🏽

wking:
[All][1], [three][2] [recent][3] failures look unrelated to this change to me.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/1027/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-ovn-single-node/1638259075357609984
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/1027/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-ovn-serial/1638259075005288448
[3]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/1027/pull-ci-openshift-cluster-etcd-operator-master-e2e-agnostic-ovn-upgrade/1638259074900430848

Elbehery:
> [All](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/1027/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-ovn-single-node/1638259075357609984), [three](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/1027/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-ovn-serial/1638259075005288448) [recent](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/1027/pull-ci-openshift-cluster-etcd-operator-master-e2e-agnostic-ovn-upgrade/1638259074900430848) failures look unrelated to this change to me.

Thanks @wking for your follow up .. It is OVN :D :D :D

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
tjungblu: just one last nitpick:

> 		klog.V(2).Infof("bootstrapping is not complete, ignoring scaling down")

this can get very spammy, especially since it's syncing every minute or so with the node status.
I would just leave this log out.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
tjungblu: all good, you fixed this with the other comment already!

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
tjungblu: > so i skip the sync loop for other strategies  .. cant get what you mean here

we just settled in-person that we rely on the BootstrapTearDownController to determine the safety of the scale-down. As long as there's a bootstrap member it's never safe to scale down. We can implicitly rely on this from this controller.

That will safe us all the math and extra checks for strategy as this is already in the other controller.

tjungblu:
lgtm, @hasbro17 do you want to take another look? otherwise I'm fine to merge

pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -196,5 +196,17 @@ func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPo
...
 	}
+
+	// check if etcd-bootstrap member is still present within the etcd cluster membership
+	membersList, err := etcdClient.MemberList(context.Background())
+	if err != nil {
+		return false, fmt.Errorf("IsBootstrapComplete couldn't list the etcd cluster members: %w", err)
+	}
+	for _, m := range membersList {
+		if m.Name == "etcd-bootstrap" {
+			return false, fmt.Errorf("(etcd-bootstrap) member is still present in the etcd cluster membership")
```
hasbro17, line 207: `IsBootstrapComplete()` is supposed to return true if complete and false if not. The error should only be returned if we have an actual error condition e.g failing to list members.

Finding that the etcd-bootstrap member is present is an expected state during bootstrapping and should not result in an error.
As it is, the usage of  `IsBootstrapComplete()` in the clustermemberremoval controller [L96-102](https://github.com/openshift/cluster-etcd-operator/blob/9cf91a0df3bfd9407b10811b5eb709c3160df371/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go#L96-L102) will always result in the error `IsBootstrapComplete failed to determine bootstrap status: (etcd-bootstrap) member is still present in the etcd cluster membership` being returned during bootstrap and that controller will go degraded.
```Go
func (c *clusterMemberRemovalController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
	// skip reconciling this controller, if bootstrapping is not completed
	bootstrapComplete, err := ceohelpers.IsBootstrapComplete(c.configMapLister, c.operatorClient, c.etcdClient)
	if err != nil {
		return fmt.Errorf("IsBootstrapComplete failed to determine bootstrap status: %w", err)
	}
	if !bootstrapComplete {
		return nil
	}
```

Same for any other controller relying on this e.g etcdendpointscontroller
https://github.com/openshift/cluster-etcd-operator/blob/9cf91a0df3bfd9407b10811b5eb709c3160df371/pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go#L97-L100

hasbro17 (changes requested):
@Elbehery Can you confirm `isBootstrapComplete()` will be returning the error message during bootstrap like I mentioned?

hasbro17:
/hold 
Need to confirm if we're inadvertently making the clustermemberremoval and etcdendpoints controllers go degraded with this change.

pkg/operator/ceohelpers/bootstrap.go
hasbro17: Having a quick look around at the [past CI runs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/1027/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-ovn-serial/1638564021026164736/artifacts/e2e-aws-ovn-serial/gather-extra/artifacts/pods/openshift-etcd-operator_etcd-operator-8b57fd5bc-t287x_etcd-operator.log) it doesn't seem like the controllers get degraded during bootstrap because `IsBootstrapComplete()` is correctly returning a `false, nil` from the other pre-checks earlier up.
However if for whatever reason we actually hit this condition of the member being present then those controllers will unnecessarily go degraded.
So this should also be returning `false, nil` like the above pre-checks when the bootstrap member is present.

And since we don't want to silently be spinning on this if the bootstrap member is not being removed for whatever reason, then we can add a debug log here to avoid being spammed.
```Go
klog.V(2).Infof("(etcd-bootstrap) member is still present in the etcd cluster membership")
```

Elbehery: +1 for both reviews 👍🏽  

I will use debug level to avoid spamming the CEO logs, i think it is fair 👍🏽

Elbehery:
> /hold Need to confirm if we're inadvertently making the clustermemberremoval and etcdendpoints controllers go degraded with this change.

Do u think the CI will show this ? .. this would happen during bootstrapping, no ? 🤔

pkg/operator/ceohelpers/bootstrap.go
hasbro17: Sorry typo earlier, debug is `klog.V(4)` I believe.

hasbro17:
/lgtm
/hold cancel

The scaling test is passing, but the job is displaying the usual pathological events test flake which is being triaged separately.

Elbehery:
> /retest-required
> 
> 
> 
> 

appreciate it buddy 🙏🙏🙏

Elbehery:
@romfreiman  @eranco74  hello ✋🏽 

`ci/prow/e2e-aws-ovn-single-node — Job failed` never builds, any idea why ? 🤔

eranco74:
> @romfreiman @eranco74 hello ✋🏽
> 
> `ci/prow/e2e-aws-ovn-single-node — Job failed` never builds, any idea why ? thinking

I"m guessing the failure is related to the changes in this PR, since the job is passing for other PRs:
https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-ovn-single-node

The job installs SNO using a bootstrap node, I'm not sure what is the scaling strategy for this case but it is unique so it's possible that a change to the bootstrap etcd-member deletion will result in a failed installation.
I'll try to take a closer look into the failure reason

pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -196,5 +196,18 @@ func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPo
...
+	membersList, err := etcdClient.MemberList(context.Background())
+	if err != nil {
+		return false, fmt.Errorf("IsBootstrapComplete couldn't list the etcd cluster members: %w", err)
+	}
+	for _, m := range membersList {
+		if m.Name == "etcd-bootstrap" {
+			klog.V(4).Infof("(etcd-bootstrap) member is still present in the etcd cluster membership")
+			return false, nil
+		}
+	}
```
Elbehery, line 210:
```suggestion
	for _, m := range membersList {
		if m.Name == "etcd-bootstrap" {
			klog.V(4).Infof("(etcd-bootstrap) member is still present in the etcd cluster membership")
			return false, nil
		}
	}
```

Thanks @eranco74 for your hint 👍🏽  

So we have added this condition to `IsBootstrapComplete`, to avoid reconciling `clustermemberremovalcontroller` whenever there is `etcd-bootstrap`. 

I assume we can skip this check for SNO strategy, and this should do, or ? 🤔

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
Elbehery: it broke the SNO :/ 

Fixing ....

Elbehery:
@eranco74  @romfreiman I have fixed the failure, we skip listing etcd membership for SNO 

Would you kindly have a quick look, maybe I am missing sth ?

eranco74:
Last run didn't run the test
/test e2e-aws-ovn-single-node

pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -196,5 +196,27 @@ func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPo
...
 		}
 	}
+
+	// skip listing etcd membership for SingleReplicaTopologyMode
+	singleNode, err := IsSingleNodeTopology(infraLister)
+	if err != nil {
+		return false, fmt.Errorf("failed to get control plane topology: %w", err)
+	}
+	if singleNode {
+		return true, nil
```
Elbehery, line 206: @eranco74 @tsorya  I added this for SNO, wdyt ?

damdo:
It looks like the override command wasn't picked up by prow.
I've seen this happening sometimes. @Elbehery @tjungblu should we retry issuing it?

PR #1035, opened 2023-03-24
BIP: use hostname in bootstrap pod name

vrutkovs:
This would make sure bootstrap pod name in bootstrap-in-place mode is not generic "etcd-bootstrap" but "etcd-<hostname>". This makes etcd-operator reuse many certs we have already generated on bootstrap mode 

etcd pod is still being restarted, as we need to mount different directories and set env vars, but it cuts downtime to ~1 minute


lance5890:
i'm a little curious about this, u mean when we set BootstrapInPlaceStrategy , the bootkube will not teardown the bootstrap etcd， and will change this bootstrap node to a master node ?

vrutkovs:
>the bootkube will not teardown the bootstrap etcd

it will still tear it down 'cause different mount points / env vars are set in the final etcd container.
This PR would later allow us to prepare etcd certs beforehand and embed them in the image to avoid autogenerating those on the fly. 

> will change this bootstrap node to a master node ?

yes, correct, see [this enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/installer/single-node-installation-bootstrap-in-place.md)

lance5890:
> 

when we install ocp in UPI， if we only have 3 nodes(1 bootstrap + 2 master), can we use the BootstrapInPlaceStrategy? so when the other two masters are ready, and then the bootstrap node teardown , but this can change the bootstrap to a master, it finally will be 3 masters

Is this possible???

vrutkovs:
BootstrapInPlace is for single node clusters only

geliu2016 (approved):
/label cherry-pick-approved

romfreiman:
> i'm a little curious about this, u mean when we set BootstrapInPlaceStrategy , the bootkube will not teardown the bootstrap etcd， and will change this bootstrap node to a master node ?

@lance5890 you should not use UPI. If you use the Assisted Installer SaaS, it already has the bootstrap in place as a built in mechanism (no need for a bootstrap node).

vrutkovs:
/hold

e2e-metal-single-node-live-iso is failing

vrutkovs:
/hold cancel
/retest

All failures seem to be flakes

Elbehery:
gr8 work prof @vrutkovs

PR #1037, opened 2023-03-28
ETCD-410: Add Controller health to CEO liveness probe

tjungblu:


pkg/operator/health/checking_sync_wrapper.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/openshift/library-go/pkg/controller/factory"
+)
+
+// CheckingSyncWrapper wraps calls to the factory.SyncFunc in order to track when it last ran successfully.
+type CheckingSyncWrapper struct {
+	syncFunc          factory.SyncFunc
+	livenessThreshold time.Duration
+
+	// lastSuccessfulRun is updated with atomics, as updates can come from different goroutines
+	lastSuccessfulRun int64
```
hasbro17, line 17: I'm probably missing something but I'm a little unclear on the need for atomics here.

Since each controller is constructing its own `CheckingSyncWrapper` struct that means the `Sync()` for each controller's goroutine would only update its own `checkingSyncWrapper.lastSuccessfulRun`.
So the controller goroutines aren't really sharing this state or variable right?

hasbro17: Oh you probably meant the parallelism from the workers of each controller.
https://github.com/openshift/cluster-etcd-operator/blob/6ecd6d294d9b1bc713e674611e64fdea48202c44/pkg/operator/starter.go#L398-L416

Seems like we're not using any parallelism in any controller (nor would some of them tolerate that I imagine) but if that is ever the case in the future then the use of atomics makes more sense.

pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
```diff
@@ -22,15 +24,21 @@ type EtcdMembersController struct {
...
 	etcdClient etcdcli.EtcdClient,
 	eventRecorder events.Recorder,
 ) factory.Controller {
 	c := &EtcdMembersController{
 		operatorClient: operatorClient,
 		etcdClient:     etcdClient,
 	}
-	return factory.New().ResyncEvery(time.Minute).WithSync(c.sync).ToController("EtcdMembersController", eventRecorder.WithComponentSuffix("member-observer-controller"))
+
+	syncer := health.NewCheckingSyncWrapper(c.sync, 5*time.Minute)
```
hasbro17, line 38: Nit: Seems like every controller is using 5 mins as the default threshold. Thoughts on passing this as an arg to all controllers in starter.go? That way we don't have to change it everywhere if we need to tweak this again.

pkg/operator/health/checking_sync_wrapper.go
tjungblu: yes and no, the value is read from the http handler called from the health probe. So that's where your concurrent access comes from. That's just to prevent stale cached reads.

pkg/operator/etcdmemberscontroller/etcdmemberscontroller.go
tjungblu: I wanted to have this next to the "ResyncEvery" code, but I can add a default value. The only controller that doesn't have the 5m set is the defrag controller.

hasbro17:
/lgtm
/retest-required

Infra setup issues on latest run.

PR #1056, opened 2023-05-20
add etcd leader change metrics

lance5890:
1. I am so glad to see the etcd metrics doc in ocp 4.12, these metrics are very helpful to etcd stability
2. From my personal experience about etcd, maybe we could add the etcd_leader_change metric to the docs, this metric is the most intuitive metric to observe the stability of etcd

### DIFF
--- a/docs/performance-metrics.md
+++ b/docs/performance-metrics.md
@@ -4,6 +4,16 @@ This document is intended to help users and support engineers to troubleshoot pe
 
 ## Common Bottlenecks
 
+### Etcd Leader change
+
+The stability of the etcd cluster is crucial to the ocp cluster. When etcd lost its leader, the cluster will be temporarily unavailable, The cluster may responds slowly, and client requests may even fail
+
+Etcd leader change metrics:
+
+> etcd_server_leader_changes_seen_total{job="etcd"}
+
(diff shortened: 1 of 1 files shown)


docs/performance-metrics.md
```diff
@@ -4,6 +4,16 @@ This document is intended to help users and support engineers to troubleshoot pe
...
 
+### Etcd Leader change
+
+The stability of the etcd cluster is crucial to the ocp cluster. When etcd lost its leader, the cluster will be temporarily unavailable, The cluster may responds slowly, and client requests may even fail
+
+Etcd leader change metrics:
+
+> etcd_server_leader_changes_seen_total{job="etcd"}
+
+When the value of this etcd leader change metrics is relatively large and continues to grow(usually this metric will not be greater than 10), we would run the following metrics to check the bottlenecks
```
tjungblu, line 15: why not use the changes function instead? Checkout our alerts that use it:
https://github.com/openshift/cluster-etcd-operator/blob/master/jsonnet/custom.libsonnet#L58-L60

lance5890: > why not use the changes function instead? Checkout our alerts that use it: https://github.com/openshift/cluster-etcd-operator/blob/master/jsonnet/custom.libsonnet#L58-L60

I think both are ok, this will give users an intuitive feeling about the etcd stability

lance5890: i got what u think , will close this PR

PR #1062, opened 2023-06-21
Integrate feature flags in CEO

tjungblu:
requirement for logic in [ETCD-436](https://issues.redhat.com/browse/ETCD-436)

### DIFF
--- a/hack/featuregate_backup.yaml
+++ b/hack/featuregate_backup.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: config.openshift.io/v1
+kind: FeatureGate
+metadata:
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/create-only: "true"
+  name: cluster
+spec:
+  customNoUpgrade:
+    enabled:
+      - AutomatedEtcdBackup
+  featureSet: CustomNoUpgrade
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"os"
 	"regexp"
 	"time"
@@ -18,6 +19,7 @@ import (
 	operatorv1informers "github.com/openshift/client-go/operator/informers/externalversions"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/health"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/staticpod"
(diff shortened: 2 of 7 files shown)


pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -50,6 +51,7 @@ const (
 	recentBackupPath         = "/etc/kubernetes/cluster-backup"
 	failedPodBackoffDuration = 30 * time.Second
 	backupDirEnvName         = "CLUSTER_BACKUP_PATH"
+	featureFlagBackupAPI     = "EtcdBackupAPI"
```
tjungblu, line 54: just fyi, I likely split this into another controller - that was just as an example for testing it

pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go
```diff
@@ -87,6 +92,7 @@ func NewUpgradeBackupController(
 		clusterVersionLister:  clusterVersionInformer.Lister(),
 		targetImagePullSpec:   targetImagePullSpec,
 		operatorImagePullSpec: operatorImagePullSpec,
+		featureGateAccessor:   featureGateAccessor,
```
hasbro17, line 95: Looks good, I'm guessing this will give us access to reading the EtcdBackupAPI feature gate e.g:
```Go
featureGates, err := featureGateAccessor.CurrentFeatureGates()
if err != nil {
	return nil, err
}
etcdBackupAPIEnabled := featureGates.Enabled(configv1.FeatureGateEtcdBackupAPI)
```

hasbro17:
/lgtm
/retest-required

We can use the featuregate accessor in the new controller(s) once we have those in place and the API actually merged.

hasbro17:
/hold

Actually need to double check if we don't have anything extraneous here for testing purposes.

hack/featuregate_backup.yaml
```diff
@@ -0,0 +1,14 @@
...
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/create-only: "true"
+  name: cluster
+spec:
+  customNoUpgrade:
+    enabled:
+      - EtcdBackupAPI
+  featureSet: CustomNoUpgrade
```
hasbro17, line 14: Was this just for testing or is this where we read the feature flag?

tjungblu: nope that's just for easier testing against cluster bot, to enable the feature so to say

tjungblu:
quick mental note to myself here, the version is not actually available during the bootstrap process. Thus we're not able to get the feature flags properly. 

Gotta do a bit more research on what is available when, otherwise we have to pull the feature gates into a dedicated controller that updates whenever the etcd clusteroperator CRD updates.

hack/featuregate_backup.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/create-only: "true"
+  name: cluster
+spec:
+  customNoUpgrade:
+    enabled:
+      - AutomatedEtcdBackup
```
hasbro17, line 13: Sorry for the nit but the feature gate name is `AutomatedEtcdBackupAPI`. 
```suggestion
      - AutomatedEtcdBackupAPI
```
https://github.com/openshift/api/blob/master/config/v1/feature_gates.go#L285

hasbro17: Nvm, this was fixed in https://github.com/openshift/api/pull/1518

PR #1063, opened 2023-06-22
[OSD-16903] Limit etcdDatabaseHighFragmentationRatio to 100MB+

robotmaxtron:
Limiting alerting of higher percentages in small databases should improve how actionable the alert will be and is present in upstream: https://github.com/etcd-io/etcd/commit/1002346ceb37f1b097d52d72823a1ab7401bfd52

### DIFF
--- a/jsonnet/jsonnetfile.lock.json
+++ b/jsonnet/jsonnetfile.lock.json
@@ -8,8 +8,8 @@
           "subdir": "contrib/mixin"
         }
       },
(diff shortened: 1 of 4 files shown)


tjungblu:
Thanks @robotmaxtron , there is a script that does that automatically for you:

https://github.com/openshift/cluster-etcd-operator/blob/master/docs/HACKING.md#update-alerts

TLDR: run  `hack/generate.sh`

PR #1066, opened 2023-06-28
ETCD-436: backup controller

tjungblu:


pkg/operator/backupcontroller/backupcontroller.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	bp.Status.Conditions = append(bp.Status.Conditions,
+		v1.Condition{
+			Type:               "BackupCompleted",
+			Reason:             "BackupCompleted",
+			Message:            fmt.Sprintf("%s", jobState),
+			Status:             v1.ConditionTrue,
+			LastTransitionTime: metav1.NewTime(time.Now()),
+		})
```
hasbro17, line 240: We do have `BackupCompleted` and `BackupFailed` as conditions for the EtcdBackup CR.
https://github.com/openshift/api/blob/master/operator/v1alpha1/types_etcdbackup.go#L81-L85

So perhaps we can map `batchv1.JobFailed` to `BackupFailed` on the EtcdBackup status.

tjungblu: perfect, let me use that.

pkg/operator/backupcontroller/backupcontroller.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	statusReason := operatorv1alpha1.BackupCompleted
+	if jobState == batchv1.JobFailed {
+		statusReason = operatorv1alpha1.BackupFailed
+	}
+
+	bp.Status.Conditions = append(bp.Status.Conditions,
+		v1.Condition{
+			Type:               "BackupCompleted",
```
hasbro17, line 256: Ah okay I get it. I initially only defined the condition types.
But we're only using `BackupCompleted` as the Type with the reasons `BackupFailed` `BackupSkipped`, and (a little confusingly) `BackupCompleted`.

Now that I see this in practice I should have done a better job of maybe defining the successful condition with Type as `BackupCompleted` with the reason as `BackupSucceeded`.

I can update the condition types later but WDYT about making `BackupSucceeded` or something similar as the reason for a successful completion?

pkg/operator/backupcontroller/backupcontroller.go
```diff
@@ -0,0 +1,364 @@
...
+func (c *BackupController) sync(ctx context.Context, _ factory.SyncContext) error {
+	if enabled, err := backuphelpers.AutoBackupFeatureGateEnabled(c.featureGateAccessor); !enabled {
+		if err != nil {
+			klog.V(4).Infof("BackupController error while checking feature flags: %v", err)
+		}
+		return nil
+	}
+
+	jobsClient := c.kubeClient.BatchV1().Jobs(operatorclient.TargetNamespace)
+	currentJobs, err := jobsClient.List(ctx, v1.ListOptions{LabelSelector: "app=" + backupLabel + ",state!=processed"})
```
hasbro17, line 78: Interesting. So we have this label marker `state=processed` to filter out jobs already reconciled. I guess the conditions tell us that as well but I think we can't filter on those via ListOptions. So this can stay for now until we have something better in mind.

pkg/operator/backupcontroller/backupcontroller.go
```diff
@@ -0,0 +1,364 @@
...
+
+	if len(backupsToRun) == 0 {
+		klog.V(4).Infof("BackupController no backups to reconcile, skipping")
+		return nil
+	}
+
+	// in case of multiple backups requested, we're trying to reconcile in order of their names (also to reduce flakiness in tests)
+	sort.Slice(backupsToRun, func(i, j int) bool {
+		return strings.Compare(backupsToRun[i].Name, backupsToRun[j].Name) < 0
+	})
```
hasbro17, line 135: For the CronJob we'll hopefully disallow this by setting a [forbid concurrency](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#concurrency-policy) policy but yes the user could still create multiple EtcdBackup requests at once or in quick succession.

hasbro17:
@tjungblu Hopefully once my client-go bump PR lands we can pull that in. Although do you think we can split out the periodic controller stub for `config.openshift.io/v1alpha1 Backup` into its own PR? That might make it easier to review and merge just the one time `EtcdBackup` controller.

pkg/operator/backupcontroller/backupcontroller.go
tjungblu: neither the cronjob nor the job doesn't really give you a guarantee that there's only ever a single pod running. The backup logic needs to be idempotent anyway.

Would you rather use Forbid or Replace as a strategy?

pkg/operator/backupcontroller/backupcontroller.go
tjungblu: oh shoot, I got confused because the type was called "...Reason":
https://github.com/openshift/api/pull/1482/files#diff-07061d3f7cd529f689aeef37c7e2cc30938d36ee18a91fdbb8f456585bb582a8R82

I personally would stay with the definition of the batchv1/job:

> if (c.Type == batchv1.JobComplete || c.Type == batchv1.JobFailed)

BackupCompleted means success, BackupFailed is failure. 
Do we need to switch the reason with the type or do you prefer to have the reason be the same as the type?

tjungblu:
@hasbro17 moved the periodic controller out, rebased over newest master branch

pkg/operator/backupcontroller/backupcontroller.go
tjungblu: we can continue this convo in the other PR, but I'll take Forbid for the time being.

pkg/operator/starter.go
```diff
@@ -449,6 +458,14 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		os.Getenv("OPERATOR_IMAGE"),
 	)
 
+	backupController := backupcontroller.NewBackupController(
+		AlivenessChecker,
+		operatorConfigClientv1Alpha1,
+		kubeClient,
+		controllerContext.EventRecorder,
+		os.Getenv("IMAGE"),
```
tjungblu, line 466: that's not the right image, it should be the operator image

pkg/operator/backupcontroller/backupcontroller.go
hasbro17: Yeah no I agree this needs to be idempotent. I just commenting on how we'd use the forbid strategy at the cronjob level on top of disallowing parallel runs at this controller's level (since there's no guarantee the user won't create multiple requests). So this is good as is.

pkg/operator/backupcontroller/backupcontroller.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	if !injected {
+		return fmt.Errorf("could not inject PVC into Job template, please check the included cluster-backup-job.yaml")
+	}
+
+	klog.Infof("BackupController starts with backup [%s] as job [%s], writing to filename [%s]", backup.Name, job.Name, backupFileName)
+	_, err = jobClient.Create(ctx, job, v1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("BackupController could create job: %w", err)
```
hasbro17, line 332:
```suggestion
		return fmt.Errorf("BackupController failed to create job: %w", err)
```

pkg/operator/backupcontroller/backupcontroller.go
```diff
@@ -0,0 +1,364 @@
...
+				Message:            fmt.Sprintf("%s", job.Name),
+				Status:             v1.ConditionTrue,
+				LastTransitionTime: metav1.NewTime(time.Now()),
+			},
+			{
+				Type:               "BackupFilename",
+				Reason:             string(operatorv1alpha1.BackupPending),
+				Message:            backupFileName,
+				Status:             v1.ConditionTrue,
+				LastTransitionTime: metav1.NewTime(time.Now()),
```
hasbro17, line 349: As mentioned above if we go with type==reason for simplicity, then this can just be one condition with `Type: BackupPending` and `Reason: BackupPending` as well.
Otherwise `BackupStatus` type could logically have multiple reasons, e.g pending, failed, completed etc.

And the filename isn't really a condition or phase of the backup right? It's more just info about where we intend to save the backup (not guaranteed to even succeed yet).
WDYT about just including that in the message for the pending condition e.g:
```suggestion
				Type:               operatorv1alpha1.BackupPending,
				Reason:           string(operatorv1alpha1.BackupPending),
				Message:        fmt.Sprintf("Executing job %s to save backup file %s", job.Name, backupFileName),
				Status:             v1.ConditionTrue,
				LastTransitionTime: metav1.NewTime(time.Now()),
			},
```

pkg/operator/backupcontroller/backupcontroller.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	bp.Status.Conditions = append(bp.Status.Conditions,
+		v1.Condition{
+			Type:               "BackupCompleted",
+			Reason:             string(statusReason),
+			Message:            fmt.Sprintf("%s", jobState),
+			Status:             v1.ConditionTrue,
+			LastTransitionTime: metav1.NewTime(time.Now()),
+		})
```
hasbro17, line 261: Also at this point we would also need to set the `BackupPending` condition as false right?
We first set it as true when we create the job and update the EtcdBackup status.
https://github.com/openshift/cluster-etcd-operator/pull/1066/files#diff-a56fd06a4d0f37ea33be039d1e2afe545083c1b773453e61ee4a75c2eb35ea22R335-R361

pkg/operator/starter.go
```diff
@@ -449,6 +458,14 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		os.Getenv("OPERATOR_IMAGE"),
 	)
 
+	backupController := backupcontroller.NewBackupController(
+		AlivenessChecker,
+		operatorConfigClientv1Alpha1,
+		kubeClient,
+		controllerContext.EventRecorder,
+		os.Getenv("OPERATOR_IMAGE"),
```
hasbro17, line 466: Nit:
So we're mapping the `OPERATOR_IMAGE` to `targetImagePullSpec` arg.
Looking around in other controllers it seems we use `OPERATOR_IMAGE => operatorImagePullSpec` and `IMAGE => targetImagePullSpec`(for the etcd image).

Perhaps we can change the arg name from `targetImagePullSpec` to `operatorImagePullSpec` so it's consistent with how other controllers use that arg?

hasbro17: Also just curious, for the backup pod itself, does it matter which image we use?
Seems like the upgradebackupcontroller's backup pod used the etcd image to run the actual backup script (and the operator image to verify storage in the init container):

https://github.com/openshift/cluster-etcd-operator/blob/a662347dba24b5c10462bf2273ee7959c2d24f0f/pkg/operator/upgradebackupcontroller/upgradebackupcontroller.go#L291-L295


https://github.com/openshift/cluster-etcd-operator/blob/a662347dba24b5c10462bf2273ee7959c2d24f0f/pkg/operator/starter.go#L438-L449

And we're using the operator image for that instead. Seems like we download etcdctl in the backup script so it doesn't matter either way.

pkg/operator/backupcontroller/backupcontroller.go
tjungblu: applied

pkg/operator/backupcontroller/backupcontroller.go
tjungblu: applied. The reason to have the filename as its own message is to have an easier time for retention later

pkg/operator/backupcontroller/backupcontroller.go
tjungblu: done

pkg/operator/starter.go
tjungblu: Thanks for raising this, that's what caused me to change the image yesterday. 
Renamed all references in that controller to operatorImagePullSpec, I leave a note on the other PR too.

> Also just curious, for the backup pod itself, does it matter which image we use?

it shouldn't, all tools are either mounted or downloaded. For some reason it was crashing with the operand image yesterday, let me try this again. Will post after the cluster is built.

tjungblu: `oc` is missing:

```
        message: |
          + /usr/local/bin/cluster-backup.sh /etc/kubernetes/cluster-backup/backup-testbackup-2023-07-19_095939
          /usr/local/bin/cluster-backup.sh: line 45: oc: command not found
          Could not find the status of the kube-apiserver. Check if the API server is running. Pass the --force flag to skip checks.

```

tjungblu: maybe the mount path has changed in the meantime?
it's in `/usr/bin` now

tjungblu: it was using --force, which skips the use of OC. I'll mimic the upgrade controller in terms of their images though, so we're at least consistent.

PR #1067, opened 2023-07-05
OCPBUGS-15327: reset snapshot default counts to avoid file already lo…

tjungblu:
…cked warnings


geliu2016 (approved):
/label cherry-pick-approved

PR #1070, opened 2023-07-10
MON-669: Add etcd ServiceMonitor, before that was managed by cluster-…

machine424:


bindata/etcd/minimal-sm.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: etcd
+  namespace: openshift-etcd-operator
```
machine424, line 5: don't know if it's ok to deploy resources on operator namespace
(still think this makes more sense on `openshift-etcd` ns, but your call)

tjungblu: absolutely fine

bindata/etcd/minimal-sm.yaml
```diff
@@ -0,0 +1,38 @@
...
+        keySecret:
+          name: etcd-metric-client
+          key: tls.key
+  jobLabel: k8s-app
+  namespaceSelector:
+    matchNames:
+    - openshift-etcd
+  selector:
+    matchLabels:
+      k8s-app: etcd
```
tjungblu, line 38: super nitpick: newline

tjungblu: same on sm.yaml below

hasbro17:
Seems to be passing other than the plaguing disruption tests flakes.

LGTM whenever the nits from @tjungblu are addressed.

bindata/etcd/minimal-sm.yaml
machine424: Done

PR #1072, opened 2023-07-11
add rev bumps and etcdutl

tjungblu:
This PR will:
* use etcdutl when available, to ease the eventual transition towards 3.6 
* add a rev bump by 1 billion and mark compacted to test the feature in our CI restore job, flagged off by default


tjungblu:
/hold

still figuring out what makes most sense

PR #1074, opened 2023-07-20
ETCD-446: add periodic backup controller

tjungblu:


pkg/operator/periodicbackupcontroller/periodicbackupcontroller.go
```diff
@@ -0,0 +1,198 @@
...
+		kubeClient:            kubeClient,
+		operatorImagePullSpec: operatorImagePullSpec,
+		featureGateAccessor:   accessor,
+	}
+
+	syncer := health.NewDefaultCheckingSyncWrapper(c.sync)
+	livenessChecker.Add("PeriodicBackupController", syncer)
+
+	return factory.New().ResyncEvery(1*time.Minute).
+		WithSync(syncer.Sync).ToController("PeriodicBackupController", eventRecorder.WithComponentSuffix("periodic-backup-controller"))
```
hasbro17, line 59: I'm thinking we should probably have this controller get triggered via `.WithInformers()` on the following objects:
- Backups CR (to reconcile on updates to the config)
- CronJobs (to update on CronJob changes)

Otherwise it would be slow to react to spec changes.

I think we forgot this for the other backup controller but that should also probably be watching for its CR, and Jobs
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/backupcontroller/backupcontroller.go#L68-L69

PR #1078, opened 2023-07-25
add informers to new backup controllers

tjungblu:
/assign @hasbro17


tjungblu:
/hold

seems the caches are not synced and the new controllers just get locked

tjungblu:
/hold cancel

starting the informers was missing

pkg/operator/backupcontroller/backupcontroller.go
```diff
@@ -65,8 +67,11 @@ func NewBackupController(livenessChecker *health.MultiAlivenessChecker,
 	syncer := health.NewDefaultCheckingSyncWrapper(c.sync)
 	livenessChecker.Add("BackupController", syncer)
 
-	return factory.New().ResyncEvery(1*time.Minute).
-		WithSync(syncer.Sync).ToController("BackupController", eventRecorder.WithComponentSuffix("backup-controller"))
+	return factory.New().
+		ResyncEvery(1*time.Minute).
+		WithInformers(backupInformer.Informer()).
```
hasbro17, line 72: This controller also needs to watch Jobs so that it can reconcile when they fail or succeed right?
```suggestion
		WithInformers(backupInformer.Informer(),
		kubeInformers.InformersFor(operatorclient.TargetNamespace).Batch().V1().Jobs().Informer()
		).
```
where kubeInformers is:
https://github.com/openshift/cluster-etcd-operator/blob/e3d686d9e7125514a9045ecdb92263a50ed961ba/pkg/operator/starter.go#L136

tjungblu: thanks and done, I'll do a quick test run still

tjungblu: glad I did, needed to start its informer separately as well. All fixed now.

tjungblu: hmm no, CI still complains about it restarting...

tjungblu: ah dang, that makes sense:

```
W0726 12:54:27.942842       1 reflector.go:533] k8s.io/client-go@v0.27.4/tools/cache/reflector.go:231: failed to list *v1alpha1.EtcdBackup: the server could not find the requested resource (get etcdbackups.operator.openshift.io)
E0726 12:54:27.942864       1 reflector.go:148] k8s.io/client-go@v0.27.4/tools/cache/reflector.go:231: Failed to watch *v1alpha1.EtcdBackup: failed to list *v1alpha1.EtcdBackup: the server could not find the requested resource (get etcdbackups.operator.openshift.io)
W0726 12:54:27.942894       1 reflector.go:533] k8s.io/client-go@v0.27.4/tools/cache/reflector.go:231: failed to list *v1alpha1.Backup: the server could not find the requested resource (get backups.config.openshift.io)
E0726 12:54:27.942902       1 reflector.go:148] k8s.io/client-go@v0.27.4/tools/cache/reflector.go:231: Failed to watch *v1alpha1.Backup: failed to list *v1alpha1.Backup: the server could not find the requested resource (get backups.config.openshift.io)
```

especially when the cluster isn't running techpreview enabled

pkg/operator/starter.go
```diff
@@ -479,10 +484,16 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
 		os.Getenv("IMAGE"),
 		os.Getenv("OPERATOR_IMAGE"),
-		featureGateAccessor)
+		featureGateAccessor,
+		etcdBackupInformer,
+		jobsInformer)
 
+	go jobsInformer.Informer().Run(ctx.Done())
+	go etcdBackupInformer.Informer().Run(ctx.Done())
+	go configBackupInformer.Informer().Run(ctx.Done())
```
hasbro17, line 493: Perhaps we can conditionally run these with the feature gate so the informer doesn't error out if the CRDs don't exist?
See e.g:
https://github.com/openshift/insights-operator/blob/0338819de275e2854d68df955c3d7beceb7f5fa7/pkg/controller/operator.go#L131C5-L141

tjungblu: @hasbro17 hacked this by just faking the HasSynced functions. Works well :package:  

> Perhaps we can conditionally run these with the feature gate so the informer doesn't error out if the CRDs don't exist?

that's also an alternative way, but then we would need to restart the entire operator when the feature flags are enabled/disabled. Otherwise it becomes a bit messy to manage the informer routines. 

Happy to also code that approach for further discussion.

hasbro17: Haha that's pretty clever 😄 

My only slight concern in bypassing/disabling `HasSynced()` is whether there would be any unintended side effects. Does this handle runtime enablement of the feature flag (i.e CRD appearing) gracefully? As in, the cache is updated with the CR as a watch event. Or do we wait until it periodically resyncs every 1 min? If not, then this is fine.

> that's also an alternative way, but then we would need to restart the entire operator when the feature flags are enabled/disabled. Otherwise it becomes a bit messy to manage the informer routines.

Oh yeah I definitely don't want us handling or restarting the informer routines at runtime. I think restarting the operator after enabling the feature flags is probably not too much to ask so we can do a clean sync of the new APIs.
I actually expected that would happen automatically based on:
> // By default, when the enabled/disabled list  of featuregates changes, os.Exit is called.

which should cause the operator's main to terminate as well (or so I think from a [quick test](https://go.dev/play/p/xlsdNVzdzG_P)).

https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/featuregates/simple_featuregate_reader.go#L87
https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/featuregates/simple_featuregate_reader.go#L151-L155

Does our operator not restart automatically if the feature gates change?

hasbro17: We can keep the `HasSynced ` method if it turns out the operator isn't automatically restarting on feature gate changes. We can address that issue later. Just surprised at what other folks might be doing for their feature gates.

tjungblu: >  Does this handle runtime enablement of the feature flag (i.e CRD appearing) gracefully? As in, the cache is updated with the CR as a watch event. 

Yes it does, the informer routine basically runs in the background the whole time and will react normal once the API is available. The drawback is that it has error messages in the logs once in a while. 

> Or do we wait until it periodically resyncs every 1 min? 

Both. The controller will still re-sync based on the schedule and once the informer is finding the API it will trigger the sync as usual, too.

>  I think restarting the operator after enabling the feature flags is probably not too much to ask so we can do a clean sync of the new APIs.

That's fair, that will also fix it. Currently the exit handler is overruled by one that just prints the change:
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/starter.go#L166-L169

> Does our operator not restart automatically if the feature gates change?

Nope. The controller only runs when the feature flag is found to be enabled, so we can get away entirely without restarts. Maybe the restart solution would be much cleaner and we could gate both informers and controllers directly with the gate.

tjungblu: here's the alternative with restarts:
https://github.com/openshift/cluster-etcd-operator/pull/1082

looks much cleaner to use, WDYT?

hasbro17: Thanks for adding the alternate version! I think that looks much easier to understand and reason about. We can close this one in favor of that.

PR #1081, opened 2023-07-26
ETCD-448: add backup pruning cmd

tjungblu:


pkg/cmd/prune-backups/prune.go
```diff
@@ -0,0 +1,226 @@
...
+func (s backupDirStat) String() string {
+	return fmt.Sprintf("Name=[%s] SizeBytes=[%d] ModTime=[%s]", s.name, s.sizeBytes, s.modTime.String())
+}
+
+func (b backupDirStats) Len() int {
+	return len(b)
+}
+
+func (b backupDirStats) Less(i, j int) bool {
+	return b[j].modTime.Before(b[i].modTime)
```
tjungblu, line 219: leaving this for discussion: WDYT about also sorting by the name? in case some filesystem doesn't have proper mod times

pkg/cmd/prune-backups/prune.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	opts.AddFlags(cmd)
+	return cmd
+}
+
+func (r *pruneOpts) AddFlags(cmd *cobra.Command) {
+	flagSet := cmd.Flags()
+	flagSet.StringVar(&r.retentionType, "type", r.retentionType, "Which kind of retention to execute, can either be None, RetentionNumber or RetentionSize.")
+	flagSet.IntVar(&r.maxNumberOfBackups, "maxNumberOfBackups", 1, "how many backups to keep when type=RetentionNumber")
```
hasbro17, line 60: Not sure if we need to set the default here or in the periodic controller template or both:
https://github.com/openshift/cluster-etcd-operator/blob/381ffb81706699cdadd0735a52f9d20379505ef7/pkg/operator/periodicbackupcontroller/periodicbackupcontroller.go#L145

But per the API the default should be `maxNumberOfBackups: 15`
https://github.com/openshift/api/blob/master/config/v1alpha1/types_backup.go#L115

pkg/cmd/prune-backups/prune.go
hasbro17: Not sure if there are filesystems that won't set mtime. But our backup files names are timestamped right?
https://github.com/openshift/cluster-etcd-operator/blob/381ffb81706699cdadd0735a52f9d20379505ef7/pkg/operator/backupcontroller/backupcontroller.go#L313

So that should give us the ordering on the creation time as well. So we could do that as well if we don't find the modtime set.

pkg/cmd/prune-backups/prune.go
tjungblu: I think we add this to the periodic controller to keep it close to the schema. Thanks for reminding me of the default value again, I was just browsing through the min/max properties

hasbro17:
/hold

/lgtm

The cmd itself looks good. Only thing missing is plugging it into the CronJob template as an init container and setting the args from the periodicbackup controller.
Similar to what I have in https://github.com/openshift/cluster-etcd-operator/pull/1084
But feel free to unhold if you want to do that as a follow up.

tjungblu:
> But feel free to unhold if you want to do that as a follow up.

yep, that was the intention.

/hold cancel

tjungblu:

unrelated tests

tjungblu:

/override ci/prow/e2e-agnostic-ovn-upgrade 

unnecessary retest :/

PR #1085, opened 2023-08-01
OCPBUGS-14673: add timeout to member list in bootstrap

tjungblu:


tjungblu:
/jira refresh 

wiping the bug for now, in case it's not actually fixing the bug

tjungblu:
doesn't fix, but turns out we have the timeouts on the lower level already:
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/etcdcli/etcdcli.go#L270-L272

closing

PR #1086, opened 2023-08-01
ETCD-458: Add etcd pod tuning configuration from api

dusk125:
This PR refactors the hardcoded tuning parameters for Heartbeat interval and leader election timeout and instead sets them via a mapping from the [tuning profile set in the api server](https://github.com/openshift/api/pull/1538)


dusk125:
[openshift/api/1538](https://github.com/openshift/api/pull/1538) needs to merge first.
/hold

pkg/etcdenvvar/envvarcontroller.go
```diff
@@ -59,9 +60,11 @@ func NewEnvVarController(
 	infrastructureInformer configv1informers.InfrastructureInformer,
 	networkInformer configv1informers.NetworkInformer,
 	eventRecorder events.Recorder,
+	etcdSpec operatorv1.EtcdSpec,
```
tjungblu, line 63: that spec can change at runtime, don't you want to use the client/informer instead?

pkg/profilehelpers/profile.go
```diff
@@ -0,0 +1,17 @@
...
+import operatorv1 "github.com/openshift/api/operator/v1"
+
+func ProfileToEnvMap(profile operatorv1.Profile) map[string]string {
+	// Default values
+	envs := map[string]string{
+		"ETCD_HEARTBEAT_INTERVAL": "100",
+		"ETCD_ELECTION_TIMEOUT":   "1000",
+	}
+	switch profile {
+	case operatorv1.SlowerProfile:
```
tjungblu, line 12: add a default case to raise an error on unknown profiles, maybe?

pkg/profilehelpers/profile.go
```diff
@@ -0,0 +1,17 @@
+package profilehelpers
+
+import operatorv1 "github.com/openshift/api/operator/v1"
```
tjungblu, line 3: also wondering, don't you have to vendor the new API with go mod?

dusk125: I was just working with it with a go workspace until the api got merged. But yeah, I'll update the vendoring when that happens

pkg/cmd/render/env.go
```diff
@@ -68,41 +69,9 @@ func getEtcdName(e *envVarData) (map[string]string, error) {
...
-			timeout = "2500"
-		}
-	}
-
-	return map[string]string{
-		"ETCD_ELECTION_TIMEOUT": timeout,
-	}, nil
-
+func getTuningProfileValues(e *envVarData) (map[string]string, error) {
+	// TODO (alray) Can the profile be set before the bootstrap is created?
```
dusk125, line 73: // TODO Remove me ;)

pkg/etcdenvvar/envvarcontroller.go
tjungblu: I'd still use the informer instead (and wire it up in the controller factory), because that way you can have the controller sync with changes. When you do a GET with the client you're at the mercy of some other resource changing at the same time.

pkg/etcdenvvar/envvarcontroller.go
```diff
@@ -75,6 +83,7 @@ func NewEnvVarController(
 			networkInformer.Informer().HasSynced,
 			kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().Endpoints().Informer().HasSynced,
```
tjungblu, line 84: don't forget to add the informer to the cachesToSync above ^^

pkg/etcdenvvar/etcd_env.go
```diff
@@ -205,52 +206,12 @@ func getEtcdURLHost(envVarContext envVarContext) (map[string]string, error) {
 	return ret, nil
 }
 
-func getHeartbeatInterval(envVarContext envVarContext) (map[string]string, error) {
-	heartbeat := "100" // etcd default
-
-	infrastructure, err := envVarContext.infrastructureLister.Get("cluster")
+func getTuningProfileValues(envVarContext envVarContext) (map[string]string, error) {
+	etcd, err := envVarContext.etcdLister.Get("cluster")
```
tjungblu, line 210: thanks!

tjungblu:
logic lgtm, let's wait for the API to merge :)

dusk125:
I'll squash before push, I just wanted an easy way to back out the API update in case it wasn't correct.
/unhold

_output/golang-versions
```diff
@@ -0,0 +1,3 @@
+1.20
```
tjungblu, line 1: I think the _output folder should be on .gitignore :)

pkg/cmd/render/env.go
tjungblu: the infra CRD should be available here:

https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/cmd/render/render.go#L260-L264

so shouldn't be hard to directly set this to the right profile. It would safe a static pod rollout, I reckon...

pkg/etcdenvvar/etcd_env.go
```diff
@@ -205,52 +207,36 @@ func getEtcdURLHost(envVarContext envVarContext) (map[string]string, error) {
...
-	}
-
-	return map[string]string{
-		"ETCD_HEARTBEAT_INTERVAL": heartbeat,
-	}, nil
-}
-
-func getElectionTimeout(envVarContext envVarContext) (map[string]string, error) {
-	timeout := "1000" // etcd default
+	speed := etcd.Spec.HardwareSpeed
```
hasbro17, line 216: How does this work by default when the `TechPreviewNoUpgrade` feature gate is not enabled?
Does this field exist and is just not set? Or do we error on trying to access this field?

If it's the latter then we probably need to check whether we have the feature gate enabled before we attempt to access this field.

Something similar to the following except we check for whether `TechPreviewNoUpgrade` is enabled on the FeatureGate status :
https://github.com/openshift/cluster-etcd-operator/blob/6c205c247a3965132d6c73d098126996f9768a95/pkg/operator/backupcontroller/backupcontroller.go#L80
https://github.com/openshift/cluster-etcd-operator/blob/6c205c247a3965132d6c73d098126996f9768a95/pkg/backuphelpers/backuphelpers.go#L10-L17

dusk125: I've added a check in the most recent update... thanks for finding this!

PR #1087, opened 2023-08-02
OCPBUGS-17199: remove revision stability check from bootstrap complet…

tjungblu:
…eness

### DIFF
--- a/pkg/operator/ceohelpers/bootstrap.go
+++ b/pkg/operator/ceohelpers/bootstrap.go
@@ -182,19 +182,13 @@ func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPo
 		return false, nil
 	}
 
-	// now run check to stability of revisions
 	_, status, _, err := staticPodClient.GetStaticPodOperatorState()
 	if err != nil {
 		return false, fmt.Errorf("failed to get static pod operator state: %w", err)
 	}
+
 	if status.LatestAvailableRevision == 0 {
 		return false, nil
-	}
-	for _, curr := range status.NodeStatuses {
-		if curr.CurrentRevision != status.LatestAvailableRevision {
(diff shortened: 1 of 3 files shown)


pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -182,20 +182,14 @@ func IsBootstrapComplete(configMapClient corev1listers.ConfigMapLister, staticPo
...
 	}
-	for _, curr := range status.NodeStatuses {
-		if curr.CurrentRevision != status.LatestAvailableRevision {
-			klog.V(4).Infof("bootstrap considered incomplete because revision %d is still in progress", status.LatestAvailableRevision)
-			return false, nil
-		}
-	}
 
 	// check if etcd-bootstrap member is still present within the etcd cluster membership
 	membersList, err := etcdClient.MemberList(context.Background())
```
tjungblu, line 195: alternatively, maybe smarter is to swap the order of the checks.

If we observe the etcd-bootstrap node is not there anymore, bootstrap is completely done and we don't need to check revisions anymore.

tjungblu:
might still be the better option, I'm currently just figuring out why we check the revisions in the first place

geliu2016 (approved):
/label cherry-pick-approved

hasbro17:
Given that we have other checks that tell us when bootstrapping is complete i.e checking the `kube-system/bootstrap` configmap and querying for the `etcd-bootstrap` member, I think we can remove the check around whether a revision rollout is in progress since that's a common operation and we would have a lot of false positive for wherever we are using `IsBootstrapComplete()` (`clusterMemberRemovalController` and `EtcdEndpointsController`) it seems.

/lgtm

tjungblu:
Thanks @hasbro17 

/hold cancel

PR #1089, opened 2023-08-03
ETCD-460: add basic backup e2e tests

tjungblu:


test/e2e/backup_test.go
```diff
@@ -0,0 +1,241 @@
...
+				klog.Infof("error while getting job list: %v", err)
+				return false, nil
+			}
+
+			// the cronjob will keep at most 5 successful runs, we're waiting until the list has that many jobs completed
+			succeededJobs := 0
+			for _, j := range jobList.Items {
+				ownedByOurCron := false
+				for _, reference := range j.OwnerReferences {
+					if reference.Kind == "CronJob" && reference.Name == backupCrd.Name {
```
Elbehery, line 111: nit: can we replace the literal string "CronJob" with a const or similar ?

test/e2e/backup_test.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func ensureHostPathPVC(t *testing.T, pvcName string) {
+	kubeConfig, err := framework.NewClientConfigForTest("")
+	require.NoError(t, err)
+
+	c, err := corev1client.NewForConfig(kubeConfig)
+	require.NoError(t, err)
+
+	pv := v1.PersistentVolume{
```
Elbehery, line 174: I cant get why we create the PV manually, isn't PVC creates a PV automatically as long as there is `StorageClass` ?

tjungblu: dynamic provisioning doesn't work with hostPaths IIRC

Elbehery: really :O .. how the hostpath CSI driver works !! .. let me check

tjungblu: is there a CSI driver? I thought that's still hardcoded somewhere in kubelet

Elbehery: yes u r right, it does not 👍🏽

test/e2e/backup_test.go
Elbehery: thanks

test/e2e/backup_test.go
tjungblu: hahaha great, well it's a demo :)

tjungblu:
/hold

I'm going to split the commits into logic fixes and the tests once we have the full loop working

PR #1095, opened 2023-08-09
add backup skip e2e test

tjungblu:


test/e2e/backup_test.go
```diff
@@ -174,6 +174,48 @@ func TestRetentionBySize(t *testing.T) {
...
+		func(ctx context.Context) (done bool, err error) {
+			list, err := c.OperatorV1alpha1().EtcdBackups().List(context.Background(), metav1.ListOptions{})
+			if err != nil {
+				klog.Infof("error while listing backup: %v", err)
+				return false, nil
+			}
+
+			for _, b := range list.Items {
+				if backupHasCondition(&b, operatorv1alpha1.BackupSkipped, metav1.ConditionTrue) {
+					return true, nil
```
Elbehery, line 210: iiuc, there could be a case when there is no backup skipped, right ?

tjungblu: it's unlikely with 5, but not impossible. Let's retest a couple of times to see :-)

PR #1096, opened 2023-08-09
add pvc validation and e2e tests

tjungblu:
When we create a backup for a pvc that doesn't exist the pod will just be stuck in Pending state forever. This PR introduces validation for it and marks the backup as failed before spawning a pod.

### DIFF
--- a/pkg/operator/backupcontroller/backupcontroller.go
+++ b/pkg/operator/backupcontroller/backupcontroller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"github.com/openshift/cluster-etcd-operator/pkg/backuphelpers"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sort"
 	"strings"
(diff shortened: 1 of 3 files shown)


Elbehery:
nice work 👍🏽  

would it make sense to attempt to create the PVC by the controller in case it does not exist ?

tjungblu:
Thanks

> would it make sense to attempt to create the PVC by the controller in case it does not exist ?

if you tell me how many gigabytes and which storage class are needed for all our 150 different cloud and deployment configurations, please :)

pkg/operator/backupcontroller/backupcontroller.go
```diff
@@ -144,6 +145,16 @@ func (c *BackupController) sync(ctx context.Context, _ factory.SyncContext) erro
...
 
 	klog.V(4).Infof("BackupController backupsToRun: %v, chooses %v", backupsToRun, backupsToRun[0])
+	valid, err := validateBackup(ctx, backupsToRun[0], c.kubeClient, backupsClient)
+	if err != nil {
+		return err
+	}
+
+	if !valid {
+		klog.V(4).Infof("BackupController deems: %v invalid, skipping", backupsToRun[0])
+		return nil
```
Elbehery, line 155: so if the backup is invalid, we just log it, no err returned 

the next sync() it will be the same, or ?

tjungblu: yes, because the backup failure state is set on the CRD, no need to degrade the operator in this case

PR #1098, opened 2023-08-11
add e2e test for cronjob reconciliation failures

tjungblu:
that tests that the operator will degrade when a faulty cronjob was pushed to the API and recovers when removed.


tjungblu:
sorry folks, git seems to be really challenged with insert unit tests after one another LOL.

PR #1105, opened 2023-09-08
OCPBUGS-18357: prepend authfile to podman create

tjungblu:
the auth file seems used as a command argument to the container creation instead of a podman option.


geliu2016 (approved):
/label cherry-pick-approved

tjungblu:
Adding the "critical" label here so we can backport this timely for the customer in 4.12. That's blocking them from taking backups of their cluster.

PR #1118, opened 2023-09-20
ETCD-187: add highOverallControlPlaneIOWait alert rule

Elbehery:
resolves https://issues.redhat.com/browse/ETCD-187

cc @wking @dgrisonnet @williamcaban

### DIFF
--- a/jsonnet/custom.libsonnet
+++ b/jsonnet/custom.libsonnet
@@ -79,6 +79,21 @@
               severity: 'critical',
             },
           },
+          {
+            alert: 'highOverallControlPlaneIOWait',
+            expr: |||
+              (sum(irate(node_cpu_seconds_total {mode="iowait"} [2m])) without (cpu)) / count(node_cpu_seconds_total) without (cpu) * 100 > 20
+              AND on (instance) label_replace( kube_node_role{role="master"}, "instance", "$1", "node", "(.+)" )
+            |||,
+            'for': '5m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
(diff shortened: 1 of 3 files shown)


jsonnet/custom.libsonnet
```diff
@@ -79,6 +79,21 @@
               severity: 'critical',
             },
           },
+          {
+            alert: 'highOverallControlPlaneIOWait',
+            expr: |||
+              (sum(irate(node_cpu_seconds_total {mode="iowait"} [2m])) without (cpu)) / count(node_cpu_seconds_total) without (cpu) * 100 > 20
```
dgrisonnet, line 85: Hi @openshift/sre-alert-sme folks, we are trying to build an alert to detect disk/network pressure causing CPU latency. Based on your experience with OCP, would 20% of CPU time waiting for IO make sense to start alerting on?

Also what could be a meaningful threshold for a critical alert?

jsonnet/custom.libsonnet
```diff
@@ -79,6 +79,21 @@
...
+              AND on (instance) label_replace( kube_node_role{role="master"}, "instance", "$1", "node", "(.+)" )
+            |||,
+            'for': '5m',
```
dgrisonnet, line 88: for a warning level alert, we can wait for 1h before firing the alert

Elbehery: i fixed this on the PR against api-operator

Elbehery:
this PR is being replaced with https://github.com/openshift/cluster-kube-apiserver-operator/pull/1553 

/close

PR #1121, opened 2023-09-21
OCPBUGS-16794: installerpod: change pod manifest mode to 0600

ingvagabund:
Pulling https://github.com/openshift/library-go/pull/1574

/hold
until https://github.com/openshift/library-go/pull/1574 merges and the vendor is resynced


tjungblu:
The restore test fails because the fix is still missing from https://github.com/openshift/origin/pull/28223
Otherwise looks good to me.

ardaguclu:
as revendored version is in here https://github.com/openshift/cluster-etcd-operator/pull/1123, I'm closing this one in favor of the other.

/close

PR #1131, opened 2023-10-05
HOSTEDCP-1228: Add Hypershift Defrag Operator

muraee:


pkg/hypershift/defragcontroller/defragcontroller.go
```diff
@@ -0,0 +1,201 @@
...
+	return factory.New().ResyncEvery(compactionInterval+1*time.Minute).
+		WithSync(syncer.Sync).
+		ToController("HypershiftDefragController", eventRecorder.WithComponentSuffix("hypershift-defrag-controller"))
+}
+
+func (c *HypershiftDefragController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	return c.runDefrag(ctx, syncCtx.Recorder())
+}
+
+func (c *HypershiftDefragController) runDefrag(ctx context.Context, recorder events.Recorder) error {
```
tjungblu, line 53: no personal offense, but if you're already copying the whole controller - you might also copy it directly into your hypershift operator codebase. 

The PR at https://github.com/openshift/hypershift/pull/2869 looks pretty good already, doesn't it?

muraee: non taken :)
We are trying to avoid dragging and maintaining so many new dependencies into our code base as you can see is the case with https://github.com/openshift/hypershift/pull/2869

The reason for copying the controller is to avoid refactoring the existing controller to work with Hypershift and end up breaking standalone.

tjungblu: is there no other place you can build a dedicated binary for it, if dependencies are the concern? The etcdclient is not that bad to carry around... maybe GRPC is annoying to deal with.

@hasbro17 wdyt?

hasbro17: > We are trying to avoid dragging and maintaining so many new dependencies into our code base as you can see is the case with https://github.com/openshift/hypershift/pull/2869

@muraee If we're copying the controller (with hypershift specific modifications that we're not quite aware of) then we're still maintaining it in two places. So we're not really reducing the maintenance burden by placing it in the cluster-etcd-operator vs the hypershift repo, since you still have to maintain it here.

Plus by sharing the underlying dependency instead of copying that too, there's a good chance we might unintentionally break hypershift's defrag cmd with future changes.

I'm not opposed to sharing dependencies where it makes sense but if we do this, then will we have alternative controllers/cmds for all the other pieces that hypershift might require from the CEO? e.g cert rotation or automated backups.

tjungblu: btw, I've completely forgot about this. Ben wrote an upstream tool that also does that as a cmdline utility: https://github.com/ahrtr/etcd-defrag

there's plenty of config options to replicate our logic in CEO. That should also get you past the dependency issues.

muraee: @tjungblu thanks for sharing the cmdline, we will look if its feasible for us to use it.

muraee: @hasbro17 ideally we wouldn't have a seperate cmd for Hypershift, but the CEO would be Hypershift compatible and configurable enough, that we would just deploy it as it is.

But since that is not the case and we need to have the defragmentation feature before GA, the fasted solution was to create a separate cmd.

btw, we already have our own backup solution :)

enxebre:
>@muraee If we're copying the controller (with hypershift specific modifications that we're not quite aware of) then we're still maintaining it in two places. So we're not really reducing the maintenance burden by placing it in the cluster-etcd-operator vs the hypershift repo, since you still have to maintain it here.

Even with convenient code duplication, I believe it makes sense to have this code here as it's cohesive to etcd expertise and CVE maintenance.

>Plus by sharing the underlying dependency instead of copying that too, there's a good chance we might unintentionally break hypershift's defrag cmd with future changes.

We might want to consider having a gating hypershift e2e test job here to prevent that from happening if we ever get to that as we do in other multiple components.

>I'm not opposed to sharing dependencies where it makes sense but if we do this, then will we have alternative controllers/cmds for all the other pieces that hypershift might require from the CEO? e.g cert rotation or automated backups.

These are all good questions that we need to answer together. Regardless of where the code base lives I'd expect the Etcd SMEs (as well as for any other Openshift component) to progressively increase ownership of Etcd impact in the Openshift hcp form factor.

My preference would be we start driving a collaboration in that direction with this PR as we've done for other components like storage, networking, mco... Otherwise If the etcd team believes this PR is not best path forward to that end we'll proceed merging the code in the hypershift repo.

imain:
I should add I have tested this a lot and it is working great.

PR #1164, opened 2023-11-23
MON-3287: manage RBAC resources for prometheus to scrape openshift-etcd

machine424:
previously was managed by CMO see https://github.com/openshift/cluster-monitoring-operator/pull/2165

/hold

### DIFF
--- a/bindata/etcd/prometheus-role.yaml
+++ b/bindata/etcd/prometheus-role.yaml
@@ -1,0 +1,32 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: prometheus-k8s
+  namespace: openshift-etcd
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - services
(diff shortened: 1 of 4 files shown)


tjungblu:
Don't worry about the failures, that doesn't seem related. Thanks for cleaning this rbac up for us!

jan--f:
/lgtm
from the CMO perspective.

jan--f:
@tjungblu we can go ahead here and then merge the CMO PR. It'd be good to get this into the 4.15 payload.

tjungblu:
then let's do it!

/hold cancel

tjungblu:
@machine424 all good, I think this is broken on something else entirely. I can override when it comes back red.

machine424:
> @machine424 all good, I think this is broken on something else entirely. I can override when it comes back red.

I saw a prometheus test failing, even though it didn't seem related I wanted to be sure.
Ok, thanks.

PR #1170, opened 2023-11-28
OCPBUGS-17199: remove revision stability check from bootstrap completeness

tjungblu:
slightly more involved checking against actual node status now, requires further testing against assisted installer and some refactoring.

### DIFF
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
@@ -89,7 +89,7 @@ func NewClusterMemberRemovalController(
 	return factory.New().
 		WithSync(syncer.Sync).
 		WithSyncDegradedOnError(operatorClient).
-		ResyncEvery(33*time.Minute). // make it slow since nodes are updated every few minutes
+		ResyncEvery(5*time.Minute). // nodes are updated several times a minute, 5m for safety on getting stuck
 		WithBareInformers(networkInformer.Informer()).
 		WithInformers(
 			masterNodeInformer,
@@ -99,6 +99,13 @@ func NewClusterMemberRemovalController(
 }
 
 func (c *clusterMemberRemovalController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	// only attempt to scale down if the machine API is functional
+	if isFunctional, err := c.machineAPIChecker.IsFunctional(); err != nil {
+		return fmt.Errorf("failed to determine whether machine api is functional: %w", err)
+	} else if !isFunctional {
+		return nil
+	}
+
 	// skip reconciling this controller, if bootstrapping is not completed
 	bootstrapComplete, err := ceohelpers.IsBootstrapComplete(c.configMapLister, c.operatorClient, c.etcdClient)
 	if err != nil {
@@ -108,10 +115,13 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, syncCtx facto
 		return nil
 	}
 
-	// only attempt to scale down if the machine API is functional
-	if isFunctional, err := c.machineAPIChecker.IsFunctional(); err != nil {
-		return err
-	} else if !isFunctional {
+	// check for stability of revisions, by also taking the node status into account
+	// note that this is an important step for the AssistedInstallers installation procedure.
+	stable, err := c.isRevisionStable()
+	if err != nil {
+		return fmt.Errorf("isRevisionStable failed to determine revision stability: %w", err)
+	}
+	if !stable {
 		return nil
 	}
 
@@ -126,7 +136,45 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, syncCtx facto
 		errs = append(errs, err)
 	}
 	return kerrors.NewAggregate(errs)
+}
 
+func (c *clusterMemberRemovalController) isRevisionStable() (bool, error) {
+	_, status, _, err := c.operatorClient.GetStaticPodOperatorState()
+	if err != nil {
+		return false, fmt.Errorf("failed to get static pod operator state: %w", err)
+	}
+
+	masterNodes, err := c.masterNodeLister.List(c.masterNodeSelector)
+	if err != nil {
+		return false, fmt.Errorf("failed to list master nodes: %w", err)
+	}
+
+	for _, node := range masterNodes {
+		nodeReady := false
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == corev1.NodeReady && condition.Status == corev1.ConditionTrue {
+				nodeReady = true
+			}
+		}
+
+		// There are cases where the installer can be quicker to create a new revision than it can detect a node going away.
+		// This causes the revision check below to infinite loop, waiting forever for the installation on that node.
+		// Subsequently, all healthy other nodes are barred from installing the new revision.
+		// Normally, the machine-api will detect this and replace said node, here we do not want to block the member removal
+		// that can get us out of that situation automatically.
+		if !nodeReady {
+			return true, nil
+		}
+	}
+
+	for _, curr := range status.NodeStatuses {
+		if curr.CurrentRevision != status.LatestAvailableRevision {
+			klog.V(4).Infof("pending revision %d/%d is still in progress on node %s", curr.CurrentRevision, status.LatestAvailableRevision, curr.NodeName)
+			return false, nil
+		}
+	}
+
+	return true, nil
 }
 
 // attemptToScaleDown attempts to remove a voting member only once we have identified that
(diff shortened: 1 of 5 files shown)


tjungblu:
current state functionally works again, let's see if AI installs fine ;)

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -129,6 +145,46 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, syncCtx facto
...
+		// TODO that might be too generic, not sure if there's any other status we can tap into
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == corev1.NodeReady && condition.Status == corev1.ConditionTrue {
+				nodeReady = true
+			}
+		}
+
+		// There are cases where the installer can be quicker to create a new revision than it can detect a node going not ready.
+		// This causes the revision check below to become stale, meaning it will wait forever for the installation to finish on the node.
+		// Or even worse, it will wait for it to install on that node, before progression an installation on the other healthy nodes.
```
tjungblu, line 170: > progressing

tjungblu:
seems to also install fine by AI. 

retrying it once more for safety
/test e2e-metal-assisted

gotta spend the rest of the day writing tests.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -108,10 +115,13 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, syncCtx facto
 		return nil
 	}
 
-	// only attempt to scale down if the machine API is functional
-	if isFunctional, err := c.machineAPIChecker.IsFunctional(); err != nil {
-		return err
-	} else if !isFunctional {
+	// check for stability of revisions, by also taking the node status into account
+	// note that this is an important step for the AssistedInstallers installation procedure.
```
hasbro17, line 119: I understand how removing the revision check in `IsBootstrapComplete()` is fixing the original issue ([OCPBUGS-1799](https://issues.redhat.com/browse/OCPBUGS-17199)) pretty much as in the first PR https://github.com/openshift/cluster-etcd-operator/pull/1087

I'm still confused on how we broke the asssited-installer's bootstrap or install flow with that fix.
Which then caused [OCPBUGS-23522](https://issues.redhat.com/browse/OCPBUGS-23522) and subsequently how this check here in `isRevisionStable()` is fixing that issue (given that `e2e-metal-assisted` is passing).

Honestly surprised to see the scale-down controller being the point of contention here as that's not part of bootstrap right? Plus I would think the scale-down controller wouldn't be running at all if the machine API wasn't there (which I think is the case for assisted-installs)

This makes me think that we previously broke the assisted-install by inadvertently running the scale-down controller and this fix makes it so we don't?

But please clarify as I think I'm a bit off track here.

tjungblu: > Plus I would think the scale-down controller wouldn't be running at all if the machine API wasn't there (which I think is the case for assisted-installs)

there are machines, even though it says it's bare metal :) 

> This makes me think that we previously broke the assisted-install by inadvertently running the scale-down controller and this fix makes it so we don't?

We inadvertently have not updated the endpoints configmap properly at the right time and thus the pod ended up misconfigured, I think that's the main culprit. The code I've added to the scale-down controller is to actually fix the reported bug we wanted to fix initially. And to keep https://issues.redhat.com/browse/OCPBUGS-10351 working as it used to. 

--- 

checkout the must-gather in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/1165/pull-ci-openshift-cluster-etcd-operator-release-4.13-e2e-metal-assisted/1729146483854282752

If you look at the endpoints CM:

```
  data:
    339f128468fd4051: 192.168.127.12
    81556f099dc4d812: 192.168.127.10
  kind: ConfigMap
  metadata:
    annotations:
      alpha.installer.openshift.io/etcd-bootstrap: 192.168.127.11
    name: etcd-endpoints
    namespace: openshift-etcd
```

you see that the bootstrap still exists as the annotation (due to the `IsBootstrapComplete()` check on revisions, which this PR removes), yet there are only two members.

The bootstrap is actually gone and removed for good fairly early:

```
bootstrap_teardown_controller.go:140] cluster-bootstrap is not yet finished - ConfigMap 'kube-system/bootstrap' not found
bootstrap_teardown_controller.go:151] Removing bootstrap member [d8403780dd26e716]
bootstrap_teardown_controller.go:158] Successfully removed bootstrap member [d8403780dd26e716]
```

Now the issue with updating the configmap is a race condition between the env var controller and the installer controller: 

> installer_controller.go:512] "test-infra-cluster-803c5056-master-1" is in transition to 4, but has not made progress because static pod is pending

This pod is simply misconfigured and can't come up, if you look at `registry-build04-ci-openshift-org-ci-op-363njqnf-stable-sha256-2020af153d5445ec4ab69e1403338f0361624af0f651b15c2dff815fb29bddd7/namespaces/openshift-etcd/pods/etcd-test-infra-cluster-803c5056-master-1/etcd-test-infra-cluster-803c5056-master-1.yaml`, the master-1 doesn't actually exists in its env vars:

```   

 /bin/sh: line 3: NODE_test_infra_cluster_803c5056_master_1_ETCD_URL_HOST: not set

...

  - name: ALL_ETCD_ENDPOINTS 
    value: https://192.168.127.10:2379,https://192.168.127.11:2379,https://192.168.127.12:2379

...

  - name: NODE_test_infra_cluster_803c5056_master_0_ETCD_NAME
    value: test-infra-cluster-803c5056-master-0
  - name: NODE_test_infra_cluster_803c5056_master_0_ETCD_URL_HOST
    value: 192.168.127.10
  - name: NODE_test_infra_cluster_803c5056_master_0_IP
    value: 192.168.127.10
  - name: NODE_test_infra_cluster_803c5056_master_2_ETCD_NAME
    value: test-infra-cluster-803c5056-master-2
  - name: NODE_test_infra_cluster_803c5056_master_2_ETCD_URL_HOST
    value: 192.168.127.12
  - name: NODE_test_infra_cluster_803c5056_master_2_IP
    value: 192.168.127.12
...    
```

From where do the env vars get populated? 
ALL_ETCD_ENDPOINTS come from the endpoints configmap, populated by `etcdendpointscontroller` and you can see it still includes the bootstrap node, but the others come from the (static pods) NodeStatus:
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/etcdenvvar/etcd_env.go#L158-L179C3

which, considers that node unknown (or well, whatever you want to call rev 0) because it has a pending installation:
> "StaticPodsAvailable: 2 nodes are active; 1 nodes are at revision 0; 2 nodes are at revision 4

Why does this PR fix this race? We're updating the configmap with the right values at the right time. If we were to take the configmap values to render the IP env, or just iterate the master nodes instead, we would've fixed this issue too. I'm not entirely sure what other consequences it has however...

tjungblu: let's see if #1172 produces some meaningful output to see it in action and sequence.

hasbro17: Thanks a lot for the walkthrough (and trip down memory lane), don't think I would have pieced all that together :)

> The code I've added to the scale-down controller is to actually fix the reported bug we wanted to fix initially. And to keep https://issues.redhat.com/browse/OCPBUGS-10351 working as it used to.

This part makes sense. So basically don't scale-down while the revision is still rolling out so we don't regress to OCPBUGS-10351.

> ALL_ETCD_ENDPOINTS come from the endpoints configmap
> Why does this PR fix this race? We're updating the configmap with the right values at the right time.

But at this point in time while the `master-1` etcd pod hasn't even run and is still pending, it's correct for the `etcd-endpoints` configmap to not have `master-1`'s IP. It's not supposed to be a member at this point in time.
Not until that pod is first Running, (and awaiting membership addition).
https://github.com/openshift/cluster-etcd-operator/blob/99be19d84dcc0f256a804b1f058e7abc1e58a449/pkg/operator/clustermembercontroller/clustermembercontroller.go#L228-L244

So I'm a little confused on whether you meant that the endpoints configmap should have `master-1`'s IP even though it hasn't been added to the membership yet? The right values for the configmap should only be `master-0` and `master-2` IPs and that seems to be the case.

Also side note but I'm scratching my head because the node manifest for `master-1`  is showing `master-1`'s IP as `192.168.145.11`
From `registry-build04-ci-openshift-org-ci-op-363njqnf-stable-sha256-2020af153d5445ec4ab69e1403338f0361624af0f651b15c2dff815fb29bddd7/cluster-scoped-resources/core/nodes/test-infra-cluster-803c5056-master-1.yaml`
```YAML
status:
  addresses:
  - address: 192.168.127.11
    type: InternalIP
  - address: test-infra-cluster-803c5056-master-1
    type: Hostname
```

Is `master-1` is the bootstrap node and also stuck at revision 0???

Or as I thought the bootstrap node already gone (as the member was removed) and we're failing because the pod yaml for `master-1` is malformed without its envvars.

hasbro17: If it's just malformed then

>This pod is simply misconfigured and can't come up

```
/bin/sh: line 3: NODE_test_infra_cluster_803c5056_master_1_ETCD_URL_HOST: not set
```
Seems like we can't get past the init container `etcd-ensure-env-vars` since `master-1`'s pod's env vars are not getting populated correctly.
And the reason for that as you've mentioned is:

> the others come from the (static pods) NodeStatus... which, considers that node unknown (or well, whatever you want to call rev 0) because it has a pending installation

Zooming in on where we loop over the NodeStatus https://github.com/openshift/cluster-etcd-operator/blob/99be19d84dcc0f256a804b1f058e7abc1e58a449/pkg/etcdenvvar/etcd_env.go#L168

So master-1 isn't showing up in the NodeStatus because it's at rev 0? Going to try and dig for the pointer to that code but that's troubling. We just want the node name and IP to populate the following env vars on the etcd static pod:
```
NODE_NODE_ENVVAR_NAME_ETCD_URL_HOST
NODE_NODE_ENVVAR_NAME_ETCD_NAME
NODE_NODE_ENVVAR_NAME_IP
```
If the NodeStatus is correctly or incorrectly not showing us master-1 then now I'm wondering whether we really need to be looking there for the node name and IP as we do in `getEtcdURLHost()`.

PR #1171, opened 2023-11-30
NO-JIRA: bump library-go and use default UnhealthyPodEvictionPolicy

atiratree:
etcd should use a default UnhealthyPodEvictionPolicy behavior corresponding to the
IfHealthyBudget policy. This policy achieves the least amount of disruption, as it
does not allow eviction when multiple etcd pods do not report readiness.
This can block node drain/maintenance. The cluster administrator should then
analyze these pods and decide which one to bring down manually.

As opposed to for example to kube-apiserver that can use more permissive eviction policy: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1579


Elbehery:
cc @hasbro17 @dusk125

dusk125:
This test is being made optional https://github.com/openshift/release/pull/47252
/override ci/prow/e2e-gcp-qe-no-capabilities

ingvagabund:
/retitle NO-JIRA: bump library-go and use default UnhealthyPodEvictionPolicy

This PR bumps library-go only. The change under pkg/operator/starter.go reflects the new API which is no-op for this operator.

Issue #1181, opened 2024-01-16
Access to a privileged container allows for breakout to the underlying host


lance5890:
1. I have seen many ocp containers have set the ""privileged": true"
```
        "terminationMessagePolicy": "FallbackToLogsOnError",
        "imagePullPolicy": "IfNotPresent",
        "securityContext": {
          "privileged": true
        }
```

2. I just wonder why should the etcd containerd should set the  "privileged": true flag. compared to the kubeadm installation, the etcd container has not this flag

3. when using standard container runtimes (for example ContainerD or CRI-O) access to a privileged container allows for easy breakout to the underlying host, which in turn allows for access to all other workloads on that host and credentials for the node agent (Kubelet)

lance5890:
@Elbehery @tjungblu

tjungblu:
> I just wonder why should the etcd containerd should set the "privileged": true flag.

because we write the etcd data dir onto a hostPath, which requires this unfortunately.

lance5890:
> > I just wonder why should the etcd containerd should set the "privileged": true flag.
> 
> because we write the etcd data dir onto a hostPath, which requires this unfortunately.

1 but even without the 'privileged: true', the openshift-apiserver can also write audit to /var/lib/etcd
2 as the etcd container run as root by default

tjungblu:
I doubt that you can just mount /var/lib/etcd on the host without it. But feel free to try, you already know how to run PRs. 
Plenty of detail around this in https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

tjungblu:
Unless you're at RedHat and can authenticate via our SSO the automation won't be able to add you :) But I'm happy to give you the ok-to-test label to try it out.

PR #1183, opened 2024-01-17
Security: no need to set privileged: true option

lance5890:
no need to set privileged:true option

related to https://github.com/openshift/cluster-etcd-operator/issues/1181

### DIFF
--- a/bindata/etcd/pod.yaml
+++ b/bindata/etcd/pod.yaml
@@ -24,8 +24,6 @@ spec:
           #!/bin/sh
           echo -n "Fixing etcd log permissions."
           chmod 0700 /var/log/etcd && touch /var/log/etcd/etcd-health-probe.log && chmod 0600 /var/log/etcd/*
-      securityContext:
-        privileged: true
       resources:
         requests:
           memory: 50Mi
@@ -66,8 +64,6 @@ spec:
         requests:
           memory: 60Mi
           cpu: 10m
-      securityContext:
-        privileged: true
       env:
 ${COMPUTED_ENV_VARS}
       - name: NODE_IP
@@ -92,8 +88,6 @@ ${COMPUTED_ENV_VARS}
         requests:
           memory: 60Mi
           cpu: 10m
-      securityContext:
-        privileged: true
       volumeMounts:
         - mountPath: /etc/kubernetes/static-pod-resources
           name: resource-dir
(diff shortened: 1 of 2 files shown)


lance5890:
@tjungblu  I think the failed e2e case has nothing to do with with PR

lance5890:
prow/e2e-gcp-qe-no-capabilities This test is being made optional

https://github.com/openshift/release/pull/47252
https://github.com/openshift/release/pull/47890

lance5890:
I feel there is something to do with the selinux  container_log_t  of /var/lib/etcd dir , I will dig this up later

```
ls -lZ /var/log/

system_u:object_r:container_log_t:s0         35 Dec  4 07:48 etcd

```

lance5890:
> @lance5890 while working on certificates right now, I stumbled upon this annotation: https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/bootkube/manifests/00_openshift-etcd-ns.yaml#L11
> 
> Maybe that's causing it to effectively run privileged.

yes, the "openshift.io/run-level: "0" label will skip scc to the pods n ccos-etcd

and the default securityContext is enough to read/write in the /var/lib/etcd dir,

there is no need to set privileged:true , this may some permissions risky

I will continue to pay attention to this issue， any progress I will let u know

lance5890:
> @lance5890 while working on certificates right now, I stumbled upon this annotation: https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/bootkube/manifests/00_openshift-etcd-ns.yaml#L11
> 
> Maybe that's causing it to effectively run privileged.

I have a general idea about this issue, please correct me if I'm wrong

1. yes ,the  openshift.io/run-level: "0" label will skip the scc setting for the pod in ccos-etcd, but this will not cause it to effectively run privileged.
2. if we do not set etcd run as privileged, why should the etcd still can read/write for the /var/lib/etcd dir
- 2.1 this was because the cri-o will not set selinuxLabel for etcd container, such as https://github.com/cri-o/cri-o/blob/a4d194b3ac0ed51f29d4702052d36305dc3d04d6/server/container_create_linux.go#L299-L307

image

- 2.2 so if we just run etcd as HostNetwork, the etcd container will still run as spc_t, which means the etcd container runs as "Super Privileged Container", such as :
 ```
system_u:system_r:spc_t:s0      root      107866 15.9  2.0 11737432 650776 ?     S<sl Jan22 650:08 etcd --logger=zap
```
3. the difference between privileged and HostNetwork when set etcd container  is that the capacities differs, but both privileged and HostNetwork can make etcd container run as spc_t

/cc @haircommander

PR #1188, opened 2024-01-25
WIP: initial test for kubelet fix on SNO

eggfoobar:
This PR is an attempt to validate another approach for resolving OCPBUGS-26489

Validating assumption that ignoring the NODE_IP for SNO deployments will allow cluster reboot to continue as normal


eggfoobar:
This won't work due to kube-api making use of downward api for host ip. This needs to be resolved another way, closing.

PR #1192, opened 2024-02-01
[WIP]add_scrnario_OCP-71253

geliu2016:
@tjungblu I added qe test scenario to e2e test, could you please take a look? 
cc: @sandeepknd


Makefile
```diff
@@ -47,4 +52,13 @@ test-e2e: GO_TEST_FLAGS += -v
 test-e2e: GO_TEST_FLAGS += -timeout 2h
 test-e2e: GO_TEST_FLAGS += -p 1
 test-e2e: test-unit
+test-e2e:
+	go test \
```
tjungblu, line 56: Thanks @geliu2016 - dunno if you saw this already, but @Elbehery added a very similar test in test/e2e/backup_script_test.go already. I believe you could reuse almost everything in there to write a testcase for that bug.

test/e2e/etcd_scenario_test.go
```diff
@@ -0,0 +1,107 @@
...
+				fmt.Println("Remove the debug pod now.")
+				debugPodStr := strings.Split(fmt.Sprintf("delete pod %s -n default", debugPodName), " ")
+				_, err := exec.Command("oc", debugPodStr...).CombinedOutput()
+				Expect(err).NotTo(HaveOccurred())
+			}()
+
+			By("Wait for debug pod to be in Running phase.")
+			err = wait.PollUntilContextTimeout(context.Background(), time.Second, 30*time.Minute, true, func(ctx context.Context) (done bool, err error) {
+				pods, err := clientSet.CoreV1Interface.Pods(debugNamespace).List(ctx, metav1.ListOptions{})
+				if err != nil {
```
lance5890, line 52: if err != nil && !apierrors.IsNotFound(err) {
  return false, err
}

Makefile
geliu2016: Thank you, @tjungblu Yes, I saw there is similar test in backup_script_test.go already, and it checking the etcd backup feature with go function mode, and QE's test are focus on more complex scenarios, and QE using ginkgo mod to design test case, as I reviewed other sub component teams' e2e, they are also organize e2e test with go functions(Dev's prefer) and ginkgo(QE's prefer), it could co-exist, and I wish this backup case is a beginning, we will contribute QE's scenarios with this ginkgo framework.

test/e2e/etcd_scenario_test.go
geliu2016: Updated, thanks

go.mod
```diff
@@ -3,31 +3,34 @@ module github.com/openshift/cluster-etcd-operator
 go 1.20
 
 require (
-	github.com/davecgh/go-spew v1.1.1
+	github.com/cert-manager/cert-manager v1.14.0
```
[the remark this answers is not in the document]
geliu2016: aha, correct it.

Makefile
tjungblu: I prefer to have ginkgo tests in origin: https://github.com/openshift/origin/tree/master/test/extended/etcd

Otherwise this causes too much confusion and I don't want to maintain two test frameworks in here.

PR #1194, opened 2024-02-02
ETCD-512: refactoring the cert signer controller

tjungblu:
Second try, containing a fix for the client secret deletion described in:
https://issues.redhat.com/browse/TRT-1485?focusedId=24060763&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24060763

----
This PR will
* replace the existing cert rotation logic with more battle tested ones from library-go
* create new signer certificates (metrics + serving) in openshift-etcd namespace, in addition to existing ones in openshift-config
* create new server certificates (peer, serving, serving-metrics)
* create new client certificates (etcd-client, etcd-metrics)
* bundle existing signer certificates with newly created CAs (to stay backward compatible)

The consequence of merging this PR is:
* an additional static pod rollout during installation and upgrades (slightly longer install/upgrade time expected)
* all existing certs are rotated with existing old and new signers, which are distributed to all nodes for actual signer rotation later on

### DIFF
--- a/docs/etcd-tls-assets.md
+++ b/docs/etcd-tls-assets.md
@@ -772,3 +772,53 @@ Eventually, when `cluster-etcd-operator` is installed, running, and
 reporting the `EtcdRunningInCluster = True` condition. This is usually
 when the etcd cluster has scaled to 4, and it is safe to remove the
 bootstrap member.
+
+# The Big CA Cert Rotation
+
+With 4.16 we want to take an initial stab at the rotation of the signer certificates.
+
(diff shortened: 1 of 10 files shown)


tjungblu:
/payload 4.16 nightly blocking

doing another run for good measure, the last one looks green

tjungblu:
@dgoodwin / @simonpasquier I think we're good again, the payloads (if they were running) were green. 

@hasbro17 is going to review this over the course the day and hopefully get this merged again - so heads-up in case there are any issues

PR #1200, opened 2024-02-08
ETCD-535: Manual CA rotation should rotate all leaf certs

tjungblu:
/hold

The cool thing is that we're now able to "swap" signers with the existing logic with:

> $ oc get secret etcd-signer -n openshift-etcd -ojson | jq 'del(.metadata["namespace","creationTimestamp","resourceVersion","selfLink","uid"])' | oc apply -n openshift-config -f -

which is effectively overwriting the new signer from `openshift-etcd` into the old signer in `openshift-config`. That works, because the bundle with the new signer is already distributed to all CP nodes. CEO will then proceed to rewrite all leaf certs, which are then rolled out together via `etcd-all-certs`. 

Manual rotation is then just two step manual process:

Generate new signer: 
>  $ oc delete secret etcd-signer -n openshift-etcd

... wait for the rollout ...

Replace the old signer with the new signer:

> $ oc get secret etcd-signer -n openshift-etcd -ojson | jq 'del(.metadata["namespace","creationTimestamp","resourceVersion","selfLink","uid"])' | oc apply -n openshift-config -f -

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/openshift/api v0.0.0-20231218131639-7a5aa77cc72d
 	github.com/openshift/build-machinery-go v0.0.0-20220913142420-e25cf57ea46d
 	github.com/openshift/client-go v0.0.0-20231218140158-47f6d749b9d9
-	github.com/openshift/library-go v0.0.0-20240305144041-18ee8279b4e3
+	github.com/openshift/library-go v0.0.0-20240312150634-4b1aed6f35f5
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.16.0
 	github.com/prometheus/common v0.44.0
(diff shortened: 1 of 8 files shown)


pkg/operator/etcdcertsigner/target_cert_creator.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// CARotatingTargetCertCreator ensures we also rotate leaf certificates when we detect a change in signer.
+// The certrotation.TargetCertCreator only assumes the bundle to change on a CA rotation, whereas we have to keep
+// the bundle around for some time for a proper static pod rollout.
+type CARotatingTargetCertCreator struct {
+	certrotation.TargetCertCreator
+}
+
+func (c *CARotatingTargetCertCreator) NeedNewTargetCertKeyPair(
```
hasbro17, line 20: So I get that we're overriding the `NeedNewTargetCertKeyPair()` method here to check the extra constraints on the signer not matching to trigger rotation.

Shouldn't we be using `CARotatingTargetCertCreator` somewhere then?
e.g:
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/tlshelpers/tlshelpers.go#L216
or
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/tlshelpers/tlshelpers.go#L109

(Unless I'm being dense and missing where we have that already).

tjungblu: LOL you're right. Sorry, I'll get that done throughout the day.

tjungblu: done

tjungblu:

unrelated OLM failure

tjungblu:
I'm going to set this label to not let a full week of CI run data go to waste - sorry it took so long to retest and eventually get here

/label acknowledge-critical-fixes-only

tjungblu:
sigh, another retest for today :yawning_face:

PR #1186, opened 2024-01-24
ETCD-517: run certsignercontroller for bootstrap render

tjungblu:

PR #1205, opened 2024-02-13
NO-JIRA: replace nodelister with master nodelister

tjungblu:


dusk125 (approved):
/lgtm

tjungblu:
/hold cancel

@dusk125 could you take another look please? I've rebased to the merged changes in #1186

dusk125 (approved):
/lgtm

PR #1206, opened 2024-02-20
OCPBUGS-29717: [4.15] Replace nodelister with master nodelister everywhere

tjungblu:


geliu2016 (approved):
/label cherry-pick-approved

tjungblu:
/cherry-pick release-4.14

@geliu2016 could you label again please? I think it didn't come through

PR #1208, opened 2024-02-21
ETCD-349: add etcd cachedmembers

Elbehery:
resolves https://issues.redhat.com/browse/ETCD-349

### DIFF
--- a/pkg/etcdcli/cached_members.go
+++ b/pkg/etcdcli/cached_members.go
@@ -10,29 +10,31 @@ import (
 const refreshPeriod = time.Minute
 
 type CachedMembers struct {
-	client  EtcdClient
-	lock    sync.RWMutex
-	members memberHealth
-
-	lastRefresh time.Time
+	client          EtcdClient
+	lock            sync.Mutex
+	members         memberHealth
+	refreshInterval time.Duration
+	lastRefresh     time.Time
 }
 
 func NewMemberCache(client EtcdClient) AllMemberLister {
+	return newMemberCache(client, refreshPeriod)
+}
+
+func newMemberCache(client EtcdClient, refreshInterval time.Duration) AllMemberLister {
 	return &CachedMembers{
-		client:      client,
-		lock:        sync.RWMutex{},
-		lastRefresh: time.UnixMilli(0),
+		client:          client,
+		lock:            sync.Mutex{},
+		refreshInterval: refreshInterval,
+		lastRefresh:     time.UnixMilli(0),
 	}
 }
 
 func (c *CachedMembers) shouldRefresh() bool {
-	return c.lastRefresh.Add(refreshPeriod).Before(time.Now())
+	return c.lastRefresh.Add(c.refreshInterval).Before(time.Now())
 }
 
 func (c *CachedMembers) refresh(ctx context.Context) error {
-	c.lock.Lock()
-	defer c.lock.Unlock()
-
 	if c.shouldRefresh() {
 		health, err := c.client.MemberHealth(ctx)
 		if err != nil {
@@ -47,9 +49,6 @@ func (c *CachedMembers) refresh(ctx context.Context) error {
 }
 
 func (c *CachedMembers) convertToProtoMemberList() []*etcdserverpb.Member {
-	c.lock.RLock()
-	defer c.lock.RUnlock()
-
 	var members []*etcdserverpb.Member
 	for _, member := range c.members {
 		members = append(members, member.Member)
@@ -58,14 +57,19 @@ func (c *CachedMembers) convertToProtoMemberList() []*etcdserverpb.Member {
 }
 
 func (c *CachedMembers) MemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
 	if err := c.refresh(ctx); err != nil {
 		return nil, err
 	}
-
 	return c.convertToProtoMemberList(), nil
 }
 
 func (c *CachedMembers) VotingMemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
 	if err := c.refresh(ctx); err != nil {
 		return nil, err
 	}
@@ -75,45 +79,56 @@ func (c *CachedMembers) VotingMemberList(ctx context.Context) ([]*etcdserverpb.M
 }
 
 func (c *CachedMembers) HealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
 	if err := c.refresh(ctx); err != nil {
 		return nil, err
 	}
-
-	c.lock.RLock()
-	defer c.lock.RUnlock()
-
 	return c.members.GetHealthyMembers(), nil
 }
 
 func (c *CachedMembers) HealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
 	if err := c.refresh(ctx); err != nil {
 		return nil, err
 	}
-
-	c.lock.RLock()
-	defer c.lock.RUnlock()
-
 	return filterVotingMembers(c.members.GetHealthyMembers()), nil
 }
 
 func (c *CachedMembers) UnhealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
 	if err := c.refresh(ctx); err != nil {
 		return nil, err
 	}
-
-	c.lock.RLock()
-	defer c.lock.RUnlock()
-
 	return c.members.GetUnhealthyMembers(), nil
 }
 
 func (c *CachedMembers) UnhealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+	return filterVotingMembers(c.members.GetUnhealthyMembers()), nil
+}
+
+func (c *CachedMembers) MemberHealth(ctx context.Context) (memberHealth, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
 	if err := c.refresh(ctx); err != nil {
 		return nil, err
 	}
 
-	c.lock.RLock()
-	defer c.lock.RUnlock()
-
-	return filterVotingMembers(c.members.GetUnhealthyMembers()), nil
+	var health memberHealth
+	for _, m := range c.members {
+		health = append(health, m)
+	}
+	return health, nil
 }
--- a/pkg/etcdcli/cached_members.go
+++ b/pkg/etcdcli/cached_members.go
@@ -1,0 +1,119 @@
+package etcdcli
+
+import (
+	"context"
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	"sync"
+	"time"
+)
+
+const refreshPeriod = time.Minute
+
+type CachedMembers struct {
+	client  EtcdClient
+	lock    sync.RWMutex
+	members memberHealth
+
+	lastRefresh time.Time
+}
+
+func NewMemberCache(client EtcdClient) AllMemberLister {
+	return &CachedMembers{
+		client:      client,
+		lock:        sync.RWMutex{},
+		lastRefresh: time.UnixMilli(0),
+	}
+}
+
+func (c *CachedMembers) shouldRefresh() bool {
+	return c.lastRefresh.Add(refreshPeriod).Before(time.Now())
+}
+
+func (c *CachedMembers) refresh(ctx context.Context) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if c.shouldRefresh() {
+		health, err := c.client.MemberHealth(ctx)
+		if err != nil {
+			return err
+		}
+
+		c.members = health
+		c.lastRefresh = time.Now()
+	}
+
+	return nil
+}
+
+func (c *CachedMembers) convertToProtoMemberList() []*etcdserverpb.Member {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	var members []*etcdserverpb.Member
+	for _, member := range c.members {
+		members = append(members, member.Member)
+	}
+	return members
+}
+
+func (c *CachedMembers) MemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	return c.convertToProtoMemberList(), nil
+}
+
+func (c *CachedMembers) VotingMemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	list := c.convertToProtoMemberList()
+	return filterVotingMembers(list), nil
+}
+
+func (c *CachedMembers) HealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.members.GetHealthyMembers(), nil
+}
+
+func (c *CachedMembers) HealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return filterVotingMembers(c.members.GetHealthyMembers()), nil
+}
+
+func (c *CachedMembers) UnhealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.members.GetUnhealthyMembers(), nil
+}
+
+func (c *CachedMembers) UnhealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return filterVotingMembers(c.members.GetUnhealthyMembers()), nil
+}
--- a/pkg/etcdcli/cached_members_test.go
+++ b/pkg/etcdcli/cached_members_test.go
@@ -1,0 +1,348 @@
+package etcdcli
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	u "github.com/openshift/cluster-etcd-operator/pkg/testutils"
+	"github.com/stretchr/testify/require"
+
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+)
+
+func TestCachedMembers_refresh(t *testing.T) {
+	client, err := NewFakeEtcdClient(defaultEtcdMembers(), WithFakeClusterHealth(&FakeMemberHealth{Healthy: 3, Unhealthy: 0}))
+	require.NoError(t, err)
+	memberCache := newMemberCache(client, 1*time.Second)
+	currentMembers, err := memberCache.MemberList(context.Background())
+	require.Equal(t, defaultEtcdMembers(), currentMembers)
+
+	err = client.MemberRemove(context.Background(), uint64(2))
+	require.NoError(t, err)
+	time.Sleep(1 * time.Second)
+	currentMembers, err = memberCache.MemberList(context.Background())
+	require.Equal(t, fakeEtcdMembers(1, 3), currentMembers)
+}
+
+func TestCachedMembers_MemberList(t *testing.T) {
+	testCases := []struct {
+		name            string
+		etcdMembersList []*etcdserverpb.Member
+		clientOpts      FakeClientOption
+		expEtcdMembers  []*etcdserverpb.Member
+	}{
+		{
+			"all healthy members",
+			defaultEtcdMembers(),
+			WithFakeClusterHealth(&FakeMemberHealth{Healthy: 3, Unhealthy: 0}),
+			defaultEtcdMembers(),
+		},
+		{
+			"one unhealthy member and two healthy members",
+			defaultEtcdMembers(),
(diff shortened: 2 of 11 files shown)


dusk125:
So since pretty much all functions call the refresh function (which write-locks the mutex), it's very unlikely that this will be able to properly use the read mutexes. So essentially you have a write-lock mutex on every function.

I wonder if it would be better to start a goroutine that write-locks and updates the cache (calling refresh) on a timer instead. That way, each of the external functions would just read-lock and could actually access the data at the same time.

tjungblu:
@dusk125  alternatively, I think it could be a normal mutex. This is not necessary a performance improvement, merely an attempt at reducing the amount of RPC calls to etcd. The remote calls take much longer than the write lock would be taken (just a few instructions to check whether it would need a refresh).

Timer is also cool with me, but needs to have some additional error handling to bubble up when a cache is known to be outdated by more than a minute.

dusk125:
Sounds good Thomas, then a normal mutex would be good, keep it simple! :D

pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -146,12 +146,12 @@ func CheckSafeToScaleCluster(
 		return fmt.Errorf("CheckSafeToScaleCluster %d nodes are required, but only %d are available", minimumNodes, nodeCount)
 	}
 
-	healthyMembers, err := etcdClient.HealthyMembers(context.Background())
+	memberHealth, err := etcdClient.MemberHealth(context.Background())
```
Elbehery, line 149: @tjungblu the tests were failing because the `etcdfakeclient` was returning always empty slice for memberhealth :/ 

i tried many fixes, but this is the cleanest, ptal

pkg/etcdcli/cached_members.go
```diff
@@ -117,3 +117,7 @@ func (c *CachedMembers) UnhealthyVotingMembers(ctx context.Context) ([]*etcdserv
 
 	return filterVotingMembers(c.members.GetUnhealthyMembers()), nil
 }
+
+func (c *CachedMembers) MemberHealth(ctx context.Context) (memberHealth, error) {
```
Elbehery, line 121: i just returned the `memberhealth` as it is always updated using the mutex iiuc

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,123 @@
...
+func NewMemberCache(client EtcdClient) AllMemberLister {
+	return &CachedMembers{
+		client:      client,
+		lock:        sync.RWMutex{},
+		lastRefresh: time.UnixMilli(0),
+	}
+}
+
+func (c *CachedMembers) shouldRefresh() bool {
+	return c.lastRefresh.Add(refreshPeriod).Before(time.Now())
```
tjungblu, line 29: before or after? Sorry I think this was my mistake. Can you add some unit tests for this refresh if you have the time please?

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,123 @@
...
+	lock    sync.RWMutex
+	members memberHealth
+
+	lastRefresh time.Time
+}
+
+func NewMemberCache(client EtcdClient) AllMemberLister {
+	return &CachedMembers{
+		client:      client,
+		lock:        sync.RWMutex{},
```
tjungblu, line 23: I thought we wanted to move this to a plain `sync.Mutex` after some discussion here?

pkg/etcdcli/cached_members.go
tjungblu: I would still return a defensive copy here. Don't we have to export the "memberHealth" struct?

pkg/etcdcli/health.go
```diff
@@ -41,6 +41,10 @@ type healthCheck struct {
 
 type memberHealth []healthCheck
 
+func GetMemberHealth(ctx context.Context, etcdMembers []*etcdserverpb.Member) memberHealth {
```
tjungblu, line 44: this seems duplicated, why not make the below method public?

pkg/etcdcli/cached_members.go
Elbehery: +1

pkg/etcdcli/health.go
Elbehery: on it 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: on it 👍🏽

Elbehery: exporting `memberHealth` will cause a lot of changes all over the package, also it duplicates with 

```
type MemberHealth interface {
	MemberHealth(ctx context.Context) (memberHealth, error)
}
```

i think using exporting the method will do 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: i think before is correct, so the `lastRefresh` + `minutes` is before `now` 

i can add the test, but it is already tested extensively in `memberhealth_test` 

is there any other test cases i can add ?

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,129 @@
...
+		lastRefresh: time.UnixMilli(0),
+	}
+}
+
+func (c *CachedMembers) shouldRefresh() bool {
+	return c.lastRefresh.Add(refreshPeriod).Before(time.Now())
+}
+
+func (c *CachedMembers) refresh(ctx context.Context) error {
+	c.lock.Lock()
```
Elbehery, line 33: @tjungblu i think we should lock and ulock if `shouldRefresh()` is true 

wdyt ?

tjungblu: we always need to lock, because reading the lastRefresh needs a memory barrier

pkg/etcdcli/cached_members.go
tjungblu: :+1: I think this whole struct needs _some_ unit testing, even if it's already covered - get creative!

Elbehery: +1 :) 👍🏽

pkg/etcdcli/cached_members.go
tjungblu: yep, sounds good. I think it's a typedef for the slice anyway, so we're fine.

pkg/etcdcli/cached_members.go
Elbehery: i cant get this ? .. what should i change ?

i am working on testing now 👍🏽

pkg/etcdcli/cached_members_test.go
```diff
@@ -0,0 +1,314 @@
...
+			require.NoError(t, err)
+			memberCache := newMemberCache(client, 1*time.Millisecond)
+
+			currentMembers, err := memberCache.UnhealthyMembers(context.Background())
+			require.Equal(t, tc.expEtcdMembers, currentMembers)
+		})
+	}
+}
+
+func TestCachedMembers_MemberHealth(t *testing.T) {
```
Elbehery, line 254: @tjungblu  this test is failing, seems the memberHealth is not responding to the fake

tjungblu: it is responding, it just always returns all members with a status: healthy true/false.

pkg/etcdcli/cached_members.go
tjungblu: I'll send you a suggestion on this PR

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+		c.members = health
+		c.lastRefresh = time.Now()
+	}
+
+	return nil
+}
+
+func (c *CachedMembers) convertToProtoMemberList() []*etcdserverpb.Member {
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
tjungblu, line 56:
```suggestion

```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+func (c *CachedMembers) shouldRefresh() bool {
+	return c.lastRefresh.Add(c.refreshInterval).Before(time.Now())
+}
+
+func (c *CachedMembers) refresh(ctx context.Context) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
tjungblu, line 39:
```suggestion

```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+
+	var members []*etcdserverpb.Member
+	for _, member := range c.members {
+		members = append(members, member.Member)
+	}
+	return members
+}
+
+func (c *CachedMembers) MemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
```
tjungblu, line 66:
```suggestion
      	c.lock.Lock()
	defer c.lock.Unlock()
	
	if err := c.refresh(ctx); err != nil {
```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, err
+	}
+
+	return c.convertToProtoMemberList(), nil
+}
+
+func (c *CachedMembers) VotingMemberList(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
```
tjungblu, line 74:
```suggestion
	c.lock.Lock()
	defer c.lock.Unlock()
	
	if err := c.refresh(ctx); err != nil {
```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, err
+	}
+
+	list := c.convertToProtoMemberList()
+	return filterVotingMembers(list), nil
+}
+
+func (c *CachedMembers) HealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
```
tjungblu, line 83:
```suggestion
	c.lock.Lock()
	defer c.lock.Unlock()
	
	if err := c.refresh(ctx); err != nil {
```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, err
+	}
+
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
tjungblu, line 88:
```suggestion

```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+	return c.members.GetHealthyMembers(), nil
+}
+
+func (c *CachedMembers) HealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
+		return nil, err
+	}
+
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
tjungblu, line 99:
```suggestion

```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return filterVotingMembers(c.members.GetHealthyMembers()), nil
+}
+
+func (c *CachedMembers) UnhealthyMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
```
tjungblu, line 105:
```suggestion
	c.lock.Lock()
	defer c.lock.Unlock()
	
	if err := c.refresh(ctx); err != nil {
```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, err
+	}
+
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
tjungblu, line 110:
```suggestion

```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return c.members.GetUnhealthyMembers(), nil
+}
+
+func (c *CachedMembers) UnhealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
```
tjungblu, line 116:
```suggestion
	c.lock.Lock()
	defer c.lock.Unlock()
	
	if err := c.refresh(ctx); err != nil {
```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, err
+	}
+
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
tjungblu, line 121:
```suggestion

```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return filterVotingMembers(c.members.GetUnhealthyMembers()), nil
+}
+
+func (c *CachedMembers) MemberHealth(ctx context.Context) (memberHealth, error) {
+	if err := c.refresh(ctx); err != nil {
```
tjungblu, line 127:
```suggestion
	c.lock.Lock()
	defer c.lock.Unlock()
	
	if err := c.refresh(ctx); err != nil {
```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, err
+	}
+
+	var health memberHealth
+	c.lock.Lock()
```
tjungblu, line 132:
```suggestion
```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+	for _, m := range c.members {
+		health = append(health, m)
+	}
+	c.lock.Unlock()
```
tjungblu, line 136:
```suggestion
```

pkg/etcdcli/cached_members_test.go
```diff
@@ -0,0 +1,314 @@
...
+	"context"
+	"testing"
+	"time"
+
+	u "github.com/openshift/cluster-etcd-operator/pkg/testutils"
+	"github.com/stretchr/testify/require"
+
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+)
+
```
tjungblu, line 13: you're still missing a test that tests the thing that's actually needed to be tested in the cache: the refresh mechanism :)

suggestion: setup with some members, assert it works, sleep (maybe find a more resilient way), change the members in the client, assert the result has changed.

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	return c.members.GetHealthyMembers(), nil
+}
+
+func (c *CachedMembers) HealthyVotingMembers(ctx context.Context) ([]*etcdserverpb.Member, error) {
+	if err := c.refresh(ctx); err != nil {
```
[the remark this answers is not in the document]
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members_test.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members_test.go
Elbehery: done 👍🏽

pkg/etcdcli/cached_members_test.go
```diff
@@ -0,0 +1,364 @@
...
+func TestCachedMembers_refresh(t *testing.T) {
+	testCases := []struct {
+		name            string
+		etcdMembersList []*etcdserverpb.Member
+		expEtcdMembers  []*etcdserverpb.Member
+	}{
+		{
+			"all healthy members",
```
tjungblu, line 21: one last nit-pick: if it's just a test, you can directly write it down without the t.Run indirection

Elbehery: But is this 'refreshTest' sufficient ?

I used `time.Sleep()` 

What would be better option ?

tjungblu: should be fine, another option would be to have some kind of callback that you could register on refresh and assert that one is called

Elbehery: Like ?

`mock or stub`

I would like to add this actually -   i used to do it java but never done it in go

Elbehery: i think it is too complex for a simple test .. lets keep it this way 👍🏽

pkg/operator/starter.go
```diff
@@ -242,7 +244,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().Namespaces().Lister(),
 		configInformers.Config().V1().Infrastructures().Lister(),
 		operatorClient,
-		etcdClient)
+		cachedMemberClient)
```
tjungblu, line 247: just going through some other controllers that could be cached:
* NewEtcdMembersController <- that's just informatory, so doesn't need consistent information
* NewDefragController <- that's also not mission critical to be consistent, it's also quite shoddy in how often it calls the member listing


anything else that springs to your mind?

Elbehery: i also had another run over CEO ... i cant find anything that will need `memberList()` periodically

i think if we utilise this in the `clientPool` it will be best however, wdyt ?

Elbehery:
the failing jobs are due to other failures, not related to this PR 

/retest-required

Elbehery:
@tjungblu `ci/prow/e2e-operator-fips` failure not related to this PR 

shall we  override ?

tjungblu:
why? it's not required

Elbehery:
> why? it's not required, or is it? 🤣

doch doch, it is required .. but the failure has nothing to do with this change https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1208/pull-ci-openshift-cluster-etcd-operator-master-e2e-operator-fips/1764984733453783040

pkg/operator/starter.go
tjungblu: do you want to update the members and defrag controller with this as well?

pkg/etcdcli/cached_members_test.go
tjungblu: sounds good!

pkg/operator/starter.go
Elbehery: 👍🏽

pkg/etcdcli/interfaces.go
```diff
@@ -31,6 +31,15 @@ type EtcdClient interface {
 	MemberUpdatePeerURL(ctx context.Context, id uint64, peerURL []string) error
 }
 
+type AllMemberLister interface {
+	MemberLister
+	HealthyMemberLister
+	UnhealthyMemberLister
+	MemberHealth
+	Status
+	Defragment
```
tjungblu, line 40:
```suggestion

```

let's keep the interface with just cached calls

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -44,7 +44,7 @@ const (
 // observations the controller will perform rolling defragmentation of each etcd member in the cluster.
 type DefragController struct {
 	operatorClient       v1helpers.OperatorClient
-	etcdClient           etcdcli.EtcdClient
+	etcdClient           etcdcli.AllMemberLister
```
tjungblu, line 47:
```suggestion
	memberLister           etcdcli.AllMemberLister
	
	defragClient              Defragment
	statusClient               Status
```

that's freestyle, hope this gets the idea across, no idea if that compiles

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -55,7 +55,7 @@ type DefragController struct {
 func NewDefragController(
 	livenessChecker *health.MultiAlivenessChecker,
 	operatorClient v1helpers.OperatorClient,
-	etcdClient etcdcli.EtcdClient,
+	etcdClient etcdcli.AllMemberLister,
```
tjungblu, line 58:
```suggestion
	memberLister etcdcli.AllMemberLister,
	statusClient Status,
	defragClient Defragment,
	
```

pkg/etcdcli/cached_members.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	var health memberHealth
+	for _, m := range c.members {
+		health = append(health, m)
+	}
+	return health, nil
+}
+
+func (c *CachedMembers) Status(ctx context.Context, target string) (*clientv3.StatusResponse, error) {
```
tjungblu, line 137: you can remove the two funcs now :)

Elbehery: done 👍🏽

tjungblu (commented):
/lgtm

thanks Mustafa :)

PR #1209, opened 2024-02-22
ETCD-518: garbage collect unused dynamic certs

tjungblu:


pkg/operator/etcdcertcleaner/etcd_cert_cleaner_controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+	// secret should NOT be deleted
+	updated := false
+	deleted := false
+	fx := cleaner.kubeClient.(*fake.Clientset)
+	for _, action := range fx.Actions() {
+		if action.Matches("delete", "secrets") {
+			require.Equal(t, "etcd-peer-master-0", action.(k8stesting.DeleteAction).GetName())
+			deleted = true
+		} else if action.Matches("update", "secrets") {
+			// secret should be relabeled with the grace annotation though
```
tjungblu, line 97: copy&paste error

pkg/operator/etcdcertcleaner/etcd_cert_cleaner_controller.go
```diff
@@ -0,0 +1,187 @@
...
+// deletion label is getting removed automatically - in the worst case the cert is getting recreated entirely.
+func (c *EtcdCertCleanerController) sync(ctx context.Context, _ factory.SyncContext) error {
+	secrets, err := c.findUnusedNodeBasedSecrets(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, secretToDelete := range secrets {
+		klog.Infof("deleting secret [%s] detected unused for %s", secretToDelete.Name, DeletionGracePeriod)
+		err := c.secretClient.Secrets(operatorclient.TargetNamespace).Delete(ctx, secretToDelete.Name, v1.DeleteOptions{})
```
hasbro17, line 90: The dynamic certs are all in the `openshift-etcd` target namespace so that makes sense. I was wondering if we could have unused certs in the other namespaces that we copy to e.g openshift-config. Don't think those are dynamic e.g client or ca-bundle, so no.

tjungblu: Yep, good point. 

The cool feature of the secret/configmap sync is that it deletes when the origin resources disappears. Maybe you remember that this caused some issues in the nightlies the other day and TRT had to revert ;)

It's enough to clear the "source of truth", the rest follows along nicely.

pkg/operator/etcdcertcleaner/etcd_cert_cleaner_controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+			obj := action.(k8stesting.UpdateAction).GetObject().(*corev1.Secret)
+			require.Equal(t, "etcd-peer-master-1", obj.Name)
+			require.Contains(t, obj.Annotations, DeletionGracePeriodAnnotation)
+			updated = true
+		}
+	}
+
+	require.Truef(t, updated, "expected secret to be deleted, but only found actions: %v", fx.Actions())
```
hasbro17, line 47:
```suggestion
	require.Truef(t, updated, "expected secret to be updated, but only found actions: %v", fx.Actions())
```

pkg/operator/etcdcertcleaner/etcd_cert_cleaner_controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+		} else if action.Matches("update", "secrets") {
+			// secret should be relabeled and have the grace annotation removed again
+			obj := action.(k8stesting.UpdateAction).GetObject().(*corev1.Secret)
+			require.Equal(t, "etcd-peer-master-0", obj.Name)
+			require.NotContains(t, obj.Annotations, DeletionGracePeriodAnnotation)
+			updated = true
+		}
+	}
+
+	require.Truef(t, updated, "expected secret to be deleted, but only found actions: %v", fx.Actions())
```
hasbro17, line 105:
```suggestion
	require.Truef(t, updated, "expected secret to be updated, but only found actions: %v", fx.Actions())
```

pkg/operator/etcdcertcleaner/etcd_cert_cleaner_controller.go
```diff
@@ -0,0 +1,187 @@
...
+		operatorClient:     operatorClient,
+		masterNodeLister:   masterNodeLister,
+		masterNodeSelector: masterNodeSelector,
+		secretInformer:     secretInformer,
+		secretLister:       secretLister,
+		secretClient:       secretClient,
+	}
+
+	// we estimate the rate of control plane scales to maybe 1-2 a day, so running this once an hour with 6h of health threshold
+	// should be more than sufficient. Note that this controller does not sync on any informers.
```
hasbro17, line 73: Sounds fair. Even if there's an abnormal amount of scaling and we have a pile up of unused secrets and you need to clean them before the hour mark, you can always manually clean them. Or just restart the operator to resync the controller.

hasbro17 (approved):
Just a nit in the unit test assertion, otherwise LGTM

pkg/operator/etcdcertcleaner/etcd_cert_cleaner_controller_test.go
tjungblu: applied, thanks for double-checking. I should stop copy and pasting :rofl:

pkg/operator/etcdcertcleaner/etcd_cert_cleaner_controller_test.go
tjungblu: applied

pkg/operator/etcdcertcleaner/etcd_cert_cleaner_controller.go
tjungblu: yep, after our v-scaling work I'd have some trouble finding a hyperscaler that could scale that fast :)

tjungblu:
as reported in the other PR, OLM seems to fail on any change in FIPS.

PR #1214, opened 2024-03-01
NO_JIRA: ensure etcd client informer sync

tjungblu:
As an extension to #1205, the client should ensure the first informers are already sync'd before starting. On CEO updates this causes flapping operator state on several controller status reports.

### DIFF
--- a/pkg/operator/operatorclient/common.go
+++ b/pkg/operator/operatorclient/common.go
@@ -1,0 +1,21 @@
+package operatorclient
+
+import (
+	"context"
+	"fmt"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+	"time"
+)
+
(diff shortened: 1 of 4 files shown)


tjungblu:
/hold

need to see whether that even installs

pkg/operator/operatorclient/common.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"context"
+	"fmt"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+	"time"
+)
+
+func AwaitInformerCacheSync(name string, inf cache.SharedIndexInformer) error {
```
tjungblu, line 11: not the best place to put, but I couldn't put it into `ceohelpers` because of cyclic imports :/ it wasn't that useful to create a new package for it either

PR #1216, opened 2024-03-05
OCPBUGS-30245: [4.13] Replace nodelister with master nodelister everywhere

tjungblu:


geliu2016 (approved):
/label cherry-pick-approved

tjungblu:
@Elbehery can you add the BPA label here please?
https://github.com/openshift/cluster-etcd-operator/pull/1216

PR #1223, opened 2024-03-13
OCPBUGS-30873: CEO aliveness check should only detect deadlocks

tjungblu:
Currently we only detect whether a controller has been running continuously into errors. Whereas we wanted to detect real deadlock situations. This change defuses the aliveness check to only declare real locking situations as problematic.

Additionally, to not create insane amounts of log traffic, this change will throttle the stack dumping to once every 15 minutes. Previously it would trigger almost immediately every health probe invocation and create multi-megabyte log spam.


tjungblu:
unrelated failures

PR #1227, opened 2024-03-14
ETCD-573: add recert cmd

tjungblu:
you can run it with:

> cluster-etcd-operator recert -o asset-out --hips master-1=192.168.2.1,master-2=192.168.2.2,master-3=192.168.2.3

### DIFF
--- a/cmd/cluster-etcd-operator/main.go
+++ b/cmd/cluster-etcd-operator/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	goflag "flag"
 	"fmt"
+	"github.com/openshift/cluster-etcd-operator/pkg/cmd/recert"
 	"io/ioutil"
 	"math/rand"
 	"os"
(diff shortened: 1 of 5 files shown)


hasbro17:
/hold

Haven't forgotten about this but still reviewing.

Not blocking this PR but just wanted to think ahead on how we actually want to run this cmd automatically once we detect etcd is down with expired certs. That may affect how we generate them here.

First is the detection of expired certs. I thinking this would be a health check or polling probe that can either query etcd locally to see a `x509: certificate has expired or is not yet valid` or just inspect the on-disk cert to check the date of expiry.
If this is a sidecar in the operator then we may not have sufficient hostpath permissions to do either of that right? And it can't be in the etcd pod as we need to run this from a single place.

And secondly the distribution step. Since we're generating everything in one place, I'm guessing we have to scp this around to all the other nodes. Not for SNO though.

Lastly since we're only modifying the on-disk cert files, that doesn't change the secrets and bundle configmaps in etcd, that are used by the installer for a new revision.
So we need to figure out how we update the cert secrets and configmaps in etcd otherwise the next revision rollout would reuse the expired signer certs in etcd, as opposed to the new ones generated on disk.

Maybe if we relaxed the constraint and assume that the signers aren't expired when the cluster is offline then we can only regenerate the peer/server and client certs on disk, distribute them, bring the cluster up, and then rotate the node cert secrets and configmaps.

Anyway, not a blocker for this PR but we can discuss and flesh that out a bit as well.

PR #1234, opened 2024-04-03
use RotatedSigningCASecret and RotatedSelfSignedCertKeySecret only in update mode

p0lyn0mial:
UseSecretUpdateOnly is intended as a short term hack for a very specific use case,
and it works in tandem with a particular carry patch applied to the openshift kube-apiserver.
(https://github.com/openshift/kubernetes/pull/1924)

we will remove this when we migrate all of the affected secret
objects to their intended type: https://issues.redhat.com/browse/API-1800

the issue is that the secret's crypto material
can be regenerated, which has serious consequences for the platform
as it can break external clients and the cluster itself.

xref: https://github.com/openshift/library-go/pull/1705
xref: https://github.com/openshift/kubernetes/pull/1924


tjungblu:
@p0lyn0mial I think we don't need to have that flag set, none of the certs that you tag are created with the SecretTLS type AFAIK

p0lyn0mial:
> @p0lyn0mial I think we don't need to have that flag set, none of the certs that you tag are created with the SecretTLS type AFAIK

I'm still testing, so far we have at least 
https://github.com/openshift/cluster-etcd-operator/blob/eeef8033eb058b0f9899c7b16aead237d5b84462/manifests/0000_20_etcd-operator_03_secret.yaml#L3

tjungblu:
That's for CEO, we can simply patch that out.

p0lyn0mial:
> That's for CEO, we can simply patch that out.

We should, we should also make sure that the secret created in the previous versions will be safely migrated to the new type.

> @p0lyn0mial I think we don't need to have that flag set, none of the certs that you tag are created with the SecretTLS type AFAIK

Vadim created a 4.6 cluster and below is the list of secrets with the old type.
However, considering the platform's long history spanning several years, I am not convinced that the secrets could have been created using only one type

https://github.com/openshift/kubernetes/pull/1929#issuecomment-2034605161

tjungblu:
Great list, I think we might be OK still, given the clients can be updated safely. Let me know how your testing goes. 

Keep in mind that most secrets outside of openshift-etcd will be simply copied through the resource controller. 
Its implementation IIRC will also simply delete+recreate, so the patching in here might not be sufficient (if we really need it).

PR #1235, opened 2024-04-04
ETCD-579: store revision at which signers are rotated

tjungblu:
This is some helper code to safer automated signer rotations in a follow-up PR.

/hold


pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -87,7 +91,7 @@ func NewEtcdCertSignerController(
 	eventRecorder = eventRecorder.WithComponentSuffix("etcd-cert-signer-controller")
 	cmInformer := kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps()
 	cmLister := cmInformer.Lister()
-	cmGetter := v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), kubeInformers)
+	cmGetter := kubeClient.CoreV1()
```
tjungblu, line 94: btw, I removed the caching here because it was causing flaky tests - that should also help with the recert command...

tjungblu: actually, funny enough: because the cache is synced on the same informer the results are always consistent :rofl: 

Otherwise the informer might be lagging behind what the client getter would return.

pkg/cmd/render/certs_test.go
```diff
@@ -133,6 +133,7 @@ func assertBundleCorrectness(t *testing.T, secrets []corev1.Secret, bundles []co
 			return bytes.Compare(bundleCerts[i].Raw, bundleCerts[j].Raw) < 0
 		})
 
+		require.Equal(t, len(signers), len(bundleCerts))
```
lance5890, line 136: redundant, just for testing ?

tjungblu: yep, I was debugging some issue where the informer-based secret listing would return only one signer secret instead of two - which then created a bundle of different size. 
It's just a more explicit assertion than the unwieldy equals output

tjungblu:
/hold cancel

works like a charm :)

pkg/cmd/render/certs_test.go
lance5890: I have also met some issue where the informer-based secret or configmap return different value, but that is found in the kube-controller-mamager-operator, and I have found the bug in the library-go:
https://github.com/openshift/library-go/pull/1532/files


Hope it can give you some inspiration

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -272,7 +294,7 @@ func (c *EtcdCertSignerController) syncAllMasterCertificates(ctx context.Context
 	// pod controller to watch. A single secret ensures that a cert change
 	// (e.g. node addition or cert rotation) triggers at most one static pod
 	// rollout. If multiple secrets were written, the static pod controller
-	// might initiate rollout before all secrets had been updated.
+	// might initiate rollout before all secrets had been signerUpdated.
```
hasbro17, line 297: Nit: assuming this wasn't intended.
```suggestion
	// might initiate rollout before all secrets had been updated.
```

pkg/tlshelpers/client_lister.go
```diff
@@ -0,0 +1,77 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+// Below lister implementations are required to make the library-go certificate logic more robust.
+// Especially in render, we were relying on a race condition between the informer cache and the client
+// to create the bundles correctly. This lister always queries the API (or its mock) to directly return the
```
hasbro17, line 14: Is this something that we need to fix back in library-go? Or is it only us that's impacted with our use of the render cmd?

hasbro17 (approved):

pkg/tlshelpers/client_lister.go
tjungblu: I think this is a side-effect of the fake libs in upstream k8s

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
tjungblu: thanks! will fix this in the follow-up PR, to not waste your lgtm :)

PR #1238, opened 2024-04-08
ETCD-580: Decrease validity period to prevent cert expiry for offline clusters

hasbro17:
Addresses https://issues.redhat.com/browse/ETCD-580 to leave enough time on the cert before it is expired in the event the cluster is shutdown/hibernated for 9 months.

This helps prevent etcd certs from becoming expired while the cluster is offline for extended periods. 

/cc @tjungblu


tjungblu:

unrelated OLM failure

Issue #1237, opened 2024-04-07
quorum check when update secret


lance5890:
1. we found the etcd has also rollouted occasionally when we remove one master from cluster(3 masters)， and the logs show as follows:
```
2024-04-07T04:06:26.674157868-04:00 stderr F I0407 08:06:26.674067       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"253380d4-7d65-496f-8214-ab89f7878550", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Warning' reason: 'MasterNodeRemoved' Observed removal of master node node3
2024-04-07T04:06:26.748289563-04:00 stderr F I0407 08:06:26.748226       1 quorumguardcleanupcontroller.go:133] 3/2 guard pods ready. Waiting until all new guard pods are ready
2024-04-07T04:06:26.762266273-04:00 stderr F I0407 08:06:26.761362       1 quorumguardcleanupcontroller.go:133] 3/2 guard pods ready. Waiting until all new guard pods are ready
2024-04-07T04:06:26.762266273-04:00 stderr F E0407 08:06:26.761749       1 base_controller.go:272] EtcdEndpointsController reconciliation failed: EtcdEndpointsController can't evaluate whether quorum is safe: CheckSafeToScaleCluster 3 nodes are required, but only 2 are available
[...]

PR #1239, opened 2024-04-09
OCPBUGS-31849: check the quorum in case the cluster is healthy or not after generating certs

lance5890:
address https://github.com/openshift/cluster-etcd-operator/issues/1240  and https://github.com/openshift/cluster-etcd-operator/issues/1237

### DIFF
--- a/pkg/operator/ceohelpers/bootstrap.go
+++ b/pkg/operator/ceohelpers/bootstrap.go
@@ -117,11 +117,6 @@ func CheckSafeToScaleCluster(
 		return nil
 	}
 
-	_, operatorStatus, _, err := staticPodClient.GetStaticPodOperatorState()
-	if err != nil {
-		return fmt.Errorf("CheckSafeToScaleCluster failed to get operator state: %w", err)
-	}
-
 	scalingStrategy, err := GetBootstrapScalingStrategy(staticPodClient, namespaceLister, infraLister)
 	if err != nil {
 		return fmt.Errorf("CheckSafeToScaleCluster failed to get bootstrap scaling strategy: %w", err)
@@ -141,11 +136,6 @@ func CheckSafeToScaleCluster(
 		return fmt.Errorf("CheckSafeToScaleCluster unrecognized scaling strategy %q", scalingStrategy)
 	}
 
-	nodeCount := len(operatorStatus.NodeStatuses)
-	if nodeCount < minimumNodes {
-		return fmt.Errorf("CheckSafeToScaleCluster %d nodes are required, but only %d are available", minimumNodes, nodeCount)
-	}
-
 	memberHealth, err := etcdClient.MemberHealth(context.Background())
 	if err != nil {
 		return fmt.Errorf("CheckSafeToScaleCluster couldn't determine member health: %w", err)
@@ -156,7 +146,7 @@ func CheckSafeToScaleCluster(
 		return err
 	}
 
-	klog.V(4).Infof("node count %d satisfies minimum of %d required by the %s bootstrap scaling strategy", nodeCount, minimumNodes, scalingStrategy)
+	klog.V(4).Infof("node count %d satisfies minimum of %d required by the %s bootstrap scaling strategy", len(memberHealth.GetHealthyMembers()), minimumNodes, scalingStrategy)
 	return nil
 }
 
--- a/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
+++ b/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
@@ -283,6 +283,16 @@ func (c *EtcdCertSignerController) syncAllMasterCertificates(ctx context.Context
 		Type: corev1.SecretTypeOpaque,
 		Data: allCerts,
 	}
+
+	//check the quorum in case the cluster is healthy or not after generating certs
+	safe, err := c.quorumChecker.IsSafeToUpdateRevision()
+	if err != nil {
+		return fmt.Errorf("EtcdCertSignerController can't evaluate whether quorum is safe: %w", err)
+	}
+
+	if !safe {
+		return fmt.Errorf("skipping EtcdCertSignerController reconciliation due to insufficient quorum")
+	}
 	_, _, err = resourceapply.ApplySecret(ctx, c.secretClient, recorder, secret)
 
 	return err
--- a/pkg/operator/ceohelpers/bootstrap_test.go
+++ b/pkg/operator/ceohelpers/bootstrap_test.go
@@ -269,15 +269,6 @@ func Test_CheckSafeToScaleCluster(t *testing.T) {
 			infraObj:           defaultInfra,
 			expectError:        nil,
 		},
-		"HA with insufficient nodes": {
-			namespace:          defaultNamespace,
-			bootstrapConfigMap: bootstrapComplete,
-			operatorConfig:     defaultOperatorConfig,
-			nodes:              twoNodesAtCurrentRevision,
-			etcdMembers:        u.DefaultEtcdMembers(),
-			infraObj:           defaultInfra,
-			expectError:        fmt.Errorf("CheckSafeToScaleCluster 3 nodes are required, but only 2 are available"),
-		},
 		"unsupported with sufficient nodes": {
 			namespace:          defaultNamespace,
 			bootstrapConfigMap: bootstrapComplete,
@@ -322,15 +313,6 @@ func Test_CheckSafeToScaleCluster(t *testing.T) {
 			etcdMembers:        u.DefaultEtcdMembers(),
 			infraObj:           defaultInfra,
 			expectError:        nil,
-		},
-		"delayed HA with insufficient nodes during steady state": {
-			namespace:          namespaceWithDelayedHAEnabled,
-			bootstrapConfigMap: bootstrapComplete,
-			operatorConfig:     defaultOperatorConfig,
-			nodes:              twoNodesAtCurrentRevision,
-			etcdMembers:        u.DefaultEtcdMembers(),
-			infraObj:           defaultInfra,
-			expectError:        fmt.Errorf("CheckSafeToScaleCluster 3 nodes are required, but only 2 are available"),
 		},
 		"HA with insufficient etcd members": {
 			namespace:          defaultNamespace,


pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -141,7 +139,11 @@ func CheckSafeToScaleCluster(
 		return fmt.Errorf("CheckSafeToScaleCluster unrecognized scaling strategy %q", scalingStrategy)
 	}
 
-	nodeCount := len(operatorStatus.NodeStatuses)
+	node, err := masterNodeLister.List(masterNodeSelector)
+	if err != nil {
+		return fmt.Errorf("CheckSafeToScaleCluster couldn't get master nodes: %w", err)
+	}
+	nodeCount := len(node)
```
tjungblu, line 146: what if there are nodes that are not in ready state? I think the node status is slightly more advanced

lance5890: - please correct me if I missed something or not

1. I'm afraid the quorum check has never cared about the nodes ready or not , it just check the length of master nodes:
https://github.com/openshift/cluster-etcd-operator/blob/3d5483e1871ba147a692736c71645175a85769c4/pkg/operator/ceohelpers/bootstrap.go#L144-L148

2. the node_controller will add or remove the nodes to the operator.nodestatus just by watching the master nodes:
https://github.com/openshift/cluster-etcd-operator/blob/3d5483e1871ba147a692736c71645175a85769c4/vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/node/node_controller.go#L47-L86

3. and the operator.nodestatus may have not updated  when we use it in the quorum check

tjungblu: What's the difference then? Besides that we have one less indirection via the node status controller? 

Below the range you've linked, there's logic for NodeReady which we're going to lose:

https://github.com/openshift/cluster-etcd-operator/blob/3d5483e1871ba147a692736c71645175a85769c4/vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/node/node_controller.go#L89-L102

tjungblu: having said that, any logic based on the node and status is more explicitly encoded in the `MemberHealth` call a couple of lines below, so I think we can (IMHO) remove all node based logic. 

I would only try to add a timeout to the memberhealth call context, otherwise we might get stuck as we've seen multiple times with GRPC.

pkg/operator/ceohelpers/qourum_check_test.go
```diff
@@ -96,13 +108,41 @@ func TestQuorumCheck_IsSafeToUpdateRevision(t *testing.T) {
...
+				u.FakeNode("master-1", u.WithMasterLabel(), u.WithNodeInternalIP("10.0.0.2")),
+			},
+			etcdMembers: []*etcdserverpb.Member{
+				u.FakeEtcdMemberWithoutServer(0),
+				u.FakeEtcdMemberWithoutServer(1),
+				u.FakeEtcdMemberWithoutServer(2),
+			},
+			safe:        false,
+			expectedErr: fmt.Errorf("CheckSafeToScaleCluster 3 nodes are required, but only 2 are available"),
+		},
```
lance5890, line 145: should we fall into this situation as the UT I added ?: Etcd cluster is healthy, all nodes are ready(Just change one node label from master to worker ) @tjungblu

tjungblu: >  all nodes are ready(Just change one node label from master to worker 

that's an interesting case, how would the static pod react to that?

lance5890: - if we only remove the master role: node-role.kubernetes.io/master, the static pods on that node could still be running as normal, the etcd cluster is healthy too,:
```
https://x.x.x.1:2379, 5df6bd632e46a20e, 3.5.11, 127 MB, false, false, 10, 2458929, 2458929,
https://x.x.x.2:2379, 1b6b8161626a4eb9, 3.5.11, 127 MB, true, false, 10, 2458929, 2458929,
https://x.x.x.3:2379, 49869657b1942ea7, 3.5.11, 128 MB, false, false, 10, 2458929, 2458929,
```

but the ceo logs show as follows because there are only two masters now:

```
I0411 01:46:23.347054       1 quorumguardcleanupcontroller.go:133] 3/2 guard pods ready. Waiting until all new guard pods are ready
I0411 01:46:23.399103       1 quorumguardcleanupcontroller.go:133] 3/2 guard pods ready. Waiting until all new guard pods are ready
I0411 01:46:23.447863       1 quorumguardcleanupcontroller.go:133] 3/2 guard pods ready. Waiting until all new guard pods are ready
I0411 01:46:23.498217       1 quorumguardcleanupcontroller.go:133] 3/2 guard pods ready. Waiting until all new guard pods are ready
I0411 01:46:23.550646       1 quorumguardcleanupcontroller.go:133] 3/2 guard pods ready. Waiting until all new guard pods are ready
I0411 01:46:24.448640       1 quorumguardcleanupcontroller.go:133] 3/2 guard pods ready. Waiting until all new guard pods are ready
E0411 01:46:25.598762       1 base_controller.go:272] TargetConfigController reconciliation failed: TargetConfigController can't evaluate whether quorum is safe: CheckSafeToScaleCluster 3 nodes are required, but only 2 are available
E0411 01:46:25.646253       1 base_controller.go:270] "TargetConfigController" controller failed to sync "", err: TargetConfigController can't evaluate whether quorum is safe: CheckSafeToScaleCluster 3 nodes are required, but only 2 are available
E0411 01:46:25.649209       1 base_controller.go:272] EtcdEndpointsController reconciliation failed: EtcdEndpointsController can't evaluate whether quorum is safe: CheckSafeToScaleCluster 3 nodes are required, but only 2 are available
E0411 01:46:25.845918       1 base_controller.go:272] EtcdCertSignerController reconciliation failed: EtcdCertSignerController can't evaluate whether quorum is safe: CheckSafeToScaleCluster 3 nodes are required, but only 2 are available
E0411 01:46:41.373595       1 base_controller.go:272] EtcdCertSignerController reconciliation failed: EtcdCertSignerController can't evaluate whether quorum is safe: CheckSafeToScaleCluster 3 nodes are required, but only 2 are available
E0411 01:46:41.374949       1 base_controller.go:272] TargetConfigController reconciliation failed: TargetConfigController can't evaluate whether quorum is safe: CheckSafeToScaleCluster 3 nodes are required, but only 2 are available
E0411 01:46:41.378822       1 base_controller.go:272] EtcdEndpointsController reconciliation failed: EtcdEndpointsController can't evaluate whether quorum is safe: CheckSafeToScaleCluster 3 nodes are required, but only 2 are available
E0411 01:46:41.381235       1 base_controller.go:270] "TargetConfigController" controller failed to sync "", err: TargetConfigController can't evaluate whether quorum is safe: CheckSafeToScaleCluster 3 nodes are required, but only 2 are available
E0411 01:46:51.639452       1 base_controller.go:272] EtcdCertSignerController reconciliation failed: EtcdCertSignerController can't evaluate whether quorum is safe: CheckSafeToScaleCluster 3 nodes are required, but only 2 are available

```

tjungblu: will one etcd pod eventually wind down? I didn't exactly found the scale-down logic here, but it's also based on the node status from:
https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/node/node_controller.go#L47-L55

If it eventually shuts down that pod, failing the sync loops is the right thing to do.

tjungblu:
we're mostly done for the 4.16 cert stuff, I'll check this out tomorrow in depth again @lance5890

pkg/operator/ceohelpers/qourum_check_test.go
lance5890: I just scale-down the cluster manually by relabeling the master node to worker, and remove the etcd  static pod yaml to simulate this situation in ocp 4.12

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -148,15 +148,6 @@ func NewEtcdCertSignerController(
 }
 
 func (c *EtcdCertSignerController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
tjungblu, line 150: I think we definitely need to keep this check (in 4.16) because an update to the certificate bundle will also cause a revision rollout. 

You can test this with an earlier version as well, by adding a new entry to openshift-config/etcd-ca-bundle configmap.

pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -141,7 +139,11 @@ func CheckSafeToScaleCluster(
 		return fmt.Errorf("CheckSafeToScaleCluster unrecognized scaling strategy %q", scalingStrategy)
 	}
 
-	nodeCount := len(operatorStatus.NodeStatuses)
+	node, err := masterNodeLister.List(masterNodeSelector)
```
tjungblu, line 142: let's remove the entire nodeStatus and node listing logic here, I think the memberHealth and quorum check below should suffice. We can run some more payload tests to ensure we don't break anything.

tjungblu (changes requested):

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
lance5890: agree

pkg/operator/ceohelpers/bootstrap.go
lance5890: how about the logic of  createNodeCertConfigs which uses the master node listing logic
https://github.com/openshift/cluster-etcd-operator/blob/9afc6795eaac93bf854f00ed03db9c243b2157fb/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go#L293-L338

tjungblu: I think that can stay for now, because we need to have the certificates before a member can join the cluster. So using the etcd client won't work when a new node joins the cluster :)

pkg/operator/ceohelpers/bootstrap.go
```diff
@@ -156,7 +145,7 @@ func CheckSafeToScaleCluster(
 		return err
 	}
 
-	klog.V(4).Infof("node count %d satisfies minimum of %d required by the %s bootstrap scaling strategy", nodeCount, minimumNodes, scalingStrategy)
+	klog.V(4).Infof("node count satisfies minimum of %d required by the %s bootstrap scaling strategy", minimumNodes, scalingStrategy)
```
tjungblu, line 148: >  len(memberHealth.GetHealthyMembers())

could also do as a replacement for node count?

pkg/operator/ceohelpers/qourum_check.go
```diff
@@ -26,11 +26,11 @@ func (c *AlwaysSafeQuorumChecker) IsSafeToUpdateRevision() (bool, error) {
...
-	configMapLister corev1listers.ConfigMapLister
-	namespaceLister corev1listers.NamespaceLister
-	infraLister     configv1listers.InfrastructureLister
-	operatorClient  v1helpers.StaticPodOperatorClient
-	etcdClient      etcdcli.AllMemberLister
+	configMapLister    corev1listers.ConfigMapLister
+	namespaceLister    corev1listers.NamespaceLister
+	infraLister        configv1listers.InfrastructureLister
+	operatorClient     v1helpers.StaticPodOperatorClient
+	etcdClient         etcdcli.AllMemberLister
```
tjungblu, line 33: that's an odd formatting, tab vs. spaces maybe?

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -283,6 +282,15 @@ func (c *EtcdCertSignerController) syncAllMasterCertificates(ctx context.Context
 		Type: corev1.SecretTypeOpaque,
 		Data: allCerts,
 	}
+
+	safe, err := c.quorumChecker.IsSafeToUpdateRevision()
```
tjungblu, line 286: please also leave a comment here on why we are checking that again

tjungblu:
I'm so free to already run the blocking payloads, the unit test failures are expected given the removal of the statements

/payload 4.16 nightly blocking

pkg/operator/ceohelpers/bootstrap.go
lance5890: done

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
lance5890: make update-gofmt  , done

pkg/operator/ceohelpers/qourum_check.go
lance5890: make update-gofmt  done

tjungblu:
besides Azure, the bare metal jobs report bootstrap issues:
https://pr-payload-tests.ci.openshift.org/runs/ci/82e77220-fccf-11ee-8b38-ace81943d6db-0

I'll take a look in a bit, brb

tjungblu:
/payload 4.16 nightly blocking

trying again, the failures don't make a whole lot of sense to me. One run does have neither the installer bundle nor the must-gather. The other run it seems there's a whole machine entirely missing.

lance5890:
> /test e2e-metal-single-node-live-iso

should we move forward this ?

tjungblu:
I'm cool with merging this, we need to see how we can backport this potentially. But let's get this into 4.16 for now, we should have some more time to see if any regressions pop up.
Thanks a lot @lance5890 

/lgtm

PR #1242, opened 2024-04-10
first try for auto-recert

tjungblu:
This PR adds:
* deprecation path for openshift-config certificates
* fully managed auto-rotation of signers and their leafs


tjungblu:
/hold 

before this merges without actually testing anything

PR #1247, opened 2024-04-15
API-1800: manifests: convert SecretTypeTLS secrets to kubernetes.io/tls

vrutkovs:
SecretTypeTLS is a deprecated type, which gets converted to kubernetes.io/tls when controller parses. Changing this type to supported to make controller skip convertion on initial install.

Ref: https://issues.redhat.com/browse/API-1800

### DIFF
--- a/manifests/0000_20_etcd-operator_03_secret.yaml
+++ b/manifests/0000_20_etcd-operator_03_secret.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 kind: Secret
-type: SecretTypeTLS
+type: kubernetes.io/tls
 metadata:
   annotations:
     include.release.openshift.io/self-managed-high-availability: "true"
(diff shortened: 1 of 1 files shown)


tjungblu:
hmm interesting case:

> E0416 04:14:42.256614       1 task.go:124] error running apply for secret "openshift-etcd-operator/etcd-client" (80 of 961): Secret "etcd-client" is invalid: [data[tls.crt]: Required value, data[tls.key]: Required value]

The bootstrap render should create the secret with the proper type in the right place already, so this secret _should_ actually be unnecessary.

vrutkovs:
`etcd-client` manifest has `release.openshift.io` which are not being set by the controller, so we probably still want to keep it.

Lets see if setting empty tls.crt and tls.key helps

vrutkovs:
>CVO re-creates it with empty values

`release.openshift.io/create-only: "true"` annotation would make sure it touches it just once

tjungblu:
then the apply is a rather mysterious failure :) let's wait for the CI results

tjungblu:

that's a known issue. Otherwise, seems the empty list did the trick, great find @vrutkovs 

/lgtm

tjungblu:
feel free to retitle whenever you feel it's ready

PR #1184, opened 2024-01-21
WIP:ETCD-519:Add reconcile etcd database size

Elbehery:
resolves https://issues.redhat.com/browse/ETCD-519

cc @tjungblu @dusk125 @hasbro17 @soltysh

PR #1248, opened 2024-04-15
ETCD-519:Add reconcile etcd database size

Elbehery:
resolves https://issues.redhat.com/browse/ETCD-519

cc @tjungblu @dusk125 @hasbro17 @soltysh

suppresses  https://github.com/openshift/cluster-etcd-operator/pull/1184


Elbehery:
is failing since `29 Feb 2024` , find [job history ](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-etcd-operator-master-e2e-operator-fips?buildId=1763312792833626112)

actually the job's cluster is never successful, more over the job run an upgrade and this feature is `TechPreviewNoUpgrade`

Elbehery:
`e2e-operator-fips` fails perma since `Feb 29 2024` 

see [Job History](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-etcd-operator-master-e2e-operator-fips?buildId=1764984733453783040) 

see this [PR](https://github.com/openshift/release/pull/51668/files)

Elbehery:
`e2e-gcp-qe-no-capabilities` fails perma since it was initially created 

see [Job History](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-etcd-operator-master-e2e-gcp-qe-no-capabilities?buildId=1737657169374875648)

see this [PR](https://github.com/openshift/release/pull/51667)

pkg/etcdenvvar/etcd_env.go
```diff
@@ -243,6 +243,24 @@ func getHardwareSpeedValues(envVarContext envVarContext) (map[string]string, err
 	return hwspeedhelpers.HardwareSpeedToEnvMap(speed)
 }
 
+func getEtcdDBSize(envVarContext envVarContext) (map[string]string, error) {
+	etcd, err := envVarContext.etcdLister.Get("cluster")
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve etcd CR: %v", err)
+	}
+
```
tjungblu, line 251: okay that's very nitpicky, but how about:

```suggestion

   backendQuotaGiB := int64(8)
   if etcd.Spec.BackendQuotaGiB != 0 {
      backendQuotaGiB = int64(etcd.Spec.BackendQuotaGiB)
   }
   
   return map[string]string{
		"ETCD_QUOTA_BACKEND_BYTES": GibibytesToBytesString(backendQuotaGiB),
   }, nil 
```

then the below logic can just be int64 based, so you don't have to worry about casting the values up or down anymore.

test/e2e/scaling_dbsize_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/openshift/cluster-etcd-operator/test/e2e/framework"
+
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	operatorNamespace      = "openshift-etcd-operator"
+	operatorDeploymentName = "etcd-operator"
+	operatorUpdatedImage   = "quay.io/melbeher/ceo-1900"
```
tjungblu, line 22: this needs an update

test/e2e/scaling_dbsize_test.go
```diff
@@ -0,0 +1,82 @@
...
+	etcdCR, err := opClientSet.OperatorV1().Etcds().Get(context.Background(), etcdCRName, metav1.GetOptions{})
+	require.NoError(t, err)
+
+	etcdCR.Spec.BackendQuotaGiB = dbSize32GB
+	etcdCR, err = opClientSet.OperatorV1().Etcds().Update(context.Background(), etcdCR, metav1.UpdateOptions{})
+	require.NoError(t, err)
+
+	var etcdPod v1.Pod
+	// wait for new etcd pods to be rolled out
+	err = wait.PollUntilContextTimeout(context.Background(), time.Second, 10*time.Minute, true, func(ctx context.Context) (done bool, err error) {
```
tjungblu, line 45: there's a nicer helper in library-go that's waiting for revision rollouts:

https://github.com/openshift/library-go/blob/52527b886e57ab47fb355a4d1234200047996004/test/library/pod_same_revision.go#L24

I believe we might be using it here for something, we definitely use it a lot in origin

vendor/github.com/openshift/api/operator/v1/types_etcd.go
```diff
@@ -43,6 +43,18 @@ type EtcdSpec struct {
...
 	HardwareSpeed ControlPlaneHardwareSpeed `json:"controlPlaneHardwareSpeed"`
+
+	// backendQuotaGiB sets the etcd backend storage size limit in gibibytes.
+	// The value should be an integer not less than 8 and not more than 32.
+	// When not specified, the default value is 8.
+	// +kubebuilder:default:=8
+	// +kubebuilder:validation:Minimum=8
+	// +kubebuilder:validation:Maximum=32
+	// +kubebuilder:validation:XValidation:rule="self>=oldSelf",message="etcd backendQuotaGiB may not be decreased"
+	// +openshift:enable:FeatureGate=EtcdBackendQuota
```
tjungblu, line 54: don't we have to check the feature gates as well? 

similar to how we did it for the backups:
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/starter.go#L451-L454

tjungblu (changes requested):

pkg/etcdenvvar/etcd_env.go
Elbehery: lgtm :D :D .. yes way easier to follow 👍🏽 💯

test/e2e/scaling_dbsize_test.go
Elbehery: +1

test/e2e/scaling_dbsize_test.go
Elbehery: yes we used it in `vertical-scaling` test 

in this current context, we just check the first etcd-pod that is `running`, we dont need all the revision to be rolled-out imho

but I change it no issue 👍🏽

tjungblu: any code that we don't maintain is good code ;)

Elbehery: lmfaoooo :D :D :D :D :D :D :D

vendor/github.com/openshift/api/operator/v1/types_etcd.go
Elbehery: yes i missed that, thanks 👍🏽

pkg/etcdenvvar/etcd_env.go
Elbehery: fixed 👍🏽

vendor/github.com/openshift/api/operator/v1/types_etcd.go
Elbehery: I actually dont think we need to add this in the `starter.go` since we dont not create a separate controller 

However, I have added these check in the `EnvVarController`, if the featuregate is not enabled, then we dont check the field within the CR, as it does not exist on the CRD on a cluster without this feature gate enabled 

pkg/etcdenvvar/etcd_env.go
```diff
@@ -243,6 +244,25 @@ func getHardwareSpeedValues(envVarContext envVarContext) (map[string]string, err
...
+	backendQuotaGiB := int64(8)
+	enabled, err := backendquotahelpers.IsBackendQuotaFeatureGateEnabled(envVarContext.featureGateAccessor)
```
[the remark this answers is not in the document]
tjungblu: looks good, I would just error out before setting any wrong value. 
```suggestion
	enabled, err := backendquotahelpers.IsBackendQuotaFeatureGateEnabled(envVarContext.featureGateAccessor)
	if err != nil {
		return nil, fmt.Errorf("failed check backend quota enabled: %w", err)
	}
```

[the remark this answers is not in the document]
Elbehery: but if we error, then we return `nil` envVarMap 

is this expected behaviour ?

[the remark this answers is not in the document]
tjungblu: I think so, it would not create a new revision:
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/etcdenvvar/envvarcontroller.go#L163-L165

vendor/github.com/openshift/api/operator/v1/types_etcd.go
tjungblu: sounds good!

pkg/etcdenvvar/etcd_env.go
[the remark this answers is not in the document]
Elbehery: done 👍🏽

tjungblu:
looks good to me! Let's wait for Allen/Haseeb for any other comments

pkg/etcdenvvar/etcd_env.go
```diff
@@ -243,6 +245,28 @@ func getHardwareSpeedValues(envVarContext envVarContext) (map[string]string, err
...
+	if err != nil {
+		return nil, fmt.Errorf("failed check backend quota enabled: %w", err)
+	}
+	if enabled {
+		if etcd.Spec.BackendQuotaGiB != 0 {
```
lance5890, line 260: what about 
```  
if enabled && etcd.Spec.BackendQuotaGiB != 0 {
} 
```

tjungblu: yeah sounds like a good suggestion

pkg/etcdenvvar/etcd_env.go
```diff
@@ -243,6 +245,26 @@ func getHardwareSpeedValues(envVarContext envVarContext) (map[string]string, err
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve etcd CR: %v", err)
+	}
+
+	backendQuotaGiB := int64(8)
+	enabled, err := backendquotahelpers.IsBackendQuotaFeatureGateEnabled(envVarContext.featureGateAccessor)
+	if err != nil {
+		return nil, fmt.Errorf("failed check backend quota enabled: %w", err)
+	}
+	if enabled && etcd.Spec.BackendQuotaGiB != 0 {
```
tjungblu, line 259: good thing that the testcase fails :) 

I think we need to also patch the feature gate into the release config like that:
https://github.com/openshift/release/commit/c978053a9dcc671e1ffbbf35a482df63c061b21e#diff-db0f271e35562d24e014bfd2f6c796daa79c27f26bd863aaaa303d59e6ce433bR134-R136

I'll send you a PR on slack

Elbehery: gotcha 👍🏽 

But should not we turn on the FeatureGate for the `e2e test` in this PR only ? 

otherwise will stay on for coming PRs. no ?

tjungblu: once the release config merges and is distributed to the build clusters, it should get picked up just fine

Elbehery: i think we can override the featuregate on the cluster after being launched programmatically from within the test case, and then after we go GA we can just trim it, wdyt ?

tjungblu:
I think this was a little too early to retest, but we have to wait for another hour now anyway :p

Elbehery:
featureGate is not enabled yet [link](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1248/pull-ci-openshift-cluster-etcd-operator-master-e2e-operator/1787531723764928512/artifacts/release/artifacts/release-payload-latest/0000_05_config-operator_02_featuregate.cr.yaml)

Elbehery:
this run has the featureGates on [link](featureGates: [EtcdBackendQuota=true, AutomatedEtcdBackup=true])👍🏽

Elbehery:
so this last run had the featureGates enabled [link](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1248/pull-ci-openshift-cluster-etcd-operator-master-e2e-operator/1787739873508069376/artifacts/e2e-operator/gather-extra/artifacts/featuregate.json) 

Prow [job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1248/pull-ci-openshift-cluster-etcd-operator-master-e2e-operator/1787739873508069376) 

I am creating a cluster to debug manually before re-testing 👍🏽

Elbehery:
@tjungblu  i found the issue :) 

`WaitForPodsToStabilizeOnTheSameRevision()`  does not wait for the new revision, it asserts on the current revision 

I executed the test manually against a TP cluster, and featuregates are on, yet the test failed 

debugging it, found that the assertion takes place on the old revision, not the new revision 

I am raising a fix now 👍🏽

test/e2e/scaling_dbsize_test.go
```diff
@@ -0,0 +1,66 @@
...
+	etcdPodsClient := clientSet.CoreV1Interface.Pods(operandNameSpace)
+
+	etcdCR, err := opClientSet.OperatorV1().Etcds().Get(context.Background(), etcdCRName, metav1.GetOptions{})
+	require.NoError(t, err)
+
+	etcdCR.Spec.BackendQuotaGiB = dbSize32GB
+	etcdCR, err = opClientSet.OperatorV1().Etcds().Update(context.Background(), etcdCR, metav1.UpdateOptions{})
+	require.NoError(t, err)
+
+	err = library.WaitForPodsToStabilizeOnTheSameRevision(t, etcdPodsClient, etcdLabelSelector, 3, 10*time.Second, 5*time.Second, 30*time.Minute)
```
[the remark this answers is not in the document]
Elbehery: thanks @lance5890 

yes I had previous checks that was waiting for the new condition 

```
// wait for new etcd pods to be rolled out
	err = wait.PollUntilContextTimeout(context.Background(), time.Second, 10*time.Minute, true, func(ctx context.Context) (done bool, err error) {
		etcdPods, err := clientSet.Pods(operandNameSpace).List(context.Background(), metav1.ListOptions{LabelSelector: etcdLabelSelector})
		if err != nil {
			if apierrors.IsNotFound(err) {
				return false, nil
			}
			return false, err
		}

		for _, pod := range etcdPods.Items {
			if pod.Status.Phase == v1.PodRunning && time.Now().Sub(pod.CreationTimestamp.Time) < 1*time.Minute {
				return true, nil
			}
		}
		return false, nil
	})
	require.NoError(t, err)
```

but using `library-go` is always best .. 

the issue is that the helpers from library-go does not wait for new-revision, i am adding a helper now that does exactly that 👍🏽

Elbehery:
here you go https://github.com/openshift/library-go/pull/1733 

cc @tjungblu @lance5890

test/e2e/scaling_dbsize_test.go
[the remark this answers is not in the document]
tjungblu: I'm not sure why we need another helper for this. Just increase the poll interval and the success threshold.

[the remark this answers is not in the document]
Elbehery: The helper makes sure that a new revision has been rolled out before assertion 

the current e2e failure is due to asserting on the old etcd-pods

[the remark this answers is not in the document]
tjungblu: it is entirely unnecessary, if you make the poll timeout 1 minute, have it finish for five successful runs it will still pass your test. Why does it take so long to update the env vars?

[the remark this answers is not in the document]
Elbehery: ok let me try 👍🏽

test/e2e/scaling_dbsize_test.go
```diff
@@ -0,0 +1,66 @@
...
+	etcdPodsClient := clientSet.CoreV1Interface.Pods(operandNameSpace)
+
+	etcdCR, err := opClientSet.OperatorV1().Etcds().Get(context.Background(), etcdCRName, metav1.GetOptions{})
+	require.NoError(t, err)
+
+	etcdCR.Spec.BackendQuotaGiB = dbSize32GB
+	etcdCR, err = opClientSet.OperatorV1().Etcds().Update(context.Background(), etcdCR, metav1.UpdateOptions{})
+	require.NoError(t, err)
+
+	err = library.WaitForPodsToStabilizeOnTheSameRevision(t, etcdPodsClient, etcdLabelSelector, 5, 1*time.Minute, 5*time.Second, 30*time.Minute)
```
Elbehery, line 39: @tjungblu like this?

tjungblu: yes, even though the 5s polling seems a bit too crazy, but it's alright

PR #1262, opened 2024-05-14
NO-JIRA:check the cluster is healthy or not after getting env var, to ensure it is safe to rollout

lance5890:
1. similar to the https://github.com/openshift/cluster-etcd-operator/pull/1239, I found the ceo could also rollout when remove one master node
2. the targetconfigcontroller  should check the cluster is safe or not when get env vars, in case the env vas have changed, but the cluster is not healthy yet


tjungblu:
ah well, just got an email that the label is being lifted. Thanks a ton @lance5890 - I check back later to see if this got merged

PR #1264, opened 2024-05-23
ETCD-620: Update api vendor libraries to get GA'ed tuning profiles

dusk125:
`go get github.com/openshift/api@e278985`


dusk125:
/retitle ETCD-620: Update api vendor libraries to get GA'ed tuning profiles

PR #1269, opened 2024-05-29
ETCD-607: gate leaf cert generation

tjungblu:
This implements a gate to avoid triggering leaf cert generation in the
same static pod revision as an update to the signer certificates (and
their respective bundles).

This PR also removes the openshift-config dependency on signer
certificates. With 4.17 we should be safe to remove any reference, as
the newly generated signer certificate in openshift-etcd should already
be distributed and ready for consumption.

In plain words: an upgrade to 4.17 will automatically rotate your etcd
signer certificates for the first time and every refresh period happily
ever after.

### DIFF
--- a/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
+++ b/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
@@ -5,11 +5,14 @@ import (
 	"crypto/x509"
 	"fmt"
 	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/library-go/pkg/operator/bootstrap"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	corev1informers "k8s.io/client-go/informers/core/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/component-base/metrics"
+	"k8s.io/klog/v2"
 	"strconv"
 	"strings"
 	"time"
@@ -70,7 +73,11 @@ type EtcdCertSignerController struct {
 	secretLister       corev1listers.SecretLister
 	secretClient       corev1client.SecretsGetter
 	configMapClient    corev1client.ConfigMapsGetter
+	configmapLister    corev1listers.ConfigMapLister
 	quorumChecker      ceohelpers.QuorumChecker
+
+	// when true we skip all checks related to the rollout of static pods, this is used in render
+	forceSkipRollout bool
 
 	certConfig *certConfig
 
@@ -93,6 +100,7 @@ func NewEtcdCertSignerController(
 	eventRecorder events.Recorder,
 	quorumChecker ceohelpers.QuorumChecker,
 	metricsRegistry metrics.KubeRegistry,
+	forceSkipRollout bool,
 ) factory.Controller {
 	eventRecorder = eventRecorder.WithComponentSuffix("etcd-cert-signer-controller")
 	cmInformer := kubeInformers.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps()
@@ -144,18 +152,21 @@ func NewEtcdCertSignerController(
 	metricsRegistry.MustRegister(signerExpirationGauge)
 
 	c := &EtcdCertSignerController{
-		eventRecorder:         eventRecorder,
-		kubeClient:            kubeClient,
-		operatorClient:        operatorClient,
-		masterNodeLister:      masterNodeLister,
-		masterNodeSelector:    masterNodeSelector,
-		secretInformer:        secretInformer,
-		secretLister:          secretLister,
-		secretClient:          secretClient,
-		configMapClient:       cmGetter,
+		eventRecorder:      eventRecorder,
+		kubeClient:         kubeClient,
+		operatorClient:     operatorClient,
+		masterNodeLister:   masterNodeLister,
+		masterNodeSelector: masterNodeSelector,
+		secretInformer:     secretInformer,
+		secretLister:       secretLister,
+		secretClient:       secretClient,
+		configMapClient:    cmGetter,
+		// this one can go through the informers, it's only used for bootstrap checks
+		configmapLister:       kubeInformers.InformersFor(operatorclient.KubeSystemNamespace).Core().V1().ConfigMaps().Lister(),
 		quorumChecker:         quorumChecker,
 		certConfig:            certCfg,
 		signerExpirationGauge: signerExpirationGauge,
+		forceSkipRollout:      forceSkipRollout,
 	}
 
 	syncer := health.NewDefaultCheckingSyncWrapper(c.sync)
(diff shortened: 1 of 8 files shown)


pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -171,16 +177,18 @@ func NewEtcdCertSignerController(
 }
 
 func (c *EtcdCertSignerController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
-	safe, err := c.quorumChecker.IsSafeToUpdateRevision()
-	if err != nil {
-		return fmt.Errorf("EtcdCertSignerController can't evaluate whether quorum is safe: %w", err)
+	_, currentStatus, _, err := c.operatorClient.GetStaticPodOperatorStateWithQuorum(ctx)
```
lance5890, line 180: what if `err != nil`

tjungblu: yep, good catch!

tjungblu:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.17-upgrade-from-stable-4.16-e2e-aws-ovn-upgrade 10

tjungblu:
/test e2e-metal-assisted

/payload-aggregate periodic-ci-openshift-release-master-nightly-4.17-e2e-metal-ovn-assisted 10

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -202,41 +234,35 @@ func (c *EtcdCertSignerController) sync(ctx context.Context, syncCtx factory.Syn
...
-	// Unfortunately, we can't make use of the certrotation.RotatedSigningCASecret here because it would immediately try to rotate the existing signer.
-	signerCaPair, err := tlshelpers.ReadConfigSignerCert(ctx, c.secretClient)
-	if err != nil {
-		return err
-	}
-
-	c.reportExpirationMetric(signerCaPair, "signer-ca")
-
-	// TODO(thomas): we need to transition that new signer as a replacement for the above - today we only bundle it
-	newSignerCaPair, _, err := c.certConfig.signerCert.EnsureSigningCertKeyPair(ctx)
```
hasbro17, line 218: Please bear with me as I try to walk through the rollout sequence to make sure I have this right :) and cut me off if I'm assuming something incorrect.

### The old way:
We had `openshift-config/etcd-signer` secret as the `old-CA` and `openshift-etcd/etcd-signer` secret as the `new-CA` both of which were bundled on all etcd pods but all server certs were only signed by the `old-CA`. 
And we updated and overwrote `openshift-etcd/etcd-signer`=>`openshift-config/etcd-signer` to regenerate the leaf certs. That way the first pod that rolled out a the new revision (with the `new-CA` signed server cert) would still be recognized by the older revision etcd members (since they already had the `new-CA` trust bundled)

### The new way:
With this change we only have `openshift-etcd/etcd-signer` as the definitive CA secret right?
So call that the `old-CA`.
If we now delete `openshift-etcd/etcd-signer` and have it get recreated with `new-CA` (e.g as Revision 2), then we won't regenerate the leaf certs until we first have all etcd pods bundled with `new-CA` at Revision 2 right?

So at revision 2 we have all etcd members trust bundled via `etcd-all-bundles` with only `new-CA` then.
But aren't the server certs still signed by the `old-CA` that is no longer in the trust bundle anymore?

So IIUC it's effectively `--cert-file=<old-CA signed server.crt>` and `--trusted-ca-file=<new-CA.crt>`.

But seems like this is working in CI so trying to figure out what I'm missing in how the cluster is working before the leaf certs are re-generated.

tjungblu: > So at revision 2 we have all etcd members trust bundled via etcd-all-bundles with only new-CA then.
> But aren't the server certs still signed by the old-CA that is no longer in the trust bundle anymore?

They are! otherwise we would've split the cluster, the bundle itself is stateful, so if we present it a new CA it would just append it. Revoking the old CA would be either a manual step, or another control loop to delete the configmap once in a while.

tjungblu:
/test e2e-aws-ovn-single-node
/test e2e-metal-assisted
/test e2e-metal-ipi-ovn-ipv6
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.17-e2e-metal-ovn-assisted 10

tjungblu:
/test e2e-aws-ovn-single-node
/test e2e-metal-assisted
/test e2e-metal-ipi-ovn-ipv6
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.17-e2e-metal-ovn-assisted 10

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -202,41 +241,35 @@ func (c *EtcdCertSignerController) sync(ctx context.Context, syncCtx factory.Syn
...
-		return fmt.Errorf("error on ensuring metrics-signer cert: %w", err)
+		return fmt.Errorf("error on ensuring bundles: %w", err)
 	}
 
-	signerBundle, metricsSignerBundle, err := c.ensureBundles(ctx, recorder,
-		[]*crypto.CA{signerCaPair, newSignerCaPair}, []*crypto.CA{metricsSignerCaPair, newMetricsSignerCaPair})
-	if err != nil {
-		return fmt.Errorf("error on ensuring bundles: %w", err)
+	if !forceSkipRollout {
+		if rolloutTriggered {
```
hasbro17, line 264: Just confirming that we'll skip here as we first trigger the rollout:
`skipping EtcdCertSignerController leaf cert generation as revision rollout has been triggered`
and then the controller will loop for a couple mins on L213 until the revision stabilizes:
`skipping EtcdCertSignerController can't get current revision. Err=revision rollout in progress, can't establish current revision`

Basically pause during rollouts.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -202,41 +241,35 @@ func (c *EtcdCertSignerController) sync(ctx context.Context, syncCtx factory.Syn
...
+			klog.Infof("skipping EtcdCertSignerController leaf cert generation as revision rollout has been triggered")
+			return nil
+		}
+
+		if currentRevision <= lastRotationRevision {
+			klog.Infof("skipping EtcdCertSignerController leaf cert generation as safe revision is not yet achieved, currently at %d - rotation happend at %d", currentRevision, lastRotationRevision)
```
hasbro17, line 270: Isn't `currentRevision` the latest available revision that all nodes have settled on (i.e no rollout in progress).
In that case, if `currentRevision = lastRotationRevision` then isn't it safe to proceed.

But I guess setting a new `lastRotationRevision` will always trigger a new `currentRevision` so we always expect `currentRevision>lastRotationRevision` if the rollout has stabilized.

hasbro17: This is a useful log line but I'm wondering if its ever even printed. Because we already skip during rollouts at L213 and see:
`skipping EtcdCertSignerController can't get current revision. Err=revision rollout in progress, can't establish current revision`

and once that's done `currentRevision>lastRotationRevision` so we're good to go again.

I should probably check the CI runs and stop speculating :)

hasbro17 (approved):
Just a couple of questions but mostly lgtm

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -358,20 +391,23 @@ func (c *EtcdCertSignerController) ensureBundles(ctx context.Context, recorder e
...
+		return nil, nil, rolloutTriggered, fmt.Errorf("could not apply bundle configmap: %w", err)
 	}
 
-	if changed {
-		// TODO(thomas): set the rollout bundle revision in the CEO status
-		// TODO(thomas): we need to notify the outside caller that it's not safe to rotate the leafs just yet
+	if rolloutTriggered {
+		// note that even when this operation fails, the triggered revision will not generate leaf certificates
+		// this is guarded by the next sync loops first check against a rolling out revision in ceohelpers.CurrentRevision
+		if err = c.updateCertRevision(ctx, currentRevision); err != nil {
```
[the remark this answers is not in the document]
tjungblu: The bundle is updated before the operator status and should cause revision 7, yes. So when it fails to persist the next sync invocation will _likely_ go into the static pod rollout detection. Once that revision is rolled out, it _should_ be safe to generate the leaf certs. 

In reality, I'm sure this is not as happy path, we know from the AI failures that sometimes revisions just can stomp right in between. 

I have an alternative solution where we store this as part of the bundle to avoid this non-transactional double object update, but that would also require library-go changes, so I held off on this for now.  The leaf certs also have the signer certificate and resource version in their annotation, so we could also try to match from the side.

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
tjungblu: correct

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
tjungblu: it is printed, but for a very short period of time, basically until the watch cache updates the CEO/etcd CRD

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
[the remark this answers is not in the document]
tjungblu: sorry, we own our bundled bundle configmap now :) so we can scrap that part of ETCD-633 and add that annotation on the apply.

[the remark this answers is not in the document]
tjungblu: checkout https://github.com/openshift/cluster-etcd-operator/pull/1275 - I'll run some tests on the other PR, so we can see whether this works better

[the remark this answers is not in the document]
tjungblu: works just as well, so we can either merge #1275 directly or have the commit as a separated PR, let me know what you prefer :)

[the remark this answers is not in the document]
hasbro17: Thanks for the walkthrough!
And yep just finished looking at https://github.com/openshift/cluster-etcd-operator/pull/1275 and I think that's much cleaner with the revision being stored on the configmap annotation rather than a separate object. So let's merge that in favor of this one.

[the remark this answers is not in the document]
tjungblu: lovely, then let's close :)

tjungblu:
/close 

taking #1275

PR #1275, opened 2024-06-14
ETCD-607: gate leaf cert generation (with bundle revision)

tjungblu:
/hold

this is a slight improvement over #1269 by not having a two-phase commit issue on storing the revision

### DIFF
--- a/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
+++ b/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/component-base/metrics"
 	"k8s.io/klog/v2"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
@@ -35,6 +36,7 @@ import (
 	"github.com/openshift/cluster-etcd-operator/pkg/tlshelpers"
 )
 
+const BundleRolloutRevisionAnnotation = "openshift.io/ceo-bundle-rollout-revision"
 const signerExpirationMetricName = "openshift_etcd_operator_signer_expiration_days"
 
 type certConfig struct {
@@ -214,7 +216,7 @@ func (c *EtcdCertSignerController) sync(ctx context.Context, syncCtx factory.Syn
 		return nil
 	}
 
-	rotationRevision, err := getCertRotationRevision(*currentStatus)
+	rotationRevision, err := c.getCertRotationRevision(ctx)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


tjungblu:
/payload 4.17 nightly blocking
/test e2e-aws-ovn-single-node
/test e2e-metal-assisted
/test e2e-metal-ipi-ovn-ipv6
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.17-e2e-metal-ovn-assisted 10

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
```diff
@@ -171,16 +184,44 @@ func NewEtcdCertSignerController(
...
+		}
+		return nil
+	}
+
+	_, currentStatus, _, err := c.operatorClient.GetStaticPodOperatorStateWithQuorum(ctx)
+	if err != nil || currentStatus == nil {
+		return fmt.Errorf("skipping EtcdCertSignerController can't get current status: %w", err)
+	}
+
+	currentRevision, err := ceohelpers.CurrentRevision(*currentStatus)
```
tjungblu, line 212: we might be able to relax this a bit and return the MIN(all current revisions) and keep the remainder of the controller running. That could save us the `hasNodeCertDiff` call above, but we can also improve this later if we need to.

tjungblu:
/payload 4.17 nightly blocking
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.17-e2e-metal-ovn-assisted 10

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
hasbro17: Ah okay, so we would no longer block blindly if there is a revision rollout. That let's us keep the controller going during vertical scaling. But we get the MIN revision that's currently lagging and only compare that against the `openshift.io/ceo-bundle-rollout-revision` value only gate the leaf certs if the MIN revision hasn't moved past the last rotation revision.

Actually now that I wrote that out, that means the controller would still block from generating leaf certs for the new node that gets added. So maybe I don't quite follow how that would help us replace `hasNodeCertDiff()` 😓

hasbro17: In any case, if you want to target this as later improvement, I'm good with the `hasNodeCertDiff()` for now. Or if you make the change, I can try and see how that works too.

hasbro17 (approved):
LGTM on having the last rotation revision stored on top of the bundle.

tjungblu:
Thanks! 
I'll wait for the aggregation results once more, the job has to retest anyway :) 

/hold

pkg/operator/etcdcertsigner/etcdcertsignercontroller.go
tjungblu: let's test this in a separate PR, otherwise we won't get anything done in here :mechanic:  
I'll have to still debug Vadim's cert rotation workflow, so we can certainly expect one or two more fixes coming out of that :)

tjungblu:
/hold cancel

all good, @dgoodwin / @neisw fyi we're going to merge this - which I hope will not cause any further issues this time around.

dgoodwin:
Thanks for investigating and fixing+testing thoroughly, sounds good to us.

PR #1279, opened 2024-06-17
ETCD-593: Enable experimental-stop-grpc-service-on-defrag

Elbehery:
This PR enable `--experimental-stop-grpc-service-on-defrag` on OCP/Etcd. 


hasbro17:
/hold

We need to test and validate whether this flag is actually stopping grpc on the degraded member or if we need to enable something on the etcd client side as well.
https://github.com/kubernetes/kubernetes/pull/124787

PR #1283, opened 2024-06-20
NO-JIRA: fix false positive bootstrap in cert signer controller

tjungblu:
on CEO restart the informer is sometimes not synchronized, which causes premature leaf cert generation as the bootstrap thinks it's not done yet.


tjungblu:
/hold cancel

/assign @Elbehery 
can you PTAL?

tjungblu:

unrelated OLM failure

PR #1190, opened 2024-01-30
OCPBUGS-27959: fix panic in health check timeouts

tjungblu:

PR #1286, opened 2024-06-28
OCPBUGS-36301: parallelize member health checks

AlexVulaj:
Currently, member health is checked in serial with a 30s timeout per member. 3 out of 4 GetMemberHealth callers had their own default 30s timeout as well for the entire process. Because of this, a slow check on one member could exhaust the timeout for the entire GetMemberHealth function, and thus cause later-checked members to report as unhealthy even though they were fine.

With this commit, I am dropping the internal 30s timeout from GetMemberHealth, and instead letting the caller set the timeout. Also, the code now checks the health of all members in parallel. This will prevent a single slow member from affecting the health reporting of other members.

I also added a timeout to the context used in IsMemberHealthy which calls GetMemberHealth. Neither Trevor nor I were sure why a default timeout wasn't present there, though one was present in all other call sites.

### DIFF
--- a/pkg/etcdcli/health.go
+++ b/pkg/etcdcli/health.go
@@ -43,34 +43,21 @@ type memberHealth []healthCheck
 
 func GetMemberHealth(ctx context.Context, etcdMembers []*etcdserverpb.Member) memberHealth {
 	memberHealth := memberHealth{}
+	resChan := make(chan healthCheck, 1)
 	for _, member := range etcdMembers {
 		if !HasStarted(member) {
 			memberHealth = append(memberHealth, healthCheck{Member: member, Healthy: false})
 			continue
 		}
 
-		const defaultTimeout = 30 * time.Second
-		resChan := make(chan healthCheck, 1)
 		go func() {
-			ctxTimeout, cancel := context.WithTimeout(ctx, defaultTimeout)
-			defer cancel()
-
-			resChan <- checkSingleMemberHealth(ctxTimeout, member)
-			// closing here to avoid late replies to panic on resChan,
-			// the result will be considered a timeout anyway
-			close(resChan)
+			resChan <- checkSingleMemberHealth(ctx, member)
 		}()
+	}
 
-		select {
-		case res := <-resChan:
-			memberHealth = append(memberHealth, res)
-		case <-time.After(defaultTimeout):
-			memberHealth = append(memberHealth, healthCheck{
-				Member:  member,
-				Healthy: false,
-				Error: fmt.Errorf("30s timeout waiting for member %s to respond to health check",
-					member.Name)})
-		}
+	for len(memberHealth) < len(etcdMembers) {
+		res := <-resChan
+		memberHealth = append(memberHealth, res)
 	}
 
 	// Purge any unknown members from the raft term metrics collector.
--- a/pkg/etcdcli/etcdcli.go
+++ b/pkg/etcdcli/etcdcli.go
@@ -416,6 +416,8 @@ func (g *etcdClientGetter) IsMemberHealthy(ctx context.Context, member *etcdserv
 	if member == nil {
 		return false, fmt.Errorf("member can not be nil")
 	}
+	ctx, cancel := context.WithTimeout(ctx, DefaultClientTimeout)
+	defer cancel()
 	memberHealth := GetMemberHealth(ctx, []*etcdserverpb.Member{member})
 	if len(memberHealth) == 0 {
 		return false, fmt.Errorf("member health check failed")


pkg/etcdcli/health.go
```diff
@@ -43,34 +43,21 @@ type memberHealth []healthCheck
...
-		resChan := make(chan healthCheck, 1)
 		go func() {
-			ctxTimeout, cancel := context.WithTimeout(ctx, defaultTimeout)
-			defer cancel()
-
-			resChan <- checkSingleMemberHealth(ctxTimeout, member)
-			// closing here to avoid late replies to panic on resChan,
-			// the result will be considered a timeout anyway
-			close(resChan)
+			resChan <- checkSingleMemberHealth(ctx, member)
```
lance5890, line 54: - what if the caller has not set the ctx timeout,  please see https://github.com/openshift/cluster-etcd-operator/blob/0a00f2ac2d578114cea7164dbdba4c98ce06c69e/pkg/etcdcli/etcdcli.go#L415-L427
-  and in extreme cases, this call would be blocked until the linux system releases this abnormal connection(which maybe 15 minutes)
- maybe u could set the caller's ctx timeout as 30 * time.Second

AlexVulaj: Hey @lance5890 , unless I'm misunderstanding your comment here my PR updates that specific call site to have a timeout of 30s as well. It was the only one I'm the entire repo I found that didn't.

lance5890: o, my mistake,  it looks good

geliu2016 (approved):
/label cherry-pick-approved

pkg/etcdcli/health.go
```diff
@@ -43,34 +43,21 @@ type memberHealth []healthCheck
...
-		const defaultTimeout = 30 * time.Second
-		resChan := make(chan healthCheck, 1)
 		go func() {
-			ctxTimeout, cancel := context.WithTimeout(ctx, defaultTimeout)
-			defer cancel()
-
-			resChan <- checkSingleMemberHealth(ctxTimeout, member)
-			// closing here to avoid late replies to panic on resChan,
-			// the result will be considered a timeout anyway
-			close(resChan)
```
tjungblu, line 61: we kinda still have to close this channel, don't we? 

fyi, that's a panic I've fixed recently:
https://issues.redhat.com//browse/OCPBUGS-27959

https://github.com/openshift/cluster-etcd-operator/pull/1190

wking: Looking at the pre-#1190 [code][1], the panics were from timelines like:

1. `checkSingleMemberHealth` goroutine launched with its own 30s `Context` duration.
2. `select` waited on a result from `resChan` _or_ a new 30s `time.After`.
3. When the `select` `time.After` won, it appended a `30s timeout waiting for member...` to `memberHealth`, and `close`d `resChan`.
4. A millisecond or two later, `checkSingleMemberHealth` would hit its 30s `Context` timeout in [the `Get` call][2], create [its own `health check failed: ...` result][3], and [push it into `resChan`][4].
5. But `resChan` was closed in step 3!  Panic!

With #1190, you dropped the `close` from step 3, and moved it to step 4, so no more panic.

But from [Go's _Range and Close_ tour][5]:

> Channels aren't like files; you don't usually need to close them. Closing is only necessary when the receiver must be told there are no more values coming, such as to terminate a range loop.

And with _this_ pull, we no longer have the receiver-side `select` or timeout.  With this pull, the receiver will block until it has a result back from each launched `checkSingleMemberHealth` goroutine, and it's up to those goroutines to respect the `Context` timeout.  So there is no chance of the `GetMemberHealth` `close`-ing the channel before a `checkSingleMemberHealth` goroutine writes, because we no longer have an explicit `close` at all, and the channel is just garbage-collected as it goes out of scope, like all local Go variables.

[1]: https://github.com/openshift/cluster-etcd-operator/blob/1e1970f23b7a588953ad47617108a683ecc4f30f/pkg/etcdcli/health.go#L61-L72
[2]: https://github.com/openshift/cluster-etcd-operator/blob/1e1970f23b7a588953ad47617108a683ecc4f30f/pkg/etcdcli/health.go#L121-L124
[3]: https://github.com/openshift/cluster-etcd-operator/blob/1e1970f23b7a588953ad47617108a683ecc4f30f/pkg/etcdcli/health.go#L136C26-L136C45
[4]: https://github.com/openshift/cluster-etcd-operator/blob/1e1970f23b7a588953ad47617108a683ecc4f30f/pkg/etcdcli/health.go#L58
[5]: https://go.dev/tour/concurrency/4

tjungblu: Thanks for the explanation, indeed kicking the select out and just looping all the values is enough :)

tjungblu:
/cherry-pick release-4.16 release-4.15 release-4.14 release-4.13 release-4.12

wking:
I think the `e2e-aws-ovn-etcd-scaling` failures are unrelated to this change, and are instead a combination of [OCPBUGS-36462](https://issues.redhat.com/browse/OCPBUGS-36462) (which I've just opened) and a need to refactor [the `etcd is able to vertically scale up and down with a single node` test case][1] to stop assuming the ControlPlaneMachineSet `status.readyReplicas` will hit 4, and instead do something else to check that the roll/recovery completed.

[1]: https://github.com/openshift/origin/blob/3deedee4ae147a03afdc3d4ba86bc175bc6fc5a8/test/extended/etcd/vertical_scaling.go#L34-L39

wking:
> ...to stop assuming the ControlPlaneMachineSet `status.readyReplicas` will hit 4...

This is now tracked in [ETCD-637](https://issues.redhat.com/browse/ETCD-637).  In the meantime, possibly worth an `/override ci/prow/e2e-aws-ovn-etcd-scaling` here?  Or keep launching retests until we get lucky?  Or wait for the CPMS and etcd work to green up the test?

tjungblu:

no doubt :)

tjungblu:
unrelated failure 

PR #1287, opened 2024-06-30
ETCD-636: add automated backup sidecar

Elbehery:
This PR add an etcd backup sidecar container to the etcd pod manifest. 

The container copies the snapshot state upon changes from the etcd data dir into backup dir. 

fixes https://issues.redhat.com/browse/ETCD-636


bindata/etcd/pod.yaml
```diff
@@ -129,6 +129,33 @@ ${COMPUTED_ENV_VARS}
...
+    image: ${IMAGE}
+    imagePullPolicy: IfNotPresent
+    terminationMessagePolicy: FallbackToLogsOnError
+    command:
+      - /bin/sh
+      - -c
+      - |
+        #!/bin/sh
+        set -euo pipefail        
+        cp --verbose --recursive --preserve --reflink=auto /var/lib/etcd/ /var/backup/etcd
```
tjungblu, line 143: I suggest you make this a command, so we have some go code we can also test properly. Eg in here: https://github.com/openshift/cluster-etcd-operator/tree/master/pkg/cmd

Didn't we want to take the snapshot with etcdctl for starters? and there's no retention and schedule either

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,88 @@
...
+	if !b.snapshotExist {
+		if err := backup(&b.backupOptions); err != nil {
+			klog.Errorf("run: backup failed: [%v]", err)
+		}
+		b.snapshotExist = true
+		klog.Infof("config-dir is: %s", b.configDir)
+		return nil
+	}
+
+	if err := b.copySnapshot(); err != nil {
```
tjungblu, line 65: why do we need to copy the snapshot? can't we just directly put it where it should be?

pkg/operator/periodicbackupcontroller/periodicbackupcontroller.go
```diff
@@ -79,6 +79,11 @@ func (c *PeriodicBackupController) sync(ctx context.Context, _ factory.SyncConte
 	if err != nil {
 		return fmt.Errorf("PeriodicBackupController could not list backup CRDs, error was: %w", err)
 	}
+	// skip if default backup enabled
```
Elbehery, line 82: @tjungblu  sooo :D

tjungblu: you're skipping the whole controller. 

What I had in mind was something like:

```suggestion
	// ignore the reconciliation of the default backup
	backups.Items = slices.DeleteFunc(backups.Items, func(b backupv1alpha1.Backup) bool {
		return b.Name == "default"
	})
```

Elbehery: achsooo .. but this way we could have other `Backup` CRs running alongside the default, correct ?

tjungblu: sure, why not?

Elbehery: yes its gr8 idea actually, updated 👍🏽

Elbehery:
failures are due to authentication to the image registry 


```
{  release "release-latest" failed: could not watch pod: the pod ci-op-0zcbppy9/release-latest failed after 1m13s (failed containers: release): ContainerFailed one or more containers exited

Container release exited with code 1, reason Error
---
29a0873cc59feb1f2f00fc818420330b4210e10021601e3f3c63ba87cc790e5e oc-mirror
info: Loading sha256:fcad3a8a4a17cbc13e8eaaa8503b2ec3c5d4c006ddc39cc798daaba62d8691d9 operator-lifecycle-manager
info: Loading sha256:2dd0da125e0e23b5530863690b8a629d524d1a393b63b6c3ef27cb46f7a0f961 openstack-cluster-api-controllers
info: Loading sha256:fd1e3fb553482d31c4831b845aaf2ef68ea4976cb595b40f14e3dc014e74b2d1 operator-marketplace
info: Loading sha256:88e955c3aaf50a4c75f61ed34542354389386463c61a1931dabac61bacb6d056 operator-framework-tools
info: Loading sha256:dd5d0691c5135ebc8173c5b3486c4fd7b187b65b0375e557bfa35115c7660e74 service-ca-operator
info: Loading sha256:85d166c140588335aa7712be55565931d2f4eccbb62e0031e43f8f4976e3633b tests
info: Loading sha256:5a3b73c0132212ad41a9e40d3c09df29e67f56b1757bdc018d1fe60df7493c1a vsphere-cluster-api-controllers
info: Loading sha256:0f036c80d66513d8125850a11f4d3117888af17b6f2037eed07148a83875a91f vsphere-problem-detector
info: Included 190 images from 72 input operators into the release
error: failed to push image registry.build03.ci.openshift.org/ci-op-0zcbppy9/release:latest: uploading the source layer sha256:7a4643f5f2a50088993f8d8f43a8f86bc0c497a96e1323a5a5eaf051bfa8dcc8 failed: Patch "https://registry.build03.ci.openshift.org/v2/ci-op-0zcbppy9/release/blobs/uploads/175ca850-94e9-4f12-96bd-2215443abd65?_state=OK6IMXDu0oEqW_1kApl-CScd7gplm-bE0vJwspCQk7t7Ik5hbWUiOiJjaS1vcC0wemNicHB5OS9yZWxlYXNlIiwiVVVJRCI6IjE3NWNhODUwLTk0ZTktNGYxMi05NmJkLTIyMTU0NDNhYmQ2NSIsIk9mZnNldCI6MCwiU3RhcnRlZEF0IjoiMjAyNC0wNy0wOVQxMjoyOTo1MC45ODg4MTkyODhaIn0%3D": http2: Transport: cannot retry err [stream error: stream ID 2349; REFUSED_STREAM; received from peer] after Request.Body was written; define Request.GetBody to avoid this error
{"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:84","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2024-07-09T12:30:21Z"}
---}
```

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,246 @@
...
+	backupOptions
+}
+
+func NewBackupNoConfigCommand(errOut io.Writer) *cobra.Command {
+	backupNoConf := &backupNoConfig{
+		snapshotExist: false,
+		backupOptions: backupOptions{errOut: errOut},
+	}
+	cmd := &cobra.Command{
+		Use:   "cluster-backup-no-config",
```
tjungblu, line 40: let's call it `backup-server`, see below as to why

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,246 @@
...
+		return err
+	}
+
+	if err = b.extractBackupSpecs(backupsClient); err != nil {
+		return err
+	}
+
+	var errs []error
+	go func() {
+		err := b.scheduleBackup()
```
tjungblu, line 80: the cronjob.Start() needs to block this command, otherwise you will have an infinitely restarting sidecar that copies backups all the time

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,246 @@
...
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}()
+	go func() {
+		err := b.scheduleBackupPrune()
```
tjungblu, line 86: can we make the prune running before or after the actual backup running? Remember we had some issues around race conditions? ;)

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,246 @@
...
+		return lErr
+	}
+
+	idx := slices.IndexFunc(backups.Items, func(backup backupv1alpha1.Backup) bool {
+		return backup.Name == "default"
+	})
+	if idx == -1 {
+		sErr := fmt.Errorf("could not find default backup CR, found [%v]", backups.Items)
+		klog.Error(sErr)
+		return sErr
```
tjungblu, line 136: that's not an error, if it's not there the user likely has it removed for a reason. We have two options here:
* the sidecar is patched out by the operator when this CRD is not available
* you keep this cron server running without any task to schedule, but you need to have some reconciliation loop to check whether the default CRD comes back eventually

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,246 @@
...
+		gcron.NewTask(b.backup()),
+	); err != nil {
+		return err
+	}
+
+	s.Start()
+	return nil
+}
+
+func (b *backupNoConfig) copySnapshot() error {
```
tjungblu, line 186: can we please remove this? I think this is going to confuse people more than this is actually helping

pkg/cmd/backuprestore/backupnoconfig.go
tjungblu: if we go with the first option, then we need to patch the CRD client out of this command entirely and use configuration args to pass the CRD (similar to the backup prune command). Otherwise we're going to have issues with race conditions on static pod rollouts when people create+delete the default CRD, which will send this sidecar crash looping.

pkg/cmd/backuprestore/backupnoconfig.go
Elbehery: yes i remember them very well :D :D

pkg/cmd/backuprestore/backupnoconfig.go
Elbehery: so if we go with second option, that is we return non-err above, why do we need to a `sync()`, as this is a cmd ? 

iiuc,  it does check CRD upon every invocation, or ?

pkg/cmd/backuprestore/backupnoconfig.go
Elbehery: done 👍🏽

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,246 @@
...
+	src := "/var/lib/etcd/member/snap"
+	dst := "/var/backup/etcd/snap"
+	if _, err := exec.Command("cp", getCpArgs(src, dst)...).CombinedOutput(); err != nil {
+		klog.Errorf("run: backup failed: [%v]", err)
+	}
+
+	return nil
+}
+
+func (b *backupNoConfig) pruneBackups() error {
```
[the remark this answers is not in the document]
Elbehery: i tried to run the cmd directly but i could not find a way to run a `cobra` cmd from another cmd !!

pkg/cmd/backuprestore/backupnoconfig.go
Elbehery: the call is non blocking, and there is no return from the `task()` to know when to unblock 

maybe i find another package better

Elbehery: so i changed the approach 

- one scheduler being created upon `Run()`
- the scheduler runs only one job `backup()`
- the `prune()` is being invoked upon successful completion of the Job (i.e. backup())
- the scheduler configs is to :
   - run only one job at a time
   - the job run is singleton
   - the job is never being re-queued

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (b *backupNoConfig) Run() error {
+	backupsClient, err := b.getBackupClient()
+	if err != nil {
+		return err
+	}
+
+	b.scheduler, _ = gcron.NewScheduler(
+		gcron.WithLimitConcurrentJobs(1, gcron.LimitModeWait),
```
Elbehery, line 73: ptal @tjungblu

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,189 @@
...
+	if _, err := b.scheduler.NewJob(
+		gcron.CronJob(
+			b.schedule,
+			false,
+		),
+		gcron.NewTask(b.backup()),
+		gcron.WithEventListeners(
+			gcron.AfterJobRuns(
+				func(jobID uuid.UUID, jobName string) {
+					b.pruneBackups()
```
Elbehery, line 153: ptal @tjungblu

pkg/cmd/backuprestore/backupnoconfig.go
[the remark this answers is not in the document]
Elbehery: done 👍🏽

pkg/cmd/backuprestore/backupnoconfig.go
tjungblu: how do you envision the lifecycle of this sidecar then? we need to run it 24/7 to keep the cron scheduler running correctly. So when it never restarts, how do we detect the CRD has changed?

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,177 @@
...
+				},
+			),
+		),
+	); err != nil {
+		return err
+	}
+
+	b.scheduler.Start()
```
tjungblu, line 161: From the documentation I can see:
```
	// Start begins scheduling jobs for execution based
	// on each job's definition. Job's added to an already
	// running scheduler will be scheduled immediately based
	// on definition. Start is non-blocking.
	Start()
```

https://pkg.go.dev/github.com/go-co-op/gocron/v2#Scheduler

so this sidecar will continue to restart every time it runs, not sure how it is able to schedule a cronjob accordingly

pkg/cmd/backuprestore/backupnoconfig.go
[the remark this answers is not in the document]
tjungblu: easy, thanks

pkg/cmd/backuprestore/backupnoconfig.go
Elbehery: >  the job will be added to the scheduler **immediately**, and will be scheduled upon **definition**
iiuc, this means the job is being added, but the execution is based on the `crontab` 

why this will restart the whole sidecar ?

tjungblu: there is no crontab here, the library will just sleep on a go channel :rofl:

Elbehery: achsoooo .. but why the container will restart every time ?

tjungblu: what's stopping it from exiting the command?

Elbehery: achsooo .. ok  .. so either we run it a daemon or have a controller that invoke the cmd upon reconciliation, or ?

tjungblu: just trying whether I'm stupid here, but this example immediately exits without running any task:

```
import (
	"fmt"
	gcron "github.com/go-co-op/gocron/v2"
)

func main() {
	scheduler, err := gcron.NewScheduler(
		gcron.WithLimitConcurrentJobs(1, gcron.LimitModeWait),
		gcron.WithGlobalJobOptions(
			gcron.WithLimitedRuns(1),
			gcron.WithSingletonMode(gcron.LimitModeWait)))
	if err != nil {
		panic(err)
	}

	_, err = scheduler.NewJob(gcron.CronJob("* * * * *", false), gcron.NewTask(func() {
		fmt.Printf("hello world\n")
	}))
	if err != nil {
		panic(err)
	}
	scheduler.Start()

}
```

tjungblu: > achsooo .. ok .. so either we run it a daemon or have a controller that invoke the cmd upon reconciliation, or ?

well, hence my question on the other thread about the life cycle you're thinking about, because this command is not going to work :'D

Elbehery: no u r right, its my bad :D .. 

but i followed the same approach as in the `requestBackup` and `prune` CMDs 

probably we need a controller that reacts only to the `default` CR and invoke the cmd 

wdyt

pkg/cmd/backuprestore/backupnoconfig.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (b *backupNoConfig) AddFlags(fs *pflag.FlagSet) {
+	b.backupOptions.AddFlags(fs)
+}
+
+func (b *backupNoConfig) Validate() error {
+	return b.backupOptions.Validate()
+}
+
+func (b *backupNoConfig) Run() error {
```
soltysh, line 64: You don't need to use this scheduler here, but rather you want to just write a simple controller, similar in concept to https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/cronjob/cronjob_controllerv2.go. 

The main bits are:
1. Use a DelayingQueue from https://github.com/kubernetes/client-go/blob/master/util/workqueue/delaying_queue.go (`TypedNewDelayingQueue`, for example, or others if you need more control). Additionally you'll register an informer on that Backup CR of yours, which will be responsible to trigger worker any time a change to a CR will happen. 
2. After that you'll create a `Run` method very similar to the cronjob controller, there you'll start the queue and your worker.
3. Inside the worker you'll either schedule the backup to run or you'll queue (using `AddAfter` method) the next execution. To calculate the actual next run you can use `github.com/robfig/cron` library which helps parsing the cron spec (`cron.ParseStandard`) .

soltysh (changes requested):

bindata/etcd/pod.yaml
```diff
@@ -335,3 +335,9 @@ ${COMPUTED_ENV_VARS}
     - hostPath:
         path: /var/log/etcd
       name: log-dir
+    - hostPath:
+        path: /var/backup/etcd
+      name: backup-dir
+    - hostPath:
+        path: /etc/kubernetes
+      name: config-dir
```
hasbro17, line 343: Still have to review to this later and see if all of this aligns with what's outlined in the enhancement, but manifest changes related to this feature like this need to be feature gated and can't be in GA by default.

Elbehery:
closing this in favor of https://github.com/openshift/cluster-etcd-operator/pull/1301

PR #1288, opened 2024-07-01
NO-JIRA: only read signer/bundles on forced leaf generation

tjungblu:
During vertical scaling we could hit a case where the signers are expired and would trigger a rotation, even though we only wanted to create a new set of peer certs. This can cause bricked clusters, as there is no safe guard for the bundle distribution in this case.

/hold 
waiting for the test suite to merge first, even though it might not be relevant


hasbro17:
/lgtm

Good catch, totally missed that the node-cert diff would permit signer rotation as well. Although it doesn't make sense to add new nodes when your signer is expired, better to be safe.

The e2e won't cover this specific case of signers expired during vertical scaling but would be good to run it to see if we haven't broken the usual rotation flow.

PR #1294, opened 2024-07-10
ETCD-574: Update TLS artifact descriptions according to registry requ…

tjungblu:
…irements


tjungblu:
fyi @vrutkovs - in case your node name adjuster for the registry needs adjusting

vrutkovs:
LGTM for me, thanks!

pkg/tlshelpers/tlshelpers.go
```diff
@@ -184,7 +182,8 @@ func CreatePeerCertificate(node *corev1.Node,
 	secretGetter corev1client.SecretsGetter,
 	recorder events.Recorder) (*certrotation.RotatedSelfSignedCertKeySecret, error) {
 	return createCertForNode(
-		fmt.Sprintf("Peer Cert for node %s", node.Name),
+		fmt.Sprintf("Peer (client and server) certificate for node %s, generated by cluster-etcd-operator for etcd. "+
+			"This certificate is valid for %d days and starts refreshing after %d days.", node.Name, durationDays(EtcdCertValidity), durationDays(EtcdCertValidityRefresh)),
```
hasbro17, line 186: Do you reckon we should mention the hostnames that our certs are configured to serve? E.g the hardcoded ones plus whatever the node IPs end up being:
https://github.com/openshift/cluster-etcd-operator/blob/767256e7feed840ccedc80d5419e31f3941ce806/pkg/tlshelpers/tlshelpers.go#L56-L67


Only reason I mention it now is to not get nagged at later once we actually update the descriptions doc:
https://github.com/openshift/origin/blob/master/tls/descriptions/descriptions.md#how-to-meet-the-requirement

> Which connections a CA bundle can be used to verify.
What kind of certificates a signer will sign for.
Which names and IPs a serving certificate terminates.
Which subject (user and group) a client certificate is created for.
Which binary and flags is this certificate wired to.

hasbro17: This is the only other example I can find that's mentioning hostnames though:
https://github.com/openshift/origin/blob/master/tls/autoregenerate-after-expiry/autoregenerate-after-expiry.md#certificates-85

tjungblu: @deads2k / @vrutkovs shall we list all those host names / IPs in the annotations?

vrutkovs: No, please don't. These will be censored when collected for TLS registry to keep it consistent - and I don't think its useful for the users

tjungblu:
I think the scaling we're going to override, it's likely not failing on our annotations here :) 

PR #1293, opened 2024-07-09
OCPBUGS-36621: add etcd-all-bundles to cvo create-only

tjungblu:
This should avoid cases where the static pod installer controller races with the cert controller and they would block each other on that resource.

PR #1295, opened 2024-07-12
Revert "OCPBUGS-36621: add etcd-all-bundles to cvo create-only"

tjungblu:
Reverts openshift/cluster-etcd-operator#1293

leaving this here as this is not effectively fixing the issue


hasbro17:
/label acknowledge-critical-fixes-only
/retest-required

We need this to revert back to the old behavior of when the etcd-all-bundles configmap is missing altogether.
Basically change the signal from https://issues.redhat.com/browse/OCPBUGS-36929 to https://issues.redhat.com/browse/OCPBUGS-36621

That way we can fix the stall in the cert controller on the missing configmap symptom.

PR #1298, opened 2024-07-23
ETCD-635: add backup server sidecar

Elbehery:
This PR adds `backup-server` sidecar template.

This is part of https://issues.redhat.com/browse/ETCD-636


pkg/cmd/backuprestore/backupserver.go
```diff
@@ -0,0 +1,127 @@
...
+	//
+	//<-doneCh
+	//return nil
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		// sleep for 292 years
+		time.Sleep(time.Duration(1<<63 - 1))
```
tjungblu, line 112: we need to somehow cancel this on the `ctx` above, otherwise we won't be able to terminate the container properly

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -0,0 +1,127 @@
...
+var shutdownSignals = []os.Signal{os.Interrupt, syscall.SIGTERM}
+
+type backupServer struct {
+	schedule  string
+	timeZone  string
+	scheduler *tasker.Tasker
+	backupOptions
+}
+
+func NewBackupNoConfigCommand(ctx context.Context, errOut io.Writer) *cobra.Command {
```
tjungblu, line 29:
```suggestion
func NewBackupServer(ctx context.Context) *cobra.Command {
```

let's name it properly already, why do you need to add the stderr if you're just using `os.Stderr` anyway?

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -0,0 +1,127 @@
...
+		defer wg.Done()
+		// sleep for 292 years
+		time.Sleep(time.Duration(1<<63 - 1))
+	}()
+	wg.Wait()
+
+	return nil
+}
+
+func (b *backupServer) scheduleBackup() error {
```
tjungblu, line 119: remove for now please

bindata/etcd/pod.yaml
```diff
@@ -311,6 +311,31 @@ ${COMPUTED_ENV_VARS}
...
+    imagePullPolicy: IfNotPresent
+    terminationMessagePolicy: FallbackToLogsOnError
+    command:
+      - /bin/sh
+      - -c
+      - |
+        #!/bin/sh
+        set -euo pipefail
+        
+        exec nice -n -18 cluster-etcd-operator backup-server
```
tjungblu, line 325: you can leave it at default niceness, no need to have it fight with grpcproxy or the networking :)
```suggestion
        command: [cluster-etcd-operator]
        args:
        - backup-server
```

tjungblu: and remove the rest of the shell script

pkg/cmd/backuprestore/backupserver.go
Elbehery: done 👍🏽

bindata/etcd/pod.yaml
Elbehery: done 👍🏽

pkg/cmd/backuprestore/backupserver.go
Elbehery: done 👍🏽

pkg/cmd/backuprestore/backupserver.go
Elbehery: done 👍🏽

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -0,0 +1,121 @@
...
+			must := func(fn func() error) {
+				if err := fn(); err != nil {
+					if cmd.HasParent() {
+						klog.Fatal(err)
+					}
+					fmt.Fprint(backupSrv.errOut, err.Error())
+				}
+			}
+
+			must(backupSrv.Validate)
```
tjungblu, line 45: I assume this will crash? :)

Elbehery: removed this 👍🏽

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,41 @@
+package backuphelpers
+
+import "sync"
+
+type BackupVar interface {
+	GetBackupVars() map[string]string
```
tjungblu, line 6:
```suggestion
	GetBackupVars() BackupConfig
```

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+type BackupConfig struct {
+	schedule  string
```
tjungblu, line 10:
```suggestion
        enabled   bool
	schedule  string
```

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -101,6 +105,11 @@ func (c TargetConfigController) sync(ctx context.Context, syncCtx factory.SyncCo
 		return fmt.Errorf("TargetConfigController missing env var values")
 	}
 
+	backupVars := c.backupVarGetter.GetBackupVars()
+	if len(backupVars) == 0 {
+		return fmt.Errorf("TargetConfigController missing backup var values")
+	}
```
tjungblu, line 111: it's fine if they're missing, no need to error the controller here

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -269,3 +281,11 @@ func (c *TargetConfigController) namespaceEventHandler() cache.ResourceEventHand
...
 	}
 }
+
+func backVarsToArgs(vars map[string]string) string {
+	var args []string
+	for k, v := range vars {
+		args = append(args, fmt.Sprintf("- --%s=%s", k, v))
+	}
+	return strings.Join(args, "\n")
+}
```
tjungblu, line 291: I'm not entirely sure yet whether we could/should add this into the `BackupConfig` struct itself, maybe that's a bit better if it's centralized

pkg/operator/targetconfigcontroller/targetconfigcontroller_test.go
```diff
@@ -162,3 +162,18 @@ func TestTargetConfigController(t *testing.T) {
...
+		"config-dir": "/etc/kubernetes",
+		"backup-dir": "/var/backup/etcd",
+	}
+
+	expected := "- --data-dir=/var/lib/etcd\n- --config-dir=/etc/kubernetes\n- --backup-dir=/var/backup/etcd\n"
+
+	act := backVarsToArgs(input)
+	// map order is non-deterministic
+	// use contains instead of equal
+	require.Contains(t, expected, act)
```
tjungblu, line 178: `require.Equal` takes care of that using reflect.DeepEquals, not sure why you think this matters

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
Elbehery: 👍🏽

pkg/operator/targetconfigcontroller/targetconfigcontroller_test.go
Elbehery: it fails, i used it first, ok will try again :D

tjungblu: then it's not the same string :rofl: Please avoid this string map stuff, let's use the proper CRD types...

pkg/backuphelpers/backupvars.go
Elbehery: I think we should keep the `map`, since the struct has other fields (e.g. sync.Mutex)

wdyt ?

tjungblu: you can also reuse the embedded CRD itself, no need to invent a new one

Elbehery: actually its a good idea, but i need to add new args anyways (ie. "data-dir", "backup-dir" )

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,53 @@
+package backuphelpers
```
tjungblu, line 1:
```suggestion
package backuphelpers

package backuphelpers

import (
	"github.com/openshift/api/config/v1alpha1"
	"sync"
)

type BackupVar struct {
	m         sync.Mutex
	enabled   bool
	backupCrd v1alpha1.EtcdBackupSpec
}

```

would be plenty enough, with a Setter and a Getter.

tjungblu: and a function that returns this as an argument list:

```
func (v BackupVar) ArgString() string {
...
}

```

pkg/backuphelpers/backupvars.go
Elbehery: i used `global mutex` on the package, so that we move the struct around without the lock

Elbehery: its not ideal i know, but its patch 

otherwise, i can make another struct for the config, and embed it into `BackupConfig` struct, 

the mutex will be on the outer struct only, but it used to mutate the inner struct state

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
Elbehery: done 👍🏽

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
Elbehery: moved. should `BackupConfig` be `Singleton` ?

actually if it is, then using a global mutex wont be an issue, it is actually better iiuc

Elbehery: moved 👍🏽

pkg/backuphelpers/backupvars.go
tjungblu: I don't understand why you need a global mutex, can you explain?

pkg/operator/targetconfigcontroller/targetconfigcontroller_test.go
Elbehery: it failed again 
```
=== RUN   TestBackupConfig_ToArgs
    backupvars_test.go:26: 
        	Error Trace:	/Users/melbeher/go/src/openshift/cluster-etcd-operator/pkg/backuphelpers/backupvars_test.go:26
        	Error:      	Not equal: 
        	            	expected: "- --data-dir=/var/lib/etcd\n- --config-dir=/etc/kubernetes\n- --backup-dir=/var/backup/etcd"
        	            	actual  : "- --backup-dir=/var/backup/etcd\n- --data-dir=/var/lib/etcd\n- --config-dir=/etc/kubernetes"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	+- --backup-dir=/var/backup/etcd
        	            	 - --data-dir=/var/lib/etcd
        	            	 - --config-dir=/etc/kubernetes
        	            	-- --backup-dir=/var/backup/etcd
        	Test:       	TestBackupConfig_ToArgs
--- FAIL: TestBackupConfig_ToArgs (0.00s)
```

the order of the values within the map generates diff string

pkg/backuphelpers/backupvars.go
Elbehery: when cloning the `BackupConfig`, will carry the mutex around 

I was using a `map` to get the args, but u asked for using the struct, thats why i wanted to use global mutex

pkg/backuphelpers/backupvars.go
Elbehery: done

pkg/operator/targetconfigcontroller/targetconfigcontroller_test.go
Elbehery: done 👍🏽

pkg/backuphelpers/backupvars.go
Elbehery: done 👍🏽

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
Elbehery: done 👍🏽

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,57 @@
...
+	return c
+}
+
+func (b *BackupConfig) ArgString() string {
+	b.mux.Lock()
+	defer b.mux.Unlock()
+
+	var args []string
+	args = append(args, fmt.Sprintf("- --%s=%s", "timezone", b.backupCR.TimeZone))
+	args = append(args, fmt.Sprintf("- --%s=%s", "schedule", b.backupCR.Schedule))
```
Elbehery, line 54: @tjungblu  these are the args for now, this is ok ?

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,66 @@
...
+		"schedule":   {},
+		"timezone":   {},
+		"data-dir":   {},
+		"config-dir": {},
+		"backup-dir": {},
+	}
+)
+
+type BackupVar interface {
+	GetBackupVars() *BackupConfig
```
tjungblu, line 23: easy fix for the mutex copy:

```suggestion
	ArgString() string
```

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (b *BackupConfig) GetBackupVars() *BackupConfig {
+	b.mux.Lock()
+	defer b.mux.Unlock()
+
+	c := NewBackupConfig(b.backupCR)
+	c.enabled = b.enabled
+	return c
+}
```
tjungblu, line 47: don't think we need this if we're exporting the args as string

```suggestion

```

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,66 @@
...
+type BackupVar interface {
+	GetBackupVars() *BackupConfig
+}
+
+type BackupConfig struct {
+	enabled  bool
+	backupCR backupv1alpha1.EtcdBackupSpec
+	mux      sync.Mutex
+}
+
```
tjungblu, line 31:
```suggestion
func NewDisabledBackupConfig() *BackupConfig {
	return &BackupConfig{
		enabled:  false,
		mux:      sync.Mutex{},
	}
}
```

let's be clear that this is not enabled

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,66 @@
...
+	c := NewBackupConfig(b.backupCR)
+	c.enabled = b.enabled
+	return c
+}
+
+func (b *BackupConfig) ArgString() string {
+	b.mux.Lock()
+	defer b.mux.Unlock()
+
+	var args []string
```
tjungblu, line 53: why so complicated?

```suggestion
        args := []string{fmt.Sprintf(""--pvc="%s", b.backupCR.PvcName), 
           ...
         }
	return strings.Join(args,"\n")
```

doesn't compile, but I hope you get the gist

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -0,0 +1,114 @@
...
+		for {
+			select {
+			case <-ctx.Done():
+				cancel()
+			case <-time.After(time.Duration(1<<63 - 1)):
+				klog.Info("time elapsed ...")
+			}
+		}
+	}()
+	wg.Wait()
```
tjungblu, line 111: I haven't tested this, but shouldn't this also work?

```suggestion
<-ctx.Done()
```

pkg/operator/starter.go
```diff
@@ -260,6 +261,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		masterNodeInformer,
 		kubeClient,
 		envVarController,
+		backuphelpers.NewBackupConfig(v1alpha1.EtcdBackupSpec{}),
```
tjungblu, line 264: use the disabled version from above, please

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -179,21 +185,24 @@ func loglevelToZap(logLevel operatorv1.LogLevel) string {
...
+	backVarLines := backupVars.ArgString()
+
 	return strings.NewReplacer(
 		"${IMAGE}", imagePullSpec,
 		"${OPERATOR_IMAGE}", operatorImagePullSpec,
 		"${VERBOSITY}", loglevelToZap(operatorSpec.LogLevel),
 		"${LISTEN_ON_ALL_IPS}", "0.0.0.0", // TODO this needs updating to detect ipv6-ness
 		"${LOCALHOST_IP}", "127.0.0.1", // TODO this needs updating to detect ipv6-ness
 		"${COMPUTED_ENV_VARS}", strings.Join(envVarLines, "\n"), // lacks beauty, but it works
+		"${COMPUTED_BACKUP_VARS}", backVarLines,
```
tjungblu, line 205: just directly call it

```suggestion
		"${COMPUTED_BACKUP_VARS}", backupVars.ArgString(),
```

pkg/backuphelpers/backupvars.go
Elbehery: +1, Bon Remark :)

pkg/backuphelpers/backupvars.go
tjungblu: yeah that's stupid, so we have to add tabs in there to match the formatting. Please add a unit test for that in the targetconfigcontroller.

I would suggest (pseudocode):

```
args := []string{}
if b.enabled {
  args = append(args, "\t--enabled=true\n")
  args = append(args, "\t--pvc=${PVC_NAME}\n")
   ...
} else {
  args = append(args, "\t--enabled=false\n")
}
```

maybe we can add the tab with the join, or some other function to prepend it for each entry.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
Elbehery: done

pkg/operator/starter.go
Elbehery: done

pkg/cmd/backuprestore/backupserver.go
Elbehery: done

pkg/backuphelpers/backupvars.go
Elbehery: ok, but i actually think we dont need the PVC since we use a pre-defined `hostPath`, no ?

pkg/backuphelpers/backupvars.go
Elbehery: done

pkg/backuphelpers/backupvars.go
Elbehery: done

pkg/backuphelpers/backupvars.go
Elbehery: done

pkg/backuphelpers/backupvars.go
tjungblu: dude, it's an example :rofl: fill what you need

Elbehery: :D :D

Elbehery: there is no other way to add the tabs :/

tjungblu: > strings.Join(args,"\n\t")  + "\n"

doesn't work? :)

Elbehery: not exactly, I had to try various ways to get it right, must have empty strings in the slice beforehand in order to get it right at the end :) 

fixed 👍🏽

Elbehery:
`e2e operator` failed due to empty args 

```
  - name: etcd-backup-server
    image: registry.build03.ci.openshift.org/ci-op-r34mdkbx/stable@sha256:2ae71921641f510e0dcd434b00560b821a9e87aea39993907a219ae573835732
    imagePullPolicy: IfNotPresent
    terminationMessagePolicy: FallbackToLogsOnError
    command: [cluster-etcd-operator]
    args:
    - backup-server

	- --timezone=
	- --schedule=
	-
```

fixing ...

tjungblu:
@Elbehery maybe it makes it easier if all list items are indented the same, you can also add the `backup-server` in the golang code instead of the template

Elbehery:
> @Elbehery maybe it makes it easier if all list items are indented the same, you can also add the `backup-server` in the golang code instead of the template

hmmm .. so the template would contain only the container definition, but not the `CMD`, correct ?

Elbehery:
yes u r right 👍🏽  thxxx

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,47 @@
...
+	ArgString() string
+}
+
+type BackupConfig struct {
+	enabled  bool
+	backupCR backupv1alpha1.EtcdBackupSpec
+	mux      sync.Mutex
+}
+
+func NewDisabledBackupConfig(backupCR backupv1alpha1.EtcdBackupSpec, enabled bool) *BackupConfig {
```
tjungblu, line 21: let's leave this without parameters, the testcases seem to directly use the struct anyway
```suggestion
func NewDisabledBackupConfig() *BackupConfig {
```

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,47 @@
...
+		backupCR: backupCR,
+		mux:      sync.Mutex{},
+	}
+}
+
+func (b *BackupConfig) ArgString() string {
+	b.mux.Lock()
+	defer b.mux.Unlock()
+
+	args := []string{"    - backup-server"}
```
tjungblu, line 33: interesting indentation, but might be github just rendering the tabs wrong

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,47 @@
...
+	if b.enabled {
+		args = append(args, fmt.Sprintf("- --%s=%v", "enabled", b.enabled))
+	}
```
tjungblu, line 36: if you do it that way, you can just ignore the condition entirely

```suggestion
   args = append(args, fmt.Sprintf("- --%s=%v", "enabled", b.enabled))
```

pkg/backuphelpers/backupvars.go
Elbehery: :D :D

pkg/backuphelpers/backupvars.go
Elbehery: i need to pass the CR as an param for testing

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,47 @@
...
+
+	if b.backupCR.TimeZone != "" {
+		args = append(args, fmt.Sprintf("- --%s=%s", "timezone", b.backupCR.TimeZone))
+	}
+
+	if b.backupCR.Schedule != "" {
+		args = append(args, fmt.Sprintf("- --%s=%s", "schedule", b.backupCR.Schedule))
+	}
```
tjungblu, line 44: we can change that later when we fill the command with logic, but there is some logical flaw. 
If it's disabled, we shouldn't set any of those fields. If they are partially empty, we need to consider the CRD as invalid or impute proper defaults

pkg/backuphelpers/backupvars_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	schedule = "0 */2 * * *"
+	timezone = "GMT"
+)
+
+func TestBackupConfig_ToArgs(t *testing.T) {
```
tjungblu, line 16: good tests!

pkg/backuphelpers/backupvars.go
Elbehery: i think defaults will do 👍🏽

tjungblu: I think the defaults will bite your a$$ in the next PR ;)

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func NewDisabledBackupConfig() *BackupConfig {
+	return &BackupConfig{
+		enabled: false,
+		mux:     sync.Mutex{},
+	}
+}
+
+func (b *BackupConfig) AddBackupSpec(spec backupv1alpha1.EtcdBackupSpec) {
```
tjungblu, line 29: that's more of a `SetBackupSpec` isn't it?

pkg/operator/periodicbackupcontroller/periodicbackupcontroller.go
```diff
@@ -80,6 +92,20 @@ func (c *PeriodicBackupController) sync(ctx context.Context, _ factory.SyncConte
 		return fmt.Errorf("PeriodicBackupController could not list backup CRDs, error was: %w", err)
 	}
 
+	// reconciliation of default backup
```
tjungblu, line 95: you are already iterating over all backups, why not just name-check there?

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	backupv1alpha1 "github.com/openshift/api/config/v1alpha1"
+)
+
+type BackupVar interface {
+	SetBackupSpec(spec backupv1alpha1.EtcdBackupSpec)
```
tjungblu, line 12: another thing we can add here is the listener we also have in EnvVar:

https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/etcdenvvar/envvarcontroller.go#L36

so you can hook into the sync queue in the targetconfigcontroller:
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L73

that way we can directly notify, instead of waiting for the 1m resync period. Not a game breaker for now, but maybe if you want to have a follow-up PR.

Elbehery: ok i do it now

Elbehery: fixed 👍🏽

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,62 @@
...
+		enabled: false,
+		mux:     sync.Mutex{},
+	}
+}
+
+func (b *BackupConfig) SetBackupSpec(spec backupv1alpha1.EtcdBackupSpec) {
+	b.mux.Lock()
+	defer b.mux.Unlock()
+
+	b.spec = spec
```
tjungblu, line 39: you have the notify the listeners that the spec has changed

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if b.spec.Schedule != "" {
+		args = append(args, fmt.Sprintf("- --%s=%s", "schedule", b.spec.Schedule))
+	}
+
+	return strings.Join(args, "\n    ")
+}
+
+func (b *BackupConfig) AddListener(listener Enqueueable) {
+	b.listeners = append(b.listeners, listener)
```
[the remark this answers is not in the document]
Elbehery: done

pkg/backuphelpers/backupvars.go
Elbehery: done

pkg/backuphelpers/backupvars.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (b *BackupConfig) SetBackupSpec(spec backupv1alpha1.EtcdBackupSpec) {
+	b.mux.Lock()
+	defer b.mux.Unlock()
+
+	b.spec = spec
+
+	for _, l := range b.listeners {
+		l.Enqueue()
+	}
```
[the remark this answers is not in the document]
Elbehery: deadlock, why ?

only one controller can hold the lock, no ?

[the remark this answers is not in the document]
tjungblu: if you notify the other controller it will call the reading argument function, which can't acquire the lock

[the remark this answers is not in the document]
tjungblu: and you're also introducing the exact same issue I've been fixing in
https://github.com/openshift/cluster-etcd-operator/pull/1300

we need to check whether the spec has changed, otherwise the other controller will hotloop around every invocation of SetBackupSpec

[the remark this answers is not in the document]
Elbehery: ahhh .. so the function wont return till it holds the lock 

but this controller will eventually release the lock, no ?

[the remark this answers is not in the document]
tjungblu: depends how it's implemented on the other side, but let's leave this if the CI is OK

Elbehery:
closing in favor of https://github.com/openshift/cluster-etcd-operator/pull/1301

Elbehery:
closing in favor of https://github.com/openshift/cluster-etcd-operator/pull/1301

PR #1300, opened 2024-07-26
NO-JIRA: only update envvar listeners on real changes

tjungblu:
While reviewing the changes in the backup PR, I've seen that I've introduced unnecessary listener updates during the lock refactoring two years ago.

This PR ensures we return early when the envvars has not changed.


Elbehery:
cherrypicked manually https://github.com/openshift/cluster-etcd-operator/pull/1302

PR #1306, opened 2024-07-29
ETCD-661: Add `etcd-backup-server` scheduling logic

Elbehery:
This PR adds etcd-backup-server scheduling logic.

This is part of https://issues.redhat.com/browse/ETCD-636

- [ ] set default schedule in case the CR has empty string [note](https://github.com/openshift/api/blob/master/config/v1alpha1/types_backup.go#L51)
- [x] Add the backup Path to static pods


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.22.0
 toolchain go1.22.1
 
 require (
+	github.com/adhocore/gronx v1.8.1
```
tjungblu, line 8: @soltysh mentioned we can also use a `DelayQueue` from k8s itself, together with a cron parser. Can you check whether this would be a viable alternative here?

Elbehery: 👍🏽

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	b.backupOptions.backupDir = backupVolume + backupFolder
+	return b.backupOptions.Validate()
+}
+
+func (b *backupServer) Run(ctx context.Context) error {
+	klog.Infof("hello from backup server Run() :) ")
+
+	b.scheduler = tasker.New(tasker.Option{
```
dusk125, line 84: If we continue with Tasker, this function could be condensed to just calling `b.scheduler.Run()`. Tasker handles the same signals by default https://github.com/adhocore/gronx/blob/main/pkg/tasker/tasker.go#L294 and can be configured to take a parent context https://github.com/adhocore/gronx/blob/main/pkg/tasker/tasker.go#L113.

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -2,15 +2,22 @@ package backuprestore
...
+	"fmt"
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
-	"github.com/spf13/cobra"
 	"k8s.io/klog/v2"
+
+	"github.com/robfig/cron"
```
tjungblu, line 14: k8s uses 	
> "github.com/robfig/cron/v3"

that's also mentioned that way in:
https://github.com/robfig/cron?tab=readme-ov-file#cron

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -52,6 +59,24 @@ func (b *backupServer) AddFlags(fs *pflag.FlagSet) {
 }
 
 func (b *backupServer) Validate() error {
+	if !b.enabled {
+		klog.Infof("backup-server is disabled")
+		return nil
+	}
+
+	if len(b.schedule) == 0 {
+		err := errors.New("missing required flag: --schedule")
```
tjungblu, line 68: if you're using cobra now, you can mark it required.

I would also assume validate to try to parse the cron and return a more explicit error here

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -61,7 +86,49 @@ func (b *backupServer) Run(ctx context.Context) error {
...
+func (b *backupServer) scheduleBackup(ctx context.Context, schedule cron.Schedule) error {
+	ticker := time.NewTicker(time.Until(schedule.Next(time.Now())))
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			err := b.runBackup()
+			if err != nil {
+				return err
```
tjungblu, line 116: should this return an error or just log? Because returning here will exit the entire sidecar container

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -61,7 +86,49 @@ func (b *backupServer) Run(ctx context.Context) error {
...
+			}
+			ticker.Reset(time.Until(schedule.Next(time.Now())))
+		case <-ctx.Done():
+			klog.Infof("context deadline has exceeded")
```
tjungblu, line 120: that could be misleading, because you don't know that this context was a deadline. I think you can safely remove that statement and just leave a `return nil` or so

pkg/cmd/backuprestore/backupserver.go
Elbehery: done 👍🏽

pkg/cmd/backuprestore/backupserver.go
Elbehery: done

pkg/cmd/backuprestore/backupserver.go
Elbehery: +1, totally agree

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -61,7 +86,49 @@ func (b *backupServer) Run(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+	}
+
 	<-ctx.Done()
+	return nil
+}
+
+func (b *backupServer) scheduleBackup(ctx context.Context, schedule cron.Schedule) error {
```
tjungblu, line 107: before I forget, we need some sensible tests for the whole command as well :)

pkg/cmd/backuprestore/backupserver.go
Elbehery: done

pkg/cmd/backuprestore/backupserver.go
Elbehery: :D :D .. will start with it now 👍🏽

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -2,21 +2,28 @@ package backuprestore
...
 
-	"github.com/spf13/cobra"
 	"k8s.io/klog/v2"
+
+	cron "github.com/robfig/cron/v3"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
+const backupVolume = "/var/backup/etcd/"
```
tjungblu, line 18:
```suggestion
const backupVolume = "/var/lib/etcd-auto-backup"
```

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -48,10 +55,32 @@ func (b *backupServer) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&b.schedule, "schedule", "", "schedule specifies the cron schedule to run the backup")
 	fs.StringVar(&b.timeZone, "timezone", "", "timezone specifies the timezone of the cron schedule to run the backup")
 
+	cobra.MarkFlagRequired(fs, "schedule")
```
Elbehery, line 58: @tjungblu we cant make this required, because if the container is disabled, which is the case initially, the staticPod will fail to launch 

[prow](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1306/pull-ci-openshift-cluster-etcd-operator-master-e2e-operator/1831311344830255104/artifacts/e2e-operator/gather-extra/artifacts/pods/openshift-etcd_etcd-ip-10-0-37-122.ec2.internal_etcd-backup-server.log)

pkg/cmd/backuprestore/backupserver.go
Elbehery: done 👍🏽

pkg/cmd/backuprestore/backupserver.go
tjungblu: makes sense

pkg/cmd/backuprestore/backupserver.go
tjungblu: maybe also think about this for a follow-up task: how do we surface any errors, do we emit a metric, alerting? Do we put something into the status field of the CRD.

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -61,7 +88,43 @@ func (b *backupServer) Run(ctx context.Context) error {
 	defer cancel()
```
dusk125, line 88: NotifyContext returns a context and cancelFunc, you don't need the above WithCancel call.
https://pkg.go.dev/os/signal#NotifyContext

pkg/cmd/backuprestore/backupserver.go
tjungblu: I think we need both :)

pkg/cmd/backuprestore/backupserver.go
Elbehery: done 👍🏽

pkg/cmd/backuprestore/backupserver.go
Elbehery: I added tests in [Backupserver_test](https://github.com/openshift/cluster-etcd-operator/pull/1306/files#diff-97951c69d5abc4a441c76a64b73793e40d36d8bf53c80d73f175db1231be3b3d )

I tested only the validate() for the following reasons 

- The backup tests are already well tested, I embedded the `backupOptions`
- The `Run()` is nothing more than a ticker, that takes the backup every tick
- The only thing I could test is the validation upon various input 

Please correct me if i am wrong

pkg/cmd/backuprestore/backupserver.go
Elbehery: ok .. so how is the alert should be firing ?

tjungblu: please create a spike ticket, we can discuss this there.

bindata/etcd/pod.yaml
```diff
@@ -390,3 +392,7 @@ ${COMPUTED_ENV_VARS}
     - hostPath:
         path: /etc/kubernetes
       name: config-dir
+    - hostPath:
+        path: /var/lib/etcd-auto-backup
+        type: DirectoryOrCreate
```
tjungblu, line 397: that's a cool feature, what permissions is it creating the directory with?

pkg/cmd/backuprestore/backupserver.go
Elbehery: I made various changes to facilitate testing 

I add a interface for `runBackup()`, and add a mock within the test to control the output 

However, I cant assert upon the tickers !! 

ptal

pkg/cmd/backuprestore/backupserver_test.go
```diff
@@ -0,0 +1,122 @@
...
+	srvr := &backupServer{
+		schedule:      validSchedule,
+		timeZone:      "",
+		enabled:       true,
+		cronSchedule:  cronSchedule,
+		backupOptions: backupOptions{},
+	}
+
+	err = srvr.scheduleBackup(context.Background(), backUpErMock{})
+	require.NoError(t, err)
```
tjungblu, line 115: that's very happy path, what's with the error cases and the actual scheduling?

pkg/cmd/backuprestore/backupserver.go
tjungblu: > However, I cant assert upon the tickers !!

why not? just record the number of invocations in your mock.

pkg/cmd/backuprestore/backupserver_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			actErr := tc.backupServer.Validate()
+			if tc.expErr != nil {
+				require.Contains(t, actErr.Error(), tc.expErr.Error())
+			} else {
+				require.Equal(t, tc.expErr, actErr)
+			}
```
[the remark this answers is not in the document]
Elbehery: so there error being returned is wrapped, since the exact error can not be always the same (i.e. the dependency could change :) ), i assert on the msg 

I can change it to the exact error by wrapping the internal errors returned but i think it is fragile

pkg/cmd/backuprestore/backupserver_test.go
Elbehery: yes this is all internal to the `runBackup()` .. I can return them for assertion, but its ugly 

any suggestions ?

bindata/etcd/pod.yaml
Elbehery: wdy mean ?!!

pkg/cmd/backuprestore/backupserver.go
Elbehery: lgtm 👍🏽

pkg/cmd/backuprestore/backupserver_test.go
Elbehery: so the errors of the cron are being testing with the validation. `scheduleBackup()` will never be called in case the schedule is invalid https://github.com/openshift/cluster-etcd-operator/blob/483610a715a42e42802be5a61dd0fde9e87130dd/pkg/cmd/backuprestore/backupserver.go#L84-L88 

What i am testing here, is that the scheduling using the ticker is working as expected .. 

I create a context with timeout of `3 minutes` and assert that the `runBackup` has been invoked at least 3 times 

the cron schedule i used is `* * * * *`, the issue is that the test will need at least 3 minutes to run, there is no cron for second iiuc

pkg/cmd/backuprestore/backupserver.go
Elbehery: worked 👍🏽 thxxxx

bindata/etcd/pod.yaml
tjungblu: `/var/lib/etcd-auto-backup` doesn't exist, so it has to be created: with what permissions?
it's kinda important that we don't have 0777 permissions on etcd backups

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -2,21 +2,45 @@ package backuprestore
...
+	cron "github.com/robfig/cron/v3"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
+const backupVolume = "/var/lib/etcd-auto-backup"
+
 var shutdownSignals = []os.Signal{os.Interrupt, syscall.SIGTERM}
 
+type backUpEr interface {
```
tjungblu, line 22:
```suggestion
type backupRunner interface {
```

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -2,21 +2,45 @@ package backuprestore
...
+	runBackup(opts *backupOptions) error
+}
+
+type backUpErImpl struct{}
```
tjungblu, line 26: sounds Java-esque, but I also don't have a better name

```suggestion
type backRunnerImpl struct{}
```

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -52,16 +76,61 @@ func (b *backupServer) AddFlags(fs *pflag.FlagSet) {
...
+	if !b.enabled {
+		klog.Infof("backup-server is disabled")
+		return nil
+	}
+
+	cronSchedule, err := cron.ParseStandard(b.schedule)
+	if err != nil {
+		pErr := fmt.Errorf("error parsing backup schedule %v: %w", b.schedule, err)
+		klog.Error(pErr)
+		return pErr
```
tjungblu, line 88:
```suggestion
		return fmt.Errorf("error parsing backup schedule %v: %w", b.schedule, err)
```

no need to log here, the root cmd does it already

pkg/cmd/backuprestore/backupserver.go
```diff
@@ -52,16 +76,61 @@ func (b *backupServer) AddFlags(fs *pflag.FlagSet) {
...
+	}
+	b.cronSchedule = cronSchedule
+
+	b.backupOptions.backupDir = backupVolume
+	err = b.backupOptions.Validate()
+	if err != nil {
+		klog.Error(err)
+		return err
```
tjungblu, line 96:
```suggestion		
		return fmt.Errorf("error validating backup %v: %w", b.backupOptions, err)
```

pkg/cmd/backuprestore/backupserver_test.go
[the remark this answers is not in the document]
tjungblu: then you can also use `require.ErrorContainsf()`

pkg/cmd/backuprestore/backupserver_test.go
```diff
@@ -0,0 +1,163 @@
...
+			expBackups: 3,
+			expErr:     nil,
+		},
+		{
+			name:       "invalid schedule",
+			schedule:   "invalid schedule",
+			timeout:    time.Minute * 3,
+			expBackups: 0,
+			expErr:     errors.New("Expected exactly 5 fields"),
+		},
```
tjungblu, line 131: please also add a case for:
* runBackup returns an error and still runs backups after that
* runBackup takes longer than your schedule

for the latter we would simply skip an invocation, correct?

bindata/etcd/pod.yaml
Elbehery: so the docs says it creates it with `0755`. 

But I can just remove this and create the dir with the correct permission as [here](https://github.com/openshift/cluster-etcd-operator/blob/518174a5a82c923029c7c375c9b6d06d03c83f30/bindata/etcd/pod.yaml#L26) 

wdyt ?

tjungblu: I think 0755 is too broad, 0640 seems good enough for backups. They shouldn't be publicly readable.

Elbehery: fixed https://github.com/openshift/cluster-etcd-operator/pull/1306/commits/fe29a0c9208151c7cf49d6017c6eb9bc8bea7278

pkg/cmd/backuprestore/backupserver_test.go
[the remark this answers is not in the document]
Elbehery: I tried, if the `expErr == nil`, then `require.ErrorContainsf()` fails as it attempts to invoke `nil` value  :/

pkg/cmd/backuprestore/backupserver_test.go
Elbehery: > runBackup takes longer than your schedule

I added this test case https://github.com/openshift/cluster-etcd-operator/pull/1306/commits/610246f3d3abde32245e84cd6f1db6ea2aba4c51

- sleep 3 secs, simulate slow backup
- timeout is 8 secs
- expected 3 backups to be taken 

> runBackup returns an error and still runs backups after that

if backup returns error, the whole process exits 

```
err := b.scheduleBackup(cCtx, bck)
		if err != nil {
			return err
		}
```

tjungblu: > if backup returns error, the whole process exits

should this be the desired behavior?

bindata/etcd/pod.yaml
```diff
@@ -23,7 +23,9 @@ spec:
         - |
           #!/bin/sh
           echo -n "Fixing etcd log permissions."
-          mkdir -p /var/log/etcd && chmod 0700 /var/log/etcd
+          mkdir -p /var/log/etcd  && chmod 0700 /var/log/etcd
+          echo -n "Fixing etcd auto backup permissions."
+          mkdir -p /var/lib/etcd-auto-backup  && chmod 0700 /var/lib/etcd-auto-backup
```
tjungblu, line 28: why set the executable bit?

pkg/cmd/backuprestore/backupserver_test.go
Elbehery: Well, we could log the error and keep on running, but how to know if the `backup()` is no functioning ?!!

bindata/etcd/pod.yaml
Elbehery: I actually wanted to make it `600` but i followed the same permission as for log dir

Elbehery: fixed 👍🏽

pkg/cmd/backuprestore/backupserver_test.go
tjungblu: let me put it the other way then, as an admin how would you notice if it were crash looping?

bindata/etcd/pod.yaml
tjungblu: I have some doubts that whoever is restoring from that backup is the exact same user that's running etcd, but okay, we can fix it later.

pkg/cmd/backuprestore/backupserver_test.go
Elbehery: - statusCondition 
- alert
- logs
- kubectl :D :D 

the only notification that does not require polling is `alert`

bindata/etcd/pod.yaml
Elbehery: ah u mean the `/var/log/etcd` .. ok i can change it now, shall i ?

Elbehery: raised https://github.com/openshift/cluster-etcd-operator/pull/1333 for the log's permission

pkg/cmd/backuprestore/backupserver_test.go
tjungblu: OK let's leave it crashing for now, we can discuss later how we can improve this

PR #1310, opened 2024-08-01
OCPBUGS-36604: tune event correlation

tjungblu:
The event correlator is too strict and throttles away important events for our invariant test suite. This fine tunes the values to try not to lose those.


tjungblu:
nope, still a thing
> {  static pod lifecycle failure - static pod: "etcd" in namespace: "openshift-etcd" for revision: 3 on node: "ip-10-0-35-140.ec2.internal" didn't show up, waited: 3m30s}

PR #1318, opened 2024-08-15
OCPBUGS-38524: Ensure all certs exists on readyZ sidecar

tjungblu:
During installation there might be a brief period where we would fail readyz checks when the certificates are not generated yet by the certsignercontroller.

This will fail the container before returning the wrong readyZ signal.


tjungblu:
/cherry-pick release-4.17 release-4.16 release-4.15 release-4.14

PR #1319, opened 2024-08-15
OCPBUGS-38573: use pooled client for etcd single member health checks

tjungblu:
This should reduce CPU consumption because of less TLS handshakes and makes it usable from other places too.

### DIFF
--- a/pkg/etcdcli/health.go
+++ b/pkg/etcdcli/health.go
@@ -41,7 +41,7 @@ type healthCheck struct {
 
 type memberHealth []healthCheck
 
-func GetMemberHealth(ctx context.Context, etcdMembers []*etcdserverpb.Member) memberHealth {
+func GetMemberHealth(ctx context.Context, cli *clientv3.Client, etcdMembers []*etcdserverpb.Member) memberHealth {
 	// while we don't explicitly mention that the returned ordering has to be the same as etcdMembers,
 	// we try to keep it that way for backward compatibility reasons
 	memberHealth := make([]healthCheck, len(etcdMembers))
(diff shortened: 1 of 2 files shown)


pkg/etcdcli/health.go
```diff
@@ -80,30 +80,17 @@ func GetMemberHealth(ctx context.Context, etcdMembers []*etcdserverpb.Member) me
...
-	}
-	cli, err := newEtcdClientWithClientOpts([]string{member.ClientURLs[0]}, skipConnectionTest)
-	if err != nil {
-		return healthCheck{
-			Member:  member,
-			Healthy: false,
-			Error:   fmt.Errorf("create client failure: %w", err)}
-	}
+func checkSingleMemberHealth(ctx context.Context, cli *clientv3.Client, member *etcdserverpb.Member) healthCheck {
+	originalEndpoints := cli.Endpoints()
```
dusk125, line 84: nit: You could do:
```go
defer cli.SetEndpoints(cli.Endpoints())
```
and cli.Endpoints() will be cached and that cached version will be used at deferred execution

dusk125 (commented):
/lgtm

pkg/etcdcli/health.go
tjungblu: I leave this for now to not have your lgtm being wasted - thanks for the research, I was curious about this too :)

tjungblu:
/cherry-pick release-4.17 release-4.16 release-4.15 release-4.14

tjungblu:

that's a known failure 
> clusteroperator/kube-storage-version-migrator should not change condition/Available

pkg/etcdcli/health.go
dusk125: No worries! Here's the proof! https://go.dev/play/p/Cgzi2FAO45G

The great package github.com/pkg/profile abuses this to make profiling easier to set up.

PR #1321, opened 2024-08-19
WIP: Replace readyz sidecar container with etcds own probes

jubittajohn:


bindata/etcd/pod.yaml
```diff
@@ -196,32 +196,32 @@ ${COMPUTED_ENV_VARS}
         cpu: 300m
     readinessProbe:
       httpGet:
-        port: 9980
-        path: readyz
+        port: 2379
+        path: readyz?exclude=data_corruption&exclude=linearized_read
         scheme: HTTPS
```
tjungblu, line 201: I think it's failing on:
> {"level":"warn","ts":"2024-08-20T03:55:49.588638Z","caller":"embed/config_logging.go:169","msg":"rejected connection","remote-addr":"10.0.36.39:41140","server-name":"","error":"tls: client didn't provide a certificate"}

since we can't supply mTLS certs in the probe configs, we could try to split the http server out with the etcd args:
> --listen-client-http-urls= 

dunno if we can simply take on the ":2378" port for starters and then see if we still need some form of mTLS along with it.

tjungblu:
(just checking something for another ticket - don't be confused)

/override ci/prow/e2e-aws-etcd-recovery

tjungblu:
let's shelve this until we have a solution for
https://github.com/etcd-io/etcd/issues/18477

/close

PR #1323, opened 2024-08-21
testing endpoint returns

tjungblu:
/hold

### DIFF
--- a/pkg/etcdcli/health.go
+++ b/pkg/etcdcli/health.go
@@ -41,7 +41,7 @@ type healthCheck struct {
 
 type memberHealth []healthCheck
 
-func GetMemberHealth(ctx context.Context, cli *clientv3.Client, etcdMembers []*etcdserverpb.Member) memberHealth {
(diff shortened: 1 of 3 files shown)


pkg/etcdcli/health.go
```diff
@@ -81,12 +81,8 @@ func GetMemberHealth(ctx context.Context, cli *clientv3.Client, etcdMembers []*e
 }
 
 func checkSingleMemberHealth(ctx context.Context, cli *clientv3.Client, member *etcdserverpb.Member) healthCheck {
-	originalEndpoints := cli.Endpoints()
-	defer func(eps []string) {
-		cli.SetEndpoints(eps...)
-	}(originalEndpoints)
-
 	// we only want to check that one member that was passed, we're restoring the original client endpoints at the end
+	defer cli.SetEndpoints(cli.Endpoints()...)
```
tjungblu, line 85: the actual problem is that the cli is shared amongst goroutines calling with different members in GetMemberHealth

PR #1326, opened 2024-08-27
ETCD-652: Store latest etcd leader revision in datadir

tjungblu:
This will query all endpoints in a sidecar and provide a local output in /var/lib/etcd/revision.json with the raftindex of all endpoints and the maximum observed.


pkg/cmd/rev/rev.go
```diff
@@ -0,0 +1,190 @@
...
+		case <-ctx.Done():
+			return nil
+		case _ = <-ticker.C:
+			trySaveRevision(r.outputFile, clientPool, r.pollInterval)
+		}
+	}
+}
+
+func trySaveRevision(outputFile string, pool *etcdcli.EtcdClientPool, timeout time.Duration) {
+	timeoutCtx, cancelFunc := context.WithTimeout(context.Background(), timeout)
```
dusk125, line 123: You could pass in the context from the caller for the parent argument here.

pkg/cmd/rev/rev.go
```diff
@@ -0,0 +1,190 @@
...
+	clientCACertFile string
+	endpoints        []string
+	pollInterval     time.Duration
+	outputFile       string
+
+	clientPool *etcdcli.EtcdClientPool
+}
+
+type outputStruct struct {
+	raftIndexByEndpoint map[string]uint64
```
dusk125, line 30: You need to export these fields

dusk125 (changes requested):

tjungblu:
/hold

might need to do some optimizations still :)

PR #1328, opened 2024-08-28
ETCD-612: Added a function to check if quorum is safe and removed the dependent quorum checks from controllers

jubittajohn:
This PR is dependent on https://github.com/openshift/library-go/pull/1781


tjungblu:
/lgtm

test failure in recovery is expected, since we need to adjust the test in origin

>  however, the following indicated by their condition types have not: [EtcdCertSignerControllerDegraded EtcdEndpointsDegraded TargetConfigControllerDegraded]

PR #1330, opened 2024-09-02
ETCD-653: Update restore-pod to use locally stored revision

tjungblu:
This updates the restore script and pod to take advantage of the revision file introduced in a previous PR.

/hold

### DIFF
--- a/bindata/etcd/cluster-restore.sh
+++ b/bindata/etcd/cluster-restore.sh
@@ -8,10 +8,10 @@ set -o errtrace
 
 # example
 # ./cluster-restore.sh $path-to-backup
-# There are several customization switches based on env variables:
-# ETCD_RESTORE_SKIP_MOVE_CP_STATIC_PODS - when set this script will not move the other (non-etcd) static pod yamls
-# ETCD_ETCDCTL_RESTORE - when set this script will use `etcdctl snapshot restore` instead of a restore pod yaml
-# ETCD_ETCDCTL_RESTORE_ENABLE_BUMP - when set this script will spawn the restore pod with a large enough revision bump
+# ETCD_ETCDCTL_RESTORE - when set this script will use `etcdctl snapshot restore` instead of a restore pod yaml,
+#                        which can be used when restoring a single member (e.g. on single node OCP).
+#                        Syncing very big snapshots (>8GiB) from the leader might also be expensive, this aids in
+#                        keeping the amount of data pulled to a minimum. This option will neither rev-bump nor mark-compact.
 
 if [[ $EUID -ne 0 ]]; then
   echo "This script must be run as root"
@@ -96,10 +96,7 @@ BACKUP_FILE=$(ls -vd "${BACKUP_DIR}"/static_kuberesources*.tar.gz | tail -1) || 
 SNAPSHOT_FILE=$(ls -vd "${BACKUP_DIR}"/snapshot*.db | tail -1) || true
 
 ETCD_STATIC_POD_LIST=("etcd-pod.yaml")
-AUX_STATIC_POD_LIST=("kube-apiserver-pod.yaml" "kube-controller-manager-pod.yaml" "kube-scheduler-pod.yaml")
-
-ETCD_STATIC_POD_CONTAINERS=("etcd" "etcdctl" "etcd-metrics" "etcd-readyz")
-AUX_STATIC_POD_CONTAINERS=("kube-controller-manager" "kube-apiserver" "kube-scheduler")
+ETCD_STATIC_POD_CONTAINERS=("etcd" "etcdctl" "etcd-metrics" "etcd-readyz" "etcd-rev" "etcd-backup-server")
 
 if [ ! -f "${SNAPSHOT_FILE}" ]; then
   echo "etcd snapshot ${SNAPSHOT_FILE} does not exist"
@@ -113,12 +110,6 @@ check_snapshot_status "${SNAPSHOT_FILE}"
 ETCD_CLIENT="${ETCD_ETCDCTL_BIN+etcdctl}"
 if [ -n "${ETCD_ETCDUTL_BIN}" ]; then
   ETCD_CLIENT="${ETCD_ETCDUTL_BIN}"
-fi
-
-# Move static pod manifests out of MANIFEST_DIR, if required
-if [ -z "${ETCD_RESTORE_SKIP_MOVE_CP_STATIC_PODS}" ]; then
-  mv_static_pods "${AUX_STATIC_POD_LIST[@]}"
-  wait_for_containers_to_stop "${AUX_STATIC_POD_CONTAINERS[@]}"
 fi
 
 # always move etcd pod and wait for all containers to exit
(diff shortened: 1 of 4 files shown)


bindata/etcd/restore-pod.yaml
```diff
@@ -61,21 +58,18 @@ spec:
           fi
         fi
         
-        BUMP_ARGS=""
-        if [[ "${ETCD_ETCDCTL_RESTORE_ENABLE_BUMP}" == "true" ]]; then
-          echo "enabling restore bump"
-          BUMP_ARGS="--bump-revision 1000000000 --mark-compacted"
-        fi
-                
+        BUMP_REV=$(jq -r ".maxRaftIndex // 1000000000" -f "/var/lib/etcd-backup/revision.json")
```
tjungblu, line 61: need to still figure out how we can teach jq to do math, we need to add some percentage on top of the current maxRaftIndex to account for timing issues

bindata/etcd/etcd-common-tools
```diff
@@ -25,6 +26,9 @@ function dl_etcdctl {
        cp ${etcdmnt}/bin/etcdutl ${ETCDCTL_BIN_DIR}/
        export ETCD_ETCDUTL_BIN=etcdutl
      fi
+     if ! [ -x "$(command -v jq)" ]; then
+       cp ${etcdmnt}/bin/jq ${ETCDCTL_BIN_DIR}/
```
tjungblu, line 30: not sure it's even in this folder, tbd

tjungblu: seems to be the case:
```
sh-5.1# ls -l /bin/ | grep jq  
-rwxr-xr-x. 1 root root    32040 Oct 25  2023 jq
```

bindata/etcd/cluster-restore.sh
```diff
@@ -145,22 +145,19 @@ if [ -z "${ETCD_ETCDCTL_RESTORE}" ]; then
...
 
   echo "starting restore-etcd static pod"
-  # ideally this can be solved with jq and a real env var, but we don't have it available at this point
-  if [ -n "${ETCD_ETCDCTL_RESTORE_ENABLE_BUMP}" ]; then
-    sed "s/export ETCD_ETCDCTL_RESTORE_ENABLE_BUMP=\"false\"/export ETCD_ETCDCTL_RESTORE_ENABLE_BUMP=\"true\"/" "${RESTORE_ETCD_POD_YAML}" > "${MANIFEST_DIR}/etcd-pod.yaml"
-  else
-     cp -p "${RESTORE_ETCD_POD_YAML}" "${MANIFEST_DIR}/etcd-pod.yaml"
-  fi
-
+  cp -p "${RESTORE_ETCD_POD_YAML}" "${MANIFEST_DIR}/etcd-pod.yaml"
```
tjungblu, line 152: >2024-09-04T11:25:18.041313005+00:00 stderr F cannot access '/var/lib/etcd/etcd_perf*': No such file or directory
> 2024-09-04T11:25:18.044277499+00:00 stdout F please delete the contents of data directory before restoring, running the restore script will do this for you

bindata/etcd/cluster-restore.sh
```diff
@@ -145,21 +136,20 @@ if [ -z "${ETCD_ETCDCTL_RESTORE}" ]; then
 
   # Copy snapshot to backupdir
   cp -p "${SNAPSHOT_FILE}" "${ETCD_DATA_DIR_BACKUP}"/snapshot.db
+  # Move the revision.json to backupdir to avoid conflicts later on
+  mv "${ETCD_REV_JSON}" "${ETCD_DATA_DIR_BACKUP}"/revision.json
```
tjungblu, line 140: mv is not idempotent

bindata/etcd/cluster-restore.sh
```diff
@@ -96,10 +96,7 @@ BACKUP_FILE=$(ls -vd "${BACKUP_DIR}"/static_kuberesources*.tar.gz | tail -1) ||
 SNAPSHOT_FILE=$(ls -vd "${BACKUP_DIR}"/snapshot*.db | tail -1) || true
 
 ETCD_STATIC_POD_LIST=("etcd-pod.yaml")
-AUX_STATIC_POD_LIST=("kube-apiserver-pod.yaml" "kube-controller-manager-pod.yaml" "kube-scheduler-pod.yaml")
-
-ETCD_STATIC_POD_CONTAINERS=("etcd" "etcdctl" "etcd-metrics" "etcd-readyz")
-AUX_STATIC_POD_CONTAINERS=("kube-controller-manager" "kube-apiserver" "kube-scheduler")
```
hasbro17, line 102: Why are we not starting the auxillary static pods like the apiserver anymore? Doesn't our existing restore process depend on that as we restore to a single member control-plane node?
Or is that being handled elsewhere now?

tjungblu: why would we need this when we bump the revisions? The caches are wiped by the watches expiring through the compaction marker.

hasbro17: Ah my bad, I see the revision bump is on by default now with `ETCD_ETCDCTL_RESTORE_ENABLE_BUMP` gone.

So with this change we'll also need to update our [default DR workflow](https://docs.openshift.com/container-platform/4.16/backup_and_restore/control_plane_backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.html#dr-scenario-2-restoring-cluster-state_dr-restoring-cluster-state) docs as well  I think to exclude that part.

```
The cluster-restore.sh script must show that etcd, kube-apiserver, kube-controller-manager, and kube-scheduler pods are stopped and then started at the end of the restore process.
```

tjungblu: yep, I'm currently rewriting the entire doc for it

hasbro17: Apologies if I'm conflating this with the newer quorum-restore/force-new-cluster approach, but for our standard existing workflow do we still need to stop the auxillary control-plane pods on the other non-recovery hosts if we have the revision bump enabled by default?

i.e the following pre-recovery step:
https://docs.openshift.com/container-platform/4.16/backup_and_restore/control_plane_backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.html#dr-scenario-2-restoring-cluster-state_dr-restoring-cluster-state
```
4. Stop the static pods on any other control plane nodes.
...
d. Move the existing kube-apiserver file out of the kubelet manifest directory by running:

$ sudo mv -v /etc/kubernetes/manifests/kube-apiserver-pod.yaml /tmp
....
```
I'm guessing the API server and other pods would have their caches wiped so probably no need to restart them.

Normally we stop those but I forget if these non-recovery etcd pods also crash-loop after we restore the recovery member?

Trying to tally all the doc changes we'll need after this change.

tjungblu: checkout
https://docs.google.com/document/d/1oIlHD-gkkskQUNN7n1EWVq5vi3UCOWJHuY3QEzDKsYY/edit#heading=h.7ebq38ydvskj

tjungblu: >  for our standard existing workflow do we still need to stop the auxillary control-plane pods on the other non-recovery hosts if we have the revision bump enabled by default?

we will only have to stop etcd, everything else should fall into place. That's only for the restore from snapshot, quorum restore will automatically do that for us by stopping raft.

> Normally we stop those but I forget if these non-recovery etcd pods also crash-loop after we restore the recovery member?

when restoring from snapshot: no, 
when restoring with force-new-cluster: yes.

dusk125 (commented):
/lgtm

tjungblu:

unrelated failure to this PR

PR #1334, opened 2024-09-09
ETCD-666: Add etcd-backup-server pruning logic

Elbehery:
resolves https://issues.redhat.com/browse/ETCD-666

### DIFF
--- a/pkg/cmd/backuprestore/backupserver.go
+++ b/pkg/cmd/backuprestore/backupserver.go
@@ -8,11 +8,12 @@ import (
 	"syscall"
 	"time"
 
+	prune "github.com/openshift/cluster-etcd-operator/pkg/cmd/prune-backups"
+
 	"k8s.io/klog/v2"
 
 	"github.com/robfig/cron/v3"
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 )
 
 const backupVolume = "/var/lib/etcd-auto-backup"
@@ -20,15 +21,20 @@ const backupVolume = "/var/lib/etcd-auto-backup"
 var shutdownSignals = []os.Signal{os.Interrupt, syscall.SIGTERM}
 
 type backupRunner interface {
-	runBackup(opts *backupOptions) error
+	runBackup(*backupOptions, *prune.PruneOpts) error
 }
 
 type backupRunnerImpl struct{}
 
-func (b backupRunnerImpl) runBackup(opts *backupOptions) error {
+func (b backupRunnerImpl) runBackup(backupOpts *backupOptions, pruneOpts *prune.PruneOpts) error {
 	dateString := time.Now().Format("2006-01-02_150405")
-	opts.backupDir = backupVolume + dateString
-	err := backup(opts)
+	backupOpts.backupDir = backupVolume + dateString
+	err := backup(backupOpts)
+	if err != nil {
+		return err
+	}
+
+	err = pruneOpts.Run()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/cmd/backuprestore/backupserver.go
```diff
@@ -42,11 +48,16 @@ type backupServer struct {
...
+	prune.PruneOpts
 }
 
 func NewBackupServer(ctx context.Context) *cobra.Command {
 	backupSrv := &backupServer{
 		backupOptions: backupOptions{errOut: os.Stderr},
+		PruneOpts: prune.PruneOpts{
+			RetentionType: "None",
+			BackupPath:    backupVolume,
+		},
```
tjungblu, line 60: do we need to set this?

pkg/cmd/prune-backups/prune.go
```diff
@@ -33,10 +33,14 @@ type PruneOpts struct {
 	RetentionType      string
 	MaxNumberOfBackups int
 	MaxSizeOfBackupsGb int
+	BackupPath         string
```
tjungblu, line 36: can you change the comment at the top to reflect that we pass a path now?

pkg/cmd/prune-backups/prune.go
```diff
@@ -33,10 +33,14 @@ type PruneOpts struct {
...
 }
 
 func NewPruneCommand() *cobra.Command {
-	opts := PruneOpts{RetentionType: "None"}
+	opts := PruneOpts{
+		RetentionType: "None",
+		BackupPath:    BasePath,
```
tjungblu, line 42: let's make this a cmdline arg of the prune opts, you can still default to BasePath

pkg/cmd/prune-backups/prune_test.go
```diff
@@ -54,15 +54,15 @@ func TestRetainByNumber(t *testing.T) {
 		time.Sleep(1 * time.Second)
 	}
 
-	require.NoError(t, retainByNumber(6))
```
tjungblu, line 57: would you mind adding some tests that will test varying folders instead of hardcoding BasePath?

tjungblu: what also works for me: we do not have to change BasePath anymore, we can simply inject the tmp folder into the arguments

pkg/cmd/prune-backups/prune.go
tjungblu: and make `BasePath` a constant then, I think we can get away without the unit tests changing the path as a variable

pkg/cmd/backuprestore/backupserver.go
Elbehery: so the backup path is const [here](https://github.com/openshift/cluster-etcd-operator/blob/735f37b8d3938a03cdf1c74e61f634a8adcf97c2/pkg/cmd/prune-backups/prune.go#L18)

In order to user Prune, and use the NoConfig backup [volume](https://github.com/openshift/cluster-etcd-operator/blob/735f37b8d3938a03cdf1c74e61f634a8adcf97c2/pkg/cmd/backuprestore/backupserver.go#L18), I override the path here

please correct me if I am wrong

pkg/cmd/prune-backups/prune.go
Elbehery: done 👍🏽

pkg/cmd/backuprestore/backupserver.go
tjungblu: but is this correct?
because you're setting the backupDir path to "backupVolume + dateString"

tjungblu: nvm, that's just the suffix for each backup invocation

Elbehery: genauuu :D :D

tjungblu:

unrelated version operator failure

PR #1340, opened 2024-09-18
ETCD-662: fix: etcd-backup-server prune path

Elbehery:
During debugging [e2e](https://github.com/openshift/origin/pull/29098), found a bug within `etcd-backup-server`. 

cc @openshift/openshift-team-etcd

PR #1341, opened 2024-09-20
ETCD-662: fix: etcd-backup-server prune path

Elbehery:
follow up on https://github.com/openshift/cluster-etcd-operator/pull/1340

### DIFF
--- a/pkg/cmd/backuprestore/backupserver.go
+++ b/pkg/cmd/backuprestore/backupserver.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"path"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/backuprestore/backupserver.go
```diff
@@ -34,7 +34,6 @@ func (b backupRunnerImpl) runBackup(backupOpts *backupOptions, pruneOpts *prune.
 		return err
 	}
 
-	pruneOpts.BackupPath = backupVolume + "/" + dateString
```
tjungblu, line 37: now that I'm reading this again, `paths.Join` would also be a safer option

Elbehery: yes u r right, should have written test :'( 

fixing ...

Elbehery: updated 👍🏽

PR #1346, opened 2024-09-30
ETCD-678: introduce a templated pod.yaml

tjungblu:
This moves the pod.yaml to a go templated one, with all values replaced by a proper templating engine intead of string replacement.

### DIFF
--- a/pkg/backuphelpers/backupvars.go
+++ b/pkg/backuphelpers/backupvars.go
@@ -18,6 +18,7 @@ type BackupVar interface {
 	AddListener(listener Enqueueable)
 	SetBackupSpec(spec *backupv1alpha1.EtcdBackupSpec)
 	ArgString() string
+	ArgList() []string
 }
 
 type BackupConfig struct {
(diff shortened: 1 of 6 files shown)


pkg/backuphelpers/backupvars.go
```diff
@@ -55,6 +56,34 @@ func (b *BackupConfig) SetBackupSpec(spec *backupv1alpha1.EtcdBackupSpec) {
 	}
 }
 
+func (b *BackupConfig) ArgList() []string {
```
hasbro17, line 59: The templating changes for etcd_assets and the targetconfigcontroller look good but is this related to that?
I can see the corresponding test for this but not sure where we're invoking `ArgList()`?
Wasn't sure if this is meant for the template FuncMap?

(Or my brain is mush right now and I can't grep it :P)

tjungblu: right here: https://github.com/openshift/cluster-etcd-operator/pull/1346/files#diff-0af15f182aea9898e784fc589175ef4c2411c9391bb5a1f41327ba7c3e2172f5R226

PR #1362, opened 2024-10-25
OCPBUGS-48153: certsignercontroller: avoid using resourceapply.ApplySecret 

vrutkovs:
This method makes an extra GET, which is costly for such a large secret (due to its size mostly). Instead it uses a version from cache, runs a single GET and uses server-side apply to update it.

Amount of GET requests does on this PR (yellow) and 4.18 RC (orange) from audit logs:
image


vrutkovs:
/hold

This causes too many invalid applies:
```
event happened 39 times, something is wrong: namespace/openshift-etcd-operator deployment/etcd-operator hmsg/e9eab99c6c - reason/SecretUpdated Updated Secret/etcd-all-certs -n openshift-etcd because it changed (10:14:01Z) result=reject
```

vrutkovs:
Yeah, this won't work. The tradeoff here is that the informer may be outdated so we'll issue more rejected applies instead.

PR #1364, opened 2024-10-31
OCPBUGS-43748: etcd pod containers do not start when tls min version is 1.3

sanchezl:
- **add ETCD_TLS_MIN_VERSION env to etcd pods**
- **add listen-tls-min-version flag to readyz cmd**
- **do not start etcd with --cipher-suites with tls1.3**

### DIFF
--- a/bindata/etcd/pod.gotpl.yaml
+++ b/bindata/etcd/pod.gotpl.yaml
@@ -131,9 +131,9 @@ spec:
       - mountPath: /var/lib/etcd/
         name: data-dir
     env:
-    {{ range $i, $k := .EnvVars -}}
-    - name: {{ $k.Name | quote }}
-      value: {{ $k.Value | quote }}
+    {{ range .EnvVars -}}
+    - name: {{ .Name | quote }}
+      value: {{ .Value | quote }}
     {{ end -}}
     - name: "ETCD_STATIC_POD_VERSION"
       value: "REVISION"
@@ -195,10 +195,10 @@ spec:
           --metrics=extensive \
           --listen-metrics-urls=https://{{.ListenAddress}}:9978 ||  mv /etc/kubernetes/etcd-backup-dir/etcd-member.yaml /etc/kubernetes/manifests
     env:
-    {{ range $i, $k := .EnvVars -}}
-    - name: {{ $k.Name | quote }}
-      value: {{ $k.Value | quote }}
-    {{ end -}}
+    {{- range .EnvVars }}
+    - name: {{ .Name | quote }}
+      value: {{ .Value | quote }}
+    {{- end }}
     - name: "ETCD_STATIC_POD_VERSION"
       value: "REVISION"
     resources:
@@ -268,11 +268,14 @@ spec:
           --cert-file /etc/kubernetes/static-pod-certs/secrets/etcd-all-certs/etcd-serving-metrics-NODE_NAME.crt \
           --cacert /etc/kubernetes/static-pod-certs/configmaps/etcd-all-bundles/server-ca-bundle.crt \
           --trusted-ca-file /etc/kubernetes/static-pod-certs/configmaps/etcd-all-bundles/metrics-ca-bundle.crt \
-          --listen-cipher-suites ${ETCD_CIPHER_SUITES}
+        {{- if .CipherSuites }}
+          --listen-cipher-suites {{ .CipherSuites }} \
+          {{ end -}}
+          --tls-min-version $(ETCD_TLS_MIN_VERSION)
     env:
-    {{ range $i, $k := .EnvVars -}}
-    - name: {{ $k.Name | quote }}
-      value: {{ $k.Value | quote }}
+    {{- range .EnvVars }}
+    - name: {{ .Name | quote }}
+      value: {{ .Value | quote }}
     {{ end -}}
     - name: "ETCD_STATIC_POD_VERSION"
       value: "REVISION"
@@ -308,7 +311,10 @@ spec:
           --client-cert-file=$(ETCDCTL_CERT) \
           --client-key-file=$(ETCDCTL_KEY) \
           --client-cacert-file=$(ETCDCTL_CACERT) \
-          --listen-cipher-suites=$(ETCD_CIPHER_SUITES)
+        {{- if .CipherSuites }}
+          --listen-cipher-suites {{ .CipherSuites }} \
+          {{ end -}}
+          --listen-tls-min-version=$(ETCD_TLS_MIN_VERSION)
     securityContext:
       privileged: true
     ports:
@@ -320,9 +326,9 @@ spec:
         memory: 50Mi
         cpu: 10m
     env:
-    {{ range $i, $k := .EnvVars -}}
-    - name: {{ $k.Name | quote }}
-      value: {{ $k.Value | quote }}
+    {{ range .EnvVars -}}
+    - name: {{ .Name | quote }}
+      value: {{ .Value | quote }}
     {{ end -}}
     volumeMounts:
       - mountPath: /var/log/etcd/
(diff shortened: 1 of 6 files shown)


pkg/etcdenvvar/etcd_env.go
```diff
@@ -304,6 +308,33 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
 	}, nil
 }
 
+func getTLSMinVersion(envVarContext envVarContext) (map[string]string, error) {
```
tjungblu, line 311: given your extraordinary gotpl skills, I'm wondering whether we should just check this inside of the cipher suite function above in conjunction.

That way we can enable/disable the suites depending on the TLS version. I'm mostly curious about how we can surface this back to a user somehow. I know some (especially FIPS) customers run compliance scanners that will alert them on not setting those ciphers.

pkg/operator/starter.go
```diff
@@ -173,7 +174,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	// we keep the default behavior of exiting the controller once a gate changes
 	featureGateAccessor.SetChangeHandler(featuregates.ForceExit)
 
-	operatorClient, dynamicInformers, err := genericoperatorclient.NewStaticPodOperatorClient(controllerContext.KubeConfig, operatorv1.GroupVersion.WithResource("etcds"), operatorv1.GroupVersion.WithKind("Etcd"), ExtractStaticPodOperatorSpec, ExtractStaticPodOperatorStatus)
+	operatorClient, dynamicInformers, err := genericoperatorclient.NewStaticPodOperatorClient(clock.RealClock{}, controllerContext.KubeConfig, operatorv1.GroupVersion.WithResource("etcds"), operatorv1.GroupVersion.WithKind("Etcd"), ExtractStaticPodOperatorSpec, ExtractStaticPodOperatorStatus)
```
tjungblu, line 177: do you mind to split out the library-go update into a separate PR? we can merge that earlier, I see there is a bunch of static pod changes there too, which we might want to separate out - just in case they mess up CI somehow.

tjungblu: oh you already got #1365 - resolved :) thanks!

pkg/etcdenvvar/etcd_env.go
sanchezl: Knowing that the setting is ignored, even if we were to pass it along, I agree.
We could output a `CIPER_SUITES` env var when tls min version == "TLS1.3"?
I would say that a scanner would have to evolve to know that using TLS1.3 sets the ciphers automatically.
I think enabling FIPS also limits the TLS1.3 ciphers. We'll have to confirm that we output the correct list.

pkg/cmd/readyz/readyz.go
```diff
@@ -178,12 +180,21 @@ func (r *readyzOpts) Run() error {
 		return err
 	}
 
+	var tlsMinVersion uint16
+	switch tlsutil.TLSVersion(r.tlsMinVersion) {
+	case tlsutil.TLSVersion13:
+		tlsMinVersion = tls.VersionTLS13
+	case tlsutil.TLSVersion12:
+		tlsMinVersion = tls.VersionTLS12
+	}
```
tjungblu, line 189: "go.etcd.io/etcd/client/pkg/v3/tlsutil" to the rescue :)

pkg/etcdenvvar/etcd_env.go
```diff
@@ -299,8 +336,20 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
...
+	observedMinTLSVersion, err := getObservedTLSMinVersion(envVarContext)
+	if err != nil {
+		return nil, fmt.Errorf("unable to compute ETCD_CIPHER_SUITES: %v", err)
+	}
+
+	envName := "ETCD_CIPHER_SUITES"
+	if observedMinTLSVersion == tlsutil.TLSVersion13 {
+		// When --tls-min-version is set to 'TLS1.3', etcd does not allow --cipher-suites to also be specified.
+		// We still outout an env var for informational purposes.
+		envName = "CIPHER_SUITES"
```
tjungblu, line 348: haha nice trick

tjungblu:
looks good to me, tests are failing though

bindata/etcd/pod.gotpl.yaml
```diff
@@ -268,11 +268,14 @@ spec:
           --cert-file /etc/kubernetes/static-pod-certs/secrets/etcd-all-certs/etcd-serving-metrics-NODE_NAME.crt \
           --cacert /etc/kubernetes/static-pod-certs/configmaps/etcd-all-bundles/server-ca-bundle.crt \
           --trusted-ca-file /etc/kubernetes/static-pod-certs/configmaps/etcd-all-bundles/metrics-ca-bundle.crt \
-          --listen-cipher-suites ${ETCD_CIPHER_SUITES}
+        {{- range .EnvVars }}{{ if eq .Name "ETCD_CIPHER_SUITES" }}
```
tjungblu, line 271: if you want to make this logic a bit easier to test, you can also set a specific variable for the template here:
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L203-L220

makes this a bit more explicit and easier to test from the targetconfigcontroller I think

pkg/cmd/readyz/readyz.go
tjungblu: https://github.com/etcd-io/etcd/blob/19aa0dbe8fd6317a237bae9b6ea52a4f1b445b19/client/pkg/tlsutil/versions.go#L32-L47

to be precise

pkg/etcdenvvar/etcd_env.go
```diff
@@ -283,6 +287,39 @@ func getUnsupportedArch(_ envVarContext) (map[string]string, error) {
...
+	v, err := crypto.TLSVersion(observedMinTLSVersion)
+	if err != nil {
+		return "", fmt.Errorf("couldn't get minTLSVersion from observedConfig: %w", err)
+	}
+	switch v {
+	case tls.VersionTLS13:
+		return tlsutil.TLSVersion13, nil
+	default:
+		return tlsutil.TLSVersion12, nil
+	}
```
tjungblu, line 310: this already exists as a dedicated function in https://github.com/etcd-io/etcd/blob/19aa0dbe8fd6317a237bae9b6ea52a4f1b445b19/client/pkg/tlsutil/versions.go#L32-L47

sanchezl: That does not seem to help me here. I'm going from a library-go `TLSVersion1x` string to an etcd `TLS1.x` string.

tjungblu: oh I see, it's not the integer- sorry!

pkg/cmd/readyz/readyz.go
sanchezl: Got it this time :-)

bindata/etcd/pod.gotpl.yaml
sanchezl: Done.

pkg/cmd/readyz/readyz.go
```diff
@@ -91,6 +92,7 @@ func (r *readyzOpts) AddFlags(cmd *cobra.Command) {
 	fs := cmd.Flags()
 	fs.Uint16Var(&r.listenPort, "listen-port", r.listenPort, "Listen on this port. Default 9980")
 	fs.StringSliceVar(&r.cipherSuites, "listen-cipher-suites", r.cipherSuites, "readyZ server TLS cipher suites.")
+	fs.StringVar(&r.tlsMinVersion, "listen-tls-min-version", r.tlsMinVersion, "Minimum TLS version supported by readyZ server. Possible values: TLS1.2, TLS1.3.")
```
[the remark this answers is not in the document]
sanchezl: I purposely left it out so that the "Go" default is used.

PR #1369, opened 2024-11-19
OCPBUGS-42810: actively move bootstrap member lead

tjungblu:
This PR will actively try to move the leadership away from the bootstrap member to another healthy member.

/hold

### DIFF
--- a/pkg/operator/bootstraptest/bootstrap_lead_test_controller.go
+++ b/pkg/operator/bootstraptest/bootstrap_lead_test_controller.go
@@ -1,0 +1,66 @@
+package bootstraptest
+
+import (
+	"context"
+	"time"
+
+	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/ceohelpers"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/health"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	"k8s.io/klog/v2"
+)
+
+type BootstrapTestController struct {
+	etcdClient etcdcli.EtcdClient
+}
+
+func NewBootstrapTestController(
+	etcdClient etcdcli.EtcdClient,
+	eventRecorder events.Recorder,
+) factory.Controller {
+
+	evc := eventRecorder.WithComponentSuffix("bootstrap-test-controller")
+	c := &BootstrapTestController{
+		etcdClient: etcdClient,
+	}
+
+	syncer := health.NewDefaultCheckingSyncWrapper(c.sync)
+	return factory.New().ResyncEvery(2*time.Minute).WithSync(syncer.Sync).
+		ToController("BootstrapTestController", evc)
+}
+
+func (c *BootstrapTestController) sync(ctx context.Context, _ factory.SyncContext) error {
+
+	members, err := c.etcdClient.MemberList(ctx)
+	if err != nil {
+		return err
+	}
+
+	var hasBootstrap bool
+	var bootstrapMember *etcdserverpb.Member
+	for _, member := range members {
+		if member.Name == "etcd-bootstrap" {
+			hasBootstrap = true
+			bootstrapMember = member
+			break
+		}
+	}
+
+	if hasBootstrap {
+		klog.Warningf("TEST_ONLY moving leader to the bootstrap member")
+		moved, err := ceohelpers.MoveLeaderToAnotherMember(ctx, c.etcdClient, bootstrapMember, members)
+		if err != nil {
+			return err
+		}
+		if moved {
+			klog.Warningf("TEST_ONLY successfully moved to the bootstrap member")
+		} else {
+			klog.Warningf("TEST_ONLY failed to move to the bootstrap member")
+		}
+	}
+
+	return nil
+}
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -19,6 +19,7 @@ import (
 	operatorversionedclient "github.com/openshift/client-go/operator/clientset/versioned"
 	operatorversionedclientv1alpha1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1alpha1"
 	operatorv1informers "github.com/openshift/client-go/operator/informers/externalversions"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/bootstraptest"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
@@ -467,6 +468,9 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		configInformers.Config().V1().Infrastructures().Lister(),
 	)
 
+	// TODO(thomas): TEST ONLY, REMOVE AGAIN
+	bootstrapTestController := bootstraptest.NewBootstrapTestController(etcdClient, controllerContext.EventRecorder)
+
 	scriptController := scriptcontroller.NewScriptControllerController(
 		AlivenessChecker,
 		operatorClient,
@@ -640,6 +644,8 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	go clusterMemberController.Run(ctx, 1)
 	go etcdMembersController.Run(ctx, 1)
 	go bootstrapTeardownController.Run(ctx, 1)
+	// TODO(thomas): remove again
+	go bootstrapTestController.Run(ctx, 1)
 	go unsupportedConfigOverridesController.Run(ctx, 1)
 	go scriptController.Run(ctx, 1)
 	go defragController.Run(ctx, 1)


tjungblu:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.18-e2e-azure-ovn-upgrade 10

tjungblu:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.18-e2e-azure-ovn-upgrade 10

tjungblu:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.18-e2e-azure-ovn-upgrade 10

tjungblu:
/retest

/payload-aggregate periodic-ci-openshift-release-master-ci-4.18-e2e-azure-ovn-upgrade 10

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -62,12 +63,18 @@ func (c *BootstrapTeardownController) sync(ctx context.Context, _ factory.SyncCo
...
-	safeToRemoveBootstrap, hasBootstrap, bootstrapID, err := c.canRemoveEtcdBootstrap(ctx, scalingStrategy)
+	safeToRemoveBootstrap, hasBootstrap, bootstrapMember, err := c.canRemoveEtcdBootstrap(ctx, scalingStrategy)
 	if err != nil {
 		return fmt.Errorf("error while canRemoveEtcdBootstrap: %w", err)
 	}
 
-	err = c.removeBootstrap(timeoutCtx, safeToRemoveBootstrap, hasBootstrap, bootstrapID)
+	if hasBootstrap {
+		if err := c.ensureBootstrapIsNotLeader(ctx, bootstrapMember); err != nil {
+			klog.Errorf("error while ensuring bootstrap is not leader: %v", err)
```
lance5890, line 73: should we return err here?

tjungblu: not really, I wouldn't want the controller to degrade when that operation fails once. Mind you, this happens very very rarely and affects our CI pass rates in very few cases of hundreds.

tjungblu:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.18-e2e-azure-ovn-upgrade 10

tjungblu:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.20-e2e-azure-ovn-upgrade 10

tjungblu:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.20-e2e-azure-ovn-upgrade 20

tjungblu:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.21-e2e-azure-ovn-upgrade 20

pkg/operator/ceohelpers/leader.go
```diff
@@ -0,0 +1,52 @@
...
+	if otherMember == nil {
+		return false, fmt.Errorf("no follower member found for the members: %v", memberList)
+	}
+
+	err := client.MoveLeader(ctx, otherMember.ID)
+	if err != nil {
+		return false, err
+	}
+
+	klog.Warningf("Moving lead from member [%x] (%s) to [%x] (%s) succesfully!", leader.ID, leader.GetClientURLs()[0], otherMember.ID, otherMember.GetClientURLs()[0])
```
hasbro17: Super pedantic nit since you're already fixing spelling mistakes in this PR
```suggestion
	klog.Warningf("Moving lead from member [%x] (%s) to [%x] (%s) successfully!", leader.ID, leader.GetClientURLs()[0], otherMember.ID, otherMember.GetClientURLs()[0])
```

pkg/operator/ceohelpers/leader.go
```diff
@@ -0,0 +1,52 @@
...
+	var otherMember *etcdserverpb.Member
+	for _, member := range memberList {
+		if member.ID != leader.ID {
+			otherMember = member
+			break
+		}
+	}
+
+	if otherMember == nil {
+		return false, fmt.Errorf("no follower member found for the members: %v", memberList)
```
hasbro17: Just some more context on the operation in case this ever does happen.
```suggestion
		return false, fmt.Errorf("no follower member found for leadership transfer: %v", memberList)
```

pkg/operator/ceohelpers/leader.go
hasbro17: Also while we're here, does `MoveLeader() == nil` mean the leadership was actually transferred, or we've only made a request to the leader, and it may still be in the process of stepping down or we're in the midst of an election.

Asking since the warning here says `Moving` instead of `Moved`.

hasbro17: Okay so it seems like it waits for the actual transfer to finish and not just a shallow request.
https://github.com/openshift/cluster-etcd-operator/blob/61b7fb5116f8672bab6fcccbb9336e197ac07602/vendor/go.etcd.io/etcd/server/v3/etcdserver/server.go#L1513-L1520

In that case wdyt about the past tense:
```suggestion
	klog.Warningf("Moved lead from member [%x] (%s) to [%x] (%s) successfully!", leader.ID, leader.GetClientURLs()[0], otherMember.ID, otherMember.GetClientURLs()[0])
```

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
```diff
@@ -147,6 +150,29 @@ func (c *BootstrapTeardownController) removeBootstrap(ctx context.Context, safeT
...
 		return err
 	}
+
+	moved, err := c.ensureBootstrapIsNotLeader(ctx, bootstrapMember)
+	if err != nil {
+		return err
+	}
+
+	// if we have just moved it, we will skip this sync iteration to backoff the controller - the next resync will happen after a minute anyway
+	if moved {
```
hasbro17, line 160: This seems okay so not disagreeing here, but for my knowledge what is the potential negative impact of removing the member right after we transfer?
At first I thought we're waiting for the transfer to happen but as I found out below `MoveLeader()` seems to be synchronous.

If we are indeed waiting for **something** to stabilize then just thinking if it's possible that may not happen by the next resync.

Or just so we get a chance to update the status first.

pkg/operator/ceohelpers/leader.go
tjungblu: yep, sounds good! thanks for proof reading this. I don't really understand why my spell checker in Goland picks up on the bOOstrap but not on the success :+1:

pkg/operator/ceohelpers/leader.go
tjungblu: thank you, applied!

pkg/operator/bootstrapteardown/bootstrap_teardown_controller.go
tjungblu: > If we are indeed waiting for something to stabilize then just thinking if it's possible that may not happen by the next resync.

great question: what are we waiting on here. My intuition was to avoid disrupting the API too much, the clients will get the LeaderChangedErr (`Error from server: etcdserver: leader changed`) on the move. Then potentially removing the member could also disrupt it further, e.g. during a static pod rollout. 

I know this is highly hypothetical... 
Unfortunately, I do not have any data yet on whether this would be really necessary. It is just too damn hard to reproduce this case. 

Let me quickly put together a controller that will always make the bootstrap node the leader during the bootstrapping phase.

tjungblu: here's a run with this controller, you can grep the logs with TEST_ONLY

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1369/pull-ci-openshift-cluster-etcd-operator-main-e2e-azure/1967558360303669248

https://grafana-loki.ci.openshift.org/explore?orgId=1&left=%7B%22datasource%22:%22PCEB727DF2F34084E%22,%22queries%22:%5B%7B%22expr%22:%22%7Binvoker%3D%5C%22openshift-internal-ci%2Fpull-ci-openshift-cluster-etcd-operator-main-e2e-azure%2F1967558360303669248%5C%22%7D%20%7C~%20%5C%22TEST_ONLY%7Cremoved%20bootstrap%5C%22%22,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221757931212276%22,%22to%22:%221757967212278%22%7D%7D

it also seems there was no actual move being done before removing, even though we have moved the leader to the bootstrap node a minute earlier.

it still kinda shows what I was worried about:
> I0915 13:00:26.367586       1 event.go:377] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"6e338218-3237-4f73-9c33-f83a5f40b8a0", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'Bootstrap member removed' successfully removed bootstrap member [32326af0e011bf0a]

> E0915 13:00:26.377135       1 event.go:359] "Server rejected event (will not retry!)" err="rpc error: code = Unknown desc = **raft: stopped**" event="&Event{ObjectMeta:{etcd-operator.1865763b3da4743d  openshift-etcd-operator    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] [] 
 

it seems that this is the case of scaling from 4->3, but during a static pod rollout, so that breaks quorum and causes raft to stop. 

Anyway, I need a bit more time to test this, so I'll pick this up again when I have some time left with RIT :)

PR #1374, opened 2024-12-03
OCPBUGS-45335: Configure removal of etcd container from static pod ma…

tjungblu:
…nifest


tjungblu:
@sandeepknd could you CPA here please? nothing to test, it's just a flag for the two node folks

PR #1399, opened 2025-03-10
OCPEDGE-1498: support updates to external etcd

clobrano:
The _EnableEtcdContainer unsupported config override_ removes the etcd container from the static pod definition when set to FALSE.

This change extends that behavior by writing a separate Pod definition containing **only** the etcd container, using the same source from which it was removed, but placing it in a location that does not trigger pod creation.

This allows an external etcd manager to run etcd independently while ensuring it has a reliable source for updates from CEO (e.g., new image, path to certificates, etc.).

### DIFF
--- a/pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
+++ b/pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
@@ -1,0 +1,192 @@
+package externaletcdsupportcontroller
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/ghodss/yaml"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	"github.com/openshift/cluster-etcd-operator/pkg/etcdenvvar"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/ceohelpers"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/etcd_assets"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/health"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
+	"github.com/openshift/cluster-etcd-operator/pkg/operator/targetconfigcontroller"
+	"github.com/openshift/cluster-etcd-operator/pkg/version"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+type ExternalEtcdEnablerController struct {
+	operatorClient v1helpers.StaticPodOperatorClient
+
+	targetImagePullSpec   string
+	operatorImagePullSpec string
+	envVarGetter          etcdenvvar.EnvVar
+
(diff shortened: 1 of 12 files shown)


dusk125:
Can we flip the logic such that we're enabling this override instead of disabling default behavior; I'd like it to be very clear that if someone were to stumble upon this, it is immediately clear that they're opting into something. This would mirror our other override `oc patch etcd/cluster --type=merge -p '{"spec": {"unsupportedConfigOverrides": {"useUnsupportedUnsafeNonHANonProductionUnstableEtcd": true}}}'`.

Also, can we put this behind the 2NO feature gate please?

clobrano:
Sure, it makes sense. I'll update the PR

clobrano:
**still testing enablement via feature gate**, but it is ready for an initial review :bow:

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
```diff
@@ -0,0 +1,192 @@
...
+func (c *ExternalEtcdEnablerController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+
+	operatorSpec, _, _, err := c.operatorClient.GetStaticPodOperatorState()
+	if err != nil {
+		return err
+	}
+
+	envVars := c.envVarGetter.GetEnvVars()
+	if len(envVars) == 0 {
+		// note this will not degrade the controller, that can happen during CEO restarts often due to cold informer caches (expected)
```
tjungblu, line 98: precisely why you should use `AddListener` and only call `GetEnvVars` after you got a callback

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
```diff
@@ -0,0 +1,192 @@
...
+		OperatorImage:       operatorImagePullSpec,
+		ListenAddress:       "0.0.0.0",   // TODO this needs updating to detect ipv6-ness
+		LocalhostAddress:    "127.0.0.1", // TODO this needs updating to detect ipv6-ness
+		LogLevel:            "info",
+		EnvVars:             nameValues,
+		EnableEtcdContainer: true,
+	}
+}
+
+func (c *ExternalEtcdEnablerController) supportExternalPod(ctx context.Context, subs *targetconfigcontroller.PodSubstitutionTemplate,
```
tjungblu, line 132: I wonder why we need to replicate all the logic here, can't you copy the rendered pod template that we "ignore" from in `manageStandardPod` of the targetconfigcontroller?

clobrano: Moreover this new function render the data in JSON format, not in YAML, as the Node does not have jq and an external etcd manager (like the pacemaker agent in TNF) needs to edit the manifest

clobrano:
Tested on OCP 4.19, feature gate included and added tests.

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
tjungblu: then I wonder why you need that length check: there must be something in that slice after that callback. Or is there an informer sync missing on the controller?

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
tjungblu: I leave this to Allen, but I personally prefer to have the templating logic not duplicated across two concurrently running controllers. We can serialize json or yaml on the other controller too.

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
clobrano: honestly, I took that check from the other controllers, assuming it was useful :grin:

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
clobrano: I could rework both functions to collect same/similar code in a third function and reuse that.
Maybe ceohelpers.go is a good fit?

clobrano:
Added a check to create the controller only if topology is DualReplicaTopology

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
clobrano: I moved the common logic into a new file under ceohelpers `pkg/operators/ceohelpers/podsubstitution.go`

dusk125:
I think it looks good, I'm going to give it another read through tomorrow morning, but non binding lgtm for now
/retest

PR #1400, opened 2025-03-18
OCPEDGE-1495: Update CEO to run pacemaker initialization commands

slintes:
This is adding new two node fencing (TNF) controllers to CEO, which handles pacemaker configuration and etcd container transition from CEO to pacemaker.

Design and more context: https://docs.google.com/document/d/1ZqRYPM2OZ6kdD_tua4lB-gf3iSL-cuAoicvqIbM10MQ/edit?usp=sharing


pkg/tnf-operator/pcs/cluster.go
```diff
@@ -0,0 +1,76 @@
...
+		return false, nil
+	}
+
+	klog.Info("Starting initial HA cluster setup")
+
+	stonithArgsNode1 := getStonithArgs(cfg, 0)
+	stonithArgsNode2 := getStonithArgs(cfg, 1)
+
+	commands := []string{
+		fmt.Sprintf("/usr/sbin/pcs cluster setup TNF %s addr=%s %s addr=%s --debug --force", cfg.NodeName1, cfg.NodeIP1, cfg.NodeName2, cfg.NodeIP2),
```
tjungblu, line 33: does this need to be executed on every node?

bindata/tnfdeployment/deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+          resources:
+            requests:
+              cpu: 50m
+              memory: 64Mi
+            limits:
+              cpu: 500m
+              memory: 128Mi
+          securityContext:
+            privileged: true
+            allowPrivilegeEscalation: true
```
tjungblu, line 35: do you guys really need root privs for this?

Dockerfile.tnf.ocp
```diff
@@ -0,0 +1,10 @@
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19 AS builder
```
tjungblu, line 1: I wonder if you guys can just grab an entire folder with a dedicated OWNERS file.

Remind me again: if you need a new container image entirely, how is that different to keep your entire operator in a separate repository? or is this just a copy+paste leftover from your PoC?

pkg/tnf-operator/pcs/cluster.go
tjungblu: > After that, things need to be executed on one node only

and you can't do this as part of the machine config itself? in a kind-of a first-one-wins manner?

I think this is also more of a job than a deployment, isn't it?

pkg/operator/starter.go
```diff
@@ -281,6 +282,24 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+		tnfDeploymentReconciler = tnfdeploymentcontroller.NewTnfDeploymentController(
+			ctx,
+			operatorClient,
+			kubeClient,
+			kubeInformersForNamespaces.InformersFor("openshift-etcd"),
+			kubeInformersForNamespaces,
+			configInformers.Config().V1().Infrastructures(),
+			envVarController,
+			controllerContext.EventRecorder,
+			os.Getenv("IMAGE"),
```
clobrano, line 299: couldn't you call `tnfDeploymentReconciler.Run(...)` here, instead than guarding the call later with a ` if tnfDeploymentReconciler != nil` ?

clobrano: I said that because a bunch of other controllers do it similarly (e.g. https://github.com/openshift/cluster-etcd-operator/blob/ba5c2e88e7c489978fa847aee8afc8fb99a30e65/pkg/operator/starter.go#L522-L524)

pkg/tnf-operator/pcs/cluster.go
tjungblu: > I think this can be a conflict at least when both nodes boot at the same time, and both try to configure pacemaker at the same time

and when you backoff and retry for a few times until it settles on a configuration? Is it really likely is it that they boot the exact same time and configure themselves _differently_ on the same nanosecond? 

> Also, fencing setup needs some credentials, which are read from a secret. I'm not sure if we can easily access the cluster from a systemd unit...

we can have the static pod controller sync that secret into the /etc/kubernetes/manifest folder if you like

bindata/tnfdeployment/deployment.yaml
tjungblu: the CEO namespace can't as far as I understand your design doc, but can the `openshift-etcd` host those? 
Creating a new namespace with those settings is going to set off a whole barrage of security scanners that I don't want in my bug backlog :/

pkg/tnf-operator/pcs/cluster.go
tjungblu: summarizing the slack thread we had:
* yes this can be done safely with a bash script in systemd
* we can ship this as part of the MCO configs
* secrets can be planted on each node via static pod controller
* how do we propagate script errors? how does kubelet do it?

tjungblu: I think this looks good

Dockerfile.tnf.ocp
tjungblu: Thank you!

bindata/tnfdeployment/authjob.yaml
```diff
@@ -0,0 +1,35 @@
...
+          securityContext:
+            privileged: true
+            allowPrivilegeEscalation: true
+      hostIPC: false
+      hostNetwork: false
+      hostPID: true
+      serviceAccountName: tnf-setup-manager
+      terminationGracePeriodSeconds: 10
+      restartPolicy: Never
+  backoffLimit: 3
```
jaypoulz, line 35: I don't know that we want a backoff limit for this job. I tested this in AWS, and it failed because of firewalls blocking pcs traffic, which was something I forgot about entirely.

Instead of a back-off, is there a way to rate-limit them so it only retrys once a minute or something?

bindata/tnfdeployment/authjob.yaml
```diff
@@ -0,0 +1,35 @@
...
+  name: tnf-auth
+spec:
+  template:
+    metadata:
+      annotations:
+        openshift.io/required-scc: "privileged"
+    spec:
+      containers:
+        - name: tnf-auth
+          image: quay.io/openshift/origin-cluster-etcd-operator
```
tjungblu, line 16: you may want to update that image

bindata/tnfdeployment/job.yaml
```diff
@@ -0,0 +1,35 @@
...
+  name: tnf-setup
+spec:
+  template:
+    metadata:
+      annotations:
+        openshift.io/required-scc: "privileged"
+    spec:
+      containers:
+        - name: tnf-setup
+          image: quay.io/openshift/origin-cluster-etcd-operator
```
tjungblu, line 16: same image concern as above

cmd/cluster-etcd-operator/main.go
```diff
@@ -9,26 +9,26 @@ import (
 	"os"
 	"time"
 
-	prune_backups "github.com/openshift/cluster-etcd-operator/pkg/cmd/prune-backups"
+	"github.com/openshift/library-go/pkg/operator/staticpod/certsyncpod"
```
tjungblu, line 12: do you mind to revert this entire file? seems its just reorganized imports

pkg/tnf/auth/runner.go
```diff
@@ -0,0 +1,93 @@
...
+	command := fmt.Sprintf("echo %q > %s", tokenValue, tokenPath)
+	_, _, err = exec.Execute(ctx, command)
+	if err != nil {
+		return err
+	}
+	command = fmt.Sprintf("chmod 0600 %s", tokenPath)
+	_, _, err = exec.Execute(ctx, command)
+	if err != nil {
+		return err
+	}
```
tjungblu, line 75: that sounds like something you can directly do in go ;)

pkg/tnf/operator/starter.go
```diff
@@ -0,0 +1,177 @@
...
+
+			// we need node names for assigning auth jobs to specific nodes
+			klog.Infof("watching for nodes...")
+			_, err := controlPlaneNodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+				AddFunc: func(obj interface{}) {
+					node, ok := obj.(*corev1.Node)
+					if !ok {
+						klog.Warningf("failed to convert node to Node %+v", obj)
+					}
+					runTnfAuthJobController(ctx, node.GetName(), controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
```
tjungblu, line 66: is this not called once per node? and then every time the node gets removed and added again later?

pkg/tnf/operator/starter.go
```diff
@@ -0,0 +1,177 @@
...
+		tnf_assets.MustAsset("tnfdeployment/authjob.yaml"),
+		controllerContext.EventRecorder,
+		operatorClient,
+		kubeClient,
+		kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Batch().V1().Jobs(),
+		[]factory.Informer{},
+		[]tnflibgo.JobHookFunc{
+			func(_ *operatorv1.OperatorSpec, job *batchv1.Job) error {
+				// set operator image pullspec
+				job.Spec.Template.Spec.Containers[0].Image = os.Getenv("OPERATOR_IMAGE")
```
tjungblu, line 134: ah I see. Can you make the yaml files aware that this will be injected? add a comment or "<image>" so it doesn't inadvertently parse as an image we try to pull

pkg/tnf/pkg/config/cluster.go
```diff
@@ -0,0 +1,127 @@
...
+	nodes, err := kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return clusterCfg, err
+	}
+
+	sort.Slice(nodes.Items, func(i, j int) bool {
+		return nodes.Items[i].Name < nodes.Items[j].Name
+	})
+
+	for i, node := range nodes.Items {
```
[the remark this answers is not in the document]
tjungblu: it's better to ensure those are ordered, please keep the sorting

pkg/tnf/pkg/config/cluster.go
```diff
@@ -0,0 +1,127 @@
...
+//			},
+//		}
+//		if values["sslInsecure"] != nil && values["sslInsecure"].(bool) {
+//			fc.FencingDeviceOptions["ssl_insecure"] = ""
+//		}
+//		fcs = append(fcs, fc)
+//	}
+//
+//	return fcs, nil
+//}
```
tjungblu, line 127: if the code is unused, please remove it

pkg/tnf/pkg/library-go/batch.go
```diff
@@ -0,0 +1,89 @@
+package library_go
```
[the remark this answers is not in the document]
tjungblu: can we split out the entire package into its own PR as well?

pkg/tnf/setup/runner.go
```diff
@@ -0,0 +1,137 @@
...
+	// get pcs cib
+	cib, err := pcs.GetCIB(ctx)
+	if err != nil {
+		return err
+	}
+
+	klog.Infof("HA setup done! CIB:\n%s", cib)
+
+	return nil
+}
```
tjungblu, line 137: are all of those steps idempotent? ;)

bindata/tnfdeployment/authjob.yaml
tjungblu: make it something that doesn't parse as an image :) `<injected>` or so

pkg/tnf/auth/runner.go
tjungblu: how is this possible that `exec.Execute` runs as root and your container does not?

pkg/tnf/operator/starter.go
jaypoulz: What would prevent us from watching for control-plane node addition/removal? Surely something in CEO is responding to those kinds of events today - at least I would assume so since we have to incorporate that new control plan nodes into the fold.

bindata/tnfdeployment/authjob.yaml
jaypoulz: I think that's fair. But does something relaunch the job after? Because otherwise if you've failed once, you're never getting out of that state.

pkg/tnf/operator/starter.go
jaypoulz: I spoke to the team about this. While we recognize that node-level change events are important and need to be handled, handling them was out of our development scope for Dev Preview across *ALL* components of TNF. For example, there is no existing logic in podman-etcd to handle the event of more than 2 control-plane nodes existing, which would be one way to replace a node. (The other would be to remove the bad node first).

At the end of the day, I do agree that this problem needs to be solved, but I would argue that there is more business value in getting this Dev Preview into a state of working installations than in holding it back to rush the design of handling CP node replaces. Especially since that design will need to be treated as a holistic operation across all parts of a TNF cluster.

@jerpeter1

clobrano: > We could document commands for recovery

It should be possible to handle it manually, by _removing_ one node first. One of the basis of this configuration was to never consider three nodes in the first place.

PR #1401, opened 2025-03-27
WIP: tweak heartbeat interval, election timeout, and IO related timeout environment variables

everettraven:


pkg/hwspeedhelpers/hwhelper.go
```diff
@@ -18,16 +18,44 @@ func HardwareSpeedToEnvMap(speed operatorv1.ControlPlaneHardwareSpeed) (envs map
 	return envs, nil
 }
 
+// Note(everettraven): OCPBUGS-50510: From what I could gleam, on our
+// "standard" hardware platforms for CI we seem to have an average RTT of ~12ms
+// between etcd peers. Looking at the telemetry I was able to gather, this
+// is about half of what the average RTT on all clusters reporting
+// telemetry tends to be (~25ms).
+// Assuming clusters out in the wild set to "standard" hardware speed
+// are roughly equivalent to our CI environments, our heartbeat interval on these platforms should
```
tjungblu, line 27: I wish they were roughly equivalent to our CI environments ;)

everettraven: Yeah, I realize this likely isn't the case :). Looking at our official recommendations for etcd performance in https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/scalability_and_performance/recommended-performance-and-scalability-practices-2#etcd-verify-hardware_recommended-etcd-practices I think it is probably reasonable to set the hb interval to ~50ms for "standard" hardware and doubling it for "slower" hardware.

everettraven: I haven't had time to circle back around to this PR, but the main intent was to just see what happens when tweaking these values

PR #1407, opened 2025-04-16
NO-JIRA: redundant unhealthy check

lance5890:
In ocp 4.16, we found the ceo logs show as follows:
```
E0416 07:21:35.302735       1 base_controller.go:268] GuardController reconciliation failed: Missing operand on node node-1
E0416 07:21:36.102619       1 guard_controller.go:285] Missing operand on node node-1
E0416 07:21:36.902893       1 base_controller.go:268] GuardController reconciliation failed: Missing operand on node node-1
E0416 07:21:37.269313       1 guard_controller.go:285] Missing operand on node node-1
W0416 07:21:37.280734       1 etcdcli.go:346] UnstartedEtcdMember found: [NAME-PENDING-10.255.10.1]
W0416 07:21:37.280756       1 etcdcli.go:351] UnhealthyEtcdMember found: [NAME-PENDING-10.255.10.1]
W0416 07:21:38.536782       1 etcdcli.go:346] UnstartedEtcdMember found: [NAME-PENDING-10.255.10.1]
W0416 07:21:38.536797       1 etcdcli.go:351] UnhealthyEtcdMember found: [NAME-PENDING-10.255.10.1]
```
If the cluster has unstarted member, the cluster is unhealthy now, and no need to output more ' UnhealthyEtcdMember found'  log

### DIFF
--- a/pkg/etcdcli/etcdcli.go
+++ b/pkg/etcdcli/etcdcli.go
@@ -344,6 +344,7 @@ func (g *etcdClientGetter) UnhealthyMembers(ctx context.Context) ([]*etcdserverp
 	unstartedMemberNames := GetUnstartedMemberNames(memberHealth)
 	if len(unstartedMemberNames) > 0 {
 		klog.Warningf("UnstartedEtcdMember found: %v", unstartedMemberNames)
+		return memberHealth.GetUnstartedMembers(), nil
 	}
 
 	unhealthyMemberNames := GetUnhealthyMemberNames(memberHealth)


pkg/etcdcli/etcdcli.go
```diff
@@ -344,6 +344,7 @@ func (g *etcdClientGetter) UnhealthyMembers(ctx context.Context) ([]*etcdserverp
 	unstartedMemberNames := GetUnstartedMemberNames(memberHealth)
 	if len(unstartedMemberNames) > 0 {
 		klog.Warningf("UnstartedEtcdMember found: %v", unstartedMemberNames)
```
tjungblu, line 346: did you intend to remove the log line instead?

lance5890: the UnstartedEtcdMember log line is meaningful, but no need to log the "UnhealthyEtcdMember found" in line 352 any more, so maybe just return when we founf the UnstartedEtcdMember

tjungblu: I disagree, the function is called "UnhealthyMembers", so it should only return the unhealthy members. I think this log line is mostly just for debugging in must-gathers, but it still seems useful for me to keep it.

PR #1417, opened 2025-04-24
OCPBUGS-55111: Remove 'meta' keyword from pcs resource create command

clobrano:
Specifying 'meta' after 'clone' defines meta attributes for the base resource, not for the clone as the agent needs for the "notify=true" meta attribute.

See: https://bugzilla.redhat.com/show_bug.cgi?id=2168155
Fixes: OCPBUGS-55111


clobrano:
It seems we cannot target 4.19 anymore. I'll change the Ticket to 4.20

PR #1419, opened 2025-04-30
OCPBUGS-55445: bump etcdHighCommitDurations alert duration

Elbehery:
resolves https://issues.redhat.com/browse/OCPBUGS-55445

### DIFF
--- a/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
+++ b/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
@@ -84,7 +84,7 @@ spec:
       expr: |
         histogram_quantile(0.99, rate(etcd_disk_backend_commit_duration_seconds_bucket{job=~".*etcd.*"}[5m]))
         > 0.25
-      for: 10m
+      for: 45m
       labels:
         severity: warning
     - alert: etcdDatabaseQuotaLowSpace


dusk125:
Is there a smaller value we could use? 10 to 45 seems like quite a jump for all platforms.

dgoodwin:
The discussion yesterday centered around data gathering to prove we are or are not regressed vs 4.18. Silencing this alert does not seem like a viable thing to do right now. Then we lose all signal on this.

Strongly recommend you do not merge this until it's safety can be proven.

/hold

geliu2016 (approved):
/label cherry-pick-approved

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -84,7 +84,7 @@ spec:
       expr: |
         histogram_quantile(0.99, rate(etcd_disk_backend_commit_duration_seconds_bucket{job=~".*etcd.*"}[5m]))
         > 0.25
```
sdodson, line 86: @wking Do you have any clever prom tricks that would maintain .25 for all platforms other than Azure and .5 for Azure?

Elbehery:
this has been replaced by https://github.com/openshift/cluster-etcd-operator/pull/1422

PR #1430, opened 2025-06-12
OCPEDGE-1920: Comply with OCP pod standards for TNF jobs

slintes:
- remove resource limits
- add priorityClass

Note: I will consider to have 1 job manifest for all TNF jobs, and set the actual name and command dynamically, in order to reduce duplication. Since this is going to be backported to 4.19, I keep the change small here.


jaypoulz (approved):
Looks good to me! :)

PR #1431, opened 2025-06-13
OCPBUGS-57372: Wait for 2 cp nodes before starting TNF jobs

slintes:
When being installed with assisted installer, the 1st auth job  is already created when 1 cp node is available only. Wait with the job creation until both nodes exist, ensure jobs are created once only, and double check node count inside jobs.

### DIFF
--- a/pkg/tnf/operator/starter.go
+++ b/pkg/tnf/operator/starter.go
@@ -69,15 +69,37 @@ func HandleDualReplicaClusters(
 	runTnfResourceController(ctx, controllerContext, kubeClient, dynamicClient, operatorClient, kubeInformersForNamespaces)
 
 	// we need node names for assigning auth and after-setup jobs to specific nodes
+	var once sync.Once
 	klog.Infof("watching for nodes...")
 	_, err := controlPlaneNodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			node, ok := obj.(*corev1.Node)
+			_, ok := obj.(*corev1.Node)
 			if !ok {
 				klog.Warningf("failed to convert added object to Node %+v", obj)
 			}
-			runTnfAuthJobController(ctx, node.GetName(), controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
-			runTnfAfterSetupJobController(ctx, node.GetName(), controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
+			// ensure we have both control plane nodes before creating jobs
+			nodeList, err := kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{
+				LabelSelector: "node-role.kubernetes.io/master",
+			})
+			if err != nil {
+				klog.Errorf("failed to list nodes while waiting to create TNF jobs: %v", err)
+				return
+			}
+			if len(nodeList.Items) != 2 {
+				klog.Info("not starting TNF jobs yet, waiting for 2 control plane nodes to exist")
+				return
+			}
+			// we can have 2 nodes on the first call of AddFunc already, ensure we create job controllers once only
+			once.Do(func() {
+				klog.Info("found 2 control plane nodes, creating TNF jobs")
+				// the order of job creation does not matter, the jobs wait on each other as needed
+				for _, node := range nodeList.Items {
+					runTnfAuthJobController(ctx, node.GetName(), controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
+					runTnfAfterSetupJobController(ctx, node.GetName(), controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
+				}
+				runTnfSetupJobController(ctx, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
+				runTnfFencingJobController(ctx, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
+			})
 		},
 	})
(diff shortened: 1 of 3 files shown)


pkg/tnf/operator/starter.go
```diff
@@ -56,24 +58,43 @@ func HandleDualReplicaClusters(
...
+			// ensure we have both control plane nodes before creating jobs
+			nodeList, err := kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{
+				LabelSelector: "node-role.kubernetes.io/master",
+			})
+			if err != nil {
+				klog.Errorf("failed to list nodes while waiting to create TNF jobs: %v", err)
+				return
+			}
+			if len(nodeList.Items) != 2 {
+				klog.Info("not starting TNF jobs yet, waiting for 2 control plane nodes to exist")
```
tjungblu, line 94: just for a potential must-gather in the future, you may want to also supply what nodes were found already

pkg/tnf/operator/starter.go
```diff
@@ -56,24 +58,43 @@ func HandleDualReplicaClusters(
...
 		AddFunc: func(obj interface{}) {
-			node, ok := obj.(*corev1.Node)
+			_, ok := obj.(*corev1.Node)
 			if !ok {
 				klog.Warningf("failed to convert node to Node %+v", obj)
 			}
-			runTnfAuthJobController(ctx, node.GetName(), controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
-			runTnfAfterSetupJobController(ctx, node.GetName(), controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
+			// ensure we have both control plane nodes before creating jobs
+			nodeList, err := kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{
```
tjungblu: I know it's premature optimization, but you can also create a lister on top of the informer indexer:

> import corev1listers "k8s.io/client-go/listers/core/v1"
> controlPlaneNodeLister := corev1listers.NewNodeLister(controlPlaneNodeInformer.GetIndexer())


so that way you can also save yourself the additional synchronous list call.

pkg/tnf/operator/starter.go
```diff
@@ -56,24 +58,43 @@ func HandleDualReplicaClusters(
...
+			})
+			if err != nil {
+				klog.Errorf("failed to list nodes while waiting to create TNF jobs: %v", err)
+				return
+			}
+			if len(nodeList.Items) != 2 {
+				klog.Info("not starting TNF jobs yet, waiting for 2 control plane nodes to exist")
+				return
+			}
+			// we can have 2 nodes on the first call of AddFunc already, ensure we create job controllers once only
```
tjungblu, line 97: I know I've pestered you with retries and idempotency before, but do you think this overall construct is still safe?

Why not just poll until the list informer has two entries and then just setup everything at once. Might simplify this a lot...

jaypoulz: I've been drawing up the plans for CP-node replacement. The scope of what we're looking to cover is in-place (i.e. same name/ip replacement).

So my follow up to this is - can we run the setup jobs each team we detect scaling from 1 to 2 control-plane nodes? Even in the ungraceful shutdown case, the second CP node is still part of the cluster (just marked-not-ready).

If we run re-run set up upon discovering 2 nodes, we should be covered for the "straightforward" replacement.

--

As an aside, while we can instruct users to never scale up to 3 nodes in TNF, we can't actually guarantee that they won't. One thing that worries me is if a user scales up to three and then down to 2. If we ran the setup job there, we'd need to destroy the pacemaker cluster in its entirety and rebuild it from nothing because the added member might not be part of the original node list. I know Fabio advised us against doing this, but I'm just pointing out that we should be ready for customers to do it. If this happens, my recommendation is we force the users to scale from 3 down to 1 (where the one node is from the original cluster), and then scale back up to two with the origin IP/name intact. /endsoapbox

pkg/tnf/operator/starter.go
slintes: Thanks for the hint with the lister 👍🏼 I don't see though how that can save the synchronous list call, it's just a nicer way to do that call, not?

pkg/tnf/operator/starter.go
slintes: > Why not just poll until the list informer has two entries and then just setup everything at once. Might simplify this a lot...

I don't see the advantage of polling instead of listening to events... 🤔 

With "setup everything at once", do you mean running 1 job only? That's not possible:
- auth needs to run on each node
- setup needs to run on 1 node only
- after-setup needs to run on each node again

Since this is a bugfix for 4.19, node replacement / scaling etc. is out of scope of this PR.

pkg/tnf/pkg/config/cluster_test.go
```diff
@@ -58,6 +65,21 @@ func TestGetClusterConfig(t *testing.T) {
...
+						Labels: map[string]string{
+							"node-role.kubernetes.io/master": "",
+						},
+					},
+				},
+			}),
+			want:    ClusterConfig{},
+			wantErr: true,
+		},
 	}
```
clobrano, line 106: Do you think it's valuable to add a test where the nodes are 2, but only one as the right label?

slintes: good idea, done

clobrano:
/lgtm

not sure if the other threads have been resolved already so
/hold

clobrano:
Sure :) 


/lgtm

PR #1441, opened 2025-07-04
Use 0700 perms for etcd data dir

shaneboulden:
Currently the cluster etcd operator creates `0755` permissions for `/var/lib/etcd` by default:

```
$ oc debug node/ip-10-0-22-132.ap-southeast-2.compute.internal
sh-5.1# ls -l /host/var/lib/ | grep etcd
drwxr-xr-x.  3 root root     41 Jul  4 03:09 etcd
```
Upstream etcd uses 0700: https://github.com/etcd-io/etcd/blob/866bc0717c0b56579514c7363f3f47f6cd4109c6/client/pkg/fileutil/fileutil.go#L47

And the OpenShift CIS benchmark requires that the `/var/lib/etcd` data dir is configured with `0700` file permissions: https://github.com/ComplianceAsCode/content/blob/24dba6b94757881c8d8e0dd21390fa733a7c3b70/applications/openshift/master/file_permissions_var_lib_etcd/rule.yml#L28

This change aligns the cluster etcd operator with upstream and the CIS benchmark for OpenShift.

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -306,13 +306,13 @@ func (r *renderOpts) Run() error {
 	certDir := filepath.Join(memberDir, "etcd-all-certs")
 
 	// Creating the cert dir recursively will create the base path too
-	err = os.MkdirAll(certDir, 0755)
+	err = os.MkdirAll(certDir, 0700)
 	if err != nil {
 		return fmt.Errorf("failed to create directory %s: %w", memberDir, err)
 	}
 	// tlsDir contains the ca bundle and client cert pair for bootkube.sh and the bootstrap apiserver
 	tlsDir := filepath.Join(r.assetOutputDir, "tls")
-	err = os.MkdirAll(tlsDir, 0755)
+	err = os.MkdirAll(tlsDir, 0700)
 
 	// Write the etcd ca bundle required by the bootstrap etcd member
 	for _, bundle := range templateData.caBundles {
--- a/pkg/operator/etcd_assets/bindata.go
+++ b/pkg/operator/etcd_assets/bindata.go
@@ -2626,7 +2626,7 @@ func RestoreAsset(dir, name string) error {
 	if err != nil {
 		return err
 	}
-	err = os.MkdirAll(_filePath(dir, filepath.Dir(name)), os.FileMode(0755))
+	err = os.MkdirAll(_filePath(dir, filepath.Dir(name)), os.FileMode(0700))
 	if err != nil {
 		return err
 	}
--- a/pkg/tnf/assets/bindata.go
+++ b/pkg/tnf/assets/bindata.go
@@ -585,7 +585,7 @@ func RestoreAsset(dir, name string) error {
 	if err != nil {
 		return err
 	}
-	err = os.MkdirAll(_filePath(dir, filepath.Dir(name)), os.FileMode(0755))
+	err = os.MkdirAll(_filePath(dir, filepath.Dir(name)), os.FileMode(0700))
 	if err != nil {
 		return err
 	}


pkg/cmd/render/render.go
```diff
@@ -306,13 +306,13 @@ func (r *renderOpts) Run() error {
 	certDir := filepath.Join(memberDir, "etcd-all-certs")
 
 	// Creating the cert dir recursively will create the base path too
-	err = os.MkdirAll(certDir, 0755)
+	err = os.MkdirAll(certDir, 0700)
```
tjungblu, line 309: are you sure that this changes the `dataDir`?

shaneboulden: This is the only location I could find in the codebase where the `/var/lib/etcd` dir is created, via the `memberDir`.

lance5890: I'm also curious about the permission of "/var/lib/etcd" which set to 0755

but I'm afraid u got the wrong place

maybe u could change the dir permission in the following place just like : 

https://github.com/openshift/cluster-etcd-operator/blob/d5259e4c64c3446a173eca154570a83830e43e7a/bindata/etcd/pod.yaml#L26

tjungblu: I think the reason is that the entire folder structure is on 0755:

```
sh-5.1# stat --format '%a' /var/    
755
sh-5.1# stat --format '%a' /var/lib/etcd/
755 
```

same with other openshift folders:

``` 
sh-5.1# stat --format '%a' /var/lib/ovn-ic/                 
755
sh-5.1# stat --format '%a' /var/lib/kubelet/
755
``` 

so I would rather advocate to change the CIS rule.

shaneboulden: @lance5890 you're right, I'm mistaken. I will take a look at an `initContainers` approach. Should I turn this into a draft / WIP?

@tjungblu 
I think you're right.

I couldn't find the `/var/lib/etcd` dir explicitly created by the operator. But, doing some digging through older issues, it sounds like the kubelet will create the dir with `0755` if it doesn't exist (not sure how much has changed since 2018...)

https://github.com/kubernetes/kubeadm/issues/1308#issuecomment-445248790

> so I would rather advocate to change the CIS rule.

Looks like `0700` is also part of the STIG: https://github.com/ComplianceAsCode/content/blob/49189d32c6039a5e4ca68f6e4a04de1f719e47b2/products/ocp4/profiles/stig-node-v1r1.profile#L62

tjungblu: @shaneboulden are you sure that the data **dir** is part of the STIG? https://stigviewer.com/stigs/kubernetes
The data files I agree and understand, but the datadir itself seems pretty strange to me.

shaneboulden: @tjungblu it is certainly in the OpenShift v1r1 STIG

https://complianceascode.github.io/content-pages/guides/ssg-ocp4-guide-stig-node-v1r1.html

Screenshot From 2025-07-07 22-47-24

https://github.com/ComplianceAsCode/content/blob/49189d32c6039a5e4ca68f6e4a04de1f719e47b2/products/ocp4/profiles/stig-node-v1r1.profile#L91

https://github.com/ComplianceAsCode/content/blob/49189d32c6039a5e4ca68f6e4a04de1f719e47b2/applications/openshift/master/file_permissions_etcd_data_dir/rule.yml#L7

tjungblu: I assume that the STIG is generated from the git repo ;) 
Oz is not around anymore for a few years, so we won't be able to ask him. I remember my old team was writing a few of those rules too.

Tracking this back to 2020: https://github.com/ComplianceAsCode/content/pull/6341 and https://github.com/ComplianceAsCode/content/pull/6341#discussion_r518807477

not sure whether this ever worked, it also seems an additional check that is not listed in the initial ticket.

PR #1443, opened 2025-07-08
OCPBUGS-55404: fix ordering on bootstrap IP selection

tjungblu:
This refactors the initial map based implementation into using slices instead. The initial ordering of what's returned by the netlink library should be preserved here, while retaining the deduplication logic we previously have gotten from the map.

Disclaimer: please be extra careful, this is AI generated

### DIFF
--- a/pkg/cmd/render/bootstrap_ip.go
+++ b/pkg/cmd/render/bootstrap_ip.go
@@ -3,6 +3,7 @@ package render
 import (
 	"fmt"
 	"net"
+	"slices"
 
 	"github.com/vishvananda/netlink"
 	"k8s.io/klog/v2"
@@ -15,8 +16,8 @@ var defaultBootstrapIPLocator BootstrapIPLocator = NetlinkBootstrapIPLocator()
 func NetlinkBootstrapIPLocator() *bootstrapIPLocator {
 	return &bootstrapIPLocator{
 		getIPAddresses: ipAddrs,
-		getAddrMap:     getAddrMap,
-		getRouteMap:    getRouteMap,
+		getAddrSlice:   getAddrSlice,
+		getRouteSlice:  getRouteSlice,
 	}
 }
 
@@ -28,10 +29,28 @@ type BootstrapIPLocator interface {
 	getBootstrapIP(ipv6 bool, machineCIDR string, excludedIPs []string) (net.IP, error)
 }
 
+// LinkAddresses represents a link and its associated addresses, preserving order
+type LinkAddresses struct {
+	Link      netlink.Link
+	Addresses []netlink.Addr
+}
+
+// LinkRoutes represents a link index and its associated routes, preserving order
+type LinkRoutes struct {
+	LinkIndex int
+	Routes    []netlink.Route
+}
+
+// addrSlice is a slice of LinkAddresses that preserves ordering
+type addrSlice []LinkAddresses
+
+// routeSlice is a slice of LinkRoutes that preserves ordering
+type routeSlice []LinkRoutes
+
 type bootstrapIPLocator struct {
 	getIPAddresses func() ([]net.IP, error)
-	getAddrMap     func() (addrMap addrMap, err error)
-	getRouteMap    func() (routeMap routeMap, err error)
+	getAddrSlice   func() (addrSlice addrSlice, err error)
+	getRouteSlice  func() (routeSlice routeSlice, err error)
 }
 
 func (l *bootstrapIPLocator) getBootstrapIP(ipv6 bool, machineCIDR string, excludedIPs []string) (net.IP, error) {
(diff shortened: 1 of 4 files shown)


pkg/cmd/render/bootstrap_ip.go
```diff
@@ -140,27 +158,36 @@ func AddressFilters(filters ...AddressFilter) AddressFilter {
 	}
 }
 
-type addrMap map[netlink.Link][]netlink.Addr
-type routeMap map[int][]netlink.Route
+// Helper function to find routes by link index in the ordered slice
```
dusk125: nit: Could replace with `slices.IndexFunc`

pkg/cmd/render/bootstrap_ip.go
```diff
@@ -178,7 +205,11 @@ func routableAddresses(addrMap addrMap, routeMap routeMap, vips []net.IP, af Add
 						klog.Infof("Checking whether address %s with route %s contains VIP %s", address, route, vip)
 						if containmentNet.Contains(vip) {
 							klog.Infof("Address %s with route %s contains VIP %s", address, route, vip)
-							matches[address.IP.String()] = address.IP
+							ipStr := address.IP.String()
+							if !seen[ipStr] {
```
dusk125: Not sure if this is performance sensitive, but I remember the go team saying that `map[string]struct{}` and `if _, has := seen[ipStr]; has` is faster than mapping booleans.

pkg/cmd/render/bootstrap_ip_linux.go
```diff
@@ -1,12 +1,14 @@
 package render
```
dusk125, line 3: nit: I realized there's a generic that does `go:build !linux` but this file doesn't provide `go:build linux`. Doesn't cause a problem right now, but IF we ever added any other archs, dunno if you wanna include a build linux to prevent

pkg/cmd/render/bootstrap_ip.go
lance5890: wow, make sense, got it

pkg/cmd/render/bootstrap_ip.go
tjungblu: done

pkg/cmd/render/bootstrap_ip.go
tjungblu: it's not sensitive at all, should be about five network interfaces and we only run this once in bootkube

I find the struct{} approach more idiomatic though, so I'll take it

pkg/cmd/render/bootstrap_ip_linux.go
tjungblu: added

pkg/cmd/render/bootstrap_ip_test.go
```diff
@@ -60,22 +61,127 @@ func TestBootstrapIPLocator(t *testing.T) {
...
+				withDevice(1, "192.168.125.20"), // Device 1 gets first IP
+				withDevice(0, "192.168.125.10"), // Device 0 gets second IP
+				withDevice(1, "192.168.125.30"), // Device 1 gets second IP (accumulated)
+				withDevice(2, "10.0.0.1"),       // Device 2 gets IP outside CIDR
+			),
+			routeSlice: newRouteSlice(
+				withRoute(1, "192.168.125.0/24", unix.RTPROT_RA),
+			),
+			exclude: []string{"192.168.125.20"},    // Exclude first IP from device 1
+			expect:  net.ParseIP("192.168.125.30"), // Should select device 1's second IP (first device in slice order)
```
tjungblu: this is a wrong generation, it should be 192.168.125.10

pkg/cmd/render/bootstrap_ip_test.go
```diff
@@ -88,16 +194,16 @@ func TestBootstrapIPLocator(t *testing.T) {
 	}
 }
 
-func newAddrMap(configs ...func(addrMap)) addrMap {
-	m := map[netlink.Link][]netlink.Addr{}
+func newAddrSlice(configs ...func(*addrSlice)) addrSlice {
```
tjungblu: this seems to be an insanely difficult translation :rofl:

geliu2016:
> @geliu2016 are you able to do the new pre-merge verification thingy? I'm going to override the CPMS jobs as they should not be impacted by this.
> 
> /override ci/prow/e2e-aws-cpms /override ci/prow/e2e-aws-ovn-etcd-scaling

Hi @tjungblu I will do some trying, and do you have any idea about "Deploy SNO via Assisted Install (ZTP) with multiple IPs assigned to the primary interface"? I will sync with assisted install team for it.

tjungblu:
@geliu2016 you can ask the bug creator, they have a SNO configuration that failed

lance5890:
> > @geliu2016 are you able to do the new pre-merge verification thingy? I'm going to override the CPMS jobs as they should not be impacted by this.
> > /override ci/prow/e2e-aws-cpms /override ci/prow/e2e-aws-ovn-etcd-scaling
> 
> Hi @tjungblu I will do some trying, and do you have any idea about "Deploy SNO via Assisted Install (ZTP) with multiple IPs assigned to the primary interface"? I will sync with assisted install team for it.

I have an idea in mind, maybe u can install the SNO by the upi model, and set multi ips on one interface, and then bootstrap the cluster

PR #1450, opened 2025-07-23
WIP: Add IsLearner check to readyz

dusk125:


pkg/cmd/readyz/readyz.go
```diff
@@ -218,6 +218,44 @@ func (r *readyzOpts) Run() error {
...
+		status, err := etcdClient.Status(ctx, r.targetEndpoint)
+		if err != nil {
+			klog.V(2).Infof("failed to get member status: %v", err)
+			http.Error(w, fmt.Sprintf("failed to get member status: %v", err), http.StatusServiceUnavailable)
+			return
+		}
+		if status.IsLearner {
+			klog.V(2).Info("member is a learner and therefore not ready")
+			http.Error(w, "member is a learner and therefore not ready", http.StatusServiceUnavailable)
+			return
```
tjungblu, line 243: interesting idea. Wouldn't that make it crash loop after a while? What if it needs to sync down a big snapshot from another member?

dusk125: Mmm yeah good point, this thought came from talking with the TNF folks about an issue where a learner was reachable and causing the "rpc not enabled on learners" error to be shown. It's most likely something else, but I was curious :D

tjungblu: this happens because apiserver hardcodes localhost as the etcd endpoint 

see one of our bugs in https://issues.redhat.com/browse/OCPBUGS-42808

PR #1456, opened 2025-07-28
ETCD-590: start to refactor the probe injection

tjungblu:
This PR makes all etcd probes injectable via gotpl configuration. We only set the existing default values for now and wire up the etcd operator CRD in all places that will be needed.

A later PR will introduce the business logic to change the liveness probes based on how much database quota was configured.


pkg/operator/ceohelpers/podsubstitution.go
```diff
@@ -46,6 +58,23 @@ func GetPodSubstitution(
 		return nil, err
 	}
 
+	// TODO(thomas): derive based on etcd.Spec.BackendQuotaGiB
+	defaultProbeConfig := ProbeConfig{
+		TimeoutSeconds:      30,
+		PeriodSeconds:       5,
+		SuccessThreshold:    1,
+		FailureThreshold:    5,
+		InitialDelaySeconds: 10,
```
lance5890: In the original etcd livenessProbe and readinessProbe configuration, initialDelaySeconds is not set (defaulting to 0). Why should we set it to 10 by default here?

tjungblu: good catch! I've set it back to zero, even though 10s might not actually make much of a difference in reality

tjungblu:

unrelated failure in daemonsets

PR #1461, opened 2025-08-07
OCPBUGS-60220:  add missing ports to svc and pod spec

tjungblu:


geliu2016 (approved):
/label cherry-pick-approved

PR #1422, opened 2025-05-12
OCPBUGS-55757: adjust alerts according to platform

Elbehery:
resolves https://issues.redhat.com/browse/OCPBUGS-55757

cc @wking @sdodson @tjungblu

PR #1462, opened 2025-08-07
Bug: Fix alert generation diff on `main`

joshbranham:
In https://github.com/openshift/cluster-etcd-operator/pull/1422, the `etcdHighCommitDurations` alert had it's expression modified directly in the generated rules yaml file. As far as I can tell, this breaks rule generation, and should be reflected in the jsonnet instead.

This, with the help of Claude, handles this specific alert override. Albeit, it's a bit ugly, but opening this to validate if you expected this to drift or not. I am working on adding some alerts and would like to start from a clean generation on `main`.

Note: The diff in the manifests now is just from, my guess, a newer `gojsontoyaml` version.

We should probably run this diff check in CI as well?

# Diff on main right now
```diff
 git diff
diff --git a/manifests/0000_90_etcd-operator_03_prometheusrule.yaml b/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
index 02f89a6e5..e77641e1f 100644
--- a/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
+++ b/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
@@ -83,12 +83,7 @@ spec:
         summary: etcd cluster 99th percentile commit durations are too high.
       expr: |
         histogram_quantile(0.99, rate(etcd_disk_backend_commit_duration_seconds_bucket{job=~".*etcd.*"}[5m]))
-        > on () group_left (type)
-        bottomk(0.25,
-        0.5 * group by (type) (cluster_infrastructure_provider{type="Azure"})
-        or
-        0.25 * group by (type) (cluster_infrastructure_provider)
-        )
+        > 0.25
       for: 10m
       labels:
         severity: warning
@@ -132,9 +127,9 @@ spec:
         histogram_quantile(0.99, sum(rate(grpc_server_handling_seconds_bucket{job="etcd", grpc_method!="Defragment", grpc_type="unary"}[10m])) without(grpc_type))
         > on () group_left (type)
         bottomk(1,
-        1.5 * group by (type) (cluster_infrastructure_provider{type="Azure"})
-        or
-        1 * group by (type) (cluster_infrastructure_provider))
+          1.5 * group by (type) (cluster_infrastructure_provider{type="Azure"})
+          or
+          1 * group by (type) (cluster_infrastructure_provider))
       for: 30m
       labels:
         severity: critical
```

### DIFF
--- a/jsonnet/main.jsonnet
+++ b/jsonnet/main.jsonnet
@@ -15,14 +15,26 @@ local excludeRules = std.map(
   }, alertingRules + promRules
 );
 
-// modifiedRules injects runbook_url to all critical alerts on all rules.
+// modifiedRules injects runbook_url to all critical alerts on all rules and overrides specific alert expressions.
 local modifiedRules = std.map(function(group) group {
-  rules: std.map(function(rule) if 'alert' in rule && !('runbook_url' in rule.annotations) && (rule.labels.severity == 'critical') then rule {
-                   annotations+: {
-                     runbook_url: 'https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/%s.md' % rule.alert,
-                   },
-                 } else rule,
-                 super.rules),
+  rules: std.map(function(rule)
+    if 'alert' in rule && rule.alert == 'etcdHighCommitDurations' then rule {
+      expr: |||
+        histogram_quantile(0.99, rate(etcd_disk_backend_commit_duration_seconds_bucket{job=~".*etcd.*"}[5m]))
+        > on () group_left (type)
+        bottomk(0.25,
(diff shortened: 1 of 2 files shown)


jsonnet/main.jsonnet
```diff
@@ -15,14 +15,26 @@ local excludeRules = std.map(
   }, alertingRules + promRules
 );
 
-// modifiedRules injects runbook_url to all critical alerts on all rules.
+// modifiedRules injects runbook_url to all critical alerts on all rules and overrides specific alert expressions.
```
joshbranham, line 18: If the intended behavior is "any alert we change from upstream is completely redefined in `custom.libsonnet` then I can just pull that change in that way instead.

PR #1463, opened 2025-08-11
OCPEDGE-1998: TNF: Use `pcmk_delay_base` on 1st fence device to prevent fencing races

slintes:
see https://access.redhat.com/solutions/54829

### DIFF
--- a/pkg/tnf/pkg/pcs/fencing.go
+++ b/pkg/tnf/pkg/pcs/fencing.go
@@ -19,6 +19,11 @@ var (
 	addressRegEx = regexp.MustCompile(`.*//(.*):(.*)(/redfish.*)`)
 )
 
+const (
+	// defaultPcmkDelayBase is the delay applied to the first fence device to prevent simultaneous fencing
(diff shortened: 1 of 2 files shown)


pkg/tnf/pkg/pcs/fencing.go
```diff
@@ -56,6 +57,10 @@ func ConfigureFencing(ctx context.Context, kubeClient kubernetes.Interface, cfg
 			klog.Errorf("Failed to get fencing config for node %s: %v", nodeName, err)
 			return fmt.Errorf("failed to get fencing config for node %s: %v", nodeName, err)
 		}
+		// Add pcmk_delay_base=10s to the first fence device only
+		if i == 0 {
+			fc.FencingDeviceOptions[PcmkDelayBase] = "10s"
```
clobrano: Wouldn't it be better to set this as constant on top of this file? For readability and maintenance.

slintes: yes, done 👍🏼

PR #1464, opened 2025-08-11
OCPBUGS-60237: Vendor latest mixin, including additional and modified alerts for `etcdDatabaseQuotaLowSpace`

joshbranham:
This PR introduces a few changes, primarily in motivation of resolving [OCPBUGS-60237](https://issues.redhat.com/browse/OCPBUGS-60237).

The following "housekeeping" changes, in the first commit:
- Vendor the latest `etcd-mixin` changes. This includes overwriting some previous manual modifications introduced in https://github.com/openshift/cluster-etcd-operator/commit/674ff5490cb98a38a6d216bf8228e9e58eca085d
- Redefine `etcdMembersDown` in our custom rules. This rule was changed to a `warning` severity, and the `for` to 20m. Per conversation with @tjungblu I reverted this back to `critical`, however kept the upstream change of 20m.

The following changes that satisfy the OCPBUG in the second commit:
- Modify the upstream `etcdDatabaseQuotaLowSpace` critical level alert to use a threshold of 85% instead of 95%.
- Add a new `etcdDatabaseQuotaLowSpace` warning level alert, set to a threshold of 75%.
- Add a new `etcdDatabaseQuotaLowSpace` info level alert, set to a threshold of 65%.

### DIFF
--- a/jsonnet/custom.libsonnet
+++ b/jsonnet/custom.libsonnet
@@ -5,6 +5,42 @@
         name: 'openshift-etcd.rules',
         rules: [
           {
+            alert: 'etcdDatabaseQuotaLowSpace',
+            expr: '(last_over_time(etcd_mvcc_db_total_size_in_bytes{job=~".*etcd.*"}[5m]) / last_over_time(etcd_server_quota_backend_bytes{job=~".*etcd.*"}[5m]))*100 > 65',
+            'for': '10m',
+            labels: {
+              severity: 'info',
+            },
+            annotations: {
+              description: 'etcd cluster "{{ $labels.job }}": database size is 65% of the defined quota on etcd instance {{ $labels.instance }}, please defrag or increase the quota as the writes to etcd will be disabled when it is full.',
+              summary: 'etcd cluster database is using >= 65% of the defined quota.',
+            },
+          },
+          {
+            alert: 'etcdDatabaseQuotaLowSpace',
+            expr: '(last_over_time(etcd_mvcc_db_total_size_in_bytes{job=~".*etcd.*"}[5m]) / last_over_time(etcd_server_quota_backend_bytes{job=~".*etcd.*"}[5m]))*100 > 75',
+            'for': '10m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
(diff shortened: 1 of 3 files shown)


jsonnet/custom.libsonnet
```diff
@@ -83,6 +83,27 @@
               severity: 'critical',
             },
           },
+          {
+            alert: 'etcdMembersDown',
```
joshbranham, line 123: The upstream alert changed this to a `warning`. Per @tjungblu we should redefine this as custom to keep it as critical. The upstream changed `for: 10m` to `20m` as well. Per our convo on Slack, I've bumped our alert to 20m as well.

manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -83,12 +66,7 @@ spec:
         summary: etcd cluster 99th percentile commit durations are too high.
       expr: |
         histogram_quantile(0.99, rate(etcd_disk_backend_commit_duration_seconds_bucket{job=~".*etcd.*"}[5m]))
-        > on () group_left (type)
```
joshbranham, line 86: This was made in a previous PR manually against the generated yaml. Letting it revert per convo on Slack.

wking: Won't that regress us vs. the slow-Azure-disk noise that 674ff5490cb98a38a6d216bf8228e9e58eca085d (which added this hunk, among other changes) was trying to de-noise?  I could see arguments for removing this hunk like "actually, even with slow Azure disks, this particular alert wasn't all that noisy" or "now Azure disks are more performant than they were a few months ago when we initially added this hunk".  And I could see arguments like "when we added this hunk, we did it manually here, and instead we should have added it via jsonnet over in some other place" for moving the hunk into other places.  But I'm a bit nervous to see it dropping without an explanation for why it no longer makes sense 😅  But also, I'm not a maintainer, so if maintainers are on board with dropping it, there's no need for the change to make sense to me.

tjungblu: I'm OK and I would like to finally have dedicated disks in Azure CI, which we hope to enable this coming release cycles. So having this helps us to clean this mess up again.

tjungblu:
/cherry-pick release-4.19 release-4.18 release-4.17 release-4.16

geliu2016 (approved):
/label cherry-pick-approved

PR #1474, opened 2025-08-28
OCPBUGS-60941: add individual context to the health check

tjungblu:
When one member is timing out, the others were also declared unhealthy due to the shared timeout being cancelled.
This adds a new context with an individual timeout to each member health check.

With this fix only the respective member is being considered unhealthy:
```
[etcd-operator-566ff8dd4-dn5qx] E0828 11:20:04.415537       1 health.go:120] health check for member (tjungblu15-dq6nb-master-0) failed: err(context deadline exceeded)
[etcd-operator-566ff8dd4-dn5qx] W0828 11:20:04.415708       1 etcdcli.go:356] UnhealthyEtcdMember found: [tjungblu15-dq6nb-master-0]
...
[etcd-operator-566ff8dd4-dn5qx] E0828 11:21:01.300834       1 base_controller.go:279] "Unhandled Error" err="DefragController reconciliation failed: cluster is unhealthy: 2 of 3 members are available, tjungblu15-dq6nb-master-0 is unhealthy"


```

### DIFF
--- a/pkg/etcdcli/health.go
+++ b/pkg/etcdcli/health.go
@@ -63,7 +63,12 @@ func GetMemberHealth(ctx context.Context, clipool *EtcdClientPool, etcdMembers [
 			}
 			defer clipool.Return(cli)
 
-			memberHealth[i] = checkSingleMemberHealth(ctx, cli, member)
+			// Create an independent timeout context for each member health check
+			// This prevents one slow member from affecting other members' health checks
+			memberCtx, cancel := context.WithTimeout(ctx, DefaultClientTimeout)
+			defer cancel()
+
+			memberHealth[i] = checkSingleMemberHealth(memberCtx, cli, member)
 		}(i)
 	}
 


pkg/etcdcli/health.go
```diff
@@ -63,7 +63,12 @@ func GetMemberHealth(ctx context.Context, clipool *EtcdClientPool, etcdMembers [
 			}
 			defer clipool.Return(cli)
 
-			memberHealth[i] = checkSingleMemberHealth(ctx, cli, member)
+			// Create an independent timeout context for each member health check
+			// This prevents one slow member from affecting other members' health checks
+			memberCtx, cancel := context.WithTimeout(ctx, DefaultClientTimeout)
+			defer cancel()
+
+			memberHealth[i] = checkSingleMemberHealth(memberCtx, cli, member)
```
lance5890, line 71: Maybe there's potential data race issue in the goroutine - it's capturing the loop variable member by reference. This can cause a data race where all goroutines end up checking the last member. we should pass member as an argument: `go func(i int, m *etcdserverpb.Member) {...}(i, member)`

tjungblu: thanks for the quick review @lance5890 - I think this should be fixed in later go versions?
https://go.dev/blog/loopvar-preview

PR #1476, opened 2025-08-29
Add priority field to prevent early shutdown

CoreyCook8:
Based on the issue described here : https://github.com/kubernetes/kubernetes/issues/133442

priorityClassName is currently ignored by Kubelet for static pod files so setting this value has no impact on the gracefulShutdown order causing the static pods to start to be killed as soon as shutdown begins.

To prevent this we must set priority explicitly


sanchezl:
/retitle OCPBUGS-63213: Add priority field to prevent early shutdown

PR #1479, opened 2025-09-04
feat: update the ip parsing for the nodes

eggfoobar:
updated the logic for the cluster info to prefer ipv6 over ipv4 for internal ip addresses on the node status


pkg/tnf/pkg/config/cluster.go
```diff
@@ -53,12 +54,39 @@ func GetClusterConfig(ctx context.Context, kubeClient kubernetes.Interface) (Clu
...
+		return ""
+	}
+
+	ipv6Address := ""
+	ipv4Address := ""
 	for _, addr := range addresses {
-		if addr.Type == corev1.NodeInternalIP {
-			return addr.Address
+		switch addr.Type {
+		case corev1.NodeInternalIP:
```
tjungblu, line 68: note that you are now picking the very last of the addresses, whereas previously you would always take the first occurrence. 

just fyi, there are always ordering issues around those IP selection functions, because people rely on the ordering of their netlink interfaces and then suddenly installations fail because we pick other IPs than before. Probably not a big concern in TNF, but I wanted to let you know.

eg. last one was https://github.com/openshift/cluster-etcd-operator/commit/c94e1c2a8ba535c2efd0cb9b73f7df5926f9274c

pkg/tnf/pkg/config/cluster_test.go
```diff
@@ -52,16 +52,100 @@ func TestGetClusterConfig(t *testing.T) {
...
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "test2",
+						Labels: map[string]string{
+							"node-role.kubernetes.io/master": "",
+						},
+					},
+					Status: corev1.NodeStatus{
+						Addresses: []corev1.NodeAddress{
+							// fallback to 1st IP incase no internal IP is set
```
clobrano: The comment seems wrong. There is only 1 IP here

Also typo: incase ⇒ in case (there are some like this, even in old code)

pkg/tnf/pkg/config/cluster_test.go
```diff
@@ -52,16 +52,100 @@ func TestGetClusterConfig(t *testing.T) {
...
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "test1",
+						Labels: map[string]string{
+							"node-role.kubernetes.io/master": "",
+						},
+					},
+					Status: corev1.NodeStatus{
+						Addresses: []corev1.NodeAddress{
+							// ensure that internal IP is used
```
clobrano: The comment could clarify also that IPv6 will be preferred, in case of 2 InternalIPs.

pkg/tnf/pkg/config/cluster_test.go
```diff
@@ -52,16 +52,100 @@ func TestGetClusterConfig(t *testing.T) {
...
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "test2",
+						Labels: map[string]string{
+							"node-role.kubernetes.io/master": "",
+						},
+					},
+					Status: corev1.NodeStatus{
+						Addresses: []corev1.NodeAddress{
+							// fallback to 1st IP incase no internal IP is set
```
clobrano: The comment is wrong, all IPs are internal

clobrano (changes requested):
I left some comments

pkg/tnf/pkg/config/cluster.go
eggfoobar: That's a great catch, thanks @tjungblu , after going back and forth, I think we'll just keep this more or less the same as it was originally. Seeing as the first IP gets us what we want regardless of ipv4/ipv6.

clobrano (approved):
/lgtm

PR #1484, opened 2025-09-13
TNF: Introduce job readiness condition for etcd container removal

slintes:
Introduce a new mechanism for the TNF setup job to signal CEO when it's ready for etcd container removal by setting a job status condition. This replaces the previous approach of using unsupportedConfigOverrides.

Main changes:

- set "ReadyForEtcdContainerRemoval" condition on setup job when pacemaker is configured to take over the etcd container
- refactor DualReplica main component:
  - keep track of cluster status
  - wait for bootstrap completed before starting job controllers (instead of doing that in the auth job, for easier status tracking)
- add job informer and dual replica status tracker to TargetConfigController and ExternalEtcdSupportController, for correct handling of the etcd container on job updates based on current dual replica cluster status


pkg/operator/starter.go
```diff
@@ -371,6 +404,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		operatorClient,
 		controllerContext.EventRecorder)
 
+	go configInformers.Config().V1().ClusterOperators().Informer().Run(ctx.Done())
```
slintes: I do not understand why we need this now, but without things get stuck with `unable to sync caches for StatusSyncer_etcd`

slintes: maybe coderabbit understands, will check next week :)
https://github.com/openshift/cluster-etcd-operator/pull/1484#discussion_r2346447184

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -113,7 +117,23 @@ func (c *TargetConfigController) sync(ctx context.Context, syncCtx factory.SyncC
...
+			// Since this can happen during a CEO update, when pacemaker is already running the etcd container,
+			// but TNF jobs were recreated and not ready yet, we should not continue here,
+			// in order to prevent re-adding CEO's etcd container.
+			klog.Infof("Dual replica cluster is in bootstrap completed state, but TNF setup job is not ready yet. Requeue in 5 seconds.")
+			time.Sleep(5 * time.Second)
+			c.enqueueFn()
+			return nil
+		}
+		shouldRemoveEtcdContainer = c.dualReplicaClusterStatus.IsReadyForEtcdRemoval()
+	}
```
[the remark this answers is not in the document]
slintes: ahhh... nice!

pkg/tnf/pkg/jobs/conditions_test.go
```diff
@@ -0,0 +1,331 @@
...
+				}
+			}
+
+			if progressing != tc.expectedProgressing {
+				t.Errorf("expected progressing=%v, got %v", tc.expectedProgressing, progressing)
+			}
+			if ready != tc.expectedReady {
+				t.Errorf("expected ready=%v, got %v", tc.expectedReady, ready)
+			}
+		})
```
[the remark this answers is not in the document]
slintes: fixed

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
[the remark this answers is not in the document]
slintes: done

pkg/operator/starter.go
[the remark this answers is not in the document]
slintes: fixed

slintes: updated: there were informers added to configInformers, so it needs to be started again. Moved that to the bottom, where other informers are started as well.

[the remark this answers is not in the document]
slintes: Never ask for shoutouts again, and never mention X again

pkg/tnf/pkg/status/status.go
```diff
@@ -0,0 +1,91 @@
...
+				klog.Errorf("Failed to check if TNF setup is ready for etcd container removal: %v; using last known status: %v", err, newStatus)
+			} else if ready {
+				newStatus = DualReplicaClusterStatusSetupReadyForEtcdRemoval
+				cs.wasReady = true
+			}
+		}
+
+	}
+	return newStatus
+}
```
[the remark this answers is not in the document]
slintes: IMHO blocking for a potential single API call is fine and your suggestion adds unneeded complexity

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
```diff
@@ -28,17 +22,25 @@ import (
...
-	operatorImagePullSpec string
-	envVarGetter          etcdenvvar.EnvVar
-	etcdLister            operatorv1listers.EtcdLister
-
-	kubeClient kubernetes.Interface
+	targetImagePullSpec      string
+	operatorImagePullSpec    string
+	envVarGetter             etcdenvvar.EnvVar
+	etcdLister               operatorv1listers.EtcdLister
+	dualReplicaClusterStatus status.ClusterStatus
```
jaypoulz: What do you think about calling this `externalEtcdClusterStatus` instead of dualReplicaClusterStatus?
I know the two are hand-in hand right now, but to me, DualReplicaClusterStatus doesn't evoke the thing we're trying to keep a status of.

pkg/operator/starter.go
```diff
@@ -278,11 +278,44 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		operatorClient,
 		cachedMemberClient)
 
+	// the configInformer has to be started before using the featureGateAccessor
+	configInformers.Start(ctx.Done())
+	configInformers.WaitForCacheSync(ctx.Done())
+	go featureGateAccessor.Run(ctx)
```
jaypoulz: Personally, I don't thing we should bother checking the feature gate. We should trust the installer to enforce that boundary, and rely on the DualReplica controlPlaneTopology. ControlPlaneTopology is technically readonly, and IIRC Joel may have added an admission policy to enforce that.

pkg/operator/starter.go
```diff
@@ -278,11 +278,44 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+	// we need the dynamic informers to be started before the using dualReplicaClusterHandler
+	// for being able to use the operator client
+	dynamicInformers.Start(ctx.Done())
+	dynamicInformers.WaitForCacheSync(ctx.Done())
+	// for the job informer we also need the kubeInformersForNamespaces to be started
+	kubeInformersForNamespaces.Start(ctx.Done())
+	kubeInformersForNamespaces.WaitForCacheSync(ctx.Done())
+	// the dualReplicaClusterHandler is needed for the targetConfigReconciler for correct handling
+	// of CEO's etcd container
+	dualReplicaClusterHandler, err := tnf.NewDualReplicaClusterHandler(ctx, operatorClient, kubeClient, featureGateAccessor, configInformers)
```
jaypoulz: Similar idea to the one above, but I'd prefer we'd treat this behavior as an `externalEtcdClusterHandler`. It's not that we have other kinds of clusters that fit this pattern at the moment, but I think of TNF as a kind of cluster config that requires an externalEtcdClusterHandler.

pkg/tnf/pkg/status/status.go
```diff
@@ -0,0 +1,92 @@
...
+
+const (
+	DualReplicaClusterStatusDisabled = iota
+	DualReplicaClusterStatusEnabled
+	DualReplicaClusterStatusBootstrapCompleted
+	DualReplicaClusterStatusSetupReadyForEtcdRemoval
+)
+
+type ClusterStatus interface {
+	IsDualReplicaTopology() bool
```
jaypoulz: I would suggest this boolean be something like `RequiresExternalEtcd`. The function can call into IsDualReplicaTopology in the CEO helpers, but it should be possible to extend the function down the line to add additional criteria to the check.

pkg/operator/targetconfigcontroller/targetconfigcontroller_test.go
```diff
@@ -108,29 +137,71 @@ func TestTargetConfigController(t *testing.T) {
 				u.FakeEtcdMemberWithoutServer(1),
 				u.FakeEtcdMemberWithoutServer(2),
 			},
+			dualReplicaStatus: &mockClusterStatus{},
 		},
 		{
-			name: "Container Removed",
+			name: "Dual Replica Cluster - Topology Enabled",
```
jaypoulz: nit: DualReplica should be one word

pkg/tnf/operator/starter_test.go
```diff
@@ -51,46 +52,59 @@ type args struct {
...
+			wantHandlerInitErr: false,
+			wantStarted:        false,
+			wantErr:            false,
 		},
 		{
-			name:        "Dual replica topology without feature gate",
-			args:        getArgs(t, true, false),
-			wantStarted: false,
-			wantErr:     true,
+			name:               "Dual replica topology without feature gate",
```
jaypoulz: nit: I know the names had a space before but can we make it `DualReplica` while we are changing these :D

pkg/operator/starter.go
jaypoulz: After looking at this further, I think having a DualReplicaClusterHandler is fine - it could implement an interface of externalEtcdClusterHandler in the future if need be.

pkg/tnf/pkg/status/status.go
slintes: Also referring to other comments about naming: Good idea.
I left everything about the handler named DualReplica because it's specific for DualReplica, but changed everything about the status to ExternalEtcd, because that's the relevant thing for the other CEO controllers. I understand that's what you aiming for?

pkg/operator/starter.go
slintes: ack, removed use of feature gates in the dual replica handler

pkg/tnf/pkg/etcd/etcd.go
```diff
@@ -2,48 +2,29 @@ package etcd
...
+func RemoveStaticContainer(ctx context.Context, operatorClient operatorversionedclient.Interface, kubeClient kubernetes.Interface) error {
+	klog.Info("Signaling CEO that TNF setup is ready for etcd container removal")
 
-	etcd, err := operatorClient.OperatorV1().Etcds().Get(ctx, "cluster", metav1.GetOptions{})
+	// Set the job status condition as signal to CEO
+	err := jobs.SetTNFReadyForEtcdContainerRemoval(ctx, kubeClient)
 	if err != nil {
-		klog.Error(err, "Failed to get Etcd")
+		klog.Errorf("Failed to set TNF ready condition: %v", err)
 		return err
```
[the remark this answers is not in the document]
slintes: this compiles fine 🤷🏼

pkg/tnf/operator/starter.go
```diff
@@ -97,9 +146,27 @@ func HandleDualReplicaClusters(ctx context.Context,
...
+					}
+					err = bootstrapteardown.WaitForEtcdBootstrap(ctx, clientConfig)
+					if err != nil {
+						klog.Errorf("failed to wait for bootstrap to complete: %v", err)
+						return
+					}
+					h.clusterStatus.SetBootstrapCompleted()
+				}
+				klog.Infof("bootstrap completed, creating TNF job controllers")
+
```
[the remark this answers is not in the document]
slintes: check the last commit, should resolve the issue

pkg/tnf/operator/starter.go
```diff
@@ -61,54 +112,24 @@ func HandleDualReplicaClusters(ctx context.Context,
...
-				for _, node := range nodeList {
-					runJobController(ctx, tools.JobTypeAuth, &node.Name, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
-					runJobController(ctx, tools.JobTypeAfterSetup, &node.Name, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
-				}
-				runJobController(ctx, tools.JobTypeSetup, nil, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
-				runJobController(ctx, tools.JobTypeFencing, nil, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
-			})
-		},
+		AddFunc: h.handleAddedNode(controllerContext, controlPlaneNodeLister, once, ctx, operatorClient, kubeClient, kubeInformersForNamespaces),
 	})
```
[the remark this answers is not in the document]
slintes: that's why no job is starting on my latest cluster 🙈

pkg/tnf/operator/starter.go
```diff
@@ -46,11 +48,60 @@ var (
...
+	kubeClient kubernetes.Interface,
+	configInformers configv1informers.SharedInformerFactory) (*DualReplicaClusterHandler, error) {
+
+	bootstrapCompleted := false
+	readyForEtcdRemoval := false
+	dualReplicaEnabled, err := ceohelpers.IsDualReplicaTopology(ctx, configInformers.Config().V1().Infrastructures().Lister())
+	if err != nil {
+		klog.Errorf("failed to determine DualReplicaTopology, aborting controller start: %v", err)
+		return nil, err
+	} else if dualReplicaEnabled {
```
jaypoulz: We technically don't need this else since it's implied. (The if statement returns, so we'd save an indentation level)

pkg/tnf/operator/starter.go
```diff
@@ -46,11 +48,60 @@ var (
...
+		klog.Infof("detected DualReplica topology")
+
+		// Check if bootstrap is completed without waiting for it.
+		// If it's completed already, CEO probably was restarted for an update.
+		_, opStatus, _, err := opClient.GetStaticPodOperatorState()
+		if err != nil {
+			klog.Errorf("failed to get static pod operator state: %v", err)
+			return nil, err
+		} else {
```
jaypoulz: This else statement is also implied, since you return if you can't find the operator state.

pkg/tnf/operator/starter.go
```diff
@@ -138,19 +159,81 @@ func HandleDualReplicaClusters(ctx context.Context,
...
+			return
+		}
+		if len(nodeList) != 2 {
+			klog.Info("not starting TNF jobs yet, waiting for 2 control plane nodes to exist")
+			return
+		}
+		klog.Infof("found 2 control plane nodes (%q, %q)", nodeList[0].GetName(), nodeList[1].GetName())
+
+		// we can have 2 nodes on the first call of AddFunc already, ensure we create job controllers once only
+		once.Do(func() {
```
jaypoulz, line 151: We will need to revisit this in this release to ensure that we want control-plane node replacement.
Basically, in the case that are doing initial deployment, we run all of the jobs once.

In the case that we are removing a node post-externaletcd happening, we would run the following on the survivor:
$ sudo pcs cluster node remove {deleted-node}

Assuming that's safe to do from podman-etcd's perspective. @clobrano 

If a second node is added post install, we need to do a few things to get it up and running:
- Run the auth job on the new node
- Recreate the etcd certs on it
- Recreate the revision.json on it
- Recreate the etcd pod.yaml on it
- Run `sudo pcs cluster node add {added-node} addr={added-node-ip} --start --enable` (from the "survivor" node)
- Run the after setup job on this new node

I expect this to be something we can do in a new patch, but I wanted to highlight it here in case we needed to tweak this in any way to accommodate that.

pkg/tnf/operator/starter_test.go
```diff
@@ -51,46 +49,45 @@ type args struct {
...
+			wantHandlerInitErr: false,
+			wantStarted:        false,
+			wantErr:            false,
 		},
 		{
-			name:        "Dual replica topology with feature gate",
-			args:        getArgs(t, true, true),
-			wantStarted: true,
-			wantErr:     false,
+			name:               "Dual replica topology",
```
jaypoulz: nit: DualReplica without space

pkg/tnf/operator/starter.go
slintes: yes, but since that's a topic on its own and outside of this PR, I left out any node replacement considerations for now. And `Once` was used before already, it's not new in this PR.

pkg/tnf/operator/starter.go
slintes: thx, fixed

pkg/tnf/operator/starter.go
slintes: we need a linter for this :D

pkg/tnf/pkg/status/status.go
jaypoulz: Yes, perfect!

pkg/tnf/operator/starter.go
jaypoulz: sounds good!

pkg/tnf/pkg/etcd/etcd.go
```diff
@@ -2,81 +2,68 @@ package etcd
...
-				klog.Infof("static etcd removed: node %s, current rev %v, latest rev %v", nodeStatus.NodeName, nodeStatus.CurrentRevision, etcd.Status.LatestAvailableRevision)
+		for _, nodeStatus := range status.NodeStatuses {
+			if nodeStatus.CurrentRevision == status.LatestAvailableRevision {
+				klog.Infof("static etcd removed: node %s, current rev %v, latest rev %v", nodeStatus.NodeName, nodeStatus.CurrentRevision, status.LatestAvailableRevision)
 			} else {
-				klog.Infof("static etcd not removed yet: node %s, current rev %v, latest rev %v", nodeStatus.NodeName, nodeStatus.CurrentRevision, etcd.Status.LatestAvailableRevision)
+				klog.Infof("static etcd not removed yet: node %s, current rev %v, latest rev %v", nodeStatus.NodeName, nodeStatus.CurrentRevision, status.LatestAvailableRevision)
 				removed = false
 			}
 		}
```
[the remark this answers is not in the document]
slintes: it's impossible to have no status status or revison 0 when this runs

pkg/tnf/operator/starter.go
```diff
@@ -46,11 +48,61 @@ var (
...
+		}
+		bootstrapCompleted = v1helpers.IsOperatorConditionTrue(opStatus.Conditions, "EtcdRunningInCluster")
+		if bootstrapCompleted {
+			klog.Infof("and bootstrap completed already")
+		}
+		readyForEtcdTransition = v1helpers.IsOperatorConditionTrue(opStatus.Conditions, etcd.OperatorConditionExternalEtcdReadyForTransition)
+		if readyForEtcdTransition {
+			klog.Infof("and ready for etcd transition already")
+		}
+	}
```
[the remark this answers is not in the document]
slintes: no clue why there wasn't a constant yet. However, out of scope of this PR

pkg/operator/starter.go
```diff
@@ -278,11 +278,24 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		operatorClient,
 		cachedMemberClient)
 
+	// we need the dynamic and config informers to be started before the using dualReplicaClusterHandler
+	// for being able to check topology and use the operator client
+	dynamicInformers.Start(ctx.Done())
```
tjungblu: so we're starting them twice? what's with the one below?

pkg/operator/starter.go
```diff
@@ -278,11 +278,24 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+	dynamicInformers.WaitForCacheSync(ctx.Done())
+	configInformers.Start(ctx.Done())
```
tjungblu: ^ same thing as for the dynamicInformer above

pkg/operator/starter.go
```diff
@@ -278,11 +278,24 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		operatorClient,
 		cachedMemberClient)
 
+	// we need the dynamic and config informers to be started before the using dualReplicaClusterHandler
```
tjungblu: if this is a problem only for the dualReplicaClusterHandler, then let's put the cache sync waiting into it, the other controllers will simply wait for the informer caches to be in sync by default

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -51,6 +51,7 @@ func NewTargetConfigController(
 	livenessChecker *health.MultiAlivenessChecker,
 	targetImagePullSpec, operatorImagePullSpec string,
 	operatorClient v1helpers.StaticPodOperatorClient,
+	externalEtcdClusterStatus status.ExternalEtcdClusterStatus,
```
tjungblu: why are you injecting the status and not the client?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -113,7 +115,14 @@ func (c *TargetConfigController) sync(ctx context.Context, syncCtx factory.SyncC
 		return err
 	}
 
-	err = c.createTargetConfig(ctx, syncCtx.Recorder(), operatorSpec, envVars, etcd)
+	// Check if we need to remove the etcd container for external etcd usage.
+	// Currently used by DualReplica aka Two Node Fencing clusters.
+	shouldRemoveEtcdContainer := false
+	if c.externalEtcdClusterStatus.IsExternalEtcdCluster() && c.externalEtcdClusterStatus.IsReadyForEtcdTransition() {
```
tjungblu: this could just be a ceohelper function, based on the input from v1helpers.StaticPodOperatorClient - that would also make your mocking 100% easier and it wouldn't be such a pain with the locking inside your status.

Plus points, you'd get the free informer cache synchronization from this controller

pkg/operator/starter.go
jaypoulz: So I tried to address this as follows:
1. I pulled all of the informers up that are used across the "guaranteed" controllers to start as soon as possible
2. I pushed the cache sync calls down into the function that creates the DualReplicaClusterHandler

Functionally speaking, it doesn't change much.
- The targetConfigReconciler still needs the ExternalEtcdClusterStatus when it's created, so so we need need that first sync to happen so we can safely create and start that controller
- There is a different pattern taken for NewIsSingleNodePlatformFn for the guardRolloutPreCheck where we just return false if we haven't synced yet.

The only alternative I can think of is to hold off on clusterStatus calculations until calls to getStatus, but even then we'd just be waiting for the cache sync to happen a few instructions later for the creation of the targetConfigReconciler.

To really avoid this sync, we'd need to structure the status like an informer.

tjungblu: The ordering of when an informer is synced should not depend on this. You can, in the targetconfigcontroller, very simply get informed when the condition on the operator changed. The controller takes care of synchronizing it for you. 

There is simply no reason to fiddle around with a mutex or with waiting for cache syncs. I can send you an alternative, if you fancy.

jaypoulz: actually, the ceohelpers route might help address this. Investigating if I can relieve the informer syncing by calling into the helpers in the controllers instead.

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
```diff
@@ -156,11 +159,12 @@ func (c *ExternalEtcdEnablerController) supportExternalEtcdOnlyPod(
 	podConfigMap.Data["forceRedeploymentReason"] = operatorSpec.ForceRedeploymentReason
 	podConfigMap.Data["version"] = version.Get().String()
 
-	enabled, err := ceohelpers.IsExternalEtcdSupport(operatorSpec)
+	// we can prepare the external etcd support already while TNF setup is still running
+	supportExternalEtcd, err := ceohelpers.IsEtcdRunningInCluster(ctx, c.operatorClient)
```
jaypoulz: @tjungblu do I need to wait for the infra informer to sync before this or is this handled by the operator client

pkg/tnf/operator/starter.go
```diff
@@ -138,19 +128,84 @@ func HandleDualReplicaClusters(ctx context.Context,
...
+		runJobController(ctx, tools.JobTypeAuth, &node.Name, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
+		runJobController(ctx, tools.JobTypeAfterSetup, &node.Name, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
 	}
-	return true, nil
+	runJobController(ctx, tools.JobTypeSetup, nil, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
+	runJobController(ctx, tools.JobTypeFencing, nil, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces)
+}
+
+func (h *DualReplicaClusterHandler) waitForEtcdBootstrapCompleted(ctx context.Context, operatorClient v1helpers.StaticPodOperatorClient) error {
+	isEtcdRunningInCluster, err := ceohelpers.IsEtcdRunningInCluster(ctx, operatorClient)
```
jaypoulz, line 183: Same question about whether we need to sync before this call

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
```diff
@@ -156,11 +163,12 @@ func (c *ExternalEtcdEnablerController) supportExternalEtcdOnlyPod(
 	podConfigMap.Data["forceRedeploymentReason"] = operatorSpec.ForceRedeploymentReason
 	podConfigMap.Data["version"] = version.Get().String()
 
-	enabled, err := ceohelpers.IsExternalEtcdSupport(operatorSpec)
+	// we can prepare the external etcd support already while TNF setup is still running
+	supportExternalEtcd, err := ceohelpers.IsExternalEtcdCluster(ctx, c.infrastructureLister)
```
jaypoulz: @tjungblu will the infra lister sync as part of the controller, or do i need to add an cache wait somewhere?

pkg/tnf/operator/starter.go
jaypoulz: In this case, I guess I'm wondering if I need to sync the operatorClient informers

jaypoulz: (This specifically checks the conditions)

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -113,7 +115,18 @@ func (c *TargetConfigController) sync(ctx context.Context, syncCtx factory.SyncC
 		return err
 	}
 
-	err = c.createTargetConfig(ctx, syncCtx.Recorder(), operatorSpec, envVars, etcd)
+	// Check if we need to remove the etcd container for external etcd usage.
+	// Currently used by DualReplica aka Two Node Fencing clusters.
+	shouldRemoveEtcdContainer := false
+	externalEtcdStatus, err := ceohelpers.GetExternalEtcdClusterStatus(ctx, c.operatorClient, c.infrastructureLister)
```
jaypoulz, line 121: @tjungblu Here, the sync question applies to both objects.

tjungblu: that's fine, the operator above is defined like this:

``` 
	WithInformers(
			operatorClient.Informer(),
            ....
           infrastructureInformer.Informer(),
```

so you are guaranteed that an invocation on SYNC is always with a pre-filled cache and you get a notification invoking sync if any of those objects change.

pkg/operator/starter.go
```diff
@@ -208,6 +208,24 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+	// - the clusterInformers are used by the staticPodBuilderControllers
+	// - the kubeInformersForNamespaces are used by multiple controllers
+	go controlPlaneNodeInformer.Run(ctx.Done())
+	go etcdNamespaceInformer.Run(ctx.Done())
+	go featureGateAccessor.Run(ctx)
+	configInformers.Start(ctx.Done())
+	dynamicInformers.Start(ctx.Done())
+	operatorInformers.Start(ctx.Done())
+	clusterInformers.Start(ctx.Done())
+	kubeInformersForNamespaces.Start(ctx.Done())
```
tjungblu: with the changes you've made, you don't have to put those here, you can just leave them at the bottom where they were originally.

pkg/operator/starter.go
```diff
@@ -278,6 +296,9 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		operatorClient,
 		cachedMemberClient)
 
+	// the dualReplicaClusterHandler handles the dual replica topology transitions to external etcd
```
tjungblu: do you still need the handler?

pkg/operator/ceohelpers/external_etcd_status_test.go
```diff
@@ -0,0 +1,302 @@
...
+		},
+	}
+
+	return v1helpers.NewFakeStaticPodOperatorClient(
+		&operatorv1.StaticPodOperatorSpec{},
+		operatorStatus,
+		nil,
+		nil,
+	)
+}
```
tjungblu: that's probably Cursor, but those very likely already exist in testutils - if not, please add them there :)

pkg/operator/starter.go
jaypoulz: I don't think so. The only extra in the struct was a passed ctx. So I just pass that directly to the function now, and call the functions directly.

pkg/operator/starter.go
jaypoulz: Restored

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
jaypoulz: Thomas answered this question already in another comment. The infralister is updated because the infra informer is part of the factory init call in the constructor of this controller.

pkg/operator/externaletcdsupportcontroller/external_etcd_support_controller.go
jaypoulz: the operator client informer is synced by via the informer listing in the constructor of this file.

pkg/tnf/operator/starter.go
jaypoulz: @tjungblu This is the line I'm still not sure of. I know the generic operator client is created in starter.go: https://github.com/openshift/cluster-etcd-operator/blob/d1dbe0994de7aaa69579539172e122c09e16c41b/pkg/operator/starter.go#L206

But I'm not sure this translates to a guarantee that the client will have synced at least once before we try to look up the StaticPodOperatorState: https://github.com/openshift/cluster-etcd-operator/blob/d1dbe0994de7aaa69579539172e122c09e16c41b/pkg/operator/ceohelpers/external_etcd_status.go#L40

jaypoulz:
/retitle OCPEDGE-2033: [TNF] Use operator condition to track etcd container removal in ExternalEtcd clusters

pkg/tnf/operator/starter.go
tjungblu: yeah, so you need to use the informer and lister facility if you want cache coherence. 

If you're using a client, then it's basically a synchronous API call every time, so there is no caching involved.

tjungblu: okay word of warning: look into the implementation. This is not always as straight :)

tjungblu: turns out, it is a mixed client. The client uses the informer cache for the operator state:
https://github.com/openshift/library-go/blob/master/pkg/operator/genericoperatorclient/dynamic_operator_client.go#L132-L140

for other calls it goes directly against the API. 

For this case, we should be OK? all of the calls to GetStaticPodOperatorState are within a Sync function of a factory controller.

pkg/operator/ceohelpers/podsubstitution.go
```diff
@@ -62,25 +63,20 @@ func calculateFailureThreshold(baseThreshold int, quotaGiB int32) int {
 }
 
 // GetPodSubstitution creates a PodSubstitutionTemplate with values derived from StaticPodOperatorSpec,
-// image pull spec and environment variables. It determines whether the Etcd container should be enabled
-// based on the operator's configuration.
+// image pull spec, environment variables and the given removeEtcdContainer flag.
 func GetPodSubstitution(
 	operatorSpec *operatorv1.StaticPodOperatorSpec,
```
tjungblu, line 68: btw, is this operatorSpec still used now?

jaypoulz: Only to set the log level: https://github.com/openshift/cluster-etcd-operator/blob/64c5ffad5f81b58eb632591e0ab7d7933e19d34e/pkg/operator/ceohelpers/podsubstitution.go#L117

If preferred, we could just pass in the log level instead of the full spec.

tjungblu: no all good, the codesearch failed me once again - please disregard

PR #1485, opened 2025-09-14
[WIP] OCPEDGE-2175: embed fencing validator into etcd scripts for TNF clusters (DualReplica FG)

Neilhamza:
wiring feature gate access into the ScriptController and ensuring that fencing-validator.sh is embedded and surfaced via the etcd-scripts ConfigMap. The script is included only when the DualReplica feature gate is enabled (TNF clusters).

Updated ScriptController to accept a FeatureGateAccess.

Embedded tnfdeployment/tnf-scripts/fencing-validator.sh into the operator binary.

Synced script into the etcd-scripts ConfigMap conditionally on DualReplica FG.

TNF clusters will now receive the fencing-validator.sh, making it available for validation and testing workflows.

** Script itself probably could be reviewed in a different PR **

TNF cluster deployed with this PR shows the script is present:
image

copying it from the node was Successful  via: 
```
oc debug node/master-0 -- chroot /host cat /usr/local/bin/fencing-validator.sh > fencing-validator.sh
chmod +x fencing-validator.sh
```

### DIFF
--- a/bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
+++ b/bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
@@ -1,0 +1,445 @@
+#!/usr/bin/env bash
+set -euo pipefail
+export LANG=C
+
+# fencing-validator
+# - Non-disruptive: verify STONITH present+enabled, both nodes online, etcd has 2 started non-learner members, Daemon status active
+# - Disruptive (optional): fence node A then node B (one pass), auto-switching the conductor to avoid self-fencing
+# - Transport: auto (prefer SSH to both), or forced ssh/ocdebug
+
+usage() {
+  cat <<'EOF'
+Usage:
+  fencing-validator [--user <ssh-user>] [--ssh-key <path>]
+                       [--kubeconfig <path>]
+                       [--transport auto|ssh|ocdebug]
+                       [--hosts "<hostA,hostB>"] [--host-a <host>] [--host-b <host>]
+                       [--disruptive] [--dry-run] [--timeout <sec>]
+
+Examples (hosts):
+  --hosts "10.0.0.10,10.0.0.11"
+  --hosts "2001:db8::a,2001:db8::b"
+
+Note: For IPv6, pass the raw address (no brackets). The script adds [ ] where needed.
+
+Timeouts:
+  --timeout <sec> / TIMEOUT
+      Maximum time (in seconds) to wait for a condition to succeed.
+      This is the overall loop timeout (e.g., waiting for a node or etcd to recover).
+      Default: 1200 seconds (20 min).
+
+  CMD_EXEC_TIMEOUT_SECS
+      Maximum time (in seconds) allowed for a single remote command to run
+      (e.g., one `podman exec`, one `pcs status` call).
+      This is enforced in `host_run` for both SSH and oc debug transports.
+      Default: 60 seconds.
+
+  OC_REQ_TIMEOUT
+      Per-request API timeout for the `oc` client when contacting the API server.
+      Applies to each HTTP request inside `oc` commands, not the overall command runtime.
+      Default: 10s.
+
+
+Env (optional): SSH_USER, SSH_KEY, KUBECONFIG, TRANSPORT, DISRUPTIVE, DRY_RUN, TIMEOUT, IP_A, IP_B, OC_BIN, OC_REQ_TIMEOUT, CMD_EXEC_TIMEOUT_SECS
+EOF
+}
+
+log(){ printf '\033[36m[INFO]\033[0m %s\n' "$*"; }
+warn(){ printf '\033[33m[WARN]\033[0m %s\n' "$*"; }
+err(){ printf '\033[31m[ERROR]\033[0m %s\n' "$*" >&2; }
+ok(){  printf '\033[32m[OK]\033[0m %s\n' "$*"; }
+
+# -------- Defaults / args --------
+SSH_USER="${SSH_USER:-core}"
+SSH_KEY="${SSH_KEY:-}"
+KUBECONFIG_PATH="${KUBECONFIG:-}"
+TRANSPORT="${TRANSPORT:-auto}"
+DISRUPTIVE="${DISRUPTIVE:-false}"
+DRY_RUN="${DRY_RUN:-false}"
+TIMEOUT="${TIMEOUT:-1200}"
+IP_A="${IP_A:-}"
+IP_B="${IP_B:-}"
+OC_BIN="${OC_BIN:-oc}"
+OC_REQ_TIMEOUT="${OC_REQ_TIMEOUT:-10s}"
+CMD_EXEC_TIMEOUT_SECS="${CMD_EXEC_TIMEOUT_SECS:-60s}"
+
+# valreq <flag> <value>
+# Returns success (0) if <value> exists and is not another option (i.e., doesn't start with '-').
+# Used in argument parsing to validate that an option expecting a value actually has one.
+valreq(){ [[ -n "${2-}" && "$2" != -* ]]; }
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --user)       valreq "$1" "${2-}" || { err "--user requires a value"; exit 1; }; SSH_USER="$2"; shift 2;;
+    --ssh-key)    valreq "$1" "${2-}" || { err "--ssh-key requires a value"; exit 1; }; SSH_KEY="$2"; shift 2;;
+    --kubeconfig) valreq "$1" "${2-}" || { err "--kubeconfig requires a value"; exit 1; }; KUBECONFIG_PATH="$2"; shift 2;;
+    --transport)  valreq "$1" "${2-}" || { err "--transport requires a value"; exit 1; }; TRANSPORT="$2"; shift 2;;
+    --timeout)    valreq "$1" "${2-}" || { err "--timeout requires a value"; exit 1; }; TIMEOUT="$2"; shift 2;;
+    --hosts)      valreq "$1" "${2-}" || { err "--hosts requires a value like 'A,B'"; exit 1; }; IFS=',' read -r IP_A IP_B <<<"$2"; shift 2;;
+    --host-a)     valreq "$1" "${2-}" || { err "--host-a requires a value"; exit 1; }; IP_A="$2"; shift 2;;
+    --host-b)     valreq "$1" "${2-}" || { err "--host-b requires a value"; exit 1; }; IP_B="$2"; shift 2;;
+    --disruptive) DISRUPTIVE=true; shift;;
+    --dry-run)    DRY_RUN=true; shift;;
+    -h|--help)    usage; exit 0;;
+    *) err "Unknown arg: $1"; usage; exit 1;;
+  esac
+done
+
+IP_A="${IP_A//[[:space:]]/}"; IP_B="${IP_B//[[:space:]]/}"
+
+[[ "$TRANSPORT" =~ ^(auto|ssh|ocdebug)$ ]] || { err "--transport must be auto|ssh|ocdebug"; exit 1; }
+[[ "$TIMEOUT" =~ ^[0-9]+$ ]] || { err "Invalid --timeout '$TIMEOUT'"; exit 1; }
+
+[[ -n "$KUBECONFIG_PATH" ]] && export KUBECONFIG="$KUBECONFIG_PATH"
+log "Checking cluster access with '$OC_BIN'..."
+$OC_BIN --request-timeout="$OC_REQ_TIMEOUT" whoami >/dev/null
+
+# -------- Basic helpers --------
+_is_ipv6(){ [[ "$1" == *:* ]]; }
+_fmt_host(){
+  local h="$1"
+  h="${h#[[]}"
+  h="${h%[]]}"
+  _is_ipv6 "$h" && echo "[$h]" || echo "$h"
+}
+
+ssh_cmd() {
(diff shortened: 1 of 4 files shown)


bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
```diff
@@ -0,0 +1,445 @@
...
+  if (( rc != 0 )); then
+    if grep -Eqi 'no such container|container state improper|not running|missing required container_id' <<<"$out"; then
+      return 1
+    fi
+    err "etcdctl failed on $tgt: ${out##*$'\n'}"
+    return 2
+  fi
+  return 1
+}
+
```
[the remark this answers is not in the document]
Neilhamza: podman exec is the reliable way of testing the etcd pod status

pkg/operator/scriptcontroller/scriptcontroller.go
```diff
@@ -114,14 +119,17 @@ func (c *ScriptController) manageScriptConfigMap(ctx context.Context, recorder e
...
-	for _, filename := range []string{
+	files := []string{
 		"etcd/cluster-restore.sh",
 		"etcd/quorum-restore.sh",
 		"etcd/cluster-backup.sh",
 		"etcd/disable-etcd.sh",
 		"etcd/etcd-common-tools",
-	} {
+	}
+	if on, err := dualreplicahelpers.DualReplicaFeatureGateEnabled(c.featureGateAccessor); err == nil && on {
```
lance5890, line 129: I'm not sure, but should we also check the `DualReplicaTopology` just like the https://github.com/openshift/cluster-etcd-operator/blob/61b7fb5116f8672bab6fcccbb9336e197ac07602/pkg/tnf/operator/starter.go#L64-L66, https://github.com/openshift/cluster-etcd-operator/blob/61b7fb5116f8672bab6fcccbb9336e197ac07602/pkg/tnf/operator/starter.go#L141-L154

Neilhamza: @slintes @jaypoulz  do i need to check both Dual Replica feature gate and Topology before appending the script? or just FG/Topology? i currently check only FG enabled but i can add topology check logic

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
```diff
@@ -0,0 +1,445 @@
...
+  done
+}
+
+# -------- Discover nodes --------
+log "Detecting control-plane nodes…"
+mapfile -t A < <(timeout "$CMD_EXEC_TIMEOUT_SECS" "$OC_BIN" get nodes -l node-role.kubernetes.io/master= -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' 2>/dev/null || true)
+mapfile -t B < <(timeout "$CMD_EXEC_TIMEOUT_SECS" "$OC_BIN" get nodes -l node-role.kubernetes.io/control-plane= -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' 2>/dev/null || true)
+
+declare -a CP_NODES=(); declare -A seen=()
+for n in "${A[@]}" "${B[@]}"; do [[ -z "${seen["$n"]+x}" ]] && { CP_NODES+=("$n"); seen["$n"]=1; }; done
```
jaypoulz, line 147: TNF only supports clusters in versions post the transition from master to control-plane as the primary control-plane label moniker. It would be sufficient to just check the later.

Another boundary you could set would be OCP v4.20 and above.

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
```diff
@@ -0,0 +1,445 @@
...
+short() { echo "${1%%.*}"; }
+resolve_pcmk_name() {
+  local n="$1"
+  if host_run "$CONDUCTOR" "command -v crm_node >/dev/null 2>&1"; then
+    host_run "$CONDUCTOR" "crm_node -l" | awk -v n="$n" -v s="$(short "$n")" '$2==n||$2==s{print $2; found=1} END{if(!found) print s}'
+  else
+    short "$n"
+  fi
+}
+PCMK_A="$(resolve_pcmk_name "$NODE_A")"
```
jaypoulz, line 194: The pacemaker names are set in the GetClusterConfig function: https://github.com/openshift/cluster-etcd-operator/blob/main/pkg/tnf/pkg/config/cluster.go#L42-L53

This means that they are guaranteed to match names of the nodes used in the cluster (unless someone modifies them after the fact).

I don't this check being here, but I did want to highlight a related problem case.
Before the nodes exist, you we need to set a name for each that will be used for the creation of the fencing credential. Naturally, we use whatever name is passed to us via the install config in the fencing block.

That said, the **USER** is responsible for these matching. We've been bitten by this before - a few times - by dev scripts. A recent issue we ran into for this is when we tried to add IPv6 support, since the hostnames generated by dev-scripts for IPv6 are FQDNs instead of just the initial prefix.

Checking if there is a fencing secret that matches the registered hostname could be something we try to provided feedback for via this script. Long term, I'd like us to ID nodes by MAC address instead of hostname since it's safer.

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
```diff
@@ -0,0 +1,445 @@
...
+log "Mapping: $NODE_A -> $PCMK_A ; $NODE_B -> $PCMK_B"
+
+# -------- Health/waits --------
+node_ready(){
+  local n="$1"
+  local s; s="$("$OC_BIN" get node "$n" --request-timeout=10s -o jsonpath='{range .status.conditions[?(@.type=="Ready")]}{.status}{end}' 2>/dev/null || true)"
+  [[ "$s" == *True* ]]
+}
+
+pcs_nodes_line(){
```
jaypoulz, line 205: Unused function?

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
```diff
@@ -0,0 +1,445 @@
...
+
+pcs_nodes_names() {
+  local out
+  out="$(host_run "$CONDUCTOR" "pcs status nodes 2>/dev/null || crm_mon -1 2>/dev/null" 2>/dev/null || true)"
+  awk '/^[[:space:]]*Online:/{ 
+    for (i=2;i<=NF;i++) { gsub(/[][]/,"",$i); printf "%s ", $i } 
+  } END{ print "" }' <<<"$out"
+}
+
+pcmkonline() {
```
jaypoulz, line 219: IDK if pacemaker sorts alphabetically by default. It might be covered by the fact that you built the node list with the same ordering as pacemaker though.

Might be worth making a comment about it at least.

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
```diff
@@ -0,0 +1,445 @@
...
+
+wait_not_ready(){
+  local n="$1" deadline=$((SECONDS+TIMEOUT))
+  log "Waiting for '$n' to become NotReady (API)"
+  while (( SECONDS < deadline )); do
+    if ! node_ready "$n"; then
+      ok "$n NotReady (API)"
+      return 0
+    fi
+    sleep 5
```
jaypoulz, line 235: I think you've selected a different sleep duration each time you call it :laughing:
Can we make a constant and default to 3 or 5 or something?

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
jaypoulz: Maybe I'm just misunderstanding and this is order insensitive.
What does the "${1%%.*}" do?

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
```diff
@@ -0,0 +1,445 @@
...
+# -------- Run --------
+log "Mode: transport=$TRANSPORT disruptive=$DISRUPTIVE dry-run=$DRY_RUN"
+log "=== Non-disruptive validation ==="
+check_stonith
+if ! ( pcmkonline "$PCMK_A" && pcmkonline "$PCMK_B" ); then
+  err "Both nodes must be ONLINE (Pacemaker)"
+  exit 2
+fi
+ok "Both nodes ONLINE"
+check_daemon_status || exit 2
```
jaypoulz, line 419: Can we capture the exit value as a CONSTANT. Might even be worth offering different kinds of exit values for different errors so they're self-documenting.

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
```diff
@@ -0,0 +1,445 @@
...
+
+  etcd_two_started "$(node_exec_target "$NODE_B")"
+  local rb
+  rb=$?
+  (( rb==0 )) && return 0
+  (( rb==2 )) && return 2
+  return 1
+}
+
+wait_etcd(){
```
jaypoulz, line 342: technically speaking, we could have the wrong voter (e.g. the bootstrap node) and a CP node and pass this check. Could we tighten it so that the matched nodes are the ones we've identified as the CP nodes?

etcd tracks that by IP, and I believe you have the internal IP captured above

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
```diff
@@ -0,0 +1,445 @@
...
+}
+
+fence(){
+  local t="$1"
+  log "[ACTION] Fencing (reboot) '$t' from '$CONDUCTOR'..."
+  local to=$(( TIMEOUT/2 )); (( to < 1 )) && to=1
+
+  if [[ "$TRANSPORT" == "ocdebug" ]]; then
+    if ! host_run "$CONDUCTOR" "command -v systemd-run >/dev/null 2>&1 && \
+        systemd-run --unit fence-$t --collect bash -lc 'pcs stonith fence $t' \
```
jaypoulz, line 383: For my own understanding, is there a reason we run this via systemd?

pkg/operator/scriptcontroller/scriptcontroller.go
jaypoulz: You only need to check the DualReplica topology, and don't need to check the FeatureGate. The topology is protected by the gate in the installer, and in future versions, the gate will be gone but the topology will remain.

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
Neilhamza: Good point — the check is order-insensitive since it just looks for membership, not position. ${1%%.*} is just stripping the domain so we can match either FQDN or short hostnames.
${1%%.*} strips off everything from the first dot (.) onward.
Example:
Input: master-0.example.com
Output: master-0

and regarding the first comment:
It’s just a substring match e.g. [[ " node-a node-b " == *" node-b "* ]] works no matter the order. So we’re safe.
$want = master-0.example.com
$s = master-0
$names = " master-0 master-1 "
Since $names contains " master-0 ", the expression returns true, even though the FQDN didn’t match.
In practice this means pcmkonline master-0.example.com will return success if Pacemaker has master-0 listed as online.

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
Neilhamza: sure! when we are in oc debug basically anything we run inside this pod will be tied to the debug pod lifetime
using it inline is unstable because of some risks
-)the script will hang if the pod hangs
-)exit status can be unclear sometimes
-)fencing action can be killed before the pod goes away (causing a loop-freeze)
and some more risks that were captured when i did use it inline

this way we have much more control over the command (capturing, stopping, output, exit status...)

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
Neilhamza: @jaypoulz updated! please view https://github.com/openshift-eng/two-node-toolbox/pull/23 PR that holds the requested changes

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
Neilhamza: updated! please view https://github.com/openshift-eng/two-node-toolbox/pull/23 PR that holds the requested changes

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
Neilhamza: updated! please view https://github.com/openshift-eng/two-node-toolbox/pull/23 PR that holds the requested changes

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
Neilhamza: updated! please view https://github.com/openshift-eng/two-node-toolbox/pull/23 PR that holds the requested changes

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
Neilhamza: updated! please view https://github.com/openshift-eng/two-node-toolbox/pull/23 PR that holds the requested changes

bindata/tnfdeployment/tnf-scripts/fencing-validator.sh
Neilhamza: updated! please view https://github.com/openshift-eng/two-node-toolbox/pull/23 PR that holds the requested changes

PR #1486, opened 2025-09-17
[WIP] disruptive fencing validation TNF

Neilhamza:
ValidateDisruptivePeerOnly flow to test fencing a peer node in DualReplica topologies.

Adds fencing validation logic (pcs stonith + etcd quorum check)

Integrates into TNF setup runner behind an environment gate (TNF_VALIDATE_PEER_ONLY)

Improves safety with self-fence guard, explicit error handling, and pcs/etcd health waits

### DIFF
--- a/pkg/tnf/pkg/pcs/fencing.go
+++ b/pkg/tnf/pkg/pcs/fencing.go
@@ -5,19 +5,19 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
+	"github.com/openshift/cluster-etcd-operator/pkg/etcdcli"
 	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/config"
 	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/exec"
 	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/tools"
 )
 
-var (
-	addressRegEx = regexp.MustCompile(`.*//(.*):(.*)(/redfish.*)`)
-)
+var addressRegEx = regexp.MustCompile(`.*//(.*):(.*)(/redfish.*)`)
 
 const (
 	// defaultPcmkDelayBase is the delay applied to the first fence device to prevent simultaneous fencing
@@ -117,11 +117,9 @@ func ConfigureFencing(ctx context.Context, kubeClient kubernetes.Interface, cfg 
 
 	klog.Info("Fencing configuration succeeded!")
 	return nil
-
 }
 
 func getFencingConfig(nodeName string, secret *corev1.Secret) (*fencingConfig, error) {
-
 	address := string(secret.Data["address"])
 	if !strings.Contains(address, "redfish") {
 		klog.Errorf("Secret %s does not contain redfish address", secret.Name)
@@ -184,7 +182,6 @@ func getStatusCommand(fc fencingConfig) string {
 }
 
 func getStonithCommand(sc StonithConfig, fc fencingConfig) string {
-
 	stonithAction := "create"
 	// check if device already exists
 	for _, p := range sc.Primitives {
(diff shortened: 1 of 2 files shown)


pkg/tnf/disruptivevalidate/runner.go
```diff
@@ -0,0 +1,249 @@
...
+			j := &jl.Items[i]
+			if j.Status.Succeeded > 0 || tools.IsConditionTrue(j.Status.Conditions, batchv1.JobComplete) {
+				completed++
+			}
+		}
+		return completed >= wantAtLeast, nil
+	})
+}
+
+func detectLocalAndPeer(_ context.Context, _ kubernetes.Interface, n1, n2 string) (string, string, error) {
```
clobrano: What is the reason to pass context and kubernetes.Interface if not used?

pkg/tnf/disruptivevalidate/runner.go
```diff
@@ -0,0 +1,249 @@
...
+	podName, err := os.Hostname()
+	if err != nil || strings.TrimSpace(podName) == "" {
+		return "", "", fmt.Errorf("get pod hostname: %w", err)
```
clobrano, line 205: If the err is `nil`, but `strings.TrimSpace(podName)` returns empty string, the resulting error message won't be very helpful

pkg/tnf/disruptivevalidate/runner.go
```diff
@@ -0,0 +1,249 @@
...
+	default:
+		return "", "", fmt.Errorf("local %q not in cluster config (%q,%q)", local, n1, n2)
+	}
+}
+
+func waitPeerValidateIfSecond(ctx context.Context, kc kubernetes.Interface, local, a, b string) error {
+	min, max := a, b
+	if strings.Compare(min, max) > 0 {
+		min, max = max, min
+	}
```
clobrano, line 237: Cool :)

pkg/tnf/disruptivevalidate/runner.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+		klog.V(2).Infof("peer %s status: succeeded=%d failed=%d conditions=%+v", target, j.Status.Succeeded, j.Status.Failed, j.Status.Conditions)
+
+		// Only treat as failed if the JobFailed condition is set
+		if tools.IsConditionTrue(j.Status.Conditions, batchv1.JobFailed) {
+			return false, fmt.Errorf("peer validate job %s failed", target)
+		}
+		// Proceed when the peer is complete
+		return j.Status.Succeeded > 0 || tools.IsConditionTrue(j.Status.Conditions, batchv1.JobComplete), nil
+	})
```
clobrano: This looks very similar to waitForLabeledJob to me (well except the logging). Any chance we can label the validation job too, and reuse the same function?

pkg/tnf/disruptivevalidate/runner.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+		return fmt.Errorf("pcs fence %s failed: %w (last line: %s)", peer, ferr, strings.TrimSpace(ls))
+	}
+	if err := waitPCSState(ctx, boolp(false), peer, timeoutPCSDown); err != nil {
+		return fmt.Errorf("peer didn't go OFFLINE: %w", err)
+	}
+	if err := waitPCSState(ctx, boolp(true), peer, timeoutPCSBackUp); err != nil {
+		return fmt.Errorf("peer didn't return ONLINE: %w", err)
+	}
+	if err := waitEtcdTwoVoters(ctx, timeoutEtcdOK); err != nil {
```
clobrano, line 114: We might also hit Kubelet errors, that would prevent the podman-etcd agent from even attempting to start. If so, the resulting error message would be misleading.

clobrano (commented):
I left some comments, but the direction seems Ok to me

pkg/tnf/disruptivevalidate/runner.go
Neilhamza: this is quite tricky since this job will be running on the node, and the ServiceAccount does not have any RBAC to get lease for peer kubelet. the only way i found for this to work (still did not try - fingers crossed) is to check the the peer kubelet is listening to the same TCP port... should this be a valuable check? do you have any suggestion

pkg/tnf/disruptivevalidate/runner.go
Neilhamza: it was for consistency and future-ready for any API usage but i will remove it for now

pkg/tnf/disruptivevalidate/runner.go
Neilhamza: true! updated

pkg/tnf/disruptivevalidate/runner.go
clobrano: As you can already run commands on the node (e.g. `podman`), I think you can run `systemctl is-active kubelet`, or some similar check. As a fallback, we can also enhance the error message suggesting to check both etcd _and_ kubelet

clobrano (changes requested):

pkg/tnf/disruptivevalidate/runner.go
Neilhamza: absolutely! however i don't think running is-active kubelet on the node that was fencing and not fenced is relevant for out case, or is it?

Neilhamza: because what we care about in the post fence is that the peer node (the one that was fenced) returns in a solid state

clobrano: yeah, you're right. I think it is ok as it is now :+1:

pkg/tnf/disruptivevalidate/runner.go
Neilhamza: i already tag the jobs just like we do with other jobs. however the label is not unique for each node, meaning i had to find a way to detect this specific job for this node
for that case i merged the logic into one function. 
since labeling each node job would require some logic change in other places which could be fragile also for the other jobs which i would rather not touch
now it should look better

clobrano (approved):
/lgtm

PR #1487, opened 2025-09-17
OCPEDGE-2097: Add pacemaker health check for ExternalEtcd clusters

jaypoulz:
Add pacemaker cluster health monitoring for Two Node Fencing deployments:

- PacemakerCluster CRD for status representation
- Status collector CronJob (parses pcs status xml, updates CR)
- Health check controller (monitors CR, sets operator conditions)
- Kubernetes events for fencing history and resource failures

The operator is marked degraded when nodes are offline, critical
resources are not running, fencing is unavailable, or status is stale.


pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,656 @@
...
+const (
+	// Time windows for detecting recent events
+	FailedActionTimeWindow = 5 * time.Minute
+	FencingEventTimeWindow = 24 * time.Hour
+
+	// Resync interval for health check controller
+	HealthCheckResyncInterval = 30 * time.Second
+
+	// Expected number of nodes in ExternalEtcd cluster
+	ExpectedNodeCount = 2
```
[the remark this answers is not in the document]
jaypoulz: ExpectedNodeCount is important to the design of Two-Node OpenShift with Fencing. Pacemaker is expected to be managing clusters of exactly 2 nodes. The only exception to this is during node replacement events. There is a brief moment where one of the nodes will be removed from pacemaker in order to replace it with a fresh node.

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,656 @@
...
+	if len(result.Nodes.Node) == 0 {
+		status.Errors = append(status.Errors, "No nodes found")
+		return
+	}
+
+	// Check if we have the expected number of nodes.
+	// This should almost always be 2, but 1 node is possible during a control-plane node replacement event.
+	if len(result.Nodes.Node) != ExpectedNodeCount {
+		status.Warnings = append(status.Warnings, fmt.Sprintf("Expected %d nodes, found %d", ExpectedNodeCount, len(result.Nodes.Node)))
+	}
```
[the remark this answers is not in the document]
jaypoulz: ExpectedNode count is part of the design of Two Node OpenShift with fencing.
If you have more than 2 nodes or 0 that's an unexpected state. If you have 1 node, the only time that is expected is during node replacement events.

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,656 @@
...
+
+	for _, node := range result.Nodes.Node {
+		if node.Online != NodeStatusOnline {
+			status.Errors = append(status.Errors, fmt.Sprintf("Node %s is not online", node.Name))
+		}
+	}
```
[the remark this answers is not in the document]
jaypoulz: This is the correct behavior for the PR. I've updated the commit message to explain that we considering standby nodes offline.

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,657 @@
...
+
+	// Check if we have the expected number of nodes
+	if len(result.Nodes.Node) != ExpectedNodeCount {
+		status.Errors = append(status.Errors, fmt.Sprintf("Expected %d nodes, found %d", ExpectedNodeCount, len(result.Nodes.Node)))
+		allNodesOnline = false
+	}
+
```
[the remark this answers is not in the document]
jaypoulz: ExpectedNodeCount is core to the design of TNF. It should not be derived dynamically, because that would normalize regularly having more or less than 2 nodes in a two node cluster.

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,657 @@
...
+}
+
+// newErrorHealthStatus creates a HealthStatus with Unknown status and an error message
+func newErrorHealthStatus(errMsg string) *HealthStatus {
+	return &HealthStatus{
+		OverallStatus: StatusUnknown,
+		Warnings:      []string{},
+		Errors:        []string{errMsg},
+	}
+}
```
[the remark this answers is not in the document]
jaypoulz: This is intentional.

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,718 @@
...
+	}
+
+	if stderr != "" {
+		klog.Warningf("pcs status xml command produced stderr: %s", stderr)
+	}
+
+	// Validate XML size to prevent XML bombs
+	if len(stdout) > maxXMLSize {
+		return newUnknownHealthStatus(fmt.Sprintf("XML output too large: %d bytes (max: %d bytes)", len(stdout), maxXMLSize)), nil
+	}
```
clobrano: Note for the future. If the XML Bomb is a genuine risk, we might want to consider modifying the `exec` package to use `io.LimitReader` somehow. The current check prevents us to parse it, and it is good, but we still have the whole XML in memory at that point.

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,718 @@
...
+		return err
+	}
+
+	// Update operator status conditions based on pacemaker status
+	if err := c.updateOperatorStatus(ctx, healthStatus); err != nil {
+		return err
+	}
+
+	// Record pacemaker health check events
+	c.recordHealthCheckEvents(healthStatus)
```
clobrano: Not sure how long does `recordHealthCheckEvents` take, but maybe we should pass the context here.

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,718 @@
...
+			for _, operation := range resourceHistory.OperationHistory {
+				// Check for failed operations (rc != "0")
+				if operation.RC != operationRCSuccess || operation.RCText != operationRCTextSuccess {
+					// Parse the timestamp
+					t, err := time.Parse(pacemakerTimeFormat, operation.LastRCChange)
+					if err != nil {
+						klog.V(4).Infof("Failed to parse timestamp '%s' for operation %s on node %s: %v",
+							operation.LastRCChange, operation.Task, node.Name, err)
+						continue
+					}
```
clobrano: Shouldn't we worry about letting these errors slip through?

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,718 @@
...
+
+	now := time.Now()
+
+	// Clean up old entries that are outside the longest deduplication window
+	maxWindow := eventDeduplicationWindowFencing
+	for key, timestamp := range c.recordedEvents {
+		if now.Sub(timestamp) > maxWindow {
+			delete(c.recordedEvents, key)
+		}
+	}
```
clobrano: This cleanup logic is currently nested inside the block that decides on item addition. Should we move it out? It feels like that cleanup step can be performed just once in `recordHealthCheckEvents` execution, not for every new item to add, did I misunderstand it?
UPDATE: now that I read it a little better, it seems to me that at each loop called in `recordHealthCheckEvents` we might delete events recorded in the previous loop.

pkg/tnf/pacemaker/healthcheck_test.go
```diff
@@ -0,0 +1,744 @@
...
+	now := time.Now()
+	recentTime := now.Add(-1 * time.Minute).UTC().Format("Mon Jan 2 15:04:05 2006")
+	recentTimeISO := now.Add(-1 * time.Minute).UTC().Format("2006-01-02 15:04:05.000000Z")
+
+	// Replace the placeholders in the template
+	xmlContent := strings.ReplaceAll(templateXML, "{{RECENT_TIMESTAMP}}", recentTime)
+	xmlContent = strings.ReplaceAll(xmlContent, "{{RECENT_TIMESTAMP_ISO}}", recentTimeISO)
+
+	return xmlContent
+}
```
clobrano: Cool! :)

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,718 @@
...
+		Type:    conditionTypeDegraded,
+		Status:  operatorv1.ConditionTrue,
+		Reason:  reasonPacemakerUnhealthy,
+		Message: message,
+	}
+
+	return c.updateOperatorCondition(ctx, condition)
+}
+
+func (c *HealthCheck) clearDegradedCondition(ctx context.Context) error {
```
slintes: would it make sense to check the current degraded condition for its reason, to make sure we don't clear a degraded condition which wasn't set by this healthcheck? Not sure what best pratice is in such situations...

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,718 @@
...
+	// Maximum XML size to prevent XML bombs (10MB)
+	maxXMLSize = 10 * 1024 * 1024
+
+	// Time formats for parsing Pacemaker timestamps
+	// Go's reference time: Mon Jan 2 15:04:05 MST 2006 (mnemonic: 1 2 3 4 5 6 7)
+	pacemakerTimeFormat      = "Mon Jan 2 15:04:05 2006"     // Format for operation history timestamps
+	pacemakerFenceTimeFormat = "2006-01-02 15:04:05.000000Z" // Format for fence event timestamps
+
+	// Pacemaker commands
+	pcsStatusXMLCommand = "sudo -n pcs status xml"
```
slintes: Huh wait... did you already try running this in cluster?
IIRC one of the reasons for using jobs for pcs commands was about permissions... 🤔

jaypoulz: Ah interesting. I'll do some testing today and get back to you

jaypoulz: So it doesn't work in an interesting way.
```
[ec2-user@aws-jpoulin-dev ~]$ oc logs pod/etcd-operator-554d7f6899-5wrp2 | grep health.go
E1015 15:16:40.477459       1 health.go:120] health check for member (master-0) failed: err(context deadline exceeded)
E1015 15:16:45.703453       1 health.go:120] health check for member (master-0) failed: err(context deadline exceeded)
E1015 15:19:04.918498       1 health.go:120] health check for member (master-1) failed: err(context deadline exceeded)
```

Why does it stop after 3 tries I wonder :P
I'll try to update via chron with privileges and see if I can get something working

bindata/tnfdeployment/cronjob.yaml
```diff
@@ -0,0 +1,47 @@
...
+                  cpu: 10m
+                  memory: 32Mi
+              securityContext:
+                privileged: true
+                allowPrivilegeEscalation: true
+          hostIPC: false
+          hostNetwork: false
+          hostPID: true
+          priorityClassName: system-node-critical
+          serviceAccountName: etcd-sa
```
slintes: Would it make sense to use the tnf-setup-manager SA. like we do in the other TNF jobs? That one also has the privileged securitycontextconstraints rule assigned, which I expect the pacemaker cronjob needs as well

pkg/tnf/pkg/jobs/cronjob_controller.go
```diff
@@ -0,0 +1,249 @@
...
+			commandBuilder.WriteString(", ")
+		}
+		commandBuilder.WriteString(fmt.Sprintf(`"%s"`, cmd))
+	}
+	commandBuilder.WriteString("]")
+	commandStr := commandBuilder.String()
+
+	// Track which line we're on to do context-aware replacement
+	lines := strings.Split(result, "\n")
+	for i, line := range lines {
```
slintes: tbh, I like using a `JobHookFunc` like in `runJobController()` much more than this string replacements... WDYT?

cmd/cluster-etcd-operator/main.go
```diff
@@ -78,6 +79,7 @@ func NewSSCSCommand(ctx context.Context) *cobra.Command {
 	cmd.AddCommand(requestbackup.NewRequestBackupCommand(ctx))
 	cmd.AddCommand(rev.NewRevCommand(ctx))
 	cmd.AddCommand(backuprestore.NewBackupServer(ctx))
+	cmd.AddCommand(pacemakerstatuscollector.NewPacemakerStatusCollectorCommand())
```
slintes: what about adding this in cmd/tnf-setup-runner instead, where all the other TNF related commands live? (yes, it's not about the setup part, but we can rename it?)

pkg/cmd/pacemakerstatuscollector/collector.go
```diff
@@ -0,0 +1,414 @@
+package pacemakerstatuscollector
```
slintes: don't we want to have this in the tnf package?

pkg/tnf/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,781 @@
...
+	}
+
+	// Update operator status conditions based on pacemaker status
+	if err := c.updateOperatorStatus(ctx, healthStatus); err != nil {
+		return err
+	}
+
+	// Record pacemaker health check events
+	c.recordHealthCheckEvents(healthStatus)
+
```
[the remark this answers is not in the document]
jaypoulz: This is intentional. We only want to update the operator status if we have a definitive answer regarding the pacemaker status. Failure to collect an up-to-date status does means that we cannot definitively determine our current state, so the safest thing to do is to leave our last known status.

pkg/tnf/pacemaker/healthcheck.go
jaypoulz: I think the safest thing to do when something fails as part of this information collection is to log it, but refuse to change the operator state. We should provide updates for definitively known state, even if it may be an incomplete truth.

pkg/tnf/pacemaker/healthcheck.go
jaypoulz: Turns out that's what we are doing, but the function names are very misleading. Updated this to be more clear.

pkg/tnf/pacemaker/healthcheck.go
jaypoulz: Yeah, this does feel a bit wonky. I'll tweak it so that the window runs once-per-check instead of once per error recording

pkg/tnf/pkg/jobs/cronjob_controller.go
jaypoulz: I was confused as to how the JobHookFunc stuff worked :D
I was like - what is this sorcery

jaypoulz: Now that I am looking at it directly, it makes way more sense. Stealing this approach.

pkg/tnf/pacemaker/healthcheck.go
clobrano: yeah, technically we might have a number of errors and they all slip here because we fail parsing the timestamp :grin:, getting no state, but I understand your poing

bindata/tnfdeployment/cronjob.yaml
```diff
@@ -0,0 +1,48 @@
...
+              imagePullPolicy: IfNotPresent
+              terminationMessagePolicy: FallbackToLogsOnError
+              command: ["<injected>"]
+              resources:
+                requests:
+                  cpu: 10m
+                  memory: 32Mi
+              securityContext:
+                privileged: true
+                allowPrivilegeEscalation: true
```
clobrano, line 47: We might want to document why we need such privilegies, just for future references.

pkg/tnf/pkg/pacemaker/statuscollector.go
```diff
@@ -0,0 +1,559 @@
...
+					continue
+				}
+
+				// Parse RC
+				rc := int32(0)
+				if operation.RC != "" {
+					if _, err := fmt.Sscanf(operation.RC, "%d", &rc); err != nil {
+						klog.Warningf("Failed to parse RC value '%s': %v", operation.RC, err)
+					}
+				}
```
clobrano: Double check this indentation :)

pkg/tnf/pkg/pacemaker/statuscollector.go
```diff
@@ -0,0 +1,559 @@
...
+			}
+
+			result := restClient.Post().
+				Resource(pacemakerResourceName).
+				Body(newStatus).
+				Do(ctx)
+
+			if result.Error() != nil {
+				return fmt.Errorf("failed to create PacemakerStatus: %w", result.Error())
+			}
```
clobrano: Double check this indentation :)

pkg/tnf/pkg/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,691 @@
...
+}
+
+// recordWarningEvent records appropriate warning events based on warning type
+func (c *HealthCheck) recordWarningEvent(warning string) {
+	switch {
+	case strings.Contains(warning, warningPrefixFailedAction):
+		c.eventRecorder.Warningf(eventReasonFailedAction, msgDetectedFailedAction, warning)
+	case strings.Contains(warning, warningPrefixFencingEvent):
+		c.eventRecorder.Warningf(eventReasonFencingEvent, msgDetectedFencing, warning)
+	default:
```
clobrano, line 873: Nitpick: as the warning already contains the prefix ("Recent failed resource action:"), I think the event message will have it twice

clobrano, line 873: Nitpick: as the warning already contains the prefix ("Recent failed resource action:"), I think the event message will have it twice?

clobrano, line 873: Nitpick: as the warning already contains the prefix ("Recent failed resource action:"), I think the event message will have it twice?

clobrano (commented):
As you said it is still a WIP, I submitted only a few (not so important) notes, meanwhile I'll try to understand the workflow better

pkg/tnf/pkg/pacemaker/doc.go
```diff
@@ -0,0 +1,32 @@
...
+//
+//  1. Status Collector (statuscollector.go): Collects raw Pacemaker cluster status by executing
+//     "pcs status xml" and storing the results in a Pacemaker custom resource. The collector
+//     operates on a "collect everything" principle - it writes all data as-is without validation,
+//     allowing the health check to make informed decisions based on available data.
+//
+//  2. Health Check Controller (healthcheck.go): Monitors the Pacemaker CR and evaluates
+//     cluster health. It is hardened to handle missing or incomplete data gracefully, treating
+//     absent information as "unknown" rather than "error". The health check updates operator
+//     conditions and records events based on cluster state.
```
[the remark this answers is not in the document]
jaypoulz: I think you have this wrong. we use `sudo pcs status xml` - the fixtures show what that returns internally, but that's not what we call.

pkg/tnf/pkg/jobs/cronjob_controller.go
```diff
@@ -0,0 +1,335 @@
...
+		if container.TerminationMessagePath == "" {
+			container.TerminationMessagePath = "/dev/termination-log"
+		}
+		if container.TerminationMessagePolicy == "" {
+			container.TerminationMessagePolicy = "File"
+		}
+		if container.ImagePullPolicy == "" {
+			// Default depends on image tag, but IfNotPresent is common
+			container.ImagePullPolicy = "IfNotPresent"
+		}
```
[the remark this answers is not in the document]
jaypoulz: don't we specificy the image that's going to be used?

pkg/tnf/pkg/pacemaker/statuscollector.go
```diff
@@ -0,0 +1,852 @@
...
+	}
+
+	_, createErr := kubeClient.CoreV1().Events(TargetNamespace).Create(ctx, event, metav1.CreateOptions{})
+	if createErr != nil {
+		if apierrors.IsAlreadyExists(createErr) {
+			// Race: another collector created it; patch instead
+			klog.V(4).Infof("Event %s already exists, attempting patch", eventName)
+			now := metav1.Now()
+			patch := fmt.Sprintf(`{"count":2,"lastTimestamp":"%s"}`, now.Format(time.RFC3339))
+			_, _ = kubeClient.CoreV1().Events(TargetNamespace).Patch(ctx, eventName, types.StrategicMergePatchType, []byte(patch), metav1.PatchOptions{})
```
fonta-rh: Do we want to add a log in case the Patch fails? I think we're just ignoring the error here. Won't be frequent, but might be hard to debug if it happens

pkg/tnf/pkg/jobs/cronjob_controller.go
fonta-rh, line 1: I guess this is easier to ask for than to do, but this file has no testing. Maybe some unit tests, AI generated from just the signatures and doc?

pkg/tnf/pkg/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,908 @@
...
+		klog.Errorf("Failed to update operator status: %v", err)
+		return err
+	}
+
+	klog.V(2).Infof("Updated operator condition: %s=%s", condition.Type, condition.Status)
+	return nil
+}
+
+// cleanupExpiredEvents removes events from the deduplication map that have exceeded their window.
+// Fencing events are kept for 1 hour, other events for 5 minutes.
```
fonta-rh: comment update: Fencing events are currently kept for 24 hours, looking at the constants file

pkg/tnf/pkg/pacemaker/statuscollector.go
```diff
@@ -0,0 +1,852 @@
...
+	shortHash := hex.EncodeToString(hash[:])[:12]
+	return fmt.Sprintf("pacemaker-%s-%s", strings.ToLower(reason), shortHash)
+}
+
+// recordEventWithDeduplication patches existing events within the window, otherwise creates new ones.
+// Old events are left to age out via K8s garbage collection.
+func recordEventWithDeduplication(ctx context.Context, kubeClient kubernetes.Interface, reason, message, eventType string, deduplicationWindow time.Duration) {
+	eventName := generateEventName(reason, message)
+
+	existingEvent, err := kubeClient.CoreV1().Events(TargetNamespace).Get(ctx, eventName, metav1.GetOptions{})
```
fonta-rh, line 709: Do we want to log the err, if it exists?

pkg/tnf/pkg/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,908 @@
...
+//
+// Warning vs Error Philosophy:
+//   - Errors: Conditions requiring immediate action or causing downstream degradation.
+//     Examples: node offline, fencing unavailable, etcd/kubelet unhealthy, maintenance mode.
+//   - Warnings: Degraded redundancy where the cluster remains functional but should be investigated.
+//     Examples: FencingHealthy=False with FencingAvailable=True (can still fence, but redundancy lost).
+//
+// The operator degrades on Errors but not Warnings. Warnings generate events for observability.
+type HealthStatus struct {
+	OverallStatus string
```
fonta-rh: Claude suggests, and I tend to agree, that this should be a Typed Constant. We would need to change the constant strings above to a new Type:

```
  type HealthStatusValue string                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                              
  const (                                                                                                                                                                                                                                                                                                                     
      StatusHealthy HealthStatusValue = "Healthy"                                                                                                                                                                                                                                                                             
      StatusWarning HealthStatusValue = "Warning"                                                                                                                                                                                                                                                                             
      StatusError   HealthStatusValue = "Error"                                                                                                                                                                                                                                                                               
      StatusUnknown HealthStatusValue = "Unknown"                                                                                                                                                                                                                                                                             
  )   
```

HealthStatus would then update to this:

```
  type HealthStatus struct {                                                                                                                                                                                                                                                                                                  
      OverallStatus HealthStatusValue  // <-- Typed field                                                                                                                                                                                                                                                                     
      Warnings      []string                                                                                                                                                                                                                                                                                                  
      Errors        []string                                                                                                                                                                                                                                                                                                  
      CRLastUpdated time.Time                                                                                                                                                                                                                                                                                                 
  }
```

This removes the possibility of setting or reading wrong status strings anywhere in the code ("healthy", "Healhy", etc) and makes maintenance and readability better

pkg/tnf/pkg/pacemaker/statuscollector.go
```diff
@@ -0,0 +1,852 @@
...
+	KubeletResource *Resource
+	EtcdResource    *Resource
+}
+
+// FencingAgentInfo tracks a fencing agent parsed from pacemaker resources.
+// Fencing agents are named like "master-0_redfish" where the prefix is the target node.
+type FencingAgentInfo struct {
+	Name       string
+	TargetNode string
+	Method     string // "redfish", "ipmi", etc.
```
fonta-rh: I see that we do the conversion to the FencingMethod API later on, but why not use the FencingMethod type directly here? Is it about decoupling the OpenShift API with Pacemaker allowed methods?

If that's the case, I would improve the string to enum function on line 390 on like this, for maintainability:

```
  func methodStringToEnum(method string) v1alpha1.FencingMethod {                                                                                                                                                                                                                                                             
      switch strings.ToLower(method) {                                                                                                                                                                                                                                                                                        
      case "ipmi", "ipmilan", "ipmitool":                                                                                                                                                                                                                                                                                     
          return v1alpha1.FencingMethodIPMI                                                                                                                                                                                                                                                                                   
      case "redfish":                                                                                                                                                                                                                                                                                                         
          return v1alpha1.FencingMethodRedfish                                                                                                                                                                                                                                                                                
      default:                                                                                                                                                                                                                                                                                                                
          // Log unknown method so it's visible, but default to Redfish                                                                                                                                                                                                                                                       
          // since that's the primary TNF fencing method                                                                                                                                                                                                                                                                      
          klog.Warningf("Unknown fencing method %q, defaulting to Redfish", method)                                                                                                                                                                                                                                           
          return v1alpha1.FencingMethodRedfish                                                                                                                                                                                                                                                                                
      }                                                                                                                                                                                                                                                                                                                       
  }  
```

jaypoulz:
Fixed some issues related to how we detect job drift, but I was able to validate:

**Pacemaker Health Check Validation - Jan 28, 2026**
✅ Healthy Baseline
✅ Node Standby  
✅ Cluster Maintenance Mode
✅ Resource Location Constraint (kubelet-clone)
✅ Resource Monitor Failure (auto-recovery)
✅ Fencing Agent Degraded (bad credentials)
✅ CR Not Found Detection
✅ Stale Timestamp Detection (5-min threshold)

All scenarios passed. No new bugs found. I will update the Jira with my results before addressing another round of comments.

pkg/tnf/pkg/jobs/cronjob_controller.go
```diff
@@ -0,0 +1,516 @@
...
+		podSpec.SecurityContext = &corev1.PodSecurityContext{}
+	}
+	if podSpec.SchedulerName == "" {
+		podSpec.SchedulerName = "default-scheduler"
+	}
+	// The deprecated ServiceAccount field is auto-populated by the API server
+	// to match ServiceAccountName. Normalize it to prevent false drift detection.
+	if podSpec.DeprecatedServiceAccount == "" && podSpec.ServiceAccountName != "" {
+		podSpec.DeprecatedServiceAccount = podSpec.ServiceAccountName
+	}
```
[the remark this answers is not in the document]
jaypoulz: I've added a comment that explains that these start as nil and stay nil, so added these fields is pointless. But the documentation will provide clarity.

pkg/tnf/pkg/jobs/cronjob_controller.go
```diff
@@ -0,0 +1,516 @@
...
+		desiredField := desiredSpec.Field(i).Interface()
+
+		if !equality.Semantic.DeepEqual(existingField, desiredField) {
+			fieldDiffs = append(fieldDiffs, fieldName)
+			// Log the actual values for this field
+			existingJSON, _ := json.Marshal(existingField)
+			desiredJSON, _ := json.Marshal(desiredField)
+			klog.V(2).Infof("CronJob %s field %s differs:\n  existing: %s\n  desired:  %s",
+				c.name, fieldName, string(existingJSON), string(desiredJSON))
+		}
```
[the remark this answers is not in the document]
jaypoulz: we don't use env, args, envFrom, volumes, or volumeMounts. The only potential secret could be from commands, but we know what we populate commands with - it's just the binary path. There is no concern here.

bindata/etcd/pacemakercluster-crd.yaml
```diff
@@ -0,0 +1,632 @@
...
+                                dots, hyphens, and underscores
+                              rule: self.matches('^[a-zA-Z0-9._-]+$')
+                        required:
+                        - conditions
+                        - method
+                        - name
+                        type: object
+                      maxItems: 8
+                      minItems: 1
+                      type: array
```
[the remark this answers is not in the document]
jaypoulz: ### Current Design Rationale

1. **TNF clusters require fencing agents by definition**

   The `minItems: 1` constraint reflects a fundamental requirement: a Two Node Fencing cluster without fencing agents is misconfigured and non-functional. Empty arrays would mask a broken setup.

2. **Parsing/config errors are visible in collector logs**

   If the collector can't parse fencing agents from `pcs status xml`, it will:
   - Log the error with context (raw XML, parse failure reason)
   - Fail to update the status (CR rejected)
   - The CronJob itself shows failure in `oc get jobs`

3. **Stale status is detectable**

   The `lastUpdated` timestamp on the status will be stale, which operators/healthchecks can detect. A missing update is itself a signal.

4. **Synthetic results would mask real problems**

   Creating fake "unhealthy" fencing agents would:
   - Pollute the status with fabricated data
   - Make it harder to distinguish "no agents found" from "agents found but unhealthy"
   - Potentially trigger wrong remediation actions

### Future Extension (out of scope for current PR)

An "Unknown" status for subcomponents (or a top-level `collectionError` condition) could be a clean way to signal "we tried to collect but couldn't parse/find this data". This would be an API extension worth considering separately.

pkg/tnf/pkg/pacemaker/statuscollector.go
```diff
@@ -0,0 +1,880 @@
...
+			state := resourceState[xmlNode.Name]
+			if state == nil {
+				state = &ResourceStatePerNode{}
+			}
+			fencingAgents := fencingAgentsByTarget[xmlNode.Name]
+
+			placeholderAddresses := []v1alpha1.PacemakerNodeAddress{{Type: v1alpha1.PacemakerNodeInternalIP, Address: "0.0.0.0"}}
+			nodeStatus := buildNodeStatus(xmlNode.Name, placeholderAddresses, &xmlNode, state, fencingAgents, now)
+			nodes = append(nodes, nodeStatus)
+		}
```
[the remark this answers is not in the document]
jaypoulz: The fallback behavior using `0.0.0.0` is **intentional by design**:

```go
// Fallback when cluster config is unavailable - uses placeholder IPs that will fail validation
placeholderAddresses := []v1alpha1.PacemakerNodeAddress{{Type: v1alpha1.PacemakerNodeInternalIP, Address: "0.0.0.0"}}
```

### Same Rationale Applies

1. **Partial data without addresses is unreliable**

   If we can get node status from `pcs status xml` but can't get IPs from `pcs cluster config`, the pacemaker installation is in a degraded state worth investigating. The IP addresses are needed for correlation with Kubernetes node objects.

2. **The update rejection is a deliberate signal**

   The comment explicitly states "uses placeholder IPs that will fail validation". This was designed to prevent partial/incomplete status from being persisted. A status update that appears successful but contains placeholder data would be misleading.

3. **Logs reveal the root cause**

   The `klog.Warningf("Cluster config unavailable, falling back to XML nodes (IP addresses may be missing)")` message logs the actual problem, allowing operators to investigate why `pcs cluster config` failed.

4. **Alternative approaches have drawbacks**

   | Approach | Problem |
   |----------|---------|
   | Allow empty addresses | Status looks valid but is missing critical correlation data |
   | Emit empty node list | Discards valid resource/fencing status we did successfully collect |
   | Derive addresses elsewhere | Adds complexity and potential inconsistency between pacemaker and derived sources |

### Conclusion

The current behavior is intentional: when we can't get complete, accurate data, we prefer to fail loudly (rejected CR + warning logs) rather than persist incomplete status that appears healthy.

pkg/tnf/operator/starter.go
eggfoobar, line 207: I think it might be better to let the static resource controller here keep track of applying the CRD and then we can remove some of the CRD apply logic below, wdyt?

pkg/tnf/pkg/jobs/cronjob_controller.go
```diff
@@ -0,0 +1,516 @@
...
+// - If tag is :latest or omitted: Always
+// - If tag is anything else: IfNotPresent
+func inferImagePullPolicy(image string) corev1.PullPolicy {
+	// Handle empty image (shouldn't happen, but be defensive)
+	if image == "" {
+		return corev1.PullAlways
+	}
+
+	// Find the last colon that's not part of a port (after last /)
+	lastSlash := strings.LastIndex(image, "/")
```
eggfoobar: Might want to use the util in library-go for parsing this, should help avoid any unforseen cases, I saw a couple of helpers in the `library-go/pkg/image` pkg

https://github.com/openshift/library-go/blob/c3c3c4609280231f5ed970222921c8fa7a8f5b18/pkg/image/reference/reference.go#L32

pkg/tnf/pkg/jobs/cronjob_controller.go
```diff
@@ -0,0 +1,516 @@
...
+		return corev1.PullAlways
+	}
+
+	// Any other tag - default is IfNotPresent
+	return corev1.PullIfNotPresent
+}
+
+// logSpecDrift logs specific fields that have drifted between existing and desired CronJob specs.
+// NOTE: This function expects NORMALIZED specs (after normalizeCronJobSpec has been applied).
+func (c *CronJobController) logSpecDrift(existing, desired *batchv1.CronJob) {
```
eggfoobar, line 323: If you haven't already given it a shot, we might want to see if we can use the `cmp` library to help capture the drift, the go.mod already pulls it in `github.com/google/go-cmp/cmp`, something like `cmp.Diff(existing, desired)`

pkg/tnf/pkg/jobs/cronjob_controller.go
```diff
@@ -0,0 +1,516 @@
...
+		klog.Warningf("CronJob %s drift detected in untracked fields", c.name)
+		c.logUnknownDrift(existing, desired)
+	}
+}
+
+// logUnknownDrift uses reflection to identify exactly which fields differ between two CronJobs.
+// This is called when equality.Semantic.DeepEqual returns false but our explicit checks don't find differences.
+func (c *CronJobController) logUnknownDrift(existing, desired *batchv1.CronJob) {
+	// Compare CronJobSpec fields using reflection
+	existingSpec := reflect.ValueOf(existing.Spec)
```
eggfoobar: Same here for using `cmp.Diff`

pkg/tnf/pkg/jobs/cronjob_controller.go
```diff
@@ -0,0 +1,516 @@
...
+		klog.Infof("CronJob %s untracked drift in CronJobSpec fields: %v", c.name, fieldDiffs)
+	} else {
+		// The difference must be in metadata or status - check JobTemplate more deeply
+		c.logJobTemplateDrift(existing, desired)
+	}
+}
+
+// logJobTemplateDrift drills into JobTemplate to find differences
+func (c *CronJobController) logJobTemplateDrift(existing, desired *batchv1.CronJob) {
+	existingJT := existing.Spec.JobTemplate
```
eggfoobar: Here too I think

pkg/tnf/pkg/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,910 @@
...
+	klog.V(4).Infof("Retrieving pacemaker status from CR...")
+
+	// Read previous status
+	c.previousMu.Lock()
+	previous := c.previous
+	c.previousMu.Unlock()
+
+	// Get the PacemakerCluster CR
+	pacemakerCR := &v1alpha1.PacemakerCluster{}
+	err := c.pacemakerRESTClient.Get().
```
eggfoobar: Since we setup the informer, can we fetch this info from there?

pkg/tnf/pkg/pacemaker/healthcheck.go
```diff
@@ -0,0 +1,910 @@
...
+		c.eventRecorder.Warningf(EventReasonWarning, msgPacemakerWarning, warning)
+	}
+}
+
+// getEventReasonForError determines the appropriate event reason based on error content.
+// Used for both deduplication (key) and event recording.
+//
+// IMPORTANT: The order of checks matters. Cases are ordered from most specific to least specific
+// to avoid accidental matches. For example, we check for exact messages before checking substrings.
+func (c *HealthCheck) getEventReasonForError(errorMsg string) string {
```
fonta-rh: This should probably be made into an ordered list that we iterate over. Makes it much easier to read and extend:

`  var errorPatternsByPriority = []struct {                                                                                                                                                                                                                                                                                    
      pattern string                                                                                                                                                                                                                                                                                                          
      reason  string                                                                                                                                                                                                                                                                                                          
  }{                                                                                                                                                                                                                                                                                                                          
      {errorSubstringCRNotFound, EventReasonCRNotFound},                                                                                                                                                                                                                                                                      
      {errorSubstringStatusStale, EventReasonStatusStale},                                                                                                                                                                                                                                                                    
      // ... ordered list                                                                                                                                                                                                                                                                                                     
  }                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                              
  func getEventReasonForError(errorMsg string) string {                                                                                                                                                                                                                                                                       
      for _, entry := range errorPatternsByPriority {                                                                                                                                                                                                                                                                         
          if strings.Contains(errorMsg, entry.pattern) {                                                                                                                                                                                                                                                                      
              return entry.reason                                                                                                                                                                                                                                                                                             
          }                                                                                                                                                                                                                                                                                                                   
      }                                                                                                                                                                                                                                                                                                                       
      return EventReasonGenericError                                                                                                                                                                                                                                                                                          
  }   `

pkg/tnf/pkg/pacemaker/statuscollector.go
```diff
@@ -0,0 +1,880 @@
...
+		// Log unexpected errors (not NotFound) for debugging
+		klog.V(4).Infof("Failed to get existing event %s: %v (will create new)", eventName, err)
+	}
+	if err == nil && existingEvent != nil {
+		if time.Since(existingEvent.LastTimestamp.Time) < deduplicationWindow {
+			now := metav1.Now()
+			patch := fmt.Sprintf(`{"count":%d,"lastTimestamp":"%s"}`, existingEvent.Count+1, now.Format(time.RFC3339))
+			_, patchErr := kubeClient.CoreV1().Events(TargetNamespace).Patch(ctx, eventName, types.StrategicMergePatchType, []byte(patch), metav1.PatchOptions{})
+			if patchErr != nil {
+				klog.V(4).Infof("Failed to patch existing event %s: %v (will create new)", eventName, patchErr)
```
fonta-rh: I would say we probably want this error as a Warning, Verbose 4 might not be enabled by default. It's ok to have the successful patching below as V4

clobrano (approved):
/lgtm

pkg/tnf/operator/starter.go
jaypoulz: It doesn't simplify much, but I did rework this :) The issue is that we still need to check if it exists before the pacemakerInformer can start.

PR #1511, opened 2025-11-06
OCPBUGS-64729: Update etcd alerts to match observed real world data

dgoodwin:
Build on #1495, justification in this comment: https://issues.redhat.com/browse/OCPBUGS-64729?focusedId=28407511&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-28407511

### DIFF
--- a/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
+++ b/manifests/0000_90_etcd-operator_03_prometheusrule.yaml
@@ -106,7 +106,7 @@ spec:
       annotations:
         description: 'etcd cluster "{{ $labels.job }}": 99th percentile commit durations {{ $value }}s on etcd instance {{ $labels.instance }}.'
         summary: etcd cluster 99th percentile commit durations are too high.
-      expr: histogram_quantile(0.99, rate(etcd_disk_backend_commit_duration_seconds_bucket{job=~".*etcd.*"}[5m])) > 0.025
+      expr: histogram_quantile(0.99, rate(etcd_disk_backend_commit_duration_seconds_bucket{job=~".*etcd.*"}[5m])) > 0.08
       for: 10m
       labels:
         severity: warning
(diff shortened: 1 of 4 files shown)


manifests/0000_90_etcd-operator_03_prometheusrule.yaml
```diff
@@ -39,28 +39,6 @@ spec:
...
-      annotations:
-        description: 'etcd cluster "{{ $labels.job }}": 99th percentile fsync durations are {{ $value }}s on etcd instance {{ $labels.instance }}.'
-        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/etcdHighFsyncDurations.md
-        summary: etcd cluster 99th percentile fsync durations are too high.
-      expr: |
-        histogram_quantile(0.99, rate(etcd_disk_wal_fsync_duration_seconds_bucket{job=~".*etcd.*"}[5m]))
-        > 0.5
-      for: 10m
-      labels:
-        severity: warning
```
simonpasquier, line 52: IIUC we're removing the warning severity for `etcdHighFsyncDurations`. Do we have another rule which can notify platform admins before the critical alert first?

dgoodwin: It's a little tough for this, we don't actually know when a cluster falls over. Just that upstream recommendations are optimistic and we have thousands of clusters running much higher. I'm estimating what level of chaos we're willing to cause to lower these down to sensible levels again with the 5% alerting rate, I could trim some off the recommendations here and call that a warning threshold, but then we might be over our 5% fleet rate.

simonpasquier: While we don't provide stability guarantees for alerting rules, I presume that some cluster admins will be puzzled by the removal of the warning severity. As stated in https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md#warning-alerts warning alerts don't require immediate action but they help identifying potential issues. We could use a higher `for` value to avoid the alerting rule triggering too often.

cc @typeid

dgoodwin: Ok how about I use these limits currently in the pr for warning, and add a critical level a little higher

dgoodwin:
Actually saw a relevant failure on single node https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1511/pull-ci-openshift-cluster-etcd-operator-main-e2e-aws-ovn-single-node/1996107709992144896

/retest

PR #1520, opened 2025-12-04
OCPBUGS-66334: remove version check from guard precheck

tjungblu:
This was inadvertently deleting guard pods during upgrades, which caused etcd quorum loss while another component drained a node during a static pod rollout.


tjungblu:
Thanks @dusk125 - I'm going to leave this here until the critical fix label is lifted again

tjungblu:
/cherry-pick release-4.22 release-4.21 release-4.20 release-4.19 release-4.18

PR #1577, opened 2026-03-24
OCPBUGS-79354: Canonicalize all IP Address comparisons

barbacbd:
1. pkg/dnshelpers/util.go
      - Added CanonicalizeIP() function (lines 37-46): Converts IP addresses to canonical form using net.ParseIP().String()
        - Normalizes IPv6 addresses (e.g., 2001:0db8::1 → 2001:db8::1)
        - Auto-converts IPv4-mapped IPv6 to IPv4 (e.g., ::ffff:192.0.2.1 → 192.0.2.1)
        - Returns original string if invalid IP
    
    2. pkg/operator/ceohelpers/common.go
      - Updated 4 functions to use canonicalization:
        - MemberToNodeInternalIP() - canonicalizes IP extracted from etcd member
        - IndexMachinesByNodeInternalIP() - canonicalizes map keys
        - FindMachineByNodeInternalIP() - canonicalizes both sides of comparison
        - VotingMemberIPListSet() - canonicalizes IPs before inserting into set
        - MemberIPSetFromConfigMap() - canonicalizes IPs from configmap
        
  3. pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go

  - Updated hasInternalIP() function (line 528): Canonicalizes both machine and member IPs before comparison

  4. pkg/operator/machinedeletionhooks/machinedeletionhooks.go

  - Updated line 149: Canonicalizes machine address before set lookup

  5. pkg/operator/ceohelpers/common_test.go

  - Added comprehensive TestCanonicalizeIP() with 12 test cases covering:
    - IPv4 addresses
    - IPv6 compressed and full forms
    - IPv6 with leading zeros and mixed case
    - IPv4-mapped IPv6 conversion
    - Loopback addresses
    - Invalid IP handling

  Impact

  This fixes potential bugs where IPv6 addresses in different representations (e.g., 2001:db8::1 vs 2001:0db8:0000:0000:0000:0000:0000:0001) would fail to match, causing:
  - Machine lookups to fail
  - Member removal decisions to be incorrect
  - Deletion hooks to be misapplied

Note: This PR is opened with claude based on findings by the openshift-installer team. Review of claude's work and instructions given by @barbacbd.


pkg/dnshelpers/util.go
lance5890: maybe we should Canonicalize these address  too

barbacbd: Thanks @lance5890 that should be fixed now.

pkg/cmd/render/bootstrap_ip.go
```diff
@@ -139,8 +139,10 @@ func ContainedByCIDR(cidr string) AddressFilter {
 
 func AddressNotIn(ips ...string) AddressFilter {
```
tjungblu, line 165: do you mind to add a unit test here? I assume we don't have any covering this function

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -479,7 +479,7 @@ func (c *clusterMemberRemovalController) getNodeForMember(memberInternalIP strin
 		if err != nil {
 			return nil, fmt.Errorf("failed to get internal IP for node: %w", err)
 		}
-		if memberInternalIP == internalNodeIP {
+		if dnshelpers.CanonicalizeIP(memberInternalIP) == dnshelpers.CanonicalizeIP(internalNodeIP) {
```
tjungblu: internalNodeIP at this point should be canonical already, right?

pkg/tlshelpers/tlshelpers.go
```diff
@@ -60,6 +60,10 @@ func GetServingMetricsSecretNameForNode(nodeName string) string {
 }
 
 func getServerHostNames(nodeInternalIPs []string) []string {
+	canonicalIPs := make([]string, 0, len(nodeInternalIPs))
+	for _, ip := range nodeInternalIPs {
+		canonicalIPs = append(canonicalIPs, dnshelpers.CanonicalizeIP(ip))
+	}
```
tjungblu: this might get a bit tricky on upgrades. Imagine we already have an etcd cluster running with a not-canonical IP as its member address coming from the node internal IP. 

When we change the IPs here, we might create a new certificate that would only work with canonical IPs. I'm not sure how smart TLS is, maybe this is not going to be a problem, but it certainly doesn't hurt to add both canonical and "raw" representation here.  
The go code seems to kinda get this with the ::1 IP already, definitely worth adding a testcase here.

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
```diff
@@ -362,7 +362,7 @@ func (c *clusterMemberRemovalController) removeMemberWithoutMachine(ctx context.
 			if err != nil {
 				return err
 			}
-			if memberIP != potentialMemberToRemoveIP {
+			if dnshelpers.CanonicalizeIP(memberIP) != dnshelpers.CanonicalizeIP(potentialMemberToRemoveIP) {
```
tjungblu: also here, memberIP is canonical already.

It seems we also just work-around the fact that the node always returns the non-canonical IP address. Maybe we can have a custom node lister or extended struct that just directly canonicalizes the IP?

That way we don't have to litter these checks everywhere and I'm certain there are a few places we have missed here.

pkg/operator/ceohelpers/common.go
```diff
@@ -133,7 +133,7 @@ func IndexMachinesByNodeInternalIP(machines []*machinev1beta1.Machine) map[strin
 	for _, machine := range machines {
 		for _, addr := range machine.Status.Addresses {
 			if addr.Type == corev1.NodeInternalIP {
-				index[addr.Address] = machine
```
tjungblu, line 136: for upgrade reasons I mentioned earlier, it would make sense to keep both representations in the index for backward compatibility. We had bugs around this in the past where we wouldn't scale down a node because we couldn't find the exact IP matching.

pkg/tlshelpers/tlshelpers.go
barbacbd: After adjusting the function to include both forms of ip addresses (for backwards compatibility) I ran an analysis with Claude to see if I missed anything. Here are the results:

Key Findings:                                                                                                                                    
                                                                                                                                                   
  1. The code already handles both forms (lines 64-66 in tlshelpers.go):                                                                           
  completeIPList = append(completeIPList, dnshelpers.CanonicalizeIP(ip), ip)
  1. This adds both canonical (::1) and raw (0:0:0:0:0:0:0:1) forms for each node IP.                                                              
  2. The comment about "::1" is correct but was unclear: When IP addresses are added to TLS certificates, they're parsed to bytes. Both "::1" and  
  "0:0:0:0:0:0:0:1" parse to the same IP bytes, so TLS verification works with either form even if only one is listed. I've updated the comment to 
  explain this better.                                                                                                                             
  3. Tests prove it works: The new tests demonstrate that:                                                                                         
    - IPv6 addresses in different formats (compressed/expanded) are recognized as the same IP                                                      
    - A certificate with "::1" accepts connections to "0:0:0:0:0:0:0:1"                                                                            
    - The code includes both forms for node IPs for maximum compatibility

@tjungblu Please see the changes .

pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
barbacbd: If an existing etcd member was started with a non-canonical IP in its PeerURL (e.g., https://[0:0:0:0:0:0:0:1]:2380), then:                      
  - The etcd-endpoints configmap stores "0:0:0:0:0:0:0:1" (non-canonical)                                                                          
  - memberIP from the member list is "::1" (canonical)                                                                                             
  - Without canonicalizing both sides, they wouldn't match!
                                                                                                                                                   
  Recommendation:                                                                                                                                  
                                                                                                                                                   
  Keep the canonicalization on both sides. While canonicalizing memberIP is redundant, it's:                                                       
  1. Harmless - CanonicalizeIP(canonical) = canonical                                                                                              
  2. Defensive - Protects against future changes to MemberToNodeInternalIP                                                                         
  3. Clear intent - Makes it obvious we're doing canonical IP comparison  
  4. Consistent - Same pattern used in hasInternalIP() at line 529 

I was inclined to think of this as harmless too just to be explicit with it too. I don't know your opinions on it but this was the claude analysis.

pkg/cmd/render/bootstrap_ip_test.go
```diff
@@ -234,3 +234,111 @@ func withRoute(linkIndex int, dst string, protocol int) LinkRoutes {
...
+		},
+		{
+			name:        "multiple exclusions with mixed valid and invalid",
+			excludedIPs: []string{"invalid", "192.168.1.1", "also-invalid"},
+			testAddr:    "192.168.1.1",
+			expected:    false,
+		},
+		{
+			name:        "IPv4-mapped IPv6 address",
+			excludedIPs: []string{"::ffff:192.168.1.1"},
```
lance5890: should this be 192.0.2.1 ?  the excludedIPs is totally equal to testAddr

as u explained in the comment:
```
Converting IPv4-mapped IPv6 addresses to IPv4 (e.g., "::ffff:192.0.2.1" → "192.0.2.1")
```

pkg/operator/ceohelpers/common.go
```diff
@@ -131,9 +131,11 @@ func HasMachineDeletionHook(machine *machinev1beta1.Machine) bool {
 func IndexMachinesByNodeInternalIP(machines []*machinev1beta1.Machine) map[string]*machinev1beta1.Machine {
```
tjungblu, line 131: spoke too soon. There are some consumers of this that are iterating over this function to establish how many voting members exist:

```  
	var votingMembersMachines []*machinev1beta1.Machine
	for memberMachineIP, memberMachine := range ceohelpers.IndexMachinesByNodeInternalIP(memberMachines) {
		if currentVotingMemberIPListSet.Has(memberMachineIP) {
			votingMembersMachines = append(votingMembersMachines, memberMachine)
		}
	}
``` 

e.g. in clustermemberremovalcontroller.go. There are similar instances where this is checked against voting members and how many learners there are.

Others, for example the code in clustermembercontroller.go does work correctly that way:

``` 

	nodeInternalIP, err := ceohelpers.MemberToNodeInternalIP(member)

	nodeIPToMachinesWithDeletionHooks := ceohelpers.IndexMachinesByNodeInternalIP(machines)

	machine, hasMachine := nodeIPToMachinesWithDeletionHooks[nodeInternalIP]
``` 

So I think we have to revert this again. 
Let me quick get back to the team today, I think there are some cases that already exists that we should fix beforehand. Sorry for the back and forth :/

tjungblu: checkout https://github.com/openshift/cluster-etcd-operator/pull/1588

pkg/cmd/render/bootstrap_ip_normalize_test.go
```diff
@@ -0,0 +1,88 @@
+package render
```
tjungblu: feel free to move that test into bootstrap_ip_test.go

pkg/etcdenvvar/envvarcontroller_test.go
```diff
@@ -66,7 +70,15 @@ var (
...
-)
+
+	// Add ETCD_UNSUPPORTED_ARCH for arm64 and s390x (matches getUnsupportedArch behavior)
+	envVars, _ := getUnsupportedArch(envVarContext{})
+	for k, v := range envVars {
+		result[k] = v
+	}
+
+	return result
+}
```
tjungblu: that seems odd? where is that coming from?

pkg/cmd/render/bootstrap_ip.go
```diff
@@ -137,9 +137,71 @@ func ContainedByCIDR(cidr string) AddressFilter {
...
+	if octet == "" || dotCount != 3 {
+		return ip // Invalid format
+	}
+	normalized := octet
+	for len(normalized) > 1 && normalized[0] == '0' {
+		normalized = normalized[1:]
+	}
+	parts = append(parts, normalized)
+
+	result := fmt.Sprintf("%s.%s.%s.%s", parts[0], parts[1], parts[2], parts[3])
```
dusk125, line 184: nit: strings.Join

pkg/cmd/render/bootstrap_ip.go
```diff
@@ -137,9 +137,71 @@ func ContainedByCIDR(cidr string) AddressFilter {
...
+
+	// Add the last octet
+	if octet == "" || dotCount != 3 {
+		return ip // Invalid format
+	}
+	normalized := octet
+	for len(normalized) > 1 && normalized[0] == '0' {
+		normalized = normalized[1:]
+	}
+	parts = append(parts, normalized)
```
dusk125, line 182: Could we make this into a helper, since it's just duplicated from the above?

dusk125 (changes requested):

pkg/cmd/render/bootstrap_ip.go
```diff
@@ -137,9 +137,71 @@ func ContainedByCIDR(cidr string) AddressFilter {
...
+	if parsed := net.ParseIP(ip); parsed != nil {
+		return ip
+	}
+
+	// Check if it looks like an IPv4 with potential leading zeros
+	parts := make([]string, 0, 4)
+	octet := ""
+	dotCount := 0
+
+	for _, ch := range ip {
```
dusk125, line 154: You could get some of the same parsing without having to do the tail capture with strings.SplitSeq

pkg/cmd/render/bootstrap_ip.go
dusk125: See below comment about strings.SplitSeq

pkg/cmd/render/bootstrap_ip.go
```diff
@@ -137,9 +137,71 @@ func ContainedByCIDR(cidr string) AddressFilter {
...
+			}
+			// Remove leading zeros from octet
+			normalized := octet
+			for len(normalized) > 1 && normalized[0] == '0' {
+				normalized = normalized[1:]
+			}
+			parts = append(parts, normalized)
+			octet = ""
+			dotCount++
+		} else if ch >= '0' && ch <= '9' {
```
dusk125, line 167: This is not correct for a standard IPv4. This would allow 999.999.999.999 through.

In addition to the SplitSeq suggestion, I would suggest taking each octet string from the Split and passing it through strconv.ParseUint to get out the valid or invalid byte (this also does range testing to ensure there aren't invalid numbers for the bitsize), then using strconv.Itoa to convert to string. Also, having this function return an error would allow you to remove an additional ParseIP call on the happy path.

dusk125 (changes requested):

PR #1600, opened 2026-04-27
OCPEDGE-2116: feat: support MAC-address based fencing credentials lookup

fracappa:
When the installer creates fencing credentials using MAC addresses instead of hostnames, the secrets are named with a SHA256 hash of the normalized MAC (e.g. fencing-credentials-{hash}).

This PR adds a multi-phase fencing secret resolution to GetFencingSecrets:

1. **Hostname**: try `fencing-credentials-{nodeName}` directly
2. **MAC hash**: read MAC addresses from node annotation  (`tnf.openshift.io/mac-addresses`), hash each, try `fencing-credentials-{hash}` by name
3. **Redfish UUID**: query each unclaimed fencing secret's Redfish  endpoint for the system UUID, match against `node.status.nodeInfo.systemUUID`

The auth job (per-node) discovers all non-loopback MAC addresses via nsenter and annotates the node, so they are available when the setup/fencing/update-setup jobs resolve secrets.

Also removes unused `machines` and `baremetalhosts` RBAC rules from the TNF clusterrole and adds `patch` on nodes for the annotation.

## Summary by CodeRabbit

* **New Features**
  * Improved cluster fencing: multi-node credential resolution and host-to-node matching for more reliable STONITH configuration.

* **Chores**
  * Updated cluster RBAC to allow retrieval/listing of machine and baremetalhost resources to support fencing workflows.

* **Tests**
  * Added unit tests covering multi-node credential resolution and host-to-node matching.

### DIFF
--- a/bindata/tnfdeployment/clusterrole.yaml
+++ b/bindata/tnfdeployment/clusterrole.yaml
@@ -13,6 +13,7 @@ rules:
       - get
       - list
       - watch
+      - patch
   - apiGroups:
       - operator.openshift.io
     resources:
@@ -64,3 +65,4 @@ rules:
       - securitycontextconstraints
     verbs:
       - use
+
--- a/pkg/tnf/auth/runner.go
+++ b/pkg/tnf/auth/runner.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"context"
+	"os"
 
 	configversionedclient "github.com/openshift/client-go/config/clientset/versioned"
 	"k8s.io/apiserver/pkg/server"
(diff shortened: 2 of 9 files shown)


fonta-rh:
**Test suggestion: cover the `IsNotFound` vs API-error branch in Phase 1 and Phase 2**

`GetFencingSecrets` distinguishes `errors.IsNotFound` (fall through to next phase) from other API errors (fail immediately) in both Phase 1 (`secrets.go:651`) and Phase 2's `matchMACHashToSecrets` (`secrets.go:743`). This routing logic is load-bearing but currently untested — the fake client returns `NotFound` for missing objects by default, so only the happy fallthrough path is exercised.

A `PrependReactor` injecting a non-404 error (e.g., 403 Forbidden) on secret `Get` would cover the hard-failure branch in each phase. One test case per phase, asserting `GetFencingSecrets` returns the API error immediately rather than falling through.

Not blocking — the pattern is standard k8s and unlikely to regress accidentally. But it documents the contract and catches future refactors that might accidentally change the error routing.

PR #1611, opened 2026-05-04
OCPEDGE-2604, OCPEDGE-2605: feat: add update-fencing-credentials.sh script for TNF fencing credentials rotation

fracappa:
Introduces a bash script to update fencing credentials (BMC username, password, address) in both pacemaker stonith devices and the corresponding Kubernetes secret. Registered in ScriptController for TNF-only deployment.

## Summary by CodeRabbit

* **New Features**
  * Added an automated fencing credential update utility for supported etcd topologies to update Pacemaker stonith settings, upsert fencing credentials, verify access, and check Pacemaker health.

* **Tests**
  * Expanded tests to validate topology-dependent inclusion of the fencing credential utility in generated script bundles.

* **Documentation**
  * Added user-facing guidance on running the utility, parameters, usage methods, and security considerations.

### DIFF
--- a/bindata/etcd/update-fencing-credentials.sh
+++ b/bindata/etcd/update-fencing-credentials.sh
@@ -1,0 +1,177 @@
+#!/usr/bin/env bash
+
+### Created by cluster-etcd-operator. DO NOT edit.
+
+set -o errexit
+set -o pipefail
+set -o errtrace
+
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root"
+    exit 1
+fi
+
+function source_required_dependency {
+    local src_path="$1"
+    if [ ! -f "${src_path}" ]; then
+        echo "required dependencies not found, please ensure this script is run on a node with a functional etcd static pod"
+        exit 1
+    fi
+    # shellcheck disable=SC1090
+    source "${src_path}"
+}
+
+source_required_dependency /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd.env
+source_required_dependency /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd-common-tools
+
+function usage {
+    echo "Usage: update-fencing-credentials.sh --node <name> --username <user> --password <password> --address <redfish-url> [--ssl-insecure]"
+    exit 1
+}
+
+function redact_credentials {
+    local input
+    input=$(cat)
+    input="${input//${USERNAME}/****}"
+    input="${input//${PASSWORD}/****}"
+    printf '%s\n' "${input}"
+}
+
+NODE=""
+USERNAME=""
+PASSWORD=""
+ADDRESS=""
+SSL_INSECURE=false
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --node)     NODE="$2";      shift 2 ;;
+        --username) USERNAME="$2";  shift 2;;
+        --password) PASSWORD="$2";  shift 2;;
+        --address)  ADDRESS="$2";   shift 2;;
+        --ssl-insecure) SSL_INSECURE=true;  shift ;;
+        *) usage ;;
+    esac
+done
+
+if [ -z "${NODE}" ] || [ -z "${USERNAME}" ] || [ -z "${PASSWORD}" ] || [ -z "${ADDRESS}" ]; then
+    usage
+fi
+
+# mirrors getFencingConfig in pkg/tnf/pkg/pcs/fencing.go
+if [[ "${ADDRESS}" != *redfish* ]]; then
+    echo "Address does not contain a valid redfish URL: ${ADDRESS}"
+    exit 1
+fi
+
+# strip the redfish+ prefix scheme (e.g. redfish+https://... -> https://..)
+PARSED_URL="${ADDRESS#redfish+}"
+
+# Parse IPv6 (bracketed) vs IPv4/hostname
+if [[ "${PARSED_URL}" =~ ^https?://\[([^]]+)\](:([0-9]+))?(/.*)? ]]; then
+    REDFISH_HOST="${BASH_REMATCH[1]}"
+    REDFISH_PORT="${BASH_REMATCH[3]}"
+    REDFISH_PATH="${BASH_REMATCH[4]}"
+    REDFISH_IP="[${REDFISH_HOST}]"
+else
+    REDFISH_HOST=$(echo "${PARSED_URL}" | sed -E 's|https?://([^/:]+).*|\1|')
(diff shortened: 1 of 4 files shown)


bindata/etcd/update-fencing-credentials.sh
```diff
@@ -0,0 +1,160 @@
...
+    ipport="${REDFISH_PORT}" \
+    systems_uri="${REDFISH_PATH}" \
+    ssl_insecure="${SSL_INSECURE_VAL}" \
+    --wait=120 2>&1 > /dev/null) || true
+
+# stderr must contain "is running"
+if [[ "${STDERR}" != *"is running"* ]]; then
+    echo "Failed to update stonith device ${DEVICE_ID}: ${STDERR}"
+    exit 1
+fi
```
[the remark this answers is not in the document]
fracappa: ` 2>&1 > /dev/null `does not discard both streams:

- `2>&1` — stderr goes to the capture pipe (where stdout currently points)
- `> /dev/null `— stdout then goes to /dev/null

So STDERR correctly captures only stderr, and stdout is discarded. The work proved it works on dev-scripts environment.

fonta-rh (commented):
Three findings from review — all actionable, noise filtered out.

Summary:
1. **`redact_credentials` can leak passwords** when credentials contain sed metacharacters (`/`, `&`, `\`)
2. **`oc apply` failure needs explicit error handling** — STONITH is already updated at that point and the user needs to know
3. **Documentation step ordering is wrong** — pre-flight validation is listed as step 5 (post-mutation) but happens at step 3 (pre-mutation)

bindata/etcd/update-fencing-credentials.sh
```diff
@@ -0,0 +1,167 @@
...
+
+function usage {
+    echo "Usage: update-fencing-credentials.sh --node <name> --username <user> --password <password> --address <redfish-url> [--ssl-insecure]"
+    exit 1
+}
+
+function redact_credentials {
+    sed -e "s/${USERNAME}/****/g" -e "s/${PASSWORD}/****/g"
+}
+
```
fonta-rh, line 39: `USERNAME` and `PASSWORD` are interpolated directly into a `sed` substitution with `/` as the delimiter. BMC passwords commonly contain `/`, `&`, or `\`.

**Concrete failure scenario:**

- On line 108 (`echo ... | redact_credentials`): `pipefail` + `errexit` aborts the script with an unhelpful sed error, swallowing the actual pre-flight error context.
- On line 119 (`$(echo "${STDERR}" | redact_credentials)` inside string interpolation): `errexit` does not apply inside `$()` in bash — sed fails silently and **credentials are printed unredacted** to the terminal.

**Suggested fix** — use bash parameter expansion, which handles all special characters:

```bash
function redact_credentials {
    local input
    input=$(cat)
    input="${input//${USERNAME}/****}"
    input="${input//${PASSWORD}/****}"
    printf '%s\n' "${input}"
}
```

bindata/etcd/update-fencing-credentials.sh
```diff
@@ -0,0 +1,167 @@
...
+    --wait=120 2>&1 > /dev/null) || true
+
+# stderr must contain "is running"
+if [[ "${STDERR}" != *"is running"* ]]; then
+    echo "Failed to update stonith device ${DEVICE_ID}: $(echo "${STDERR}" | redact_credentials)"
+    exit 1
+fi
+echo "Stonith device ${DEVICE_ID} updated successfully"
+
+SECRET_NAME="fencing-credentials-${NODE}"
```
fonta-rh, line 135: At this point, the STONITH device has already been updated with new credentials (line 113). If this `oc apply` pipeline fails (API server unreachable, RBAC, quota), `errexit` aborts the script with a raw `oc` error and no context about the now-inconsistent state.

The user doesn't know that:
- STONITH has the **new** credentials
- The secret still has the **old** credentials
- The CEO operator may overwrite STONITH back to old creds on next reconciliation

Keeping the current order (STONITH first) is correct — credential rotation must work even without API access. But the failure path needs explicit handling:

```bash
if ! oc create secret generic "${SECRET_NAME}" \
    --namespace="${NAMESPACE}" \
    --from-literal=address="${ADDRESS}" \
    --from-literal=username="${USERNAME}" \
    --from-literal=password="${PASSWORD}" \
    --from-literal=certificateVerification="${CERT_VERIFICATION}" \
    --dry-run=client -o yaml | oc apply -f -; then
    echo "WARNING: Stonith device ${DEVICE_ID} was updated successfully,"
    echo "but the Kubernetes secret could not be updated."
    echo "The CEO operator may revert the stonith device on next reconciliation."
    echo "When API access is restored, re-run this script to update the secret."
    exit 1
fi
```

docs/update-fencing-credentials.md
```diff
@@ -0,0 +1,103 @@
...
+## What the script does
+
+1. **Validates prerequisites** -- checks root privileges, required dependency files (`etcd.env`, `etcd-common-tools`), and all required
+arguments.
+
+2. **Parses the Redfish URL** -- extracts host, port, and API path from the `--address` value; strips the `redfish+` scheme prefix.
+
+3. **Updates the Pacemaker stonith device** -- runs `pcs stonith update <node>_redfish` with the new credentials (username, password, ip, ipport,
+systems_uri, ssl_insecure) and waits up to 120 seconds for the update to apply.
+
```
fonta-rh, line 70: The step ordering here doesn't match the script. The script does pre-flight validation (runs `fence_redfish` with the new credentials) **before** any mutations — between steps 2 and 3 as documented here. The doc lists it as step 5, after STONITH and secret updates.

This matters because the current doc implies a dangerous design (mutate then verify), when the script actually has a safe design (verify then mutate).

**Suggested fix** — reorder to match the script:

1. Validates prerequisites
2. Parses the Redfish URL
3. **Pre-flight validation** — runs `fence_redfish` with the new credentials to confirm they work before making any changes
4. Updates the Pacemaker stonith device
5. Creates or updates the Kubernetes secret
6. Checks Pacemaker health

Also change "updated credentials" to "new credentials" — nothing has been updated at the point where validation runs.

fracappa:
@fonta-rh thanks for the feedbacks!

All your comments should be addressed now! :smile:

fonta-rh (approved):
All three comments addressed — LGTM.

- `redact_credentials` now uses bash parameter expansion instead of sed — safe for metacharacters
- `oc apply` failure path has explicit state warning and remediation hint
- Documentation step ordering matches the script's actual execution flow

/lgtm

dusk125:
I understand that the CEO helps manage the TNF transition, but I'm concerned with non-etcd related items making their way into this operator because of that reason. I worry that this will balloon because "the CEO is where X happens and it's tightly coupled with Y", which leads to "the CEO is where Y happens and it's tightly coupled to Z, etc."

I would prefer this live in a different project

PR #1614, opened 2026-05-08
OCPBUGS-81500: fix: udpate healthcheck previous state only after successful operator status update

fracappa:
Previously, getPacemakerStatus updated c.previous before updateOperatorStatus was called. If the status update failed, subsequent syncs would skip processing because c.previous already reflected the new CR timestamp, leaving PacemakerHealthCheckDegraded stuck a True indefinitely.

## Bug Fixes
* Improved Pacemaker health check status tracking for more accurate grace-period timing and state transitions.
* Prevented unknown status values from replacing meaningful health information.
* Improved consistency between reported health status and recorded health events.
* Preserved status timestamps when health conditions remain unchanged.
* Improved persistence of health information between consecutive checks.

### DIFF
--- a/pkg/tnf/pkg/pacemaker/healthcheck.go
+++ b/pkg/tnf/pkg/pacemaker/healthcheck.go
@@ -245,6 +245,13 @@ func (c *HealthCheck) sync(ctx context.Context, syncCtx factory.SyncContext) err
 
 	c.recordHealthCheckEvents(currentStatus, previousStatus)
 
+	// Only mark as processed after status was successfully updated
+	if currentStatus.OverallStatus != statusUnknown {
+		c.previousMu.Lock()
+		c.previous = currentStatus
+		c.previousMu.Unlock()
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


fonta-rh (approved):
LGTM — the fix correctly moves `c.previous` assignment from `getPacemakerStatus()` to after `updateOperatorStatus()` succeeds in `sync()`, so a failed status write no longer poisons the skip-optimization cache.

One minor nit: the doc comment on `getPacemakerStatus()` (line 254) still says "For Unknown status, previous is not updated" — this referred to `c.previous` being written inside the function, which no longer happens for any status. Consider removing that line or replacing with something like:

```go
// Note: c.previous is updated by sync() after successful status write, not here.
```

Non-blocking, cosmetic only.

pkg/tnf/pkg/pacemaker/healthcheck.go
```diff
@@ -265,6 +265,13 @@ func (c *HealthCheck) sync(ctx context.Context, syncCtx factory.SyncContext) err
 
 	c.recordHealthCheckEvents(currentStatus, previousStatus)
 
+	// Only mark as processed after status was successfully updated
+	if currentStatus.OverallStatus != statusUnknown {
+		c.previousMu.Lock()
+		c.previous = currentStatus
+		c.previousMu.Unlock()
+	}
+
```
fracappa: @coderabbitai this feedback might be outdated

PR #1618, opened 2026-05-13
CNTRLPLANE-3460: Refactor defrag controller to lower impact of defrag

dusk125:
## Summary
- Refactors the defrag controller to defrag one member per sync cycle instead of all members at once
- Members are sorted by fragmentation descending; the most fragmented member is defragged first
- The health check at the start of each sync ensures the previously defragged member recovered before the next one is defragged
- Removes the blocking `wait.Poll` recovery loop and leader-last ordering — with upstream etcd defrag improvements ([openshift/etcd#378](https://github.com/openshift/etcd/pull/378)) reducing disruption from ~30s to ~100ms, the simpler one-per-sync approach is sufficient

## Test plan
- [x] Existing unit tests updated for multi-sync behavior and pass
- [x] `TestNewDefragControllerMultiSyncs` validates degraded/recovery across sync cycles
- [ ] Manual verification in a test cluster with fragmented etcd members

## Summary by CodeRabbit

* **Improvements**
  * Defragmentation now targets at most one fragmented member per sync cycle, prioritized by highest backend fragmentation.
  * Member selection skips learner and unstarted members; when no fragmentation is found, degraded-state tracking is cleared and failure counters reset.
  * Defragmentation now fails fast on the first encountered defrag error, updating degraded-state progress immediately.
* **Tests**
  * Updated controller tests to support repeated sync loops and adjusted expected success/error behavior.
  * Enhanced the fake client defragmentation behavior used by tests to better reflect status changes.

### DIFF
--- a/pkg/operator/defragcontroller/defragcontroller.go
+++ b/pkg/operator/defragcontroller/defragcontroller.go
@@ -1,9 +1,12 @@
 package defragcontroller
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"math"
+	"slices"
+	"strings"
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -16,7 +19,6 @@ import (
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/wait"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 
@@ -27,12 +29,13 @@ import (
 
 const (
 	minDefragBytes                 int64   = 100 * 1024 * 1024 // 100MB
-	minDefragWaitDuration                  = 36 * time.Second
 	maxFragmentedPercentage        float64 = 45
-	pollWaitDuration                       = 2 * time.Second
-	pollTimeoutDuration                    = 60 * time.Second
 	compactionInterval                     = 10 * time.Minute
 	maxDefragFailuresBeforeDegrade         = 3
+
+	// defaultDefragSettleTime is the minimum time to wait after a leader transfer
+	// to allow the cluster to stabilize before proceeding with defrag.
+	defaultDefragSettleTime = 10 * time.Second
 
 	defragDisabledCondition    = "DefragControllerDisabled"
 	defragDisableConfigmapName = "etcd-disable-defrag"
@@ -47,11 +50,17 @@ type DefragController struct {
 	memberLister         etcdcli.AllMemberLister
 	defragClient         etcdcli.Defragment
 	statusClient         etcdcli.Status
+	leaderMover          etcdcli.LeaderMover
 	infrastructureLister configv1listers.InfrastructureLister
 	configmapLister      corev1listers.ConfigMapLister
 
-	numDefragFailures  int
-	defragWaitDuration time.Duration
+	// defragSettleTime is the minimum time to wait after a leader transfer or defrag
+	// to allow the cluster to stabilize before proceeding.
+	defragSettleTime time.Duration
+
+	numDefragFailures int
+	// defragTargets tracks the ids of members that need to be defragged during the current cycle
+	defragTargets []uint64
 }
 
 func NewDefragController(
@@ -60,6 +69,7 @@ func NewDefragController(
 	memberLister etcdcli.AllMemberLister,
 	defragClient etcdcli.Defragment,
 	statusClient etcdcli.Status,
+	leaderMover etcdcli.LeaderMover,
 	infrastructureLister configv1listers.InfrastructureLister,
 	eventRecorder events.Recorder,
 	kubeInformers v1helpers.KubeInformersForNamespaces) factory.Controller {
@@ -68,14 +78,15 @@ func NewDefragController(
 		memberLister:         memberLister,
 		defragClient:         defragClient,
 		statusClient:         statusClient,
+		leaderMover:          leaderMover,
 		infrastructureLister: infrastructureLister,
 		configmapLister:      kubeInformers.ConfigMapLister(),
-		defragWaitDuration:   minDefragWaitDuration,
+		defragSettleTime:     defaultDefragSettleTime,
 	}
 	syncer := health.NewCheckingSyncWrapper(c.sync, 3*compactionInterval+1*time.Minute)
 	livenessChecker.Add("DefragController", syncer)
 
-	return factory.New().ResyncEvery(compactionInterval+1*time.Minute).WithInformers( // attempt to sync outside of etcd compaction interval to ensure maximum gain by defragmentation.
+	return factory.New().ResyncEvery(compactionInterval+1*time.Minute).WithBareInformers( // attempt to sync outside of etcd compaction interval to ensure maximum gain by defragmentation.
 		operatorClient.Informer(),
 	).WithSync(syncer.Sync).ToController("DefragController", eventRecorder.WithComponentSuffix("defrag-controller"))
 }
@@ -90,7 +101,7 @@ func (c *DefragController) sync(ctx context.Context, syncCtx factory.SyncContext
 		return nil
 	}
 
-	return c.runDefrag(ctx, syncCtx.Recorder())
+	return c.runDefrag(ctx, syncCtx)
 }
 
 func (c *DefragController) checkDefragEnabled(ctx context.Context, recorder events.Recorder) (bool, error) {
@@ -124,7 +135,17 @@ func (c *DefragController) checkDefragEnabled(ctx context.Context, recorder even
 	return true, nil
 }
 
-func (c *DefragController) runDefrag(ctx context.Context, recorder events.Recorder) error {
+type StatusMember struct {
+	Status *clientv3.StatusResponse
+	Member *etcdserverpb.Member
+}
+
+func (sm StatusMember) IsLeader() bool {
+	return sm.Status.Leader == sm.Member.ID
+}
+
+func (c *DefragController) runDefrag(ctx context.Context, syncCtx factory.SyncContext) error {
+	recorder := syncCtx.Recorder()
 	// Do not defrag if any of the cluster members are unhealthy.
 	memberHealth, err := c.memberLister.MemberHealth(ctx)
 	if err != nil {
@@ -139,120 +160,138 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
 		return err
 	}
 
-	// filter out learner members since they don't support the defragment API call
-	var etcdMembers []*etcdserverpb.Member
-	for _, m := range members {
-		if !m.IsLearner {
-			etcdMembers = append(etcdMembers, m)
-		}
-	}
-
-	var endpointStatus []*clientv3.StatusResponse
-	var leader *clientv3.StatusResponse
-	for _, member := range etcdMembers {
-		if len(member.ClientURLs) == 0 {
-			// skip unstarted member
+	var (
+		isNewCycle    = len(c.defragTargets) == 0
+		statusMembers = make(map[uint64]StatusMember)
+	)
+	for _, member := range members {
+		// filter out learner members since they don't support the defragment API call
+		// and filter out unstarted members
+		if member.IsLearner || len(member.ClientURLs) == 0 {
 			continue
 		}
+
 		status, err := c.statusClient.Status(ctx, member.ClientURLs[0])
 		if err != nil {
 			return err
+		} else if status == nil {
+			return fmt.Errorf("endpoint status returned nil for member %q (%s)", member.Name, member.ClientURLs[0])
 		}
-		if leader == nil && status.Leader == member.ID {
-			leader = status
-			continue
+
+		sm := StatusMember{
+			Status: status,
+			Member: member,
 		}
-		endpointStatus = append(endpointStatus, status)
+
+		statusMembers[member.ID] = sm
+
+		if isNewCycle && isEndpointBackendFragmented(member, status) {
+			c.defragTargets = append(c.defragTargets, member.ID)
+		}
 	}
 
-	// Leader last if possible.
-	if leader != nil {
-		klog.V(4).Infof("Appending leader last, ID: %x", leader.Header.MemberId)
-		endpointStatus = append(endpointStatus, leader)
+	if !isNewCycle {
+		c.defragTargets = slices.DeleteFunc(c.defragTargets, func(targetID uint64) bool {
+			statusMember, has := statusMembers[targetID]
+			if !has {
+				// Remove any defrag targets that we don't have a status for.
+				return true
+			}
+
+			// Remove any members that no longer meet the conditions for defrag.
+			return !isEndpointBackendFragmented(statusMember.Member, statusMember.Status)
+		})
 	}
 
-	successfulDefrags := 0
-	var errors []error
-	for _, status := range endpointStatus {
-		member, err := getMemberFromStatus(etcdMembers, status)
-		if err != nil {
-			errors = append(errors, err)
-			continue
+	if len(c.defragTargets) == 0 {
+		targets := make([]string, len(c.defragTargets))
+		for i, targetID := range c.defragTargets {
+			target := statusMembers[targetID]
+			percent := checkFragmentationPercentage(target.Status.DbSize, target.Status.DbSizeInUse)
+			targets[i] = fmt.Sprintf("%s={id: %d, fragmentation: %.2f%%, sizeInUse: %d}", target.Member.Name, target.Member.ID, percent, target.Status.DbSize)
+		}
+		klog.V(4).Infof("Defrag skipped: no etcd members meet the conditions for defragmentation:\n%s", strings.Join(targets, ", "))
+		return nil
+	}
+
+	// Sort fragmented members so we defragment the most fragmented member first while defragging the leader last
+	slices.SortFunc(c.defragTargets, func(a, b uint64) int {
+		aIsLeader, bIsLeader := statusMembers[a].IsLeader(), statusMembers[b].IsLeader()
+		if aIsLeader && !bIsLeader {
+			return 1
+		} else if !aIsLeader && bIsLeader {
+			return -1
+		}
+		return sortByMostFragmented(statusMembers[a], statusMembers[b])
+	})
+
+	defragTarget := statusMembers[c.defragTargets[0]]
+	defragTargetStatus, defragTargetMember := defragTarget.Status, defragTarget.Member
+
+	// Preemptively attempt to move the leadership away from the current defrag target to another valid follower.
+	// We try this to avoid multiple leader elections in the case where defragging the leader causes leadership
+	// to move to a member we've yet to defrag, which could in turn lose leadership, etc. causing a lot of churn.
+	// We record any error that occurs while attempting this, but we do not halt defrag if the move fails.
+	if defragTarget.IsLeader() && len(statusMembers) > 1 {
+		followers := make([]StatusMember, 0, len(statusMembers))
+		for id, member := range statusMembers {
+			if defragTargetMember.ID == id {
+				continue
+			}
+			followers = append(followers, member)
 		}
 
-		// Check each member's status which includes the db size on disk "DbSize" and the db size in use "DbSizeInUse"
-		// compare the % difference and if that difference is over the max diff threshold and also above the minimum
-		// db size we defrag the members state file. In the case where this command only partially completed controller
-		// can clean that up on the next sync. Having the db sizes slightly different is not a problem in itself.
-		if isEndpointBackendFragmented(member, status) {
-			recorder.Eventf("DefragControllerDefragmentAttempt", "Attempting defrag on member: %s, memberID: %x, dbSize: %d, dbInUse: %d, leader ID: %d", member.Name, member.ID, status.DbSize, status.DbSizeInUse, status.Leader)
-			if _, err := c.defragClient.Defragment(ctx, member); err != nil {
-				// Defrag can timeout if defragmentation takes longer than etcdcli.DefragDialTimeout.
-				errMsg := fmt.Sprintf("failed defrag on member: %s, memberID: %x: %v", member.Name, member.ID, err)
-				recorder.Eventf("DefragControllerDefragmentFailed", errMsg)
-				errors = append(errors, fmt.Errorf("%s", errMsg))
+		slices.SortFunc(followers, sortByLeastFragmented)
+
+		for _, newLeader := range followers {
+			err := c.leaderMover.MoveLeader(ctx, defragTargetMember, newLeader.Member.ID)
+			if err != nil {
+				recorder.Warningf("DefragControllerLeaderTransferAttemptFailed", "Failed to move leader away from member %s to member %s before defrag: %v", defragTargetMember.Name, newLeader.Member.Name, err)
 				continue
 			}
 
-			recorder.Eventf("DefragControllerDefragmentSuccess", "etcd member has been defragmented: %s, memberID: %d", member.Name, member.ID)
-			successfulDefrags++
-
-			// Give cluster time to recover before we move to the next member.
-			if err := wait.Poll(
-				pollWaitDuration,
-				pollTimeoutDuration,
-				func() (bool, error) {
-					// Ensure defragmentation attempts have clear observable signal.
-					klog.V(4).Infof("Sleeping to allow cluster to recover before defrag next member: %v", c.defragWaitDuration)
-					time.Sleep(c.defragWaitDuration)
-
-					memberHealth, err := c.memberLister.MemberHealth(ctx)
-					if err != nil {
-						klog.Warningf("failed checking member health: %v", err)
-						return false, nil
-					}
-					if !etcdcli.IsClusterHealthy(memberHealth) {
-						klog.Warningf("cluster is unhealthy: %s", memberHealth.Status())
-						return false, nil
-					}
-					return true, nil
-				}); err != nil {
-				errors = append(errors, fmt.Errorf("timeout waiting for cluster to stabilize after defrag: %w", err))
-			}
-		} else {
-			// no fragmentation needed is also a success
-			successfulDefrags++
+			recorder.Eventf("DefragControllerLeaderTransferSuccess", "Moved leadership away from member %s (memberID: %x) to member %s (memberID: %x) before defrag, requeueing to allow etcd to settle", defragTargetMember.Name, defragTargetMember.ID, newLeader.Member.Name, newLeader.Member.ID)
+			syncCtx.Queue().AddAfter(syncCtx.QueueKey(), c.defragSettleTime)
+			return nil
 		}
+		recorder.Warningf("DefragControllerLeaderTransferFailed", "Failed to move leader away from member %s, continuing with blocking leader defrag", defragTargetMember.Name)
 	}
 
-	if successfulDefrags != len(endpointStatus) {
+	recorder.Eventf("DefragControllerDefragmentAttempt", "Attempting defrag on member: %s, memberID: %x, dbSize: %d, dbInUse: %d, leader ID: %d", defragTargetMember.Name, defragTargetMember.ID, defragTargetStatus.DbSize, defragTargetStatus.DbSizeInUse, defragTargetStatus.Leader)
+	if _, err := c.defragClient.Defragment(ctx, defragTargetMember); err != nil {
+		errMsg := fmt.Sprintf("failed defrag on member: %s, memberID: %x: %v", defragTargetMember.Name, defragTargetMember.ID, err)
+		recorder.Warningf("DefragControllerDefragmentFailed", errMsg)
+		klog.Errorf("%s", errMsg)
 		c.numDefragFailures++
-		recorder.Eventf("DefragControllerDefragmentPartialFailure",
-			"only %d/%d members were successfully defragmented, %d tries left before controller degrades",
-			successfulDefrags, len(endpointStatus), maxDefragFailuresBeforeDegrade-c.numDefragFailures)
-
 		if c.numDefragFailures >= maxDefragFailuresBeforeDegrade {
-			_, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
-				Type:    defragDegradedCondition,
-				Status:  operatorv1.ConditionTrue,
-				Reason:  "Error",
-				Message: fmt.Sprintf("degraded after %d attempts at defragmenting all etcd members", c.numDefragFailures),
-			}))
-			if updateErr != nil {
-				recorder.Warning("DefragControllerUpdatingStatus", updateErr.Error())
-			}
+			c.setDegraded(ctx, recorder)
 		}
-
-		// return all errors here for the sync loop to retry immediately
-		return v1helpers.NewMultiLineAggregate(errors)
+		syncCtx.Queue().AddAfter(syncCtx.QueueKey(), c.defragSettleTime)
+		return nil
 	}
 
-	if len(errors) > 0 {
-		klog.Warningf("found errors even though all members have been successfully defragmented: %s",
-			v1helpers.NewMultiLineAggregate(errors).Error())
-	}
-
+	recorder.Eventf("DefragControllerDefragmentSuccess", "etcd member has been defragmented: %s, memberID: %d", defragTargetMember.Name, defragTargetMember.ID)
 	c.numDefragFailures = 0
+	c.clearDegraded(ctx, recorder)
+
+	c.defragTargets = c.defragTargets[1:]
+	syncCtx.Queue().AddAfter(syncCtx.QueueKey(), c.defragSettleTime)
+	return nil
+}
+
+func (c *DefragController) setDegraded(ctx context.Context, recorder events.Recorder) {
+	_, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+		Type:    defragDegradedCondition,
+		Status:  operatorv1.ConditionTrue,
+		Reason:  "Error",
+		Message: fmt.Sprintf("degraded after %d attempts at defragmenting etcd members", c.numDefragFailures),
+	}))
+	if updateErr != nil {
+		recorder.Warning("DefragControllerUpdatingStatus", updateErr.Error())
+	}
+}
+
+func (c *DefragController) clearDegraded(ctx context.Context, recorder events.Recorder) {
 	_, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient,
 		v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
 			Type:   defragDegradedCondition,
@@ -262,8 +301,6 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
 	if updateErr != nil {
 		recorder.Warning("DefragControllerUpdatingStatus", updateErr.Error())
 	}
-
-	return updateErr
 }
 
 func (c *DefragController) ensureControllerDisabledCondition(ctx context.Context, desiredStatus operatorv1.ConditionStatus, recorder events.Recorder) error {
@@ -293,10 +330,6 @@ func (c *DefragController) ensureControllerDisabledCondition(ctx context.Context
 // This can happen if the operator starts defrag of the cluster but then loses leader status and is rescheduled before
 // the operator can defrag all members.
 func isEndpointBackendFragmented(member *etcdserverpb.Member, endpointStatus *clientv3.StatusResponse) bool {
-	if endpointStatus == nil {
-		klog.Errorf("endpoint status validation failed: %v", endpointStatus)
-		return false
-	}
 	fragmentedPercentage := checkFragmentationPercentage(endpointStatus.DbSize, endpointStatus.DbSizeInUse)
 	if fragmentedPercentage > 0.00 {
 		klog.Infof("etcd member %q backend store fragmented: %.2f %%, dbSize: %d", member.Name, fragmentedPercentage, endpointStatus.DbSize)
@@ -310,14 +343,16 @@ func checkFragmentationPercentage(ondisk, inuse int64) float64 {
 	return math.Round(fragmentedPercentage*100) / 100
 }
 
-func getMemberFromStatus(members []*etcdserverpb.Member, endpointStatus *clientv3.StatusResponse) (*etcdserverpb.Member, error) {
-	if endpointStatus == nil {
-		return nil, fmt.Errorf("endpoint status validation failed: %v", endpointStatus)
-	}
-	for _, member := range members {
-		if member.ID == endpointStatus.Header.MemberId {
-			return member, nil
-		}
-	}
-	return nil, fmt.Errorf("no member found in MemberList matching ID: %v", endpointStatus.Header.MemberId)
+func sortByLeastFragmented(a, b StatusMember) int {
+	return cmp.Compare(
+		checkFragmentationPercentage(a.Status.DbSize, a.Status.DbSizeInUse),
+		checkFragmentationPercentage(b.Status.DbSize, b.Status.DbSizeInUse),
+	)
+}
+
+func sortByMostFragmented(a, b StatusMember) int {
+	return cmp.Compare(
+		checkFragmentationPercentage(b.Status.DbSize, b.Status.DbSizeInUse),
+		checkFragmentationPercentage(a.Status.DbSize, a.Status.DbSizeInUse),
+	)
 }
--- a/pkg/operator/defragcontroller/defragcontroller_test.go
+++ b/pkg/operator/defragcontroller/defragcontroller_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"regexp"
 	"strings"
 	"testing"
 	"time"
@@ -69,6 +68,7 @@ func TestNewDefragController(t *testing.T) {
 		staticPodStatus     *operatorv1.StaticPodOperatorStatus
 		objects             []runtime.Object
 		clusterSize         int
+		syncLoops           int
 		memberHealth        *etcdcli.FakeMemberHealth
 		dbInUse             int64
 		dbSize              int64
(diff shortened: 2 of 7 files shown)


pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +150,127 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-		}
-	}
-
-	var endpointStatus []*clientv3.StatusResponse
-	var leader *clientv3.StatusResponse
-	for _, member := range etcdMembers {
-		if len(member.ClientURLs) == 0 {
-			// skip unstarted member
+	var (
+		etcdMembers    []*etcdserverpb.Member
```
atiratree: this could be a hashmap

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +150,127 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
 		if leader == nil && status.Leader == member.ID {
 			leader = status
 			continue
 		}
 		endpointStatus = append(endpointStatus, status)
 	}
 
+	// Sort non-leader members by fragmentation percentage descending so we
+	// defrag the most fragmented member first each cycle.
+	slices.SortFunc(endpointStatus, func(a, b *clientv3.StatusResponse) int {
```
atiratree: can you please add a unit test for this sort?

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -90,6 +96,11 @@ func (c *DefragController) sync(ctx context.Context, syncCtx factory.SyncContext
 		return nil
 	}
 
+	if !c.lastDefragTime.IsZero() && time.Since(c.lastDefragTime) < c.defragCooldown {
+		klog.V(4).Infof("Defrag cooldown active, %v remaining", c.defragCooldown-time.Since(c.lastDefragTime))
+		return nil
```
atiratree: we could schedule a next defrag if we know the remaining time here

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -90,6 +96,11 @@ func (c *DefragController) sync(ctx context.Context, syncCtx factory.SyncContext
 		return nil
 	}
 
+	if !c.lastDefragTime.IsZero() && time.Since(c.lastDefragTime) < c.defragCooldown {
```
atiratree: Could it be a problem that a member could exceed the fragmentation limit for 1h in high churn cluster, since this backfoff / cooldown is member agnostic?

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +150,127 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-				Reason:  "Error",
-				Message: fmt.Sprintf("degraded after %d attempts at defragmenting all etcd members", c.numDefragFailures),
-			}))
-			if updateErr != nil {
-				recorder.Warning("DefragControllerUpdatingStatus", updateErr.Error())
-			}
+		recorder.Eventf("DefragControllerDefragmentSuccess", "etcd member has been defragmented: %s, memberID: %d", member.Name, member.ID)
+
+		// Give cluster time to recover before the next sync defrags another member.
+		if err := wait.PollUntilContextTimeout(ctx, pollWaitDuration, pollTimeoutDuration, false,
```
atiratree: This is not the best pattern for controllers. Could we just select a single member try to defrag it and use a queue for any backoff checks like checking the health? 

I suppose the defrag is not that often so a 10m between each member defrag shouldn't be an issue?

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -52,6 +55,8 @@ type DefragController struct {
 
 	numDefragFailures  int
 	defragWaitDuration time.Duration
+	defragCooldown     time.Duration
+	lastDefragTime     time.Time
```
tjungblu: you will have to persist this somewhere (operator conditions?), imagine you're restarting CEO within that hour.

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +150,127 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-			if _, err := c.defragClient.Defragment(ctx, member); err != nil {
-				// Defrag can timeout if defragmentation takes longer than etcdcli.DefragDialTimeout.
-				errMsg := fmt.Sprintf("failed defrag on member: %s, memberID: %x: %v", member.Name, member.ID, err)
-				recorder.Eventf("DefragControllerDefragmentFailed", errMsg)
-				errors = append(errors, fmt.Errorf("%s", errMsg))
-				continue
+		recorder.Eventf("DefragControllerDefragmentAttempt", "Attempting defrag on member: %s, memberID: %x, dbSize: %d, dbInUse: %d, leader ID: %d", member.Name, member.ID, status.DbSize, status.DbSizeInUse, status.Leader)
+		if _, err := c.defragClient.Defragment(ctx, member); err != nil {
+			// Defrag can timeout if defragmentation takes longer than etcdcli.DefragDialTimeout.
+			errMsg := fmt.Sprintf("failed defrag on member: %s, memberID: %x: %v", member.Name, member.ID, err)
```
tjungblu: it might not indicate a failure in this case. We have two options: increase the timeout (plus some slack) based on the db size, or just always use some crazy value like 10 minutes.

Not sure it makes sense to poll the status instead, so have a short timeout always and just poll until we're back up and the size has been reduced.

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +150,127 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-				Status:  operatorv1.ConditionTrue,
-				Reason:  "Error",
-				Message: fmt.Sprintf("degraded after %d attempts at defragmenting all etcd members", c.numDefragFailures),
-			}))
-			if updateErr != nil {
-				recorder.Warning("DefragControllerUpdatingStatus", updateErr.Error())
-			}
+		recorder.Eventf("DefragControllerDefragmentSuccess", "etcd member has been defragmented: %s, memberID: %d", member.Name, member.ID)
+
+		// Give cluster time to recover before the next sync defrags another member.
```
tjungblu: if the defrag call came back, then the cluster should be in healthy state and respond. I think the polling is a bit misplaced here.

pkg/operator/defragcontroller/defragcontroller.go
tjungblu: okay I see you added the polling on success - but we need the polling on failure 😄

pkg/operator/defragcontroller/defragcontroller.go
tjungblu: line 144 already has that 	
> if !etcdcli.IsClusterHealthy(memberHealth) 

so we can have the backoff work through the sync loop for some period of time

pkg/operator/defragcontroller/defragcontroller.go
tjungblu: hmm maybe we can frame this cooldown differently.... e.g after a defrag, the cluster must have no unhealthy members for at least 10m before defragmenting the next member

dusk125:
With the improvements from https://github.com/openshift/etcd/pull/378, the current implementation should be sufficient to lower the impact of defrag. The changes to etcd can take defrag disruption from 30s to 100ms. The current implementation of the changes to the defrag controller run a single defrag, on the most fragmented member, then check the member health on the next sync (once every 11 minutes) and only continue defragging if all members are healthy.

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +145,114 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-		// can clean that up on the next sync. Having the db sizes slightly different is not a problem in itself.
-		if isEndpointBackendFragmented(member, status) {
-			recorder.Eventf("DefragControllerDefragmentAttempt", "Attempting defrag on member: %s, memberID: %x, dbSize: %d, dbInUse: %d, leader ID: %d", member.Name, member.ID, status.DbSize, status.DbSizeInUse, status.Leader)
-			if _, err := c.defragClient.Defragment(ctx, member); err != nil {
-				// Defrag can timeout if defragmentation takes longer than etcdcli.DefragDialTimeout.
-				errMsg := fmt.Sprintf("failed defrag on member: %s, memberID: %x: %v", member.Name, member.ID, err)
-				recorder.Eventf("DefragControllerDefragmentFailed", errMsg)
-				errors = append(errors, fmt.Errorf("%s", errMsg))
+	// If the leader is in the first slot, and we have other members to defrag, defrag the others first,
+	// so we limit the number of times we move the leader during defrag.
```
bhperry: in a high churn environment, wouldn't it be possible for leader to keep getting skipped while other members are repeatedly defragged and refragged? Especially since with this change defrag happens more slowly.

Perhaps there should be a threshold at which the leader is selected for defrag even when there are other candidates.

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +145,114 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
+		followers := make([]StatusMember, 0, len(statusMembers))
+		for id, member := range statusMembers {
+			if defragTargetMember.ID == id {
 				continue
 			}
+			followers = append(followers, member)
+		}
+
+		if len(followers) > 0 {
+			slices.SortFunc(followers, sortByLeastFragmented)
```
bhperry: why make followers slice and re-sort in the opposite order of defragTargets? Couldn't you just take the last element of defragTargets as the least fragmented?

bhperry (changes requested):

pkg/operator/defragcontroller/defragcontroller.go
bhperry: oh nevermind I understand, they don't get added to defragTargets unless they are above the defrag threshold.

pkg/operator/defragcontroller/defragcontroller.go
dusk125: This is a good point, let me have a little think about this... it's tangentially related to us being slower to react to the database getting close to quota. I've been noodling on a "fast mode" or something that detects these scenarios and does a requeue faster so it's not ~30m to defrag all 3 members

dusk125: I'm open to suggestions as well

bhperry: My first thought is setting a "critical threshold" above the standard 45% (say 65% for now) which will force the leader to defrag. But with high churn that could end up always selecting the leader and forcing it to move on every sync if each member is above 65% by the time it runs. No good.

Seems like it would need to record which members have already been defragged recently in order to ensure the leader isn't starved while also avoiding selecting the leader too eagerly.

dusk125: If we're not overly concerned with recovering the exact state on CEO restart, we could just store that data like the failed defrag count?

bhperry: Yeah that's what I was thinking

dusk125: I actually decided to go back to the loop implementation; however, with the preemptive leader move and making sure that the cluster is stable before continuing. There was a bug where a defrag would fail and immediately try to defrag the next member.

I think the loop implementation is simpler overall when the "keep track of who's been defragged" would've been necessary to keep from never defragging the leader.

Now, we defrag serially, but we ensure the cluster stability at each step, and still move the leader so that we avoid write disruptions.

dusk125:
Considering a different approach, before merging, I'll squash if the loop is the way we want to go

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,111 +146,129 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-		}
-	}
-
-	var endpointStatus []*clientv3.StatusResponse
-	var leader *clientv3.StatusResponse
-	for _, member := range etcdMembers {
-		if len(member.ClientURLs) == 0 {
-			// skip unstarted member
+	var (
+		statusMembers = make(map[uint64]StatusMember)
```
bhperry, line 165: It doesn't look like StatusMember is ever looked up in the map anymore with the new implementation, so I don't think there's a need for a separate map and followers slice in the fragmented leader case, maybe this should just be a slice with leader filtered out from the start. Sort by least fragmented later if needed.

pkg/operator/defragcontroller/defragcontroller.go
bhperry: I was beginning to think it looked a lot like the original looped version with extra steps, so that's a good call. Certainly much more straightforward this way.

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,111 +146,129 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-					}
-					return true, nil
-				}); err != nil {
-				errors = append(errors, fmt.Errorf("timeout waiting for cluster to stabilize after defrag: %w", err))
-			}
-		} else {
-			// no fragmentation needed is also a success
+		// Give cluster time to recover before we move to the next member.
+		klog.V(4).Infof("Sleeping to allow cluster to recover before defrag next member")
+		if err := c.waitForClusterStability(ctx); err != nil {
```
atiratree: IMO, we could do without the for loop and the synchronous wait, if we don't need to invoke the next defrag immediately. We could just do AddAfter(time.Minute) and pick the next target in the next sync

It could simplify the code and error handling

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,111 +146,129 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
 
-	if successfulDefrags != len(endpointStatus) {
+	if successfulDefrags != len(defragTargets) {
 		c.numDefragFailures++
 		recorder.Eventf("DefragControllerDefragmentPartialFailure",
 			"only %d/%d members were successfully defragmented, %d tries left before controller degrades",
-			successfulDefrags, len(endpointStatus), maxDefragFailuresBeforeDegrade-c.numDefragFailures)
+			successfulDefrags, len(defragTargets), maxDefragFailuresBeforeDegrade-c.numDefragFailures)
 
 		if c.numDefragFailures >= maxDefragFailuresBeforeDegrade {
```
atiratree, line 266: How often do we expect these failures to happen? Could we degrade after the first one?

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,111 +146,129 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-				recorder.Eventf("DefragControllerDefragmentFailed", errMsg)
-				errors = append(errors, fmt.Errorf("%s", errMsg))
-				continue
+			if len(followers) > 0 {
+				slices.SortFunc(followers, sortByLeastFragmented)
+				newLeader := followers[0]
+
+				err := c.leaderMover.MoveLeader(ctx, newLeader.Member.ID)
+				if err != nil {
+					recorder.Eventf("DefragControllerLeaderTransferFailed", "Failed to move leader away from member %s to member %s before defrag: %v", targetMember.Name, newLeader.Member.Name, err)
```
atiratree: Maybe we should use a Warning event here? How often does this happen? Is it okay to defrag the leader? Can we move the leader elsewhere when this fails?

pkg/operator/defragcontroller/defragcontroller.go
atiratree: > There was a bug where a defrag would fail and immediately try to defrag the next member.

I am a bit confused here as to where this bug occurs.

> keep track of who's been defragged

Don't we always have this information via `isEndpointBackendFragmented`?

bhperry: This was in the previous implementation in the [first commit](https://github.com/openshift/cluster-etcd-operator/pull/1618/changes/dffee92126a79ab4110a3ecbf0788a93e59318ab) in which a single member was defragged each sync. In that case you don't know which members were previously defragged, only which members are currently over the fragmentation threshold. So if a cluster is fragmenting fast enough it would be possible for the same members to keep going over the threshold by the next sync, and the leader would always get skipped.

pkg/operator/defragcontroller/defragcontroller.go
dusk125: Moved to warning. There are some small cases where it can happen and there's not really a problem with the wider cluster, but in a health cluster (that we just checked before the cycle), this should more or less always succeed.

We can defrag the leader, we'll just accept a disruption to writes while the leader is being defragged. I've made it so that the leadership will be attempted across all followers sequentially if it fails.

pkg/operator/defragcontroller/defragcontroller.go
dusk125: Done

pkg/operator/defragcontroller/defragcontroller.go
dusk125: Probably not often, but I believe this is protecting against a single transient failure to defrag degrading the controller and causing noise. A single failure to defrag isn't indicative of a problem, but a couple in a row without a success to reset is.

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +155,125 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-				Status:  operatorv1.ConditionTrue,
-				Reason:  "Error",
-				Message: fmt.Sprintf("degraded after %d attempts at defragmenting all etcd members", c.numDefragFailures),
-			}))
-			if updateErr != nil {
-				recorder.Warning("DefragControllerUpdatingStatus", updateErr.Error())
-			}
+			c.setDegraded(ctx, recorder)
 		}
+		syncCtx.Queue().AddAfter(syncCtx.QueueKey(), defragSettleTime)
```
atiratree: Just double checking here. We had a 1 minute settle time in the last version of the PR. Is 10s enough for settling and are we okay to fall into err backoff after that if it lasts longer?

pkg/operator/defragcontroller/defragcontroller.go
atiratree: sounds good and the events will certainly help if there is an issue

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +155,125 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
+			syncCtx.Queue().AddAfter(syncCtx.QueueKey(), defragSettleTime)
+			return nil
 		}
+		recorder.Warningf("DefragControllerLeaderTransferFailed", "Failed to move leader away from member %s, continuing with blocking leader defrag", defragTargetMember.Name)
 	}
 
-	if successfulDefrags != len(endpointStatus) {
+	recorder.Eventf("DefragControllerDefragmentAttempt", "Attempting defrag on member: %s, memberID: %x, dbSize: %d, dbInUse: %d, leader ID: %d", defragTargetMember.Name, defragTargetMember.ID, defragTargetStatus.DbSize, defragTargetStatus.DbSizeInUse, defragTargetStatus.Leader)
+	if _, err := c.defragClient.Defragment(ctx, defragTargetMember); err != nil {
+		// Defrag can timeout if defragmentation takes longer than etcdcli.DefragDialTimeout.
```
atiratree: This sounds significant with the increased DB size. Is there a way to circumvent this?

pkg/operator/defragcontroller/defragcontroller.go
atiratree: I wonder how likely that is - maybe we could create an alert for such high fragmentation rate?

I suppose we could also add failsafe and prefer a leader when it goes over a high percentage? e.g. 85%?

pkg/operator/defragcontroller/defragcontroller.go
atiratree: Thanks, it is easier to follow now. I wonder if we need to be concerned with https://github.com/openshift/cluster-etcd-operator/pull/1618#discussion_r3597095966?

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +155,125 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
+		errMsg := fmt.Sprintf("failed defrag on member: %s, memberID: %x: %v", defragTargetMember.Name, defragTargetMember.ID, err)
+		recorder.Eventf("DefragControllerDefragmentFailed", errMsg)
```
atiratree:
```suggestion
		recorder.Warningf("DefragControllerDefragmentFailed", errMsg)
```

pkg/operator/defragcontroller/defragcontroller.go
bhperry: > I wonder how likely that is - maybe we could create an alert for such high fragmentation rate?

Probably not common, but while running churn tests with kube-burner I have seen fragmentation jump up ~3GB in under a minute.

> I suppose we could also add failsafe and prefer a leader when it goes over a high percentage? e.g. 85%?

My worry with a hard threshold to always select the leader is that if all members are highly fragmented it would end up selecting the leader for defrag every time. Or at the very least, for the first two defrags.

Sync 1:
- etcd1 (leader, 86%) - etcd2 (86%) - etcd3 (86%)
- defrag etcd1 since it is over critical threshold, leader moves to etcd2

Sync 2:
- etcd1 (0%) - etcd2 (leader, 86%) - etcd3 (86%)
- defrag etcd2 since it is over critical threshold, leader moves to etcd1 (in the best case)

...

pkg/operator/defragcontroller/defragcontroller.go
dusk125: A couple things have changed. I changed the member client to be uncached, the cached member client has a TTL of 1 minute. So 10s should be enough for the individual member to consider itself unhealthy, then in the next sync, the health check will take care of the gating.

pkg/operator/defragcontroller/defragcontroller.go
dusk125: I'm looking into this now. Currently, not really, unless we increase the defrag timeout (which we may need to do as a 9GiB db can take ~40s: tested on cluster bot)

bhperry: In my testing, on a 16GB database at ~50% fragmentation it takes around 80s. But I also haven't seen it timeout in CEO even though DefragDialTimeout is only 60s

bhperry: > // Defrag can timeout if defragmentation takes longer than etcdcli.DefragDialTimeout.

I think this comment is misleading and could use clarification. As far as I can tell (and as the name implies), DefragDialTimeout only bounds the initial connection to the member for defrag, it is not a timeout on the total defrag time. I have regularly seen defrag take 2 minutes in CEO with no timeouts.

dusk125: Yeah, just doing a grep, it is only used in the Defragment wrapper in the CEO as the dial timeout, not the defragment timeout. I'll update the comment to clarify this.

edit: I looked into this a bit more, and that comment isn't really that help to begin with, I'm just going to remove it.

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +154,119 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-		if len(member.ClientURLs) == 0 {
-			// skip unstarted member
+	var (
+		isNewCycle    = len(c.defragTargets) == 0
+		statusMembers = make(map[uint64]StatusMember)
+	)
+	for _, member := range members {
+		// filter out learner members since they don't support the defragment API call
+		// and filter out unstarted members
+		if member.IsLearner || len(member.ClientURLs) == 0 {
```
bhperry, line 170: Is there any possibility of a member having either of these conditions (e.g. becoming a learner, not sure how that works) after it was already added to defragTargets in a previous sync?

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +154,119 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
+	})
+
+	defragTarget := statusMembers[c.defragTargets[0]]
+	defragTargetStatus, defragTargetMember := defragTarget.Status, defragTarget.Member
+
+	// Preemptively attempt to move the leadership away from the current defrag target to another valid follower.
+	// We try this to avoid the write block that occurs while the leader is being defragmented.
+	// We record any error that occurs while attempting this, but we do not halt defrag if the move fails;
+	// we just accept the write block.
+	if defragTarget.IsLeader() && len(statusMembers) > 1 {
```
bhperry, line 235: We've now learned that write blocks are not just from the leader defragmenting, correct? Guessing this is an outdated comment.

pkg/operator/defragcontroller/defragcontroller.go
dusk125: Maybe....? it would be unlikely unless the cluster is scaling or something like that where we have quorum but additional members that haven't been started, or are new and haven't been promoted yet.

How come?

bhperry: Here they are intentionally not being added to the list. But if they are already on the list and then later become a learner then they may attempt to defrag and then fail, possibly becoming a blocker that is stuck in the list.

bhperry: May not be a possible scenario. More generally, I am wondering what happens if there is a member in the defragTargets list that repeatedly fails to defragment, or if a member was removed from the cluster but is still in the list. Seems like that could prevent other members from even being considered for defragmentation.

pkg/operator/defragcontroller/defragcontroller.go
atiratree: Cool, I missed the uncached client part.

pkg/operator/defragcontroller/defragcontroller.go
atiratree: Thanks for digging down. Btw what happens when the connection drops? Is the defrag stopped, or continues without sending back a result? For example during a CEO restart? Does CEO initiates a new degrag immediately afterwards?

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +154,119 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
+			return nil
 		}
+		recorder.Warningf("DefragControllerLeaderTransferFailed", "Failed to move leader away from member %s, continuing with blocking leader defrag", defragTargetMember.Name)
 	}
 
-	if successfulDefrags != len(endpointStatus) {
+	recorder.Eventf("DefragControllerDefragmentAttempt", "Attempting defrag on member: %s, memberID: %x, dbSize: %d, dbInUse: %d, leader ID: %d", defragTargetMember.Name, defragTargetMember.ID, defragTargetStatus.DbSize, defragTargetStatus.DbSizeInUse, defragTargetStatus.Leader)
+	if _, err := c.defragClient.Defragment(ctx, defragTargetMember); err != nil {
+		errMsg := fmt.Sprintf("failed defrag on member: %s, memberID: %x: %v", defragTargetMember.Name, defragTargetMember.ID, err)
+		recorder.Eventf("DefragControllerDefragmentFailed", errMsg)
```
atiratree: I still see the event suggestion unresolved

pkg/operator/defragcontroller/defragcontroller.go
dusk125: Ahhh good catch, I don't think it's likely, but it's easy enough to protect against.

bhperry:
Some changes we were discussing in slack:

- If there are existing defragTargets in the list, they should be checked against the defrag threshold again before running defrag. Fragmentation can naturally decrease over time in an active cluster, so if a member ID has been in the list for a couple cycles it is possible that it is well below the fragmentation threshold and running defrag could result in a large spike in memory (worst case the full size of the database). Currently the status is only used to resort the list, not to check threshold.
- Ensure that all IDs in the list belong to current etcd members. If a member is removed from the cluster while it is already in the list that could lead to unexpected behavior such as a nil pointer deref or silently failing and never rebuilding the defragTargets list.

dusk125:
/testwith openshift/cluster-etcd-operator/main/e2e-operator https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/configmap-scale https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-agnostic-ovn https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-serial-1of2 https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-serial-2of2 https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-agnostic-ovn-upgrade https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-single-node https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator-disruptive https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-metal-ipi-ovn-ipv6 https://github.com/openshift/etcd/pull/378

dusk125:
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-serial https://github.com/openshift/etcd/pull/378

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +154,132 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
 		}
-		endpointStatus = append(endpointStatus, status)
 	}
 
-	// Leader last if possible.
-	if leader != nil {
-		klog.V(4).Infof("Appending leader last, ID: %x", leader.Header.MemberId)
-		endpointStatus = append(endpointStatus, leader)
+	if !isNewCycle {
+		c.defragTargets = slices.DeleteFunc(c.defragTargets, func(i uint64) bool {
```
atiratree: nit, this suggests i is an index
```suggestion
		c.defragTargets = slices.DeleteFunc(c.defragTargets, func(targetID uint64) bool {
```

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +154,132 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-	successfulDefrags := 0
-	var errors []error
-	for _, status := range endpointStatus {
-		member, err := getMemberFromStatus(etcdMembers, status)
-		if err != nil {
-			errors = append(errors, err)
-			continue
-		}
+	if len(c.defragTargets) == 0 {
+		klog.V(4).Info("Defrag skipped: no etcd members meet the conditions for defragmentation")
```
atiratree: nit: for debugging (level 4) it would be nice to print all the target fragmentation percentages

pkg/operator/defragcontroller/defragcontroller.go
atiratree: +1, nice find and the fix looks reasonable

dusk125:
/testwith openshift/cluster-etcd-operator/main/e2e-operator https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/configmap-scale https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-agnostic-ovn https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-serial https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-agnostic-ovn-upgrade https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-single-node https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator-disruptive https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-metal-ipi-ovn-ipv6 https://github.com/openshift/etcd/pull/378

dusk125:
/hold for https://github.com/openshift/etcd/pull/378 to merge first
This currently relies on the new featuregate in etcd itself so CI will fail until that etcd feature gate exists

dusk125:
/testwith openshift/cluster-etcd-operator/main/e2e-operator https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/configmap-scale https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-agnostic-ovn https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-serial https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-agnostic-ovn-upgrade https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-single-node https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator-disruptive https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-metal-ipi-ovn-ipv6 https://github.com/openshift/etcd/pull/378

dusk125:
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-single-node https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator-disruptive https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-metal-ipi-ovn-ipv6 https://github.com/openshift/etcd/pull/378

dusk125:
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator-disruptive https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-metal-ipi-ovn-ipv6 https://github.com/openshift/etcd/pull/378

dusk125:
/testwith openshift/cluster-etcd-operator/main/e2e-operator https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/configmap-scale https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-agnostic-ovn https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-serial https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-agnostic-ovn-upgrade https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-single-node https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator-disruptive https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-metal-ipi-ovn-ipv6 https://github.com/openshift/etcd/pull/378

dusk125:
/testwith openshift/cluster-etcd-operator/main/e2e-operator https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/configmap-scale https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-agnostic-ovn https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-serial https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-agnostic-ovn-upgrade https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-single-node https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-gcp-operator-disruptive https://github.com/openshift/etcd/pull/378
/testwith openshift/cluster-etcd-operator/main/e2e-metal-ipi-ovn-ipv6 https://github.com/openshift/etcd/pull/378

dusk125:
/hold cancel
etcd PR has merged

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -310,14 +338,17 @@ func checkFragmentationPercentage(ondisk, inuse int64) float64 {
...
-	}
-	return nil, fmt.Errorf("no member found in MemberList matching ID: %v", endpointStatus.Header.MemberId)
+func sortByLeastFragmented(a, b StatusMember) int {
+	return cmp.Compare(
+		checkFragmentationPercentage(a.Status.DbSize, a.Status.DbSizeInUse),
+		checkFragmentationPercentage(b.Status.DbSize, b.Status.DbSizeInUse),
+	)
+}
+
+func sortByMostFragmented(statusMembers map[uint64]StatusMember, a, b uint64) int {
```
atiratree: nit: it would be better if this had the same function definition as `sortByLeastFragmented`

pkg/operator/defragcontroller/defragcontroller.go
```diff
@@ -139,120 +155,138 @@ func (c *DefragController) runDefrag(ctx context.Context, recorder events.Record
...
-			errors = append(errors, err)
-			continue
+	if len(c.defragTargets) == 0 {
+		targets := make([]string, len(c.defragTargets))
+		for i, targetID := range c.defragTargets {
+			target := statusMembers[targetID]
+			percent := checkFragmentationPercentage(target.Status.DbSize, target.Status.DbSizeInUse)
+			targets[i] = fmt.Sprintf("%s={id: %d, fragmentation: %.2f%%, sizeInUse: %d}", target.Member.Name, target.Member.ID, percent, target.Status.DbSize)
 		}
+		klog.V(4).Infof("Defrag skipped: no etcd members meet the conditions for defragmentation: %s", strings.Join(targets, ", "))
```
atiratree:
```suggestion
		klog.V(4).Infof("Defrag skipped: no etcd members meet the conditions for defragmentation:\n%s", strings.Join(targets, ", "))
```

pkg/operator/defragcontroller/defragcontroller_test.go
```diff
@@ -413,6 +411,181 @@ func TestNewDefragControllerMultiSyncs(t *testing.T) {
...
+
+	syncCtx := factory.NewSyncContext("defrag-controller", eventRecorder)
+
+	// First sync: leader is the most fragmented, so leadership is transferred and sync requeues.
+	err := controller.sync(context.TODO(), syncCtx)
+	require.NoError(t, err)
+
+	var leaderTransferEvents int
+	for _, event := range eventRecorder.Events() {
+		if strings.HasPrefix(event.Message, "Moved leadership away from member") {
```
atiratree: nit: it might be better to just use `Contains` here

pkg/operator/defragcontroller/defragcontroller_test.go
```diff
@@ -413,6 +411,181 @@ func TestNewDefragControllerMultiSyncs(t *testing.T) {
...
+	}
+	assert.Equal(t, 1, leaderTransferEvents, "expected exactly one leader transfer event")
+
+	// Second sync: the former leader is now a follower and gets defragged.
+	err = controller.sync(context.TODO(), syncCtx)
+	require.NoError(t, err)
+
+	var defragSuccessEvents int
+	for _, event := range eventRecorder.Events() {
+		if strings.HasPrefix(event.Message, "etcd member has been defragmented") {
```
atiratree: and so on

pkg/operator/defragcontroller/defragcontroller_test.go
```diff
@@ -413,6 +411,181 @@ func TestNewDefragControllerMultiSyncs(t *testing.T) {
...
+		defragClient:         fakeEtcdClient,
+		statusClient:         fakeEtcdClient,
+		leaderMover:          fakeEtcdClient,
+		infrastructureLister: configv1listers.NewInfrastructureLister(indexer),
+		configmapLister:      corev1listers.NewConfigMapLister(indexer),
+	}
+
+	syncCtx := factory.NewSyncContext("defrag-controller", eventRecorder)
+
+	// First sync: leader is the most fragmented, so leadership is transferred and sync requeues.
```
atiratree, line 475: One concerning thing is that it seems we do not test whether the resyncs occur and we always trigger syncs manually.

Could we set the `defragSettleTime` to a low value and pass it to the controller in at least one of the tests that does multiple resyncs?

pkg/operator/defragcontroller/defragcontroller_test.go
dusk125: But this test is ensuring that there's only a single event, Contains wouldn't give the same information

atiratree: I meant `strings.Contains`, so that the message format is less hardcoded. Easier to add a new sentence to the beginning.

dusk125: Oh lol, gotcha and yep on it

pkg/operator/defragcontroller/defragcontroller_test.go
```diff
@@ -413,6 +411,185 @@ func TestNewDefragControllerMultiSyncs(t *testing.T) {
...
+			leaderTransferEvents++
+		}
+	}
+	assert.Equal(t, 1, leaderTransferEvents, "expected exactly one leader transfer event")
+
+	// Wait for the settle time to elapse and the requeue to fire.
+	time.Sleep(10 * time.Millisecond)
+
+	// Second sync: the former leader is now a follower and gets defragged.
+	err = controller.sync(context.TODO(), syncCtx)
```
atiratree, line 493: shouldn't we now remove the sync and let queue handle it?

pkg/operator/defragcontroller/defragcontroller_test.go
```diff
@@ -413,6 +411,187 @@ func TestNewDefragControllerMultiSyncs(t *testing.T) {
...
+		if strings.Contains(event.Message, "Moved leadership away from member") {
+			leaderTransferEvents++
+		}
+	}
+	assert.Equal(t, 1, leaderTransferEvents, "expected exactly one leader transfer event")
+
+	// Wait for the requeue item to appear on the queue (added by AddAfter after leader transfer).
+	item, shutdown := syncCtx.Queue().Get()
+	require.False(t, shutdown, "queue should not be shut down")
+	syncCtx.Queue().Done(item)
```
atiratree, line 490: We will have to timeout the whole test if we ever break this - which is fine I guess.

atiratree (commented):
/lgtm
Thx!

PR #1619, opened 2026-05-14
OCPEDGE-2476: Added TNF taint and annotation removal logic in CEO

vimauro:
## Summary by CodeRabbit

* **New Features**
  * Operator now proactively removes "out-of-service" taints and their associated annotation when nodes transition to Ready, and updates operator status (sets/clears a degraded condition) on taint-removal failures or successes.

* **Tests**
  * Added unit tests covering detection and cleanup behavior, including successful removal and no-op scenarios.

### DIFF
--- a/pkg/tnf/operator/starter.go
+++ b/pkg/tnf/operator/starter.go
@@ -96,6 +96,9 @@ func HandleDualReplicaClusters(
 				return
 			}
 
+			// clean up Pacemaker's out-of-service taint if the node recovered while CEO was down
+			go tools.RemoveOutOfServiceTaintIfNeeded(ctx, kubeClient, node)
+
 			// this potentially needs some time when we wait for etcd bootstrap to complete, so run it in a goroutine,
 			// to not block the event handler
 			klog.Infof("node added and ready: %s", node.GetName())
@@ -114,6 +117,8 @@ func HandleDualReplicaClusters(
 			newReady := tools.IsNodeReady(newNode)
 			if !oldReady && newReady {
 				klog.Infof("node %s transitioned to ready state", newNode.GetName())
+				// clean up Pacemaker's out-of-service taint now that the failed node is back
+				go tools.RemoveOutOfServiceTaintIfNeeded(ctx, kubeClient, newNode)
 				// this potentially needs some time when we wait for etcd bootstrap to complete, so run it in a goroutine,
 				// to not block the event handler
 				go handleNodesWithRetry(controllerContext, controlPlaneNodeLister, ctx, operatorClient, kubeClient, kubeInformersForNamespaces, etcdInformer)
--- a/pkg/tnf/operator/starter.go
+++ b/pkg/tnf/operator/starter.go
@@ -97,7 +97,7 @@ func HandleDualReplicaClusters(
 			}
 
 			// clean up Pacemaker's out-of-service taint if the node recovered while CEO was down
-			go tools.RemoveOutOfServiceTaintIfNeeded(ctx, kubeClient, node)
+			go tools.RemoveOutOfServiceTaintIfNeeded(ctx, kubeClient, operatorClient, node)
 
 			// this potentially needs some time when we wait for etcd bootstrap to complete, so run it in a goroutine,
 			// to not block the event handler
@@ -112,13 +112,12 @@ func HandleDualReplicaClusters(
 				return
 			}
 
-			// only handle if node transitioned from not ready to ready
 			oldReady := tools.IsNodeReady(oldNode)
 			newReady := tools.IsNodeReady(newNode)
 			if !oldReady && newReady {
 				klog.Infof("node %s transitioned to ready state", newNode.GetName())
 				// clean up Pacemaker's out-of-service taint now that the failed node is back
-				go tools.RemoveOutOfServiceTaintIfNeeded(ctx, kubeClient, newNode)
+				go tools.RemoveOutOfServiceTaintIfNeeded(ctx, kubeClient, operatorClient, newNode)
 				// this potentially needs some time when we wait for etcd bootstrap to complete, so run it in a goroutine,
 				// to not block the event handler
 				go handleNodesWithRetry(controllerContext, controlPlaneNodeLister, ctx, operatorClient, kubeClient, kubeInformersForNamespaces, etcdInformer)
--- a/pkg/tnf/pkg/tools/taint.go
+++ b/pkg/tnf/pkg/tools/taint.go
@@ -10,6 +10,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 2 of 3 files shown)


vimauro:
/verified by testing the changes on a live 5.0 cluster:

1. Positive - both taint and annotation present: Applied node.kubernetes.io/out-of-service=nodeshutdown:NoExecute taint and node.kubernetes.io/out-of-service-applied-by=pacemaker annotation to master-0. After CEO
  restart (to trigger informer re-list via AddFunc), both were removed within seconds. Confirmed via CEO logs.
  2. Negative - taint only, no annotation: Applied the out-of-service taint without the pacemaker annotation. CEO correctly left the taint in place (no log output, taint still present after restart).
  3. Negative - annotation only, no taint: Applied the pacemaker annotation without the taint. CEO correctly left the annotation in place.

vimauro:
/verified by CI and a live 5.0 cluster

pkg/tnf/operator/starter.go
```diff
@@ -96,6 +96,9 @@ func HandleDualReplicaClusters(
 				return
 			}
 
+			// clean up Pacemaker's out-of-service taint if the node recovered while CEO was down
+			go tools.RemoveOutOfServiceTaintIfNeeded(ctx, kubeClient, node)
```
jaypoulz: I'm wondering if there are any kinds of errors from this routine that we want to bubble up to the user. Should we be tracking this functionality in as a condition in the etcd/cluster CR so that we can bubble up degraded status to the operator if it's failing?

pkg/tnf/pkg/tools/taint.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		if !hasOutOfServiceTaint(freshNode) {
+			klog.V(4).Infof("node %s no longer has out-of-service taint after re-read, skipping taint removal", node.Name)
+			return nil
+		}
+
+		freshNode.Spec.Taints = slices.DeleteFunc(freshNode.Spec.Taints, isOutOfServiceTaint)
+
+		_, err = kubeClient.CoreV1().Nodes().Update(ctx, freshNode, metav1.UpdateOptions{})
```
jaypoulz, line 71: Can we just add this removal to the patch to ensure that both changes are processed together? An atomic operation would prevent mismatched state and reduce the number of calls.

pkg/tnf/pkg/tools/taint.go
```diff
@@ -0,0 +1,92 @@
...
+	klog.Infof("node %s has out-of-service taint and pacemaker annotation, removing both", node.Name)
+
+	// Use read-modify-update with conflict retry for taints to avoid clobbering concurrent changes
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		freshNode, err := kubeClient.CoreV1().Nodes().Get(ctx, node.Name, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get node %s: %w", node.Name, err)
+		}
+
+		if !hasOutOfServiceTaint(freshNode) {
```
jaypoulz: for the sake of atomicity (see below), we probably need to check both the taint and annotation.

pkg/tnf/pkg/tools/taint_test.go
```diff
@@ -0,0 +1,319 @@
...
+			node: &corev1.Node{},
+			want: false,
+		},
+		{
+			name: "node with other taints only",
+			node: &corev1.Node{
+				Spec: corev1.NodeSpec{
+					Taints: []corev1.Taint{
+						{
+							Key:    "node.kubernetes.io/not-ready",
```
jaypoulz: nit: there's gotta be some string constant for these somewhere? Right? If not, might be a time to find shared constants home for stuff like this.

jaypoulz (commented):
Overall, I'm happy with the the direction of the patch. There are a few things that I want to call out.

1. We should avoid removing out-of-service unless we added it. I think your approach to using the annotation was a clever way of working around that, but I think we need to sharpen our strategy to make sure that we remove both atomically.
2. We need to decide what to do when a node has the annotation but not the taint. I'd suggest we remove both if EITHER are present and the node is ready.
3. We should probably look for a reasonable way to let the user know if we aren't able to remove the taint for some reason. That said, I'm not sure what the best mechanism is for this. If we keep this code here, the obvious way is to set a condition on the operator like "TNFTaintFencedNodesServiceHealthy" = false when you detect an error and then restore on the first successful run.
4. As a team, we probably need to discuss whether this (and other TNF-specific-but-not-etcd-specific) code actually belongs in CEO. CEO is really about managing etcd. PacemakerHealthCheck falls into this grey area but it ultimately there to help inform the user about etcd from the perspective of pacemaker. This feature feels like it's here because etcd manages pacemaker and it's the closest thing to a home we have. It's the pragmatic choice, but it does raise questions about where we draw the line

vimauro:
> Overall, I'm happy with the the direction of the patch. There are a few things that I want to call out.
> 
>     1. We should avoid removing out-of-service unless we added it. I think your approach to using the annotation was a clever way of working around that, but I think we need to sharpen our strategy to make sure that we remove both atomically.
> 
>     2. We need to decide what to do when a node has the annotation but not the taint. I'd suggest we remove both if EITHER are present and the node is ready.
> 
>     3. We should probably look for a reasonable way to let the user know if we aren't able to remove the taint for some reason. That said, I'm not sure what the best mechanism is for this. If we keep this code here, the obvious way is to set a condition on the operator like "TNFTaintFencedNodesServiceHealthy" = false when you detect an error and then restore on the first successful run.
> 
>     4. As a team, we probably need to discuss whether this (and other TNF-specific-but-not-etcd-specific) code actually belongs in CEO. CEO is really about managing etcd. PacemakerHealthCheck falls into this grey area but it ultimately there to help inform the user about etcd from the perspective of pacemaker. This feature feels like it's here because etcd manages pacemaker and it's the closest thing to a home we have. It's the pragmatic choice, but it does raise questions about where we draw the line

1 - Addressed. Taint and annotations are now removed atomatically in a single node.Update call.
2 - Partially addressed. If the annotation is present, we remove both taint and annotation (even if the taint is already gone). However, if only the taint is present without the annotation, we intentionally skip it since the annotation is our provenance marker, and without it we can't confirm the taint is ours to remove. If we removed any out-of-service taint unconditionally, we'd risk interfering with taints set by other controllers/administrators
3 - Addressed. On failure, we now set a TNFTaintRemovalControllerDegraded=True condition with the error message, following the same pattern used by PacemakerHealthCheck
4 - We'll keep this in CEO for now, however we have started an internal discussion to evaluate different options since there are other TNF-specific areas that might not be 100% related to CEO.

pkg/tnf/pkg/tools/taint.go
vimauro: Done !

pkg/tnf/operator/starter.go
vimauro: Fixed !

pkg/tnf/pkg/tools/taint_test.go
vimauro: Fixed by using` corev1.TaintNodeNotReady`

pkg/tnf/pkg/tools/taint.go
```diff
@@ -0,0 +1,101 @@
...
+		return
+	}
+
+	klog.Infof("node %s has pacemaker annotation, cleaning up out-of-service taint and annotation", node.Name)
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		freshNode, err := kubeClient.CoreV1().Nodes().Get(ctx, node.Name, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get node %s: %w", node.Name, err)
+		}
```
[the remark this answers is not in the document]
vimauro: Fixed

PR #1620, opened 2026-05-15
OCPBUGS-84336: fix(tnf): gate dual-replica setup and keep retrying

jaypoulz:
TNF startup was completely refactored in 4.22. To avoid having to backport that change, I've opted to propose a minimal set of changes to get 4.21 (and 4.20) installing consistentally.

- Start TNF job setup when two nodes have master or control-plane role labels; no node informer UpdateFunc for Ready
- Poll every 10s up to 10m for those nodes to be Ready with role labels and for etcd revision stability before starting job controllers
- Clear TNFJobControllersDegraded when a setup wave succeeds after an earlier failure
- Ignore fencing secret-driven job restarts until ExternalEtcdHasCompletedTransition
- Retry handleNodes with exponential backoff per wave; on wave timeout or unexpected backoff error, set TNFJobControllersDegraded, checkpoint, then start a new wave

### DIFF
--- a/pkg/tnf/operator/starter.go
+++ b/pkg/tnf/operator/starter.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"sync"
@@ -36,6 +37,7 @@ import (
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/ceohelpers"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/externaletcdsupportcontroller"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
+	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/etcd"
 	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/jobs"
 	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/tools"
 )
@@ -46,17 +48,46 @@ var (
 	// fencingUpdateMutex is used to make usage of fencingUpdateTriggered thread safe
 	fencingUpdateMutex sync.Mutex
 
-	// handleNodesFunc is a variable to allow mocking in tests
-	handleNodesFunc = handleNodes
+	handleNodesFunc = handleNodes // test hook
 
-	// retryBackoffConfig allows customizing retry behavior for tests
+	// Per wave: exponential backoff, then Degraded + checkpoint (OCPBUGS-84336).
 	retryBackoffConfig = wait.Backoff{
 		Duration: 5 * time.Second,
 		Factor:   2.0,
-		Steps:    9, // ~10 minutes total: 5s + 10s + 20s + 40s + 80s + 120s + 120s + 120s + 120s
+		Steps:    9,
 		Cap:      2 * time.Minute,
 	}
+
+	// Between waves after backoff exhaustion (OCPBUGS-84336); tests shorten.
+	tnfJobControllerSetupCheckpointInterval = 10 * time.Minute
 )
+
+const (
+	masterNodeLabelKey       = "node-role.kubernetes.io/master"
+	controlPlaneNodeLabelKey = "node-role.kubernetes.io/control-plane"
+)
+
+// hasControlPlaneRoleLabels is true when the node has master or control-plane role labels.
+// Arbiter nodes are not full control-plane nodes and do not carry these labels; other call sites must union arbiter when needed.
+func hasControlPlaneRoleLabels(node *corev1.Node) bool {
+	if node == nil {
+		return false
+	}
+	l := node.Labels
+	_, master := l[masterNodeLabelKey]
+	_, controlPlane := l[controlPlaneNodeLabelKey]
+	return master || controlPlane
+}
+
+func filterControlPlaneNodesForTNF(nodes []*corev1.Node) []*corev1.Node {
+	out := make([]*corev1.Node, 0, len(nodes))
+	for _, n := range nodes {
+		if hasControlPlaneRoleLabels(n) {
+			out = append(out, n)
+		}
+	}
+	return out
+}
 
 // HandleDualReplicaClusters checks feature gate and control plane topology,
 // and handles dual replica aka two node fencing clusters
@@ -112,13 +143,13 @@ func HandleDualReplicaClusters(
 	klog.Infof("watching for secrets...")
 	_, err = kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().Secrets().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			handleFencingSecretChange(ctx, kubeClient, nil, obj)
+			handleFencingSecretChange(ctx, operatorClient, kubeClient, nil, obj)
 		},
 		UpdateFunc: func(oldObj, newObj interface{}) {
-			handleFencingSecretChange(ctx, kubeClient, oldObj, newObj)
+			handleFencingSecretChange(ctx, operatorClient, kubeClient, oldObj, newObj)
 		},
 		DeleteFunc: func(obj interface{}) {
-			handleFencingSecretChange(ctx, kubeClient, nil, obj)
+			handleFencingSecretChange(ctx, operatorClient, kubeClient, nil, obj)
 		},
 	})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jaypoulz:
/payload-job periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-techpreview periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-recovery-techpreview

jaypoulz:
/payload-job periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-techpreview periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-recovery-techpreview

jaypoulz:
/payload-job periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-techpreview periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-recovery-techpreview

pkg/tnf/operator/starter.go
```diff
@@ -146,26 +177,49 @@ func handleAddedNode(
...
-
-		// we can have 2 nodes on the first call of AddFunc already, ensure we create job controllers once only
-		once.Do(func() {
-			// this potentially needs some time when we wait for etcd bootstrap to complete, so run it in a goroutine,
-			// to not block the event handler
-			go handleNodesWithRetry(nodeList, ctx, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces, etcdInformer)
-		})
-	}
+		go handleNodesWithRetry(fresh, ctx, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces, etcdInformer)
+	})
```
fonta-rh, line 214: The re-list inside `once.Do` adds failure modes without adding safety.

`controlPlaneNodeLister.List()` reads from the informer cache (in-memory `ThreadSafeStore`), so a "transient API error" is near-impossible — but `len(fresh) != 2` could theoretically fire if a node label changed between the outer and inner check. Since `sync.Once` is permanently consumed regardless of the closure's return, any early return here permanently deadlocks TNF startup with no Degraded condition set.

The inner re-list is redundant anyway: `handleNodesWithRetry` → `waitForTnfControlPlaneReadyAndStableRevision` already re-validates node readiness, labels, and revision stability with polling and timeout. This matches the original 4.21 pattern (pre-PR) where `once.Do` unconditionally launched the goroutine.

```suggestion
	once.Do(func() {
		go handleNodesWithRetry(nodes, ctx, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces, etcdInformer)
	})
}
```

pkg/tnf/operator/starter.go
```diff
@@ -394,6 +537,18 @@ func handleFencingSecretChange(ctx context.Context, client kubernetes.Interface,
...
 
+	// Defer fencing job restart from secrets until external etcd transition completes (OCPBUGS-84695).
+	transitionComplete, err := ceohelpers.HasExternalEtcdCompletedTransition(ctx, operatorClient)
+	if err != nil {
+		klog.Errorf("failed to check external etcd transition for fencing secret handler: %v", err)
+		return
+	}
+	if !transitionComplete {
+		klog.V(4).Infof("skipping fencing job restart from fencing secret change until %s is True",
+			etcd.OperatorConditionExternalEtcdHasCompletedTransition)
```
fonta-rh: V(4) means this drop is invisible at default log verbosity. If a fencing secret changes during the transition window, the event is silently lost with no replay mechanism. An operator diagnosing "why didn't my fencing config update take effect" would see no evidence.

```suggestion
		klog.Infof("skipping fencing job restart from fencing secret change until %s is True",
			etcd.OperatorConditionExternalEtcdHasCompletedTransition)
```

fonta-rh (commented):
Added a nit and a not-so-nit: I'm not sure a lot of the added code on the once.Do is needed (as explained below). I'll post this a a Comment review, but I think this objection needs a closer look

pkg/tnf/operator/starter.go
jaypoulz: nice catch. it can be hard to keep track of where the checks are done, and this is a classic case of knowing that the logic needs to happen, but losing track that it already did :) appreciate the care in the review!

pkg/tnf/operator/starter.go
jaypoulz: Good catch with this as well. I'll bump it up one level.

jaypoulz:
/payload-job periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-techpreview periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-recovery-techpreview

pkg/tnf/operator/starter.go
fonta-rh: Honestly, I wish I had the time to manually check the logic of all of this. As it is, I do the dive if Claude flags it, which seems like it usually does. Should get better with time, though, so no harm done.

jaypoulz:
/payload-job periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-recovery-techpreview

jaypoulz:
/verified by https://github.com/openshift/cluster-etcd-operator/pull/1620#issuecomment-4508911465 (installs are passing)

jaypoulz:
/verified by https://github.com/openshift/cluster-etcd-operator/pull/1620#issuecomment-4508911465 (installs are passing)

jaypoulz:
the update was just updating the bug reference

pkg/tnf/operator/starter.go
```diff
@@ -260,6 +343,58 @@ func handleNodes(
...
+			if !ready {
+				klog.V(4).Infof("TNF gate: node %q not Ready", ref.Name)
+				return false, nil
+			}
+		}
+
+		stable, err := ceohelpers.IsRevisionStable(operatorClient)
+		if err != nil {
+			klog.Errorf("IsRevisionStable: %v", err)
+			return false, err
```
fonta-rh: Both `IsRevisionStable` and the `Nodes().Get()` non-NotFound path (line 365) return `(false, err)`, which terminates `PollUntilContextTimeout` immediately. The error propagates through `handleNodes` to the outer `ExponentialBackoffWithContext`, burning an entire backoff step (5s → 10s → 20s → …) instead of re-polling in 10 seconds.

`IsRevisionStable` calls `GetStaticPodOperatorState()` which can fail transiently during informer resyncs or API server rollouts. Since the poll is *designed* for transient failures (10s interval, 10m timeout), these should be retryable within the poll — same treatment as `IsNotFound` gets on line 363.

```suggestion
		if err != nil {
			klog.Warningf("IsRevisionStable check failed (will retry): %v", err)
			return false, nil
```

Same for line 365 — consider:
```go
			klog.Warningf("TNF gate: node %q API error (will retry): %v", ref.Name, err)
			return false, nil
```

pkg/tnf/operator/starter.go
```diff
@@ -146,26 +177,41 @@ func handleAddedNode(
...
-		}
-		if len(nodeList) != 2 {
-			klog.Info("not starting TNF jobs yet, waiting for 2 control plane nodes to exist")
-			return
+func tryStartTnfJobControllers(
+	controllerContext *controllercmd.ControllerContext,
+	controlPlaneNodeLister corev1listers.NodeLister,
+	once *sync.Once,
+	ctx context.Context,
+	operatorClient v1helpers.StaticPodOperatorClient,
```
fonta-rh, line 189: nit: If this `UpdateStatus` fails, `TNFJobControllersDegraded=True` stays permanently — `once.Do` prevents re-entry, so nothing will retry the clear. `v1helpers.UpdateStatus` handles conflicts internally so this is unlikely, but a 2-3 attempt retry here would be cheap insurance against a confusing stale-Degraded diagnostic on bare metal.

jaypoulz: I don't really want to change this for tech preview minimal fix to get CI off the floor. Since the flow was completely changed in 4.22+, best to focus these kinds of fixes on the supported versions.

jaypoulz:
/payload-job periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-techpreview periodic-ci-openshift-release-main-nightly-4.21-e2e-metal-ovn-two-node-fencing-recovery-techpreview

jaypoulz:
/verified by https://github.com/openshift/cluster-etcd-operator/pull/1620#issuecomment-4555835162 The clusters installed and had higher-than-normal pass rates for the pre-release TNF.

fonta-rh:
/lgtm
next one is 3.5$

PR #1625, opened 2026-05-29
OCPEDGE-2711: register pacemaker alert agents for fencing events to taint/untaint nodes

vimauro:
## Summary by CodeRabbit

* **New Features**
  * Two pacemaker fencing alert agents are now automatically configured during cluster setup and update; setup/update will abort if alert registration fails. Existing alert definitions are validated and replaced as needed, including verification that alert scripts are present and executable and optional XML select-filters are applied.

* **Tests**
  * Added unit test coverage verifying the two expected alert configurations.

### DIFF
--- a/pkg/tnf/pkg/pcs/alerts.go
+++ b/pkg/tnf/pkg/pcs/alerts.go
@@ -2,7 +2,9 @@ package pcs
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	osexec "os/exec"
 	"strings"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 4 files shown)


pkg/tnf/pkg/pcs/alerts.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to check existing alert %q: %w", ac.id, err)
+	}
+
+	if exists {
+		klog.Infof("Alert %q already exists, deleting for re-creation", ac.id)
+		deleteCmd := fmt.Sprintf("/usr/sbin/pcs alert delete %s", ac.id)
+		if _, stdErr, err := exec.Execute(ctx, deleteCmd); err != nil {
+			return fmt.Errorf("failed to delete existing alert %q: %s: %w", ac.id, stdErr, err)
+		}
```
fonta-rh, line 81: **suggestion:** The delete-then-create pattern has a (small) race window where the alert is absent if `pcs alert create` or `cibadmin --modify` fails after the delete succeeds.

Since `applyAlertSelectFilter` (line 94) already uses `cibadmin --modify` with the full `<alert>` element — which atomically updates the existing alert matched by `id` — you could use the same approach here for existing alerts:

- If the alert exists: `cibadmin --modify` with the full alert XML (path + select + meta) — atomic in-place update, no delete needed
- If it doesn't exist: `pcs alert create` (current behavior)

This eliminates the delete step entirely and collapses the select filter application into the same operation.

fonta-rh (commented):
Cross-repo review of CEO #1625 + MCO #6092 (fencing taint/untaint alert agents). One finding for this PR survived adversarial vetting (four others are on MCO #6092).

pkg/tnf/pkg/pcs/alerts.go
vimauro: Good point, fixed it!

jaypoulz (commented):
I think the other thing we should incorporate as part of this, or as a follow-up PR is integration with the pacemakercluster / health check system. Since we're adding alert agents, those should be visibile and reported on as healthy. These can also contribute to the overall health of the cluster, so users can know if auto-tainting is not working properly. Unlike the other things we bubble up to the cluster-operator, this should be reported but may not be sufficient to mark the operator as degraded. CEO should only be degraded when something poses a risk to etcd. In this case, broken taints probably doesn't, but we should look for waits to raise it to the user through our existing channels.

vimauro:
/verified by @vimauro  by running a TNF 4.22 cluster and running extensive tests, including the install and recovery of Piraeus/Linstor storage solution.
This PR was also tested along with the MCO pr https://github.com/openshift/machine-config-operator/pull/6092

PR #1626, opened 2026-05-31
OCPBUGS-77345: validate snapshot before destructive operations in cluster-restore-tnf.sh

Neilhamza:
## Summary

`cluster-restore-tnf.sh` performs destructive operations (disable etcd, clear CRM attributes, delete `/var/lib/etcd`) before attempting `etcdutl snapshot restore`. If the restore fails, the cluster is left in an unrecoverable state with no rollback.

This adds a dry-run restore to a temporary directory before any destructive operations, and passes `--skip-hash-check` on both the dry-run and the real restore so learner snapshots (hash-less raw BoltDB copies from `podman-etcd`) are accepted alongside standard `etcdctl snapshot save` snapshots.

## Bugs

- [OCPBUGS-77345](https://issues.redhat.com/browse/OCPBUGS-77345) — Pre-validate snapshot integrity before destructive operations to prevent unrecoverable cluster state.
- [OCPBUGS-79664](https://issues.redhat.com/browse/OCPBUGS-79664) — No rollback on restore failure leaves TNF cluster broken (duplicate of 77345).
- [OCPBUGS-79662](https://issues.redhat.com/browse/OCPBUGS-79662) — Missing `--skip-hash-check` causes restore to reject valid learner snapshots on etcd v3.6.

## Related

- [OCPEDGE-2299](https://issues.redhat.com/browse/OCPEDGE-2299) — Parent story: Backup and Restore testing (where all three bugs were discovered).
- RHEL-145622, RHEL-145626, RHEL-145627, [RHEL-145628](https://redhat.atlassian.net/browse/RHEL-145628) — Upstream resource-agents fixes to produce hash-compatible learner snapshots.

## Testing

Verified on OCP 4.22.0-0.nightly-2026-05-28-111510, TNF topology (2 masters, dev-scripts IPI):

| Test | Snapshot Type | Expected | Result |
|------|--------------|----------|--------|
| Corrupt snapshot | `dd if=/dev/urandom` | Abort before destructive ops | PASS — caught at `check_snapshot_status`, etcd stayed running |
| Learner snapshot (hash-less) | `podman-etcd` raw copy from rejoin | Dry-run passes, restore succeeds | PASS — both dry-run and real restore succeeded |
| Primary snapshot (hash-bearing) | `etcdctl snapshot save` | Dry-run passes, restore succeeds | PASS — `--skip-hash-check` is a no-op when hash is present |

## Summary by CodeRabbit

* **Bug Fixes**
  * Added pre-restore snapshot validation: snapshots are first restored in a temporary environment to verify integrity before any destructive cluster actions. Validation failures abort immediately with clear error messaging to prevent data loss.
  * Actual restore still fails fast on restore errors to preserve cluster safety.

### DIFF
--- a/bindata/etcd/cluster-restore-tnf.sh
+++ b/bindata/etcd/cluster-restore-tnf.sh
@@ -204,6 +204,8 @@ echo "starting snapshot restore through etcdctl..."
 # We are never going to rev-bump here to ensure we don't cause a revision split between the
 # remainder of the running cluster and this restore member. Imagine your non-restore quorum members run at rev 100,
 # we would attempt to rev bump this with snapshot at rev 120, now this member is 20 revisions ahead and RAFT is confused.
+# --skip-hash-check: learner snapshots from podman-etcd lack the trailing
+# integrity hash that etcdctl snapshot save appends (OCPBUGS-79662).
 if ! ${ETCD_CLIENT} snapshot restore "${SNAPSHOT_FILE}" --skip-hash-check "${ETCDCTL_RESTORE_FLAGS[@]}"; then
     echo "Snapshot restore failed. Aborting!"
     exit 1
--- a/bindata/etcd/cluster-restore-tnf.sh
+++ b/bindata/etcd/cluster-restore-tnf.sh
@@ -163,6 +163,22 @@ if [ -n "${ETCD_ETCDUTL_BIN}" ]; then
   ETCD_CLIENT="${ETCD_ETCDUTL_BIN}"
 fi
 
+# Validate that the snapshot can be restored before performing any destructive
+# operations. A failed dry-run here is safe — etcd is still running, no state
+# has been modified.
+RESTORE_DRYRUN_DIR=$(mktemp -d)
+trap "rm -rf ${RESTORE_DRYRUN_DIR}" EXIT
+echo "validating snapshot restorability..."
+if ! ${ETCD_CLIENT} snapshot restore "${SNAPSHOT_FILE}" \
+    --data-dir="${RESTORE_DRYRUN_DIR}/data" \
+    --skip-hash-check 2>&1; then
+  rm -rf "${RESTORE_DRYRUN_DIR}"
+  echo "Snapshot validation failed: the snapshot cannot be restored."
+  echo "Aborting before any cluster modifications."
+  exit 1
+fi
+rm -rf "${RESTORE_DRYRUN_DIR}"
+
 if [ ! -d "${ETCD_DATA_DIR_BACKUP}" ]; then
   mkdir -p "${ETCD_DATA_DIR_BACKUP}"
 fi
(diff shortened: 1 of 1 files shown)


bindata/etcd/cluster-restore-tnf.sh
```diff
@@ -163,6 +163,22 @@ if [ -n "${ETCD_ETCDUTL_BIN}" ]; then
...
 
+# Validate that the snapshot can be restored before performing any destructive
+# operations. A failed dry-run here is safe — etcd is still running, no state
+# has been modified.
+RESTORE_DRYRUN_DIR=$(mktemp -d)
+trap "rm -rf ${RESTORE_DRYRUN_DIR}" EXIT
+echo "validating snapshot restorability..."
+if ! ${ETCD_CLIENT} snapshot restore "${SNAPSHOT_FILE}" \
+    --data-dir="${RESTORE_DRYRUN_DIR}/data" \
+    --skip-hash-check 2>&1; then
```
fonta-rh, line 174: **Question:** `check_snapshot_status` (called at line 159) runs `etcdutl snapshot status` **without** `--skip-hash-check`. When OCP moves to etcd v3.6 (where `--skip-hash-check` defaults to `false`), will `snapshot status` also reject hash-less learner snapshots?

If yes, learner snapshots would be rejected at `check_snapshot_status` before this dry-run ever runs — making the dry-run + `--skip-hash-check` dead code for that exact scenario.

Did you test this against etcd v3.6 specifically, or was the 4.22 nightly still on etcd 3.5.x? If `snapshot status` does need the flag too, that's a change in `etcd-common-tools` and would affect `cluster-restore.sh` callers as well.

bindata/etcd/cluster-restore-tnf.sh
```diff
@@ -188,7 +204,7 @@ echo "starting snapshot restore through etcdctl..."
 # We are never going to rev-bump here to ensure we don't cause a revision split between the
 # remainder of the running cluster and this restore member. Imagine your non-restore quorum members run at rev 100,
 # we would attempt to rev bump this with snapshot at rev 120, now this member is 20 revisions ahead and RAFT is confused.
-if ! ${ETCD_CLIENT} snapshot restore "${SNAPSHOT_FILE}" "${ETCDCTL_RESTORE_FLAGS[@]}"; then
+if ! ${ETCD_CLIENT} snapshot restore "${SNAPSHOT_FILE}" --skip-hash-check "${ETCDCTL_RESTORE_FLAGS[@]}"; then
```
fonta-rh, line 209: nit: The PR description explains the `--skip-hash-check` rationale really well — worth capturing a one-liner here too so it travels with the code. Something like:

```bash
# --skip-hash-check: learner snapshots from podman-etcd lack the trailing
# integrity hash that etcdctl snapshot save appends (OCPBUGS-79662).
```

Helps future readers understand this isn't a workaround to remove later.

bindata/etcd/cluster-restore-tnf.sh
Neilhamza: Good question — tested on **etcd v3.6.5** (OCP 4.22.0-0.nightly-2026-05-28-111510). `snapshot status` does **not** reject hash-less learner snapshots. Here's the actual output from the learner snapshot test:

```
etcdctl version: 3.6.5
API version: 3.6
{"hash":2453523510,"revision":1000048799,"totalKey":11158,"totalSize":89800704,"version":"3.6.0"}
validating snapshot restorability...
```

`check_snapshot_status` passed, then the dry-run ran and also passed with `--skip-hash-check`.

The key distinction: the `hash` field in `snapshot status` output is the **BoltDB internal CRC** — computed from the database pages, always present. The hash that `--skip-hash-check` controls is the **trailing SHA-256 integrity hash** that `etcdctl snapshot save` appends to the end of the file. These are two different things. `snapshot status` reads BoltDB metadata and never checks the trailing hash. Only `snapshot restore` checks it.

So `check_snapshot_status` is fine as-is, and the dry-run is not dead code — it catches failures that `snapshot status` cannot (e.g., a valid BoltDB that fails WAL generation during restore).

bindata/etcd/cluster-restore-tnf.sh
Neilhamza: Good call — added in 59040935c. Thanks for the review!

PR #1630, opened 2026-06-10
 OCPEDGE-2736: feat: support MAC-based fencing secrets in update-fencing-credentials.sh

fracappa:
The installer can now create fencing secrets keyed by MAC address (fencing-credentials-<sha256>) instead of hostname. Add an auto-discovery step which would look for matching fencing credentials secret if none is found by --node parameter. Firstly, it looks for K8s Nodes tnf.openshift.io/mac-addresses annotation, then use a fallback based on the --address parameter.

## Summary by CodeRabbit

* **New Features**
  * Enhanced fencing secret discovery to dynamically select the correct Kubernetes secret (tries hostname-based name first, then MAC-normalized SHA-256 derived candidates, and finally falls back to matching the stonith device address against existing fencing secrets).
  * Validates that the configured IP/URI exactly matches the detected Redfish target before proceeding, and updates the detected secret (no longer relies on a fixed secret name).

* **Documentation**
  * Updated the fencing credential rotation guide to reflect the revised detection workflow and the idempotent secret update approach in the `openshift-etcd` namespace.

### DIFF
--- a/bindata/etcd/update-fencing-credentials.sh
+++ b/bindata/etcd/update-fencing-credentials.sh
@@ -29,6 +29,74 @@ function usage {
     exit 1
 }
 
+# mirrors hashMAC in pkg/tnf/pkg/tools/mac.go
+function normalize_and_hash_mac {
+    local mac="$1"
+    local normalized
+    normalized=$(echo "${mac}" | tr '[:upper:]' '[:lower:]' | tr -d ':-')
+    if ! [[ "${normalized}" =~ ^[0-9a-f]{12}$ ]]; then
+        echo "Invalid MAC address: ${mac}" >&2
+        return 1
+    fi
+    echo -n "${normalized}" | sha256sum | awk '{print $1}'
+}
+
+# mirrors Phases 1-2 of GetFencingSecrets in pkg/tnf/pkg/tools/secrets.go;
+# Phase 3 uses stonith device address matching instead of Redfish UUID query
+function detect_fencing_secret {
+    local node="$1"
+    local namespace="openshift-etcd"
+    local mac_annotation_key="tnf.openshift.io/mac-addresses"
+
+    if ! oc get namespace "${namespace}" &>/dev/null; then
+        echo "ERROR: Cannot reach Kubernetes API server" >&2
+        return 1
+    fi
+
+    # Phase 1: hostname-based lookup
+    local hostname_secret="fencing-credentials-${node}"
+    if oc get secret "${hostname_secret}" --namespace="${namespace}" &>/dev/null; then
+        echo "${hostname_secret}"
+        return 0
+    fi
+
+    # Phase 2: MAC hash lookup from node annotation
+    local mac_annotation
+    mac_annotation=$(oc get node "${node}" -o jsonpath="{.metadata.annotations.${mac_annotation_key//\./\\.}}" 2>/dev/null) || true
+    if [ -n "${mac_annotation}" ]; then
+        IFS=',' read -ra macs <<< "${mac_annotation}"
+        for mac in "${macs[@]}"; do
+            mac=$(echo "${mac}" | tr -d '[:space:]')
+            [ -z "${mac}" ] && continue
+            local hash
+            hash=$(normalize_and_hash_mac "${mac}") || continue
+            local mac_secret="fencing-credentials-${hash}"
+            if oc get secret "${mac_secret}" --namespace="${namespace}" &>/dev/null; then
+                echo "${mac_secret}"
+                return 0
+            fi
+        done
+    fi
+
+    # Phase 3: targeted address matching (no bulk secret dump)
+    local stonith_ip="$2"
+    local stonith_uri="$3"
+    if [ -n "${stonith_ip}" ] && [ -n "${stonith_uri}" ]; then
+        local name addr
+        for name in $(oc get secrets -n "${namespace}" -o name 2>/dev/null | grep fencing-credentials-); do
+            addr=$(oc get "${name}" -n "${namespace}" -o jsonpath='{.data.address}' 2>/dev/null | base64 -d) || continue
+            if [[ "${addr}" == *"://${stonith_ip}:"* ]] || [[ "${addr}" == *"://[${stonith_ip}]:"* ]]; then
+                if [[ "${addr}" == *"${stonith_uri}" ]]; then
+                    echo "${name#secret/}"
+                    return 0
+                fi
+            fi
+        done
+    fi
+
+    return 1
+}
+
 function redact_credentials {
     local input
     input=$(cat)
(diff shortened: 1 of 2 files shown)


bindata/etcd/update-fencing-credentials.sh
```diff
@@ -99,6 +165,27 @@ fi
...
+
+echo "Detecting fencing secret for node ${NODE}..."
+if ! SECRET_NAME=$(detect_fencing_secret "${NODE}" "${DEVICE_ID}"); then
+    echo "No fencing secret found for node ${NODE} (tried hostname, MAC hash, and stonith device address lookup)"
+    exit 1
+fi
+echo "Detected fencing secret: ${SECRET_NAME}"
+
+STONITH_IP=$(pcs stonith config "${DEVICE_ID}" 2>/dev/null | sed -n 's/.*[[:space:]]ip=\([^[:space:]]*\).*/\1/p' | head -1)
+STONITH_URI=$(pcs stonith config "${DEVICE_ID}" 2>/dev/null | sed -n 's/.*[[:space:]]systems_uri=\([^[:space:]]*\).*/\1/p' | head -1)
```
fonta-rh: **Bug: `errexit + pipefail` crash** — these two lines crash the script if the stonith device doesn't exist.

The script runs with `set -o errexit` + `set -o pipefail` (lines 5-6). If `pcs stonith config` returns non-zero (device not created yet, pcsd down), `pipefail` propagates the exit code and `errexit` kills the script immediately — before reaching the `[ -n "${STONITH_IP}" ]` guard on line 179 that was written to handle exactly this case.

The identical pattern inside `detect_fencing_secret` (lines 79-80) is safe because it runs under `if !` on line 171, which suppresses `errexit` per POSIX. And the `pcs stonith update` on line 213 already has `|| true` for the same reason. These two lines just missed it.

**When it crashes:** Phase 1 or Phase 2 finds the secret (so `detect_fencing_secret` succeeds), but the stonith device hasn't been created yet (operator hasn't reconciled). Script crashes silently at line 177 with no error message.

**Fix:** append `|| true` to both lines:
```bash
STONITH_IP=$(pcs stonith config "${DEVICE_ID}" 2>/dev/null | sed -n 's/.*[[:space:]]ip=\([^[:space:]]*\).*/\1/p' | head -1) || true
STONITH_URI=$(pcs stonith config "${DEVICE_ID}" 2>/dev/null | sed -n 's/.*[[:space:]]systems_uri=\([^[:space:]]*\).*/\1/p' | head -1) || true
```

bindata/etcd/update-fencing-credentials.sh
```diff
@@ -29,6 +29,72 @@ function usage {
...
+        local stonith_ip stonith_uri
+        stonith_ip=$(pcs stonith config "${device_id}" 2>/dev/null | sed -n 's/.*[[:space:]]ip=\([^[:space:]]*\).*/\1/p' | head -1)
+        stonith_uri=$(pcs stonith config "${device_id}" 2>/dev/null | sed -n 's/.*[[:space:]]systems_uri=\([^[:space:]]*\).*/\1/p' | head -1)
+        if [ -n "${stonith_ip}" ] && [ -n "${stonith_uri}" ]; then
+            local match
+            match=$(oc get secrets --namespace="${namespace}" -o json | \
+                jq -r --arg ip "${stonith_ip}" --arg uri "${stonith_uri}" \
+                '.items[] | select(.metadata.name | startswith("fencing-credentials-")) |
+                 select((.data.address // "") | @base64d | (contains($ip) and contains($uri))) |
+                 .metadata.name' 2>/dev/null | head -1)
```
fonta-rh: **Bug: `contains()` does substring matching on IP — can match the wrong secret.**

jq's `contains()` is substring-based, so `contains("192.168.1.1")` also matches addresses containing `192.168.1.10`, `192.168.1.100`, etc. Sequential BMC IPs are the norm in bare-metal clusters.

The `contains($uri)` check doesn't help much either — Dell iDRAC (and similar BMCs) all use the same fixed `systems_uri` like `/redfish/v1/Systems/System.Embedded.1`, so the URI won't disambiguate.

With `head -1`, the script picks whichever secret `oc get secrets` returns first, which may be the wrong one. The wrong secret gets updated, causing silent credential corruption.

**Suggestion:** Parse the address field and compare IP and URI exactly instead of using substring matching. For example:
```jq
select((.data.address // "") | @base64d |
  capture("https?://(?<h>[^:/]+)(?::(?<p>[0-9]+))?(?<u>/.*)") |
  .h == $ip and .u == $uri)
```
Or at minimum, anchor the IP match to avoid substring collisions (e.g., check that the IP appears between `://` and `:` or `/`).

bindata/etcd/update-fencing-credentials.sh
```diff
@@ -29,6 +29,72 @@ function usage {
...
+
+# mirrors GetFencingSecrets in pkg/tnf/pkg/tools/secrets.go
+function detect_fencing_secret {
+    local node="$1"
+    local namespace="openshift-etcd"
+    local mac_annotation_key="tnf.openshift.io/mac-addresses"
+
+    # Phase 1: hostname-based lookup
+    local hostname_secret="fencing-credentials-${node}"
+    if oc get secret "${hostname_secret}" --namespace="${namespace}" &>/dev/null; then
```
fonta-rh, line 58: nit: All `oc` errors in the three-phase fallback are silently swallowed (`&>/dev/null` here, `2>/dev/null || true` on line 59). If the API server is temporarily unreachable, the user sees "No fencing secret found" — implying the secret doesn't exist — when the real issue is API connectivity.

A lightweight pre-check before entering the fallback would help:
```bash
if ! oc get namespace "${namespace}" &>/dev/null; then
    echo "ERROR: Cannot reach Kubernetes API server" >&2
    return 1
fi
```

bindata/etcd/update-fencing-credentials.sh
```diff
@@ -29,6 +29,72 @@ function usage {
...
+    local normalized
+    normalized=$(echo "${mac}" | tr '[:upper:]' '[:lower:]' | tr -d ':-')
+    if ! [[ "${normalized}" =~ ^[0-9a-f]{12}$ ]]; then
+        echo "Invalid MAC address: ${mac}" >&2
+        return 1
+    fi
+    echo -n "${normalized}" | sha256sum | awk '{print $1}'
+}
+
+# mirrors GetFencingSecrets in pkg/tnf/pkg/tools/secrets.go
```
fonta-rh: nit: This comment is accurate for Phases 1-2 but not for Phase 3. Go's Phase 3 (`matchRedfishUUIDsToSecrets`) queries each secret's Redfish endpoint for a system UUID and matches against `node.status.nodeInfo.systemUUID`. This Phase 3 does stonith device address matching instead — an intentional divergence per the Jira ticket, but the comment doesn't reflect it.

Suggestion:
```bash
# mirrors Phases 1-2 of GetFencingSecrets in pkg/tnf/pkg/tools/secrets.go;
# Phase 3 uses stonith device address matching instead of Redfish UUID query
```

fonta-rh (changes requested):
Thanks for this — the three-phase fallback design is solid and the approach is well motivated by the Jira ticket. Two issues need fixing before merge, plus two minor nits.

**Blockers (2):**
1. `errexit + pipefail` crash on lines 177-178 — script dies before reaching the empty-value guard
2. Phase 3 `contains()` does substring matching on IP — `192.168.1.1` falsely matches `192.168.1.10`

**Nits (2):**
3. API errors silently swallowed in detect_fencing_secret — misleading "no secret found" when API is unreachable
4. "mirrors GetFencingSecrets" comment is inaccurate for Phase 3

bindata/etcd/update-fencing-credentials.sh
fracappa: That's true, thanks for noticing this! :-)

bindata/etcd/update-fencing-credentials.sh
```diff
@@ -99,6 +165,27 @@ fi
...
+NAMESPACE="openshift-etcd"
+
+echo "Detecting fencing secret for node ${NODE}..."
+if ! SECRET_NAME=$(detect_fencing_secret "${NODE}" "${DEVICE_ID}"); then
+    echo "No fencing secret found for node ${NODE} (tried hostname, MAC hash, and stonith device address lookup)"
+    exit 1
+fi
+echo "Detected fencing secret: ${SECRET_NAME}"
+
+STONITH_IP=$(pcs stonith config "${DEVICE_ID}" 2>/dev/null | sed -n 's/.*[[:space:]]ip=\([^[:space:]]*\).*/\1/p' | head -1)
```
dhensel-rh: Why is this being called again when it was called at line 79 ? I know they are locally scoped the first time. Just curious if this can be done in the main just once and reused ?

Actually could `pcs stonith config` only be called once, and then the variable parsed ? 


```suggestion
  STONITH_CFG=$(pcs stonith config "${DEVICE_ID}" 2>/dev/null || true)

  # Extract once
  STONITH_IP=$(echo "${STONITH_CFG}" | sed -n 's/.*[[:space:]]ip=\([^[:space:]]*\).*/\1/p' | head -1)
  STONITH_URI=$(echo "${STONITH_CFG}" | sed -n 's/.*[[:space:]]systems_uri=\([^[:space:]]*\).*/\1/p' | head -1)

```

bindata/etcd/update-fencing-credentials.sh
dhensel-rh: Why is this being called again when it was called at line 80 ? I know they are locally scoped the first time. Just curious if this can be done in the main just once and reused ?

fracappa:
@dhensel-rh thanks for your inputs. It's rather helpful to reduce pcs interactions in order to be faster. 
I should have applied the edits you suggested.

fracappa:
@fonta-rh thanks for your review! I should have addressed your requests but if I missed something let me know.

fonta-rh (approved):
All four findings from the previous review are addressed:

1. **errexit+pipefail crash** — Stonith device pre-check + single `pcs stonith config` call with `|| true` eliminates the crash path
2. **`contains()` IP substring match** — Phase 3 rewritten with anchored bash pattern matching (`://${ip}:` / `://[${ip}]:`) — prevents `192.168.1.1` matching `192.168.1.10`
3. **API errors silently swallowed** — namespace pre-check added
4. **Misleading comment** — updated to note Phase 3 divergence

The single-call stonith config and per-phase failure diagnostics are nice additions.

/lgtm

PR #1634, opened 2026-06-18
OCPEDGE-2705: Add TNF Pacemaker metrics controller

fonta-rh:
## Summary

- Adds a metrics controller that projects PacemakerCluster CR conditions into Prometheus gauges (20 metric families, ~53 series on a 2-node cluster)
- Shares the existing `pacemakerInformer` with HealthCheck — zero additional API server load
- Uses table-driven sync logic with `GaugeVec.Reset()` to prevent stale data from removed nodes

## Details

New package `pkg/tnf/pkg/metriccontroller/` with three files:

| File | Purpose |
|------|---------|
| `metrics.go` | 20 metric definitions (3 cluster, 9 node, 8 resource), mapping tables, `conditionToFloat64` helper |
| `controller.go` | Controller struct, constructor with `metrics.KubeRegistry` injection, sync logic |
| `controller_test.go` | 11 test cases covering healthy cluster, node offline, resource stopped, missing CR, nil nodes, missing condition, node removal, multi-resource, label correctness, idempotency, and edge cases |

Wiring in `pkg/tnf/operator/starter.go` — controller created after HealthCheck, sharing `pacemakerInformer`.

### Key design decisions

- **Registry injection** for test isolation (matches `EtcdCertSignerController` pattern)
- **Missing CR → gauges reset to 0** — makes "unhealthy" the safe default for alerting rules (`== 0` vs `absent()`)
- **No operator condition** — failure modes overlap with HealthCheck; self-health metric deferred to [OCPEDGE-2707](https://redhat.atlassian.net/browse/OCPEDGE-2707)
- **`StabilityLevel: metrics.ALPHA`** — explicit on all metrics

### Metrics exposed

| Level | Count | Type | Labels |
|-------|-------|------|--------|
| Cluster | 3 | `Gauge` | none |
| Node | 9 | `GaugeVec` | `node` |
| Resource | 8 | `GaugeVec` | `node`, `resource` |

## Test plan

- [x] `go build ./...` — all binaries compile
- [x] `go test ./pkg/tnf/pkg/metriccontroller/ -v -count=5` — 55/55 pass, no registration pollution
- [x] `go test ./pkg/tnf/... -count=1` — 454 existing TNF tests pass (no regressions)
- [x] `make build` — clean
- [x] `make test-unit` — all tests pass with `-race`
- [x] E2E: Deployed on fencing-IPI cluster, 53 series confirmed on `:8443/metrics` and in Prometheus

### E2E validation (11 scenarios on live cluster)

17/20 metrics dynamically validated via reversible `pcs` operations (maintenance mode, unmanage, disable, standby, fence disable, cluster stop, force-stop, double-ban, virsh destroy, kubelet kill). 3 remaining are not dynamically testable: `node_ready` (too-transient, 1-3s vs 60s poll), `node_clean` (API blind spot — quorum loss blocks CR writes), `cluster_node_count_as_expected` (structurally unreachable — configured membership fixed at 2). All code paths covered by unit tests.

## Related

- **Jira**: [OCPEDGE-2705](https://issues.redhat.com/browse/OCPEDGE-2705)
- **Epic**: [OCPEDGE-2091](https://issues.redhat.com/browse/OCPEDGE-2091) (TNF monitoring)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for `PacemakerCluster`, including cluster condition gauges, per-node online status, and per-resource state.
  * Metrics are maintained by a dedicated controller and reset appropriately when cluster or node details are unavailable.

* **Tests**
  * Added unit tests validating metric outputs across healthy, unhealthy/offline, stopped/started, missing data, label correctness, node removal, and Sync idempotency.
  * Updated health check tests to use shared fake informer utilities.

### DIFF
--- a/pkg/tnf/internal/testutil/fakeinformer.go
+++ b/pkg/tnf/internal/testutil/fakeinformer.go
@@ -1,0 +1,55 @@
+package testutil
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	pacmkrv1 "github.com/openshift/api/etcd/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+func CreateFakeInformer(obj *pacmkrv1.PacemakerCluster) cache.SharedIndexInformer {
+	store := cache.NewStore(func(obj any) (string, error) {
+		if pc, ok := obj.(*pacmkrv1.PacemakerCluster); ok {
+			return pc.Name, nil
+		}
+		return "", fmt.Errorf("object is not a PacemakerCluster")
+	})
+	if obj != nil {
+		_ = store.Add(obj)
+	}
+	return &fakeInformer{store: store}
+}
+
+type fakeInformer struct {
+	store cache.Store
+}
+
+func (f *fakeInformer) AddEventHandler(handler cache.ResourceEventHandler) (cache.ResourceEventHandlerRegistration, error) {
+	return nil, nil
+}
+func (f *fakeInformer) AddEventHandlerWithResyncPeriod(handler cache.ResourceEventHandler, resyncPeriod time.Duration) (cache.ResourceEventHandlerRegistration, error) {
+	return nil, nil
+}
+func (f *fakeInformer) AddEventHandlerWithOptions(handler cache.ResourceEventHandler, options cache.HandlerOptions) (cache.ResourceEventHandlerRegistration, error) {
+	return nil, nil
+}
+func (f *fakeInformer) RemoveEventHandler(handle cache.ResourceEventHandlerRegistration) error {
+	return nil
+}
+func (f *fakeInformer) GetStore() cache.Store           { return f.store }
+func (f *fakeInformer) GetController() cache.Controller { return nil }
+func (f *fakeInformer) Run(stopCh <-chan struct{})      {}
+func (f *fakeInformer) RunWithContext(ctx context.Context) {
+}
+func (f *fakeInformer) HasSynced() bool                                            { return true }
+func (f *fakeInformer) LastSyncResourceVersion() string                            { return "" }
+func (f *fakeInformer) SetWatchErrorHandler(handler cache.WatchErrorHandler) error { return nil }
+func (f *fakeInformer) SetWatchErrorHandlerWithContext(handler cache.WatchErrorHandlerWithContext) error {
+	return nil
+}
+func (f *fakeInformer) SetTransform(handler cache.TransformFunc) error { return nil }
+func (f *fakeInformer) IsStopped() bool                                { return false }
+func (f *fakeInformer) AddIndexers(indexers cache.Indexers) error      { return nil }
+func (f *fakeInformer) GetIndexer() cache.Indexer                      { return nil }
--- a/pkg/tnf/operator/starter.go
+++ b/pkg/tnf/operator/starter.go
@@ -24,6 +24,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/cluster-etcd-operator/bindata"
@@ -32,6 +34,7 @@ import (
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/externaletcdsupportcontroller"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
 	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/jobs"
+	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/metriccontroller"
 	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/pacemaker"
 	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/tools"
 )
@@ -313,6 +316,16 @@ func runPacemakerControllers(ctx context.Context, controllerContext *controllerc
 		// Start the healthcheck controller
 		klog.Infof("starting Pacemaker healthcheck controller")
 		go healthCheckController.Run(ctx, 1)
+
+		// Create and start the metrics controller, sharing the same informer
+		klog.Infof("creating Pacemaker metrics controller")
+		metricsController := metriccontroller.NewPacemakerMetricsController(
+			pacemakerInformer,
+			controllerContext.EventRecorder,
+			legacyregistry.DefaultGatherer.(metrics.KubeRegistry),
+		)
+		klog.Infof("starting Pacemaker metrics controller")
+		go metricsController.Run(ctx, 1)
 
 		// Start the status collector CronJob
 		klog.Infof("starting Pacemaker status collector CronJob")
--- a/pkg/tnf/pkg/metriccontroller/controller.go
+++ b/pkg/tnf/pkg/metriccontroller/controller.go
@@ -1,0 +1,121 @@
+package metriccontroller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/component-base/metrics"
+	"k8s.io/klog/v2"
+
+	pacmkrv1 "github.com/openshift/api/etcd/v1"
+	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/pacemaker"
+)
+
+const resyncInterval = 30 * time.Second
+
+type pacemakerMetricsController struct {
+	pacemakerInformer cache.SharedIndexInformer
+}
+
+func NewPacemakerMetricsController(
+	pacemakerInformer cache.SharedIndexInformer,
+	recorder events.Recorder,
+	metricsRegistry metrics.KubeRegistry,
+) factory.Controller {
+	c := &pacemakerMetricsController{
+		pacemakerInformer: pacemakerInformer,
+	}
+
+	for _, g := range clusterGauges {
+		metricsRegistry.MustRegister(g)
+	}
+	for _, gv := range nodeGaugeVecs {
+		metricsRegistry.MustRegister(gv)
+	}
+	for _, gv := range resourceGaugeVecs {
+		metricsRegistry.MustRegister(gv)
+	}
+
+	return factory.New().
+		ResyncEvery(resyncInterval).
+		WithSync(c.sync).
+		WithInformers(pacemakerInformer).
+		ToController("PacemakerMetricsController", recorder)
+}
+
+func (c *pacemakerMetricsController) sync(ctx context.Context, _ factory.SyncContext) error {
+	klog.V(4).Infof("syncing Pacemaker metrics")
+
+	item, exists, err := c.pacemakerInformer.GetStore().GetByKey(pacemaker.PacemakerClusterResourceName)
+	if err != nil {
+		return err
+	}
+
+	if !exists {
+		klog.V(4).Infof("PacemakerCluster CR not found, resetting metrics")
+		resetAllMetrics()
+		return nil
+	}
+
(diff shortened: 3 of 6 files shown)


pkg/tnf/operator/starter.go
```diff
@@ -314,6 +317,16 @@ func runPacemakerControllers(ctx context.Context, controllerContext *controllerc
...
 
+		// Create and start the metrics controller, sharing the same informer
+		klog.Infof("creating Pacemaker metrics controller")
+		metricsController := metriccontroller.NewPacemakerMetricsController(
+			pacemakerInformer,
+			controllerContext.EventRecorder,
+			legacyregistry.DefaultGatherer.(metrics.KubeRegistry),
+		)
+		klog.Infof("starting Pacemaker metrics controller")
+		go metricsController.Run(ctx, 1)
```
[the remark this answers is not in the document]
fonta-rh: Won't fix — matches existing codebase pattern. `pkg/operator/starter.go:440` uses the identical bare assertion. `DefaultGatherer` is always a `*kubeRegistry` — this cannot fail at runtime.

pkg/tnf/internal/testutil/fakeinformer.go
```diff
@@ -0,0 +1,55 @@
...
+func CreateFakeInformer(obj *pacmkrv1.PacemakerCluster) cache.SharedIndexInformer {
+	store := cache.NewStore(func(obj any) (string, error) {
+		if pc, ok := obj.(*pacmkrv1.PacemakerCluster); ok {
+			return pc.Name, nil
+		}
+		return "", fmt.Errorf("object is not a PacemakerCluster")
+	})
+	if obj != nil {
+		_ = store.Add(obj)
+	}
```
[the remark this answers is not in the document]
fonta-rh: Won't fix — error is unreachable. The function signature constrains input to `*pacmkrv1.PacemakerCluster`, which matches the key function's type assertion. Propagating an impossible error to 5 call sites adds ceremony for zero benefit.

pkg/tnf/pkg/pacemaker/healthcheck_test.go
```diff
@@ -45,66 +44,18 @@ func createTestHealthCheck() *HealthCheck {
...
-
 // createTestHealthCheckWithMockStatus creates a HealthCheck with a mocked Pacemaker CR in the informer cache
 func createTestHealthCheckWithMockStatus(t *testing.T, mockStatus *pacmkrv1.PacemakerCluster) *HealthCheck {
 	return &HealthCheck{
 		operatorClient:    createFakeOperatorClient(),
 		kubeClient:        fake.NewSimpleClientset(),
 		eventRecorder:     events.NewInMemoryRecorder("test", clock.RealClock{}),
-		pacemakerInformer: createFakeInformer(mockStatus),
+		pacemakerInformer: testutil.CreateFakeInformer(mockStatus),
 		recordedEvents:    make(map[string]time.Time),
```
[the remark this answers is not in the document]
fonta-rh: Won't fix — coupling is already explicit. Every test fixture in `healthcheck_test.go` sets `Name: "cluster"`. Adding a DeepCopy + default-name guard protects against a scenario no test exercises.

pkg/tnf/pkg/metriccontroller/metrics.go
```diff
@@ -0,0 +1,193 @@
...
+	{pacmkrv1.ResourceInServiceConditionType, resourceInServiceGauge},
+	{pacmkrv1.ResourceManagedConditionType, resourceManagedGauge},
+	{pacmkrv1.ResourceEnabledConditionType, resourceEnabledGauge},
+	{pacmkrv1.ResourceOperationalConditionType, resourceOperationalGauge},
+	{pacmkrv1.ResourceActiveConditionType, resourceActiveGauge},
+	{pacmkrv1.ResourceStartedConditionType, resourceStartedGauge},
+	{pacmkrv1.ResourceSchedulableConditionType, resourceSchedulableGauge},
+}
+
+func conditionToFloat64(cond *metav1.Condition) float64 {
```
jaypoulz, line 228: One thing to note is that this choice eliminates the distinction between status unknown (nil) and status unhealthy (0).

That said, I'm pretty sure this choice was made at the API level as well. The only time I'm not sure this should apply is during initial setup, before the pacemakercluster CR is populated.

pkg/tnf/pkg/metriccontroller/controller.go
```diff
@@ -0,0 +1,121 @@
...
+		WithSync(c.sync).
+		WithInformers(pacemakerInformer).
+		ToController("PacemakerMetricsController", recorder)
+}
+
+func (c *pacemakerMetricsController) sync(ctx context.Context, _ factory.SyncContext) error {
+	klog.V(4).Infof("syncing Pacemaker metrics")
+
+	item, exists, err := c.pacemakerInformer.GetStore().GetByKey(pacemaker.PacemakerClusterResourceName)
+	if err != nil {
```
jaypoulz, line 54: should we log this error?

pkg/tnf/operator/starter.go
```diff
@@ -314,6 +317,16 @@ func runPacemakerControllers(ctx context.Context, controllerContext *controllerc
 		klog.Infof("starting Pacemaker healthcheck controller")
 		go healthCheckController.Run(ctx, 1)
 
+		// Create and start the metrics controller, sharing the same informer
+		klog.Infof("creating Pacemaker metrics controller")
+		metricsController := metriccontroller.NewPacemakerMetricsController(
+			pacemakerInformer,
+			controllerContext.EventRecorder,
+			legacyregistry.DefaultGatherer.(metrics.KubeRegistry),
```
jaypoulz, line 325: Is default gatherer guaranteed to be a KubeRegistry? if not, it might be worth check if it is first, so that we can catch the error and log it.

pkg/tnf/pkg/metriccontroller/metrics.go
```diff
@@ -0,0 +1,193 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/component-base/metrics"
+
+	pacmkrv1 "github.com/openshift/api/etcd/v1"
+)
+
+var (
+	clusterHealthyGauge = metrics.NewGauge(&metrics.GaugeOpts{
+		Name:           "tnf_cluster_healthy",
```
jaypoulz: This is chai-bot, but it claims:
`Using Namespace: "tnf" + Subsystem: "cluster" in GaugeOpts is the k8s convention and would produce the same wire name while making the code more structured`

I'm not one way or another but I feel like this would be low effort for slightly cleaner structure.

pkg/tnf/pkg/metriccontroller/controller_test.go
```diff
@@ -0,0 +1,409 @@
...
+		withResourceCondition("master-1", pacmkrv1.PacemakerClusterResourceNameKubelet,
+			pacmkrv1.ResourceStartedConditionType, metav1.ConditionFalse),
+	))
+
+	require.NoError(t, testutil.GatherAndCompare(registry, strings.NewReader(`
+# HELP tnf_resource_started [ALPHA] Whether a TNF resource is started (1=started, 0=stopped)
+# TYPE tnf_resource_started gauge
+tnf_resource_started{node="master-0",resource="Etcd"} 0
+tnf_resource_started{node="master-0",resource="Kubelet"} 1
+tnf_resource_started{node="master-1",resource="Etcd"} 1
```
jaypoulz, line 352: etcd cannot be active while kubelet is dead :) probably not relevant but something I noticed

pkg/tnf/pkg/metriccontroller/controller_test.go
```diff
@@ -0,0 +1,409 @@
...
+tnf_resource_started{node="master-1",resource="Kubelet"} 0
+`), "tnf_resource_started"))
+}
+
+func TestLabelCorrectness(t *testing.T) {
+	registry := setupAndSync(t, buildCluster(withNodes("node-alpha", "node-beta")))
+
+	require.NoError(t, testutil.GatherAndCompare(registry, strings.NewReader(`
+# HELP tnf_node_online [ALPHA] Whether a TNF node is online (1=online, 0=offline)
```
jaypoulz, line 361: I'm not crazy about the string fixture approach. Feels like there's maybe a more programmatic way of synthesizing these, or we throw them in a fixtures file to make this more readable.

jaypoulz (commented):
Overall this is a great extension to the pacemaker healthcheck family of controllers. :)

As a standalone patch that gathers up these stats, I think this is quite reasonable. My only concern is - what happens if a node was populating information and then suddenly stops. PacemakerCluster would drop it entirely. Maybe an improvement here would be to initiate the node list from the k8s  node list, and then populate what's available from the CR. This way my lifecycle manager work plays nicely as it adds and removes nodes.

pkg/tnf/pkg/metriccontroller/metrics.go
fonta-rh: The CR is always created with conditions populated — StatusCollector writes `Status: *status` in the initial create (`statuscollector.go:825`), then immediately PUTs the status subresource. There's no window where the CR exists without conditions.

Before the CR exists, the informer store is empty, so `sync()` hits the "not found" path and resets all gauges to 0. That's the intended behavior — during initial setup, nothing is healthy yet, and 0 is the safe default for alerting (`== 0` fires without needing `absent()`).

pkg/tnf/pkg/metriccontroller/metrics.go
fonta-rh: Didn't know about this convention — good point. I was going with the existing CEO pattern from their metric (`signerExpirationGauge`). All the fields in GaugeOpts are concatenated, so the final wire name is still `tnf_cluster_healthy`, etc. — no breaking change.

Adopted it: `Namespace: "tnf"` with `Subsystem` per tier (`cluster`, `node`, `resource`).

pkg/tnf/pkg/metriccontroller/controller.go
fonta-rh: The factory controller framework (`library-go`) already logs returned errors and retries with exponential backoff — adding a log here would duplicate the message on each retry attempt.

pkg/tnf/operator/starter.go
fonta-rh: This matches the existing pattern at `pkg/operator/starter.go:440` (non-TNF starter). `DefaultGatherer` is initialized by `legacyregistry` as a `KubeRegistry` — it can't be a different type at runtime.

Happy to add a guard, but this is the only file we touch outside of our owned scope (`pkg/tnf/`), so it's the one I want to be most consistent with the existing codebase in.

pkg/tnf/pkg/metriccontroller/controller_test.go
fonta-rh: Yeah, not a realistic combination — but the test is just exercising the per-resource mapping logic, not simulating a real failure. The fixture sets one resource unhealthy per node to verify conditions don't bleed across resources.

pkg/tnf/pkg/metriccontroller/controller_test.go
fonta-rh: These are the standard k8s `testutil.GatherAndCompare` assertions — they require Prometheus text exposition format, so there's no programmatic builder to use here. Each string is only 3-6 lines and is tightly coupled to the `buildCluster()` options in the same test, so I think keeping them inline makes each test easier to read as a self-contained unit. Moving them to fixture files would separate the "what we set up" from the "what we expect" without reducing the overall line count.

pkg/tnf/pkg/metriccontroller/metrics.go
jaypoulz: Excellent. Glad we've thought this through.

pkg/tnf/pkg/metriccontroller/controller.go
jaypoulz: Awesome - good that the error is logged. :)

pkg/tnf/operator/starter.go
jaypoulz: Agreed. I'd follow the existing pattern and minimize the diff, like you said.

jaypoulz (approved):
/lgtm

PR #1639, opened 2026-06-23
CNTRLPLANE-3727: Agentic context docs and contributing guide

bhperry:
## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive contributor/agent guidance covering operator scope, contributor workflow rules, build/verify commands, and common do/don’t practices.
  * Added detailed architecture documentation describing the static pod lifecycle approach, TLS/certificate rotation flow, endpoints and health behavior, and testing strategy.
  * Added/expanded contribution guidelines covering coding and testing expectations plus pull request and verification workflows.
  * Centralized contributor instructions via a documentation link/symlink for easier discovery.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,282 @@
+# AI agent instructions for the cluster-etcd-operator codebase.
+
+## What This Repo Is
+
+The cluster-etcd-operator (CEO) manages the etcd cluster in OpenShift, handling:
+- etcd cluster scaling during bootstrap and steady state operation
+- TLS certificate provisioning and rotation
+- Backup and disaster recovery
+- Static pod lifecycle management via library-go framework
+- etcd health monitoring and defragmentation
+- Bootstrap teardown and scaling strategies (HA, Delayed-HA, Two-Node, etc.)
+
+See [ARCHITECTURE.md](ARCHITECTURE.md) for detailed architecture documentation.
+
+## Critical Rules
+
+1. **Never edit `vendor/` files** - managed by `go mod vendor`
+2. **Never edit generated files** - anything in `vendor/` or matching `zz_generated.*` patterns
+3. **Always run `make verify` before committing** - includes linting, generated file checks, and vendor verification
+4. **Always run `go mod tidy && go mod vendor`** after dependency changes
+5. **Never modify manifests without regenerating** - alerts/dashboards are generated from jsonnet (see [Alerts and Dashboards](#alerts-and-dashboards))
+6. **Always use `make build`** not `go build` - for proper version injection
+
+## Repository Structure
+
+```
+cluster-etcd-operator/
+├── bindata/             # Embedded YAML manifests and scripts
+│   ├── bootkube/        # Bootstrap manifests
+│   ├── etcd/            # etcd manifests and scripts (cluster backup, cluster restore, quorum restore)
+│   └── tnfdeployment/   # Two Node Fencing (TNF) manifests
+├── cmd/
+│   ├── cluster-etcd-operator/            # Main operator binary
+│   ├── cluster-etcd-operator-tests-ext/  # OTE test binary
+│   ├── tnf-monitor/                      # Two Node Fencing monitor
+│   └── tnf-setup-runner/                 # Two Node Fencing setup
+├── docs/                # Documentation (HACKING, FAQ, overview)
+├── hack/                # Build and generation scripts
+├── jsonnet/             # Alert and dashboard definitions
+├── manifests/           # CVO-managed YAML (deployment, RBAC, monitoring)
+├── pkg/
+│   ├── operator/        # Core operator logic (20+ controllers)
+│   ├── etcdcli/         # etcd client wrapper
+│   ├── tlshelpers/      # TLS certificate utilities
+│   └── dnshelpers/      # DNS resolution helpers
+└── test/e2e/            # End-to-end tests
+```
+
+## Key Patterns to Follow
+
+### Controller Pattern
+
+Controllers follow the library-go factory pattern:
+
+```go
+func NewMyController(
+    livenessChecker *health.MultiAlivenessChecker,
+    operatorClient v1helpers.StaticPodOperatorClient,
+    eventRecorder events.Recorder,
+) factory.Controller {
+    c := &MyController{
+        operatorClient: operatorClient,
+    }
+
+    syncer := health.NewDefaultCheckingSyncWrapper(c.sync)
+    livenessChecker.Add("MyController", syncer)
+
+    return factory.New().
+        ResyncEvery(time.Minute).
+        WithInformers(
+            operatorClient.Informer(),
+        ).
+        WithSync(syncer.Sync).
+        ToController("MyController", eventRecorder.WithComponentSuffix("my-controller"))
+}
+```
+
+Key elements:
+- Constructor function named `NewXxxController`
+- Wraps sync function with health checker
+- Uses library-go factory builder
+- Returns `factory.Controller` interface
+- **All resources the controller reads must have a corresponding informer registered via `.WithInformers()`** — missing informers are one of the most common bugs in this codebase. When adding or modifying a controller, verify that every lister/getter used in the sync function has its informer wired up in the factory builder, otherwise the controller will operate on stale data
+
+### Status Conditions
+
+Use `v1helpers.UpdateStatus` to set conditions:
+
+```go
+_, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient,
+    v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+        Type:    "MyControllerDegraded",
+        Status:  operatorv1.ConditionTrue,
+        Reason:  "SyncError",
+        Message: err.Error(),
+    }),
+)
+```
+
+Guidelines for status conditions:
+- **Use `Degraded` very sparingly** — only set it after a meaningful amount of time has passed (etcd moves slowly due to static pod machinery) and only when it requires **actionable** human intervention
+- **Avoid conditions that flap** — always add an inertia component (e.g., require the error state to persist for several sync cycles before reporting Degraded). The operator has historically bumped inertia from 5 to 10 minutes to reduce flapping during rollouts.
+
+### Bootstrap Scaling Strategies
+
+Openshift supports anywhere from 1 to 5 control-plane nodes depending on the level of high availability needed. The bootstrap scaling strategy describes the invariants which will be enforced when scaling the etcd cluster.
+
+The operator enforces different quorum requirements during bootstrap:
+- `HAScalingStrategy` - requires at least 3 nodes to scale up (default)
+- `DelayedHAScalingStrategy` - allows 2 nodes during bootstrap, then requires 3
+- `TwoNodeScalingStrategy` - requires 2 nodes (for Two Node OpenShift with Fencing deployments)
+- `DelayedTwoNodeScalingStrategy` - allows 1 node during bootstrap, then requires 2 (for Two Node OpenShift with Fencing deployments)
+- `BootstrapInPlaceStrategy` - used with assisted installer where the bootstrap node is repurposed as the final master node, allowing a 3-node cluster to be deployed with only 3 total machines (no separate bootstrap node). The bootstrap node performs a live ISO pivot to become a control-plane member. Particularly useful for bare metal installs
+- `UnsafeScalingStrategy` - scaling without regard to node count or quorum
+
+Check `pkg/operator/ceohelpers/bootstrap.go` for implementation.
+
+## Important Constraints
+
+1. **DNS Independence** - The `EtcdEndpointsController` must never depend on DNS (directly or transitively) as it maintains the IP-based ConfigMap that DNS depends on. DNS has no guarantee that a single static IP is returned. Previously, member nodes confused their identity by alternating between IPs. Using only IPs is a core design choice in this operator
+2. **CVO Resource Management** - Resources in `manifests/` are managed by CVO; removal requires special handling (see [CVO Resource Removal](#cvo-resource-removal)). The numeric ordering of file names in `manifests/` determines CVO deployment order — incorrect ordering has caused outages in CI (e.g., when a manifest was deployed before its dependency)
+3. **Never Remove Controllers** - An agent should never remove a controller. Controller removals are extremely rare (only one in the history of the project) and require cleaning up stale status conditions to prevent upgrade blockers. This is a human decision
+4. **Static Pod Revisions** - Follow library-go static pod revision pattern; don't invent custom update logic
+5. **Namespace** - Primary namespace for the operator resources is `openshift-etcd-operator`; target namespace for the operand is `openshift-etcd`. Constants are found in `pkg/operator/operatorclient/interfaces.go` and should always be used
+
+## Build and Test
+
+### Common Commands
+
+```bash
+# Build all binaries (operator + test extension)
+make build
+
+# Run unit tests
+make test-unit
+
+# Run e2e tests (requires OpenShift cluster)
+make test-e2e
+
+# Verify linting, generated files, vendor
+make verify
+
+# Update dependencies
+go mod tidy && go mod vendor
+```
+
+### Live Debugging
+
+When debugging live systems, use commands in [profiling and debugging](docs/profiling_and_debugging.md).
+
+To run CEO locally against a cluster:
+```sh
+# Build executable
+make build
+
+# Get config
+kubectl -n openshift-etcd-operator get configmap etcd-operator-config -o jsonpath='{.data.config\.yaml}' > ./config.yaml
+
+# Scale down cluster-etcd-operator deployment in cluster and exclude from CVO
+./hack/unmanage.sh
+
+# "Missing Version" markers
+export OPERATOR_IMAGE_VERSION=0.0.1-snapshot
+export OPERAND_IMAGE_VERSION=0.0.1-snapshot
+
+# Run local operator
+./cluster-etcd-operator \
+    --config ./config.yaml
+    --namespace openshift-etcd-operator \
+    --kubeconfig /path/to/kubeconfig
+```
+
+### OTE Test Framework
+
+This repo uses [OpenShift Tests Extension (OTE)](https://github.com/openshift-eng/openshift-tests-extension):
+
+```bash
+# Build test binary
+make build  # produces cluster-etcd-operator-tests-ext
+
+# List test suites
+./cluster-etcd-operator-tests-ext list suites
+
+# Run specific suite with concurrency
+./cluster-etcd-operator-tests-ext run-suite openshift/cluster-etcd-operator/all -c 1
+
+# Run with JUnit output
+./cluster-etcd-operator-tests-ext run-suite openshift/cluster-etcd-operator/all \
+    --junit-path=/tmp/junit-results/junit.xml
+```
+
+Test files are in `test/e2e/` and registered via the OTE framework in `cmd/cluster-etcd-operator-tests-ext/`.
+
+## Alerts and Dashboards
+
+Prometheus alerts and Grafana dashboards are **generated from jsonnet**, not hand-edited. In general, avoid adding new alerts — they are difficult to maintain in CI and we add them very rarely.
+
+### Jsonnet Structure
+
+Almost all files in the `jsonnet/` directory can be changed, but the lock file (`jsonnetfile.lock.json`) and `vendor/` folder are off limits.
+
+- `main.jsonnet` - Imports upstream etcd mixin alerts and merges with custom overrides. Contains the `excludedAlerts` list for upstream alerts we replace or disable. **The exclusion list must be in sorted order** — the comparison uses a sort-dedupe, so unsorted entries will silently break
+- `custom.libsonnet` - Used for overrides of upstream alerts and addition of new downstream-specific alerts
+- `dashboard.jsonnet` - Dashboard definitions. Custom dashboards/panels should be their own dedicated files (e.g., `cpu_iowait_dashboard.json`) and merge-appended into `dashboard.jsonnet`
+
+### Update Process
+
+1. **If changing upstream alerts** (e.g., different thresholds): update upstream alerts first, then exclude the original in `main.jsonnet` and add the override in `custom.libsonnet`
+2. **For new downstream alerts**: add them in `custom.libsonnet`
+    - New alerts are added only very rarely. They should be avoided unless absolutely necessary
+3. Regenerate manifests: `hack/generate.sh`
+4. This updates:
+   - `manifests/0000_90_etcd-operator_03_prometheusrule.yaml`
+   - `manifests/0000_90_cluster-etcd-operator_01-dashboards.yaml`
+5. **Verify the alert is reflected correctly in the generated output** — always diff the generated YAML to confirm your changes took effect
+
+**Never** edit the generated YAML files directly.
+
+### Prerequisites
+
+Requires: `jsonnet`, `jb` (jsonnet-bundler), `gojsontoyaml`
+See https://github.com/google/jsonnet
+
+## CVO Resource Removal
+
+Resources in `manifests/` are managed by CVO. To remove one:
+
+1. **Don't just delete the manifest file** - it won't be removed from clusters
+2. Add the `release.openshift.io/delete: "true"` annotation to the resource
+3. CVO will garbage collect it on next update
+4. Keep the annotation in manifests for 1 release, then remove the file
+
+See:
+- https://github.com/openshift/enhancements/blob/master/enhancements/update/object-removal-manifest-annotation.md
+- https://github.com/openshift/cluster-etcd-operator/pull/1016
+
+## Known Bug Patterns
+
+1. **Missing or wrong informer scope** — Always verify informer scope matches what the controller actually needs, and wait for caches to sync (e.g. using the general node lister instead of master-only node lister caused controllers to see non-master nodes and make incorrect scaling decisions)
+2. **Insufficient inertia on status conditions** — controllers reacting too quickly to transient state changes cause flapping
+3. **Bootstrap member management and quorum** — non-deterministic IP selection, revision stability vs. member removal timing, and quorum safety during member add/remove are the single most-fixed area. Changes here interact differently across release branches
+4. **Health check sensitivity** — Liveness probe timeouts and failure thresholds need careful tuning. Too aggressive kills healthy pods. Too relaxed misses real failures
+5. **Shell script fragility** — `cluster-backup.sh` and `cluster-restore.sh` have been sources of repeated bugs (caching issues, path errors, retention logic). Changes to these scripts need extra scrutiny
+6. **Static pod revision tracking** — The static pod revision mechanism is the critical path for config rollout. Getting the balance wrong between "wait for stability" and "don't block bootstrap" causes cascading issues. Each release branch has subtly different behavior, making backports risky
+7. **Upgrade path considerations** — resources and conditions from N-1 releases may still exist. Never assume a clean slate
+
+## TLS and Cipher Suites
+
+TLS certificates are managed by the `etcdcertsigner` and `etcdcertcleaner` controllers. Cipher suite configuration is a known pain point:
+
+- **Bootstrap ciphers** are defined in `pkg/cmd/render/env.go` based on the `TLSProfileIntermediateType` profile from `configv1.TLSProfiles`
+- **Runtime ciphers** are sourced from `observedConfig` in `pkg/etcdenvvar/etcd_env.go` and filtered through `tlshelpers.SupportedEtcdCiphers()` which validates each cipher against what etcd actually supports
+- **Not all OpenShift ciphers are supported by etcd** — `SupportedEtcdCiphers()` in `pkg/tlshelpers/tlshelpers.go` filters out unsupported ciphers using `tlsutil.GetCipherSuite()`. When OpenShift adds new ciphers, they may silently be dropped if etcd doesn't support them
+- When modifying cipher handling, verify that the resulting cipher list is non-empty and that both bootstrap and runtime paths produce consistent results
+
+## UnsupportedConfigOverrides
+
+The operator supports an escape-hatch mechanism via `UnsupportedConfigOverrides` on the operator spec. This is used for unsupported, non-production overrides (e.g., `useUnsupportedUnsafeNonHANonProductionUnstableEtcd`). See `pkg/operator/ceohelpers/unsupported_override.go`.
+
+This mechanism exists because the operator historically had severe stability problems and needed a way to override behavior without a full release cycle. For new feature development, prefer using proper feature gates instead of adding new unsupported config keys.
+
+## Fencing Scripts
+
+Fencing scripts in `bindata/etcd/` (shell scripts) and `bindata/tnfdeployment/` (deployment manifests) should be reviewed for injection vulnerabilities when modified.
+
+## Key Files
+
+- `pkg/operator/starter.go` - Operator entrypoint, initializes all controllers
+- `pkg/operator/ceohelpers/bootstrap.go` - Bootstrap scaling logic
+- `pkg/operator/etcdendpointscontroller/` - Critical DNS-independent controller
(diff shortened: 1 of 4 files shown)


AGENTS.md
```diff
@@ -0,0 +1,260 @@
...
+2. **Backup Credentials** - stored in secrets, never logged
+3. **Escape Hatches** - documented in docs/overview/overview.md (Secret Escape Hatches section)
+4. **Fencing Scripts** - in bindata/etcd/, reviewed for injection vulnerabilities
+
+## Key Files
+
+- `pkg/operator/starter.go` - Operator entrypoint, initializes all controllers
+- `pkg/operator/ceohelpers/bootstrap.go` - Bootstrap scaling logic
+- `pkg/operator/etcdendpointscontroller/` - Critical DNS-independent controller
+- `pkg/etcdcli/etcdclient.go` - etcd client abstraction
```
dusk125: `pkg/etcdcli/etcdcli.go` is the correct filename

AGENTS.md
```diff
@@ -0,0 +1,260 @@
...
+│   ├── cluster-etcd-operator-tests-ext/  # OTE test binary
+│   ├── tnf-monitor/                      # Two Node Fencing monitor
+│   └── tnf-setup-runner/                 # Two Node Fencing setup
+├── docs/                # Documentation (HACKING, FAQ, overview)
+├── hack/                # Build and generation scripts
+├── jsonnet/             # Alert and dashboard definitions
+├── manifests/           # CVO-managed YAML (deployment, RBAC, monitoring)
+├── pkg/
+│   ├── operator/        # Core operator logic (20+ controllers)
+│   │   ├── etcdendpointscontroller/
```
dusk125: I dunno if it's worth enumerating all of the controllers here; having only a subset may mislead the bots and some of the omitted are important as well.

ARCHITECTURE.md
```diff
@@ -0,0 +1,240 @@
...
+**Safety:** QuorumChecker blocks revisions if quorum unhealthy; bundle-rollout gating (4.17+) prevents cert updates before CA bundle rollout completes.
+
+## etcd Client Abstraction
+
+CEO's etcd client (`pkg/etcdcli/interfaces.go`) composes 10+ interfaces (MemberAdder, MemberRemover, Defragment, etc.). Two implementations:
+
+- **etcdcli.EtcdClient** - Direct client using TLS certs
+- **etcdcli.MemberCache** - Caching wrapper (60s TTL); not used for 2-node quorum checks
+
+**Connection fallback:** IP-based `openshift-etcd/endpoints` → DNS-based `openshift-etcd/etcd-endpoints` (with staleness detection) → direct node IPs.
```
dusk125: This isn't correct, it should be something like:
`Connection fallback: etcd-endpoints ConfigMap (IP-based, maintained by EtcdEndpointsController) → direct node internal IPs (from Node objects via GetPreferredInternalIPAddressForNodeName).`

ARCHITECTURE.md
```diff
@@ -0,0 +1,238 @@
...
+
+See `pkg/operator/starter.go` `RunOperator()` for full list and initialization order.
+
+## Key Design Decisions
+
+### DNS Independence for EtcdEndpointsController
+
+**Decision:** The EtcdEndpointsController must never depend on DNS, directly or transitively.
+
+**Rationale:** This controller maintains `openshift-etcd/endpoints` ConfigMap with IP addresses that DNS resolution depends on. A DNS dependency would create a circular bootstrap problem.
```
dusk125: `openshift-etcd/etcd-endpoints` is the configmap name. If you want to be explicit about the namespace `openshift-etcd` and configmap name `etcd-endpoints`

dusk125 (changes requested):

AGENTS.md
bhperry: good point, will leave it at high level dir

CONTRIBUTING.md
```diff
@@ -0,0 +1,174 @@
...
+We largely follow the [Kubernetes Code Conventions](https://github.com/kubernetes/community/blob/main/contributors/guide/coding-conventions.md#code-conventions).
+
+Review both the Kubernetes Code Conventions and the ones specified here.
+There will be some overlap. If any conventions are at odds with one another, prefer the conventions explicitly documented here.
+
+### Bash
+
+- Follow the [shell styleguide](https://google.github.io/styleguide/shellguide.html).
+- Use [`shellcheck`](https://github.com/koalaman/shellcheck) to identify common mistakes or caveats.
+- Ensure that all scripts run consistently across Linux and MacOS.
```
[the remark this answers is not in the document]
bhperry: MacOS is referrenced throughout the repo, leaving as is.

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+│   ├── cluster-etcd-operator/            # Main operator binary
+│   ├── cluster-etcd-operator-tests-ext/  # OTE test binary
+│   ├── tnf-monitor/                      # Two Node Fencing monitor
+│   └── tnf-setup-runner/                 # Two Node Fencing setup
+├── docs/                # Documentation (HACKING, FAQ, overview)
+├── hack/                # Build and generation scripts
+├── jsonnet/             # Alert and dashboard definitions
+├── manifests/           # CVO-managed YAML (deployment, RBAC, monitoring)
+├── pkg/
+│   ├── operator/        # Core operator logic (20+ controllers)
```
dusk125, line 42: Maybe bring back `ceohelpers` and controllers in the form `*controller`... although there are some other non-"controller" controllers ("bootstrapteardown", etc).

What do you think?

dusk125: Maybe leaving it as is and going into more complete detail in the Architecture file - going through all of the controllers?

bhperry: I think may not be necessary to explain ceohelpers. If agent lists the dir it's fairly self explanatory what that is.

bhperry: Enumerating the controllers with descriptions in ARCHITECTURE.md makes sense to me

dusk125: Both sound good to me!

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+        WithSync(syncer.Sync).
+        ToController("MyController", eventRecorder.WithComponentSuffix("my-controller"))
+}
+```
+
+Key elements:
+- Constructor function named `NewXxxController`
+- Wraps sync function with health checker
+- Uses library-go factory builder
+- Returns `factory.Controller` interface
```
tjungblu, line 82: a critical rule for me would be to double check whether all informers are covered (probably one of our 80% "bug" problem).

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+_, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient,
+    v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+        Type:    "MyControllerDegraded",
+        Status:  operatorv1.ConditionTrue,
+        Reason:  "SyncError",
+        Message: err.Error(),
+    }),
+)
+```
+
```
tjungblu, line 99: this is very tribal for this but useful for agents:
* use the degraded scenario very sparingly, only after a meaningful amount of time has passed (etcd moves slow due to the static pod machinery) and only if it really requires **actionable** human intervention. 
* avoid cases where the conditions can flap, always add an inertia component

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+### Bootstrap Scaling Strategies
+
+The operator enforces different quorum requirements during bootstrap:
+- `HAScalingStrategy` - requires 3 nodes (default)
+- `DelayedHAScalingStrategy` - allows 2 nodes during bootstrap, then requires 3
```
tjungblu, line 110: this could use a bit more context. IIRC that is for ass. installer where the bootstrap node is repurposed as the last master node. Meaning you can get away with only three nodes in total (especially useful in bare metal).

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+- `TwoNodeScalingStrategy` - requires 2 nodes (for Two Node OpenShift with Fencing deployments)
+- `DelayedTwoNodeScalingStrategy` - allows 1 node during bootstrap, then requires 2 (for Two Node OpenShift with Fencing deployments)
+- `BootstrapInPlaceStrategy` - single-node bootstrap where no separate bootstrap node exists (live ISO pivot)
+- `UnsafeScalingStrategy` - scaling without regard to node count or quorum
+
+Check `pkg/operator/ceohelpers/bootstrap.go` for implementation.
+
+## Important Constraints
+
+1. **DNS Independence** - The `EtcdEndpointsController` must never depend on DNS (directly or transitively) as it maintains the IP-based ConfigMap that DNS depends on
```
tjungblu: why is that: DNS has no guarantee that only a single IP and always the same is returned, hence it was decided to switch to IPs. 

Previously it happened that a member node thought it was a different member node (also flapping between IPs). Using only IPs is a core design choice in this operator and that very likely is not changing.

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+2. **CVO Resource Management** - Resources in `manifests/` are managed by CVO; removal requires special handling (see [CVO Resource Removal](#cvo-resource-removal))
```
tjungblu: it is also very important that the ordering is correctly reflected in the file names. See the recent OVN/networking problem where a manifest was deployed before the other and that caused outages in CI.

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+3. **Controller Removal** - Removing a controller requires cleaning up stale status conditions to prevent upgrade blockers
```
tjungblu: I would even say: an agent should never remove a controller. Besides the UpgradeBackup Controller we had, we didn't really have a removal in all those years.

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+4. **Static Pod Revisions** - Follow library-go static pod revision pattern; don't invent custom update logic
+5. **Namespace** - Primary namespace is `openshift-etcd`; target namespace constant is in `pkg/operator/operatorclient`
```
tjungblu:
```suggestion
5. **Namespace** - Primary namespace for the operator resources is `openshift-etcd-operator`; target namespace for the operand is `openshift-etcd`. Constants are found in `pkg/operator/operatorclient/interfaces.go` and should always be used 
```

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+
+Test files are in `test/e2e/` and registered via the OTE framework in `cmd/cluster-etcd-operator-tests-ext/`.
+
+## Alerts and Dashboards
+
+Prometheus alerts and Grafana dashboards are **generated from jsonnet**, not hand-edited.
+
+### Update Process
+
+1. Edit alert rules in `jsonnet/custom.libsonnet`
```
tjungblu: Okay this is a bit more complex and never documented.  Here's just the stuff from the top of my head:
* all files in this directory can be changed, only the lock file and the vendor folder are off limits 
* custom.libsonnet can be used for overrides of upstream alerts or addition of new alerts (downstream)
* If we're changing upstream alerts, e.g for different thresholds, we need to exclude them in main.jsonnet - note that the exclusion must be in sorted order, otherwise the comparison does not work (it's doing a sort dedupe). An agent should verify that the alert is reflected correctly in the generated output. 
* custom dashboards/panels we add should be the own dedicated file and be merge-appended into dashboard.jsonnet
* in general we avoid adding new alerts, because they are a nightmare to maintain in CI - so we very very very rarely add one


When changing any alert, ensure that we update the upstream alerts first. We may want to spawn an agent to keep them rebased at certain invervals in the future

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+        "EtcdMemberIPMigratorDegraded",  // removed in 4.5
+    },
+    operatorClient,
+    controllerContext.EventRecorder,
+)
+```
+
+See:
+- https://github.com/openshift/cluster-etcd-operator/pull/517
+- https://bugzilla.redhat.com/show_bug.cgi?id=1851351
```
tjungblu: I would kinda leave this section out, because we don't really do that anymore. But props to digging out that 4.5 issue lol

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+
+## What NOT to Do
```
tjungblu: most of those below seem redundant with the critical rules

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+7. **Don't remove controllers without cleanup logic** - stale conditions block upgrades
+8. **Don't edit go.sum** - update dependencies in go.mod and then run `go mod tidy && go mod vendor`
+
+## Common Mistakes to Avoid
+
+1. **Assuming 3-node quorum everywhere** - respect the bootstrap scaling strategy (see `ceohelpers/bootstrap.go`)
+2. **Not handling bootstrap teardown** - controllers may run during bootstrap with incomplete state
+3. **Ignoring static pod revision mechanics** - follow library-go patterns for updates
+4. **Breaking IP-based fallback** - etcd must work when DNS is down
+5. **Not considering upgrade paths** - resources/conditions from N-1 may still exist
```
tjungblu: they are not terrible, but I asked claude to scan our commits and see what failures we're mostly prone to. I think this reflects also my own experience:


```  
  Top 5 Recurring Fix Patterns

  1. Bootstrap, Member Management & Quorum (42 commits)

  The single most-fixed area. Recurring themes:

  - Revision stability vs. member removal timing — OCPBUGS-74151 alone has 5 commits across releases. The tension is: do you wait for revision stability before
  removing members, or does that block bootstrap completion? This was fixed, reverted, and re-fixed across 4.13–4.16 (OCPBUGS-14673, 17199), reverted again, then
  re-approached from a different angle.
  - Bootstrap IP ordering — OCPBUGS-55404 (3 commits): non-deterministic IP selection caused inconsistent bootstrap behavior.
  - Quorum safety during bootstrap — OCPBUGS-42809/42810: quorum wasn't always ensured, and leader transfer wasn't happening before member removal.
  - Scale-down flakes — TestAttemptToScaleDown was fixed 3 separate times for the same flaky test.

  Root cause pattern: Distributed state machine logic with many edge cases around when it's safe to add/remove members. Changes in one release branch get
  backported but interact differently with other release-specific code.

  2. Health Checks, Liveness & Readiness (40 commits)

  - Panics in health check timeouts — OCPBUGS-27959 required 5 commits across releases to fix nil pointer panics when health checks timed out.
  - Liveness probe too aggressive — OCPBUGS-7729 (3 commits): timeout/failure thresholds were too low, causing unnecessary pod restarts. A separate attempt to add
  controller health to liveness (OCPBUGS-11683) was reverted entirely.
  - Aliveness check scope creep — OCPBUGS-30873 (3 commits): the CEO aliveness check was triggering on things other than deadlocks.
  - Client timeout configuration — OCPBUGS-30169: added a timeout to etcd client.New, then reverted it because it caused other issues. Pooled client approach
  (OCPBUGS-38573) was also reverted.
  - Health check ordering — OCPBUGS-36462/37820: non-deterministic member health check ordering caused flapping status.

  Root cause pattern: Getting the sensitivity right is hard — too aggressive kills healthy pods, too relaxed misses real failures. Fixes in one direction tend to
  create problems in the other.

  3. Backup & Restore (40 commits)

  - etcdctl caching on backup — OCPBUGS-19052 (9 commits, the single most-backported bug): cluster-backup.sh was caching etcdctl binaries, causing version
  mismatches.
  - Backup schedule/CR handling — OCPBUGS-43687/43676: empty schedule fields in the EtcdBackup CR weren't handled, causing panics or no-ops.
  - Prune path and retention — multiple fixes to the backup-server pruning logic: wrong paths, wrong defaults, wrong retention counts.
  - Snapshot file locking — OCPBUGS-15327 (4 commits): default snapshot counts weren't reset, causing "file already locked" warnings.
  - Debug pod namespace — The backup e2e test debug pod namespace was fixed 3 separate times.

  Root cause pattern: Shell script fragility (cluster-backup.sh, cluster-restore.sh) and incomplete validation of CR fields. The backup server was built
  incrementally and each piece (scheduling, pruning, retention) had its own round of fixes.

  4. Revision Rollout & Static Pod Management (34 commits)

  - revision.json persistence — OCPBUGS-60273 (4 commits): revision.json didn't survive ungraceful shutdowns, causing revision tracking to desync.
  - Bundle revision rollout — ETCD-606: batch bundle rollout was implemented, reverted, and re-landed.
  - Revision stability checks — the same "remove revision stability check from bootstrap completeness" fix (OCPBUGS-14673/17199) was applied, reverted across 3
  release branches, then re-applied. This single issue accounts for ~10 commits including reverts.
  - Race conditions during cert render — OCPBUGS-32217: concurrent cert creation during static pod render caused races.

  Root cause pattern: The static pod revision mechanism is the critical path for config rollout. Getting the balance wrong between "wait for stability" and "don't
  block bootstrap" causes cascading issues. Each release branch has subtly different behavior, making backports risky.

  5. Race Conditions, Ordering & Informer/Cache Issues (29 commits)

  - Missing informer syncs — OCPBUGS-44570 (2 commits): informers weren't fully synced before controllers started making decisions, leading to stale data.
  - Wrong nodelister scope — OCPBUGS-29717/30245 (4 commits): using the general nodelister instead of master-only nodelister caused controllers to see non-master
  nodes, leading to incorrect scaling decisions.
  - Controller inertia — OCPBUGS-38659 (4 commits): controllers reacted too quickly to transient state changes, requiring inertia configuration to dampen flapping
  (bumped from 5 to 10 minutes).
  - Shared informer conflicts — dedicated informers had to be created to avoid conflicts between controllers watching the same resources.
  - Test flakiness — defrag controller race, scale-down test flakes, and general UT flakiness each required dedicated fixes.

  Root cause pattern: Kubernetes informer/lister caching semantics are subtle. Using the wrong lister scope or not waiting for sync is easy to miss in review. The
  inertia issue shows that even correct logic can misbehave when controllers react to intermediate states during rollouts.

  ---
  Notable cross-cutting observation

  19 reverts in this period — a high rate. The most-reverted areas are health check changes (3 reverts), revision stability logic (3 reverts), and bundle rollout
  (1 revert). This suggests that changes in the etcd lifecycle management space are hard to validate fully before merge, and the blast radius of getting them wrong
  is high enough that reverting is the safe default.


```

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Security Notes
+
+1. **TLS Certificates** - managed by `etcdcertsigner` and `etcdcertcleaner` controllers
```
tjungblu: another super annoying part are the TLS ciphers. Maybe you can prompt a section for handling that more specifically. But this is a moving target as well, there are some changes inbound

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+2. **Backup Credentials** - stored in secrets, never logged
```
tjungblu: that's a hallucination

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+3. **Escape Hatches** - documented in docs/overview/overview.md (Secret Escape Hatches section)
```
tjungblu: not so secret anymore, is it :) 

I think the better way to deal with such a thing would be more of a feature gate. We didn't have that back then and because of severe stability problems, hence the escape, we added a flag configmap.

Maybe we can devote a section for it, that may be relevant for feature development in this repo

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+4. **Fencing Scripts** - in bindata/etcd/ (shell scripts) and bindata/tnfdeployment/ (deployment manifests), reviewed for injection vulnerabilities
```
tjungblu: @jaypoulz would be that relevant for your teams day2day agentic workflow?

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+
```
tjungblu, line 265: just to confirm my own comments, asked claude to summarize our security vulnerabilities from the commit history:

``` 
  Security & CVE Fix Patterns

  Overview

  ~50 security-related commits since 2023, falling into 5 distinct categories. The dominant pattern is transitive dependency CVEs requiring backports across many
  release branches.

  ---
  1. Transitive Go Dependency CVEs (30+ commits)

  This is the overwhelming majority of security work. Three CVEs account for the bulk:

  ┌────────────────┬───────────────────────────────────┬───────────────────────────────┬────────────────┬─────────────────────┐
  │      CVE       │              Library              │             Issue             │    Commits     │ Branches Backported │
  ├────────────────┼───────────────────────────────────┼───────────────────────────────┼────────────────┼─────────────────────┤
  │ CVE-2025-30204 │ golang-jwt/jwt/v4 (4.4.2 → 4.5.2) │ JWT parsing vulnerability     │ 8              │ 14 (4.12–5.1)       │
  ├────────────────┼───────────────────────────────────┼───────────────────────────────┼────────────────┼─────────────────────┤
  │ CVE-2023-39325 │ golang.org/x/net/http2            │ HTTP/2 rapid reset DoS        │ 6 (1 reverted) │ 15 (4.11–5.1)       │
  ├────────────────┼───────────────────────────────────┼───────────────────────────────┼────────────────┼─────────────────────┤
  │ CVE-2022-41717 │ golang.org/x/net (via library-go) │ HTTP/2 memory exhaustion      │ 2              │ 2                   │
  ├────────────────┼───────────────────────────────────┼───────────────────────────────┼────────────────┼─────────────────────┤
  │ CVE-2025-65637 │ sirupsen/logrus (1.8.1 → 1.9.3)   │ Logging library vulnerability │ 1              │ 1 (so far)          │
  └────────────────┴───────────────────────────────────┴───────────────────────────────┴────────────────┴─────────────────────┘

  The pain: These are not bugs in this repo's code — they're bumps to indirect dependencies. But each bump must be backported to every supported release branch
  independently, because each branch pins different versions. CVE-2025-30204 alone produced 8 commits across 14 branches. CVE-2023-39325 hit 15 branches and even
  had to be reverted on 4.11 because the dependency bump broke the older Go version.

  In addition, golang.org/x/net was updated 5 more times in April 2024 (OCPBUGS-319xx series) across 5 release branches — likely for another x/net advisory that
  wasn't given a CVE in the commit messages.

  Root cause: No automated dependency update mechanism across release branches. Each branch is a manual effort with its own go.mod, and older branches have tighter
  Go version constraints that can make a straightforward bump impossible.

  ---
  2. TLS/Cipher Suite Configuration (10 commits, 1 revert)

  A multi-year saga:

  1. Dec 2022: Hardcoded safe TLS 1.2/1.3 cipher suites in etcd and CEO
  2. Aug 2024 (OCPBUGS-38341): Switched from hardcoded ciphers to environment variable — hardcoding broke when the platform's TLS policy changed
  3. May 2025 (OCPBUGS-56834): etcd pods wouldn't start when TLS min version was set to 1.3 — the cipher suite list and the readyz probe didn't support TLS
  1.3-only mode. Required touching 6 files including the pod template, readyz handler, and env var controller
  4. Oct 2025 (ensure etcd starts with TLS 1.3): Another fix for TLS 1.3 compatibility
  5. Sep 2025: "use secure cipher suites" committed
  6. Mar 2026: Reverted — the secure cipher suite change broke something

  Root cause: The cipher suite / TLS version configuration sits at the intersection of three systems: the platform's crypto policy (apiserver TLS profile), etcd's
  own TLS config, and the operator's readyz/liveness probes. Changes to one system expose incompatibilities in the others. The pattern of commit-then-revert shows
  that testing TLS configuration across all supported profiles is difficult.

  ---
  3. Missing Container Auth Credentials (6 commits)

  OCPBUGS-18357 (prepend authfile to podman create) was backported 6 times across release branches. The cluster-backup.sh script was running podman create without
  the --authfile flag, meaning on nodes where registry auth is required, the backup container couldn't pull images.

  Root cause: Shell scripts don't have the same level of review scrutiny as Go code, and the authfile flag is easy to forget since podman works without it in many
  dev/test environments that have pre-cached images.

  ---
  4. Network Policy Hardening (4 commits, 2025–2026)

  A relatively recent addition:

  - CNTRLPLANE-2663: Added network policies to the operator
  - CNTRLPLANE-2683: Added network policies to openshift-etcd pods
  - TNF jobs added to network policy exceptions
  - E2e tests added for network policy

  Root cause: The etcd namespace previously ran without network policies, relying on the cluster's default-deny being absent. When default-deny became standard,
  etcd communication broke unless explicit policies existed. This was a gap rather than a bug — the policies should have existed from the beginning.

  ---
  5. Operator Restart on TLS Config Changes (1 commit, late addition)

  a79ec721a: Restart operator when TLS config changes — the operator wasn't watching for changes to the platform TLS configuration. When an admin changed the
  crypto policy, the operator kept running with the old TLS settings until it was manually restarted.

  Root cause: The TLS config was treated as immutable at startup rather than being watched as a dynamic config source.

  ---
  Cross-cutting observations

  - Backport tax is the dominant cost. The actual fix for each CVE is typically a one-line go.mod bump. The work is doing it 8–15 times across release branches,
  each with its own dependency graph. CVE-2023-39325 showed this can actively backfire — the 4.11 backport was reverted because it was incompatible with that
  branch's Go version.
  - TLS is the only area where this repo's own code has recurring security issues. Everything else is dependency management. The cipher suite saga (4 years, still
  getting reverted) suggests the configuration surface is too complex for the testing that exists.
  - Shell scripts are a blind spot. The authfile fix is a security issue that wouldn't have been caught by Go-level tooling (linting, go vet, etc.). The
  backup/restore scripts interact with container runtimes and need their own security review process.



```

AGENTS.md
tjungblu: I also think, because it's gotten so big in the meantime, we should devote an entire TNF section here too...

CONTRIBUTING.md
```diff
@@ -0,0 +1,173 @@
...
+### Interacting with reviewers
+
+When interacting with reviewers/approvers:
+
+- Be professional.
+- Be respectful of differing opinions, viewpoints, and experiences.
+- Gracefully give and receive constructive feedback.
+- Focus on what is best for the product/organization, not just us as individuals.
+
+A special note on the usage of AI - to respect the time of those that are reviewing your contribution, please do not use AI to respond to review comments.
```
tjungblu, line 173: it's okay, we don't have to be respectful around here 🤣

ARCHITECTURE.md
```diff
@@ -0,0 +1,315 @@
...
+                           ▼
+            ┌──────────────────────────────┐
+            │  etcd static pods (per node) │
+            │  ┌─────────┐  ┌─────────┐    │
+            │  │  etcd   │  │  etcd-  │    │
+            │  │ :2379   │  │ metrics │    │
+            │  │ :2380   │  │ :9979   │    │
+            │  └─────────┘  └─────────┘    │
+            └──────────────────────────────┘
+```
```
tjungblu, line 29: can we maybe make this a mermaid chart instead?

ARCHITECTURE.md
```diff
@@ -0,0 +1,315 @@
...
+
+## Controllers
+
+All controllers are initialized in `pkg/operator/starter.go` `RunOperator()`. They follow the library-go factory pattern, wrapping their sync logic with `health.NewDefaultCheckingSyncWrapper()` for liveness monitoring via a centralized `MultiAlivenessChecker`.
```
tjungblu: maybe to add why we're doing this:
* our operator was deadlocked very frequently
* we couldn't figure it out for a long time, even with thread dumps
* we accidentally caught one stuck in grpc in etcdclient, but there are some other places for sure

So this is band-aid to restart the operator in order for it to continue processing. If you're old enough, this is kinda like the daily Tomcat/Apache restart ;-)

ARCHITECTURE.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Cluster Membership Controllers
```
tjungblu, line 42: I personally don't mind to have a shorter technical summary of every controller - but it has nothing to do with architecture. 

To me the architectural considerations in this operator are the following:
* quorum safety first - we favor etcd availability ALWAYS and on any SKU, anything that causes downtime must be avoided at all costs
* one concern per controller (needs a bit more nuance, but let's leave this here for now)
* always use informers and cached clients, avoid directly hitting APIs and etcd, we want a low resource control plane footprint
* always refactor into reusable helpers and interfaces as building blocks
* must-gather readability - debugging CI and customer cases is difficult. We want to be generally verbose in logging, only use status conditions and events when you want to communicate actionable feedback to the admin or the CI timeline

AGENTS.md
tjungblu: oh and more general openshift knowledge won't hurt: we support all control plane sizes now from 1 node to 5 nodes.

Plus microshift, which actually has some special cases that we need to be aware of here. @dusk125 I think that was mostly around the embedding into the binary, so we can't ever be ahead in go version and we have to be careful about the dependencies.

AGENTS.md
tjungblu: I saw this is also covered below more succinct as "no circular dependency on coreDNS"

AGENTS.md
```diff
@@ -0,0 +1,258 @@
...
+1. **DNS Independence** - The `EtcdEndpointsController` must never depend on DNS (directly or transitively) as it maintains the IP-based ConfigMap that DNS depends on
+2. **CVO Resource Management** - Resources in `manifests/` are managed by CVO; removal requires special handling (see [CVO Resource Removal](#cvo-resource-removal))
+3. **Controller Removal** - Removing a controller requires cleaning up stale status conditions to prevent upgrade blockers
+4. **Static Pod Revisions** - Follow library-go static pod revision pattern; don't invent custom update logic
+5. **Namespace** - Primary namespace is `openshift-etcd`; target namespace constant is in `pkg/operator/operatorclient`
+
+## Build and Test
+
+### Common Commands
```
tjungblu, line 128: one last open ended request for you - sorry for all the comments :) 

I always use agents to debug live systems for me, so it would be quite useful to tell it about https://github.com/openshift/cluster-etcd-operator/blob/main/docs/profiling_and_debugging.md

and https://github.com/openshift/cluster-etcd-operator/blob/main/hack/unmanage.sh

basically to show how to:
* run CEO locally, connect to remote cluster (means ./hack/unmanage.sh && build && run) 
* run a local build of CEO inside the cluster (like the above, just with pushing and updating the image)

But I can add this later if you're running out of time.

bhperry: Appreciate the thorough review. Happy to add those in.

ARCHITECTURE.md
bhperry: Agreed, I wanted to avoid cluttering the main AGENTS.md context too much but still thought a description of the controllers could be useful. ARCHITECTURE.md seemed like a reasonable place for that, but we could also make a separate docs/controllers.md file linked from architecture if you would prefer.

These bullet points are helpful, will add them.

AGENTS.md
bhperry: Added in a rough overview of running CEO locally. Was able to get it running with this, but there were some errors still, not sure if I need to also fetch all the certs locally or if I'm missing something else.

tjungblu: Thank you! I wanted to also say that we can do it iteratively and I can write a few sections over the course of the next sprint as well. 

On the error topic: yes, that's a bit annoying because you can't reach the etcd members locally, so it may not just be a matter of certificates, but also port forwarding and re-configuring the etcdclient accordingly. Then the certs may not match anymore? 

My own rule of thumb is: any controller that doesn't need the etcdclient is debuggable from local, everything else needs to be deployed in cluster. But maybe there is a better way to proxy into the host network nowadays.

ARCHITECTURE.md
tjungblu: Architecture.md is a good place for it, no doubts. Thanks for adding the bullets.

AGENTS.md
bhperry: Agreed. This is just the starting point.

It would be really nice to be able to dev locally against a remote cluster with etcd proxied. Something to think on for later.

tjungblu: yeah I can do that. Do you consider this ready for a v1? I can label it and we can override the failing job.

AGENTS.md
dusk125: Yeah that's correct, and that fact `so we can't ever be ahead in go version and we have to be careful about the dependencies.` gently reared its head again in the intern project we're currently working on. 

This is _currently_ true, but may, in the (hopefully short) future, it won't be - ART is looking into this so we could be "ahead"

PR #1640, opened 2026-06-23
NO-JIRA: quote bootstrap IP in bootkube etcd endpoints manifest

Rusty-Gopher:
During bootstrap, the rendered etcd-endpoints ConfigMap writes the bootstrap
IP into an annotation.

IPv4 bootstrap IPs render safely either way, but IPv6 addresses can contain
colon sequences such as `::`. When emitted as an unquoted YAML plain scalar,
a value like this can fail YAML parsing:

    alpha.installer.openshift.io/etcd-bootstrap: fd3b:8433:2a9:138::

Quote the rendered annotation value so the bootstrap IP is always treated as
a string. This preserves the rendered value for IPv4 and fixes IPv6 bootstrap
addresses.

Changed:

    alpha.installer.openshift.io/etcd-bootstrap: {{ .BootstrapIP }}

to:

    alpha.installer.openshift.io/etcd-bootstrap: "{{ .BootstrapIP }}"

Tested:

    go test ./pkg/cmd/render

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed YAML quoting for the etcd bootstrap configuration so the bootstrap endpoint renders as the intended interpolated value.
* **Tests**
  * Added an IPv6-specific rendering test to validate manifests when an IPv6 bootstrap IP is used.
  * Enhanced the render test setup to use a configurable bootstrap IP and verify the generated ConfigMap contents.

### DIFF
--- a/bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
+++ b/bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: openshift-etcd
 {{- if ne .BootstrapScalingStrategy "BootstrapInPlaceStrategy" }}
   annotations:
-    alpha.installer.openshift.io/etcd-bootstrap: {{ .BootstrapIP }}
+    alpha.installer.openshift.io/etcd-bootstrap: "{{ .BootstrapIP }}"
 {{- else }}
 data:
   {{ .EtcdEndpointConfigmapData }}
--- a/pkg/cmd/render/render_test.go
+++ b/pkg/cmd/render/render_test.go
@@ -15,6 +15,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/api/features"
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/ceohelpers"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
@@ -320,10 +321,30 @@ func TestRenderIpv6(t *testing.T) {
 		infraConfig:          infraConfig,
 		clusterConfigMap:     clusterConfigMap,
 	}
-	testRender(t, config)
+	dir := renderManifests(t, config)
+
+	etcdEndpointsPath := filepath.Join(dir, "manifests", "00_etcd-endpoints-cm.yaml")
+	etcdEndpointsBytes, err := os.ReadFile(etcdEndpointsPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	etcdEndpoints := &corev1.ConfigMap{}
+	if err := yaml.Unmarshal(etcdEndpointsBytes, etcdEndpoints); err != nil {
+		t.Fatal(err)
+	}
+
+	expectedBootstrapIP := "2001:db8::1"
+	if got := etcdEndpoints.Annotations["alpha.installer.openshift.io/etcd-bootstrap"]; got != expectedBootstrapIP {
+		t.Errorf("etcd bootstrap annotation want: %q got: %q", expectedBootstrapIP, got)
+	}
 }
 
 func testRender(t *testing.T, tc *testConfig) {
+	renderManifests(t, tc)
+}
+
+func renderManifests(t *testing.T, tc *testConfig) string {
 	var errOut io.Writer
 	dir := t.TempDir()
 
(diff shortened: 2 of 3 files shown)


Rusty-Gopher:
Hey, 

While working on this fix for my usecase, i noticed that testrender() was actually never called with IPv6 bootstrap it, which was exactly my usecase.

The IPv6 network configs were added in 876221af0 but only for testTemplateData() — the full render pipeline was never exercised with an IPv6 address.

Now i have added a TestRenderIPv6 to cover this gap.
Please let me know if you have any feedback on this approach.

Thanks

bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   namespace: openshift-etcd
 {{- if ne .BootstrapScalingStrategy "BootstrapInPlaceStrategy" }}
   annotations:
-    alpha.installer.openshift.io/etcd-bootstrap: {{ .BootstrapIP }}
+    alpha.installer.openshift.io/etcd-bootstrap: "{{ .BootstrapIP }}"
```
tjungblu, line 8: is this also a problem here?

https://github.com/openshift/cluster-etcd-operator/blob/main/pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go#L124

pkg/cmd/render/render_test.go
```diff
@@ -310,7 +311,40 @@ func TestRenderIpv4(t *testing.T) {
...
+
+	expectedBootstrapIP := "2001:db8::1"
+	if got := etcdEndpoints.Annotations["alpha.installer.openshift.io/etcd-bootstrap"]; got != expectedBootstrapIP {
+		t.Errorf("etcd bootstrap annotation want: %q got: %q", expectedBootstrapIP, got)
+	}
+}
+
 func testRender(t *testing.T, tc *testConfig) {
+	renderManifests(t, tc)
+}
```
tjungblu: no need to have a function just to call another function. Let's keep it at `testRender`

pkg/cmd/render/render_test.go
```diff
@@ -310,7 +311,40 @@ func TestRenderIpv4(t *testing.T) {
 	testRender(t, config)
 }
 
+func TestRenderIpv6(t *testing.T) {
```
tjungblu, line 309: funny that we don't already have a render test that looks at the generated yaml. Good addition.

pkg/cmd/render/render_test.go
```diff
@@ -310,7 +311,40 @@ func TestRenderIpv4(t *testing.T) {
...
+	orig := defaultBootstrapIPLocator
+	defaultBootstrapIPLocator = &fakeBootstrapIPLocator{ip: net.ParseIP("2001:db8::1")}
+	defer func() { defaultBootstrapIPLocator = orig }()
```
tjungblu: that looks strange, especially when we run tests in parallel. Do you mind to refactor the testConfig with it?

I don't really get why this needs to be on TestMain, I think we can remove that

tjungblu: this probably is a larger refactoring, but it would be useful to have it injected through TemplateData directly

tjungblu: this probably is a larger refactoring, but it would be useful to have it injected through TemplateData directly

bindata/bootkube/manifests/00_etcd-endpoints-cm.yaml
Rusty-Gopher: from my prespective, no, this one uses the k8s api client, not raw yaml files, so ipv6 colons are fine in json.
not the same problem.

pkg/cmd/render/render_test.go
Rusty-Gopher: Done, i have moved bootstrapIP into testConfig and removed TestMain.

Also i have added bootstrapIPLocator as a field on renderOpts, so the fake locator in injected through this structt, avoiding any gloabal state mutation.

Atleast from my side. tests are fully isolated and safe to run in parallel

Can you please check it and let me know if thats ok for you, thanks for the hint!

tjungblu: yeah works for me, thanks :)

tjungblu:
/lgtm
/approve

Thank you! I'll tag verified once everything passes in CI.

pkg/cmd/render/render_test.go
Rusty-Gopher: Thanks to you, that was quick, appericiate it.
If anything is needed from my side, just let me know :)

tjungblu:
/retitle NO-JIRA: quote bootstrap IP in bootkube etcd endpoints manifest

tjungblu:
/override ci/prow/e2e-aws-ovn-serial-1of2

unrelated and known failure in both jobs, no bootstrap error. Neither of them runs ipv6 bootstrap anyway.

PR #1593, opened 2026-04-15
CNTRLPLANE-2847: Support PKI customization

rh-roman:
Add support for ConfigurablePKI in both operational and installer paths:

  - Thread PKIProfileProvider through EtcdCertSignerController for runtime certificate rotation, reading configuration from cluster PKI CR
  - Add getPKIProfileProvider in render command to extract PKI config from install-config for bootstrap certificate generation

Switch `createCertForNode` from `ServingRotation` with a manual `CertificateExtensionFn` to `PeerRotation`, which handles dual ClientAuth+ServerAuth ExtKeyUsages internally. `PeerRotation` also enables proper PKI profile resolution using `CertificateTypePeer` (stronger of serving and client key configs).

When ConfigurablePKI is disabled or unconfigured, falls back to library-go defaults.

PR #1647, opened 2026-07-09
CNTRLPLANE-2847: render: Resolve featureSet into individual feature gates during bootstrap

hasbro17:
## Summary

  - Fix `getFeatureGatesStatus` to resolve `featureSet` (e.g., `TechPreviewNoUpgrade`) into individual feature gates during bootstrap rendering, so that `ConfigurablePKI` is correctly detected and bootstrap
  etcd-signer certs use the configured key algorithm
  - Add tests for `featureSet` resolution, `featureGates` override on top of `featureSet`, and an end-to-end test that mirrors the real CI install-config structure

  ## Background

  PR #1593 added ConfigurablePKI support to the etcd-operator, threading a `PKIProfileProvider` through both the bootstrap render path and the runtime controller. The installer counterpart
  (openshift/installer#10595) creates a PKI CR at install time with the desired key configuration (e.g., RSA-4096).

  CI job `e2e-aws-ovn-pki-rsa-techpreview` on the installer PR showed the etcd-signer cert was still RSA-2048 instead of the configured RSA-4096. All 7 other signers (managed by kube-apiserver-operator,
  machine-config-operator, etc.) correctly used RSA-4096, confirming the PKI CR was created correctly and the feature gate was active.

  ## Root cause

  The install-config in CI has:

  ```yaml
  featureSet: TechPreviewNoUpgrade
  pki:
    signerCertificates:
      key:
        algorithm: RSA
        rsa:
          keySize: 4096
  ```

  There are no explicit `featureGates` entries — the gates are implied by `featureSet: TechPreviewNoUpgrade`.

  `getFeatureGatesStatus()` in `pkg/cmd/render/render.go` only parsed explicit `featureGates: [Key=Value]` strings. It did not resolve `featureSet` into individual gates. So the enabled set was empty, the
  `ConfigurablePKI` guard was never entered, `pkiProfileProvider` stayed nil, and the bootstrap certs were created with the legacy fallback (`MakeSelfSignedCAConfigForDuration` → RSA-2048).

  At runtime, the operator correctly detected `ConfigurablePKI` from the cluster's FeatureGate CR status, but `needNewSigningCertKeyPair` only checks time-based conditions (expiry, refresh) — it doesn't detect
  key algorithm mismatches. So the already-valid RSA-2048 cert was never regenerated.

  ## Fix

  Use `features.FeatureSets()` from `openshift/api` to resolve the `featureSet` field into its constituent enabled/disabled gates, then layer explicit `featureGates` overrides on top (for `CustomNoUpgrade`
  support). This ensures `ConfigurablePKI` is detected during bootstrap rendering when installed with `featureSet: TechPreviewNoUpgrade`.

  ## Test plan

  - [x] New `Test_getFeatureGates` cases for `featureSet: TechPreviewNoUpgrade` and `featureSet` + `featureGates` override
  - [x] New `TestTemplateDataWithCustomPKIFeatureSet` end-to-end test using `featureSet: TechPreviewNoUpgrade` with PKI config (mirrors real CI install-config)
  - [x] Existing `TestTemplateDataWithCustomPKI` (explicit `featureGates`) still passes
  - [x] `make verify` passes
  - [ ] Re-run `e2e-aws-ovn-pki-rsa-techpreview` on installer PR with this change in the payload

## Summary by CodeRabbit

* **New Features**
  * Feature-gate detection now respects the cluster’s feature set, improving how enabled and disabled capabilities are determined.
  * Rendered output now better reflects custom PKI settings when a feature set is configured.

* **Bug Fixes**
  * Fixed an issue where explicit feature-gate settings could be skipped when a feature set was present.
  * Preserved previously determined feature gates instead of dropping them when feature-gate overrides are absent.

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -803,20 +803,32 @@ func getFeatureGatesStatus(installConfig map[string]any) (sets.Set[configv1.Feat
 	necessaryFeatureGates := []configv1.FeatureGateName{"ShortCertRotation"}
 	disabled.Insert(necessaryFeatureGates...)
 
-	featureGates, found := installConfig["featureGates"]
-	if !found {
-		return enabled, disabled
+	if featureSetRaw, found := installConfig["featureSet"]; found {
+		featureSet := configv1.FeatureSet(featureSetRaw.(string))
+		if resolved := features.FeatureSets(0, features.SelfManaged, featureSet); resolved != nil {
+			for _, fg := range resolved.Enabled {
+				enabled.Insert(fg.FeatureGateAttributes.Name)
+				disabled.Delete(fg.FeatureGateAttributes.Name)
+			}
+			for _, fg := range resolved.Disabled {
+				if !enabled.Has(fg.FeatureGateAttributes.Name) {
+					disabled.Insert(fg.FeatureGateAttributes.Name)
+				}
+			}
+		}
 	}
 
-	for _, featureGate := range featureGates.([]any) {
-		key := strings.Split(featureGate.(string), "=")[0]
-		value := strings.Split(featureGate.(string), "=")[1]
-		if value == "true" {
-			enabled = enabled.Insert(configv1.FeatureGateName(key))
-			disabled = disabled.Delete(configv1.FeatureGateName(key))
-		} else if value == "false" {
-			enabled = enabled.Delete(configv1.FeatureGateName(key))
-			disabled = disabled.Insert(configv1.FeatureGateName(key))
+	if featureGates, found := installConfig["featureGates"]; found {
+		for _, featureGate := range featureGates.([]any) {
+			key := strings.Split(featureGate.(string), "=")[0]
+			value := strings.Split(featureGate.(string), "=")[1]
+			if value == "true" {
+				enabled = enabled.Insert(configv1.FeatureGateName(key))
+				disabled = disabled.Delete(configv1.FeatureGateName(key))
+			} else if value == "false" {
+				enabled = enabled.Delete(configv1.FeatureGateName(key))
+				disabled = disabled.Insert(configv1.FeatureGateName(key))
+			}
 		}
 	}
 
--- a/pkg/cmd/render/render_test.go
+++ b/pkg/cmd/render/render_test.go
@@ -349,6 +349,64 @@ data:
           ecdsa:
             curve: P521
 `
+
+	clusterConfigMapWithCustomPKIAndFeatureSet = `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cluster-config-v1
+  namespace: kube-system
+data:
+  install-config: |
+    apiVersion: v1
+    baseDomain: gcp.devcluster.openshift.com
+    compute:
+    - architecture: amd64
+      hyperthreading: Enabled
+      name: worker
+      platform: {}
+      replicas: 3
+    controlPlane:
+      architecture: amd64
+      hyperthreading: Enabled
+      name: master
+      platform:
+        gcp:
+          osDisk:
+            DiskSizeGB: 128
+            DiskType: pd-ssd
+          type: n1-standard-4
+          zones:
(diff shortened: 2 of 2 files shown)


hasbro17:
Waiting for `e2e-aws-ovn-pki-rsa-techpreview` to validate that the operator picks up the `ConfigurablePKI` featuregate from TechPreview.

pkg/cmd/render/render.go
```diff
@@ -803,20 +803,32 @@ func getFeatureGatesStatus(installConfig map[string]any) (sets.Set[configv1.Feat
...
+				if !enabled.Has(fg.FeatureGateAttributes.Name) {
+					disabled.Insert(fg.FeatureGateAttributes.Name)
+				}
+			}
+		}
+	}
+
+	if featureGates, found := installConfig["featureGates"]; found {
+		for _, featureGate := range featureGates.([]any) {
+			key := strings.Split(featureGate.(string), "=")[0]
```
dusk125, line 823: Could do `key, value, found := strings.Cut(featureGate.(string), "=")` and that would address CodeRabbit's complaint

hasbro17:
Need a multi PR test to include the installer PKI support.

hasbro17:
/testwith e2e-aws-ovn-pki-rsa-techpreview https://github.com/openshift/installer/pull/10595

hasbro17:
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-pki-rsa-techpreview https://github.com/openshift/installer/pull/10595

sanchezl (commented):
The CEO render should not parse the feature gates from the raw install-config. Looking at other operator render (CKASO, CKCMO, MCO) , they read the pre-resolved `FeatureGate` CR from `/assets/manifests/` via `--rendered-manifest-files` and `--payload-version`. The `api-render` step in `bootkube.sh` already runs before `etcd-render` and produces the fully-resolved CR with `featureSet` expanded into individual gates, the CEO just never consumed it.

That architectural gap is why this bug exists in the first place: the other operators don't have this class of bug because they don't resolve feature gates themselves.

I've opened an alternative set of PRs that aligns CEO with the standard pattern, eliminating the entire class of feature-gate-resolution bugs rather than patching `getFeatureGatesStatus` for each new case:

1. **[#1648](https://github.com/openshift/cluster-etcd-operator/pull/1648)** — Adds `--rendered-manifest-files` and `--payload-version` support to the render command. When provided, reads the FeatureGate CR from pre-rendered manifests (the same pattern CKASO/CKCMO use). Includes a fallback that resolves `featureSet` from the install-config (covering this bug) so it **can land independently** without waiting for the installer change.

2. **[openshift/installer#10679](https://github.com/openshift/installer/pull/10679)** — Passes the two new flags to the `etcd-render` invocation in `bootkube.sh`, matching what's already done for `kube-apiserver-render`, `kube-controller-manager-render`, and `config-render`.

3. **[#1649](https://github.com/openshift/cluster-etcd-operator/pull/1649)** — Cleanup: removes the install-config fallback and makes both flags required, once the installer PR has merged.

A few additional notes on this PR if it moves forward instead:

- `features.FeatureSets(0, ...)` passes version `0`. Other callers (the installer, MCO tests) use the real major version (`4`). Today this is a no-op because no gate uses `inVersion()`, but if any gate adds a version constraint in the future, version `0` would silently exclude it.
- The unchecked type assertion on `featureSetRaw.(string)` (line 807) can panic on malformed install-config YAML — should use the comma-ok pattern.
- `validateECDSAP521Signer` is duplicated between `TestTemplateDataWithCustomPKI` and `TestTemplateDataWithCustomPKIFeatureSet` — worth extracting to a shared helper.

sanchezl (changes requested):
See https://github.com/openshift/cluster-etcd-operator/pull/1647#pullrequestreview-4668285659

hasbro17:
Replaced by https://github.com/openshift/cluster-etcd-operator/pull/1648

PR #1648, opened 2026-07-10
CNTRLPLANE-2847: render: Read FeatureGate CR from rendered manifests

sanchezl:
## Summary

- Replace manual install-config feature gate parsing with the standard operator render pattern: read the pre-resolved FeatureGate CR from rendered manifests via `--rendered-manifest-files` and `--payload-version`
- Include a fallback that resolves `featureSet` and `featureGates` from the install-config when the new flags are not yet available
- Fix the root cause of the `featureSet` resolution gap that [PR #1647](https://github.com/openshift/cluster-etcd-operator/pull/1647) addresses

## Background

The CEO render was the only operator that parsed feature gates directly from the raw install-config (`featureGates: [Key=Value]` strings). Every other operator render (CKASO, CKCMO, MCO) reads the pre-resolved `FeatureGate` CR from `/assets/manifests/` — the `api-render` step in `bootkube.sh` runs first and resolves `featureSet` (e.g. `TechPreviewNoUpgrade`) into individual gates, writing the result into the CR status.

Because the CEO was parsing the raw install-config, it missed `featureSet` resolution entirely, causing `ConfigurablePKI` to not be detected when using `featureSet: TechPreviewNoUpgrade` instead of explicit `featureGates` entries.

## Approach

**Primary path** (when `--rendered-manifest-files` and `--payload-version` are provided): Read the pre-resolved FeatureGate CR from rendered manifests, matching the pattern used by CKASO and CKCMO. This eliminates the entire class of feature-gate-resolution bugs.

**Fallback path** (when the flags are not provided): Resolve `featureSet` into individual gates via `features.FeatureSets()` from openshift/api, then layer explicit `featureGates` overrides on top. This fallback uses `strings.Cut` for safe parsing and handles the `featureSet` → `featureGates` override correctly. It should be removed once the installer carries the coordinated change.

## Changes

**`pkg/cmd/render/render.go`:**
- Add optional `--rendered-manifest-files` and `--payload-version` flags
- Add `getFeatureGates()` dispatcher that tries the CR path, falls back to install-config
- Add `getFeatureGatesFromManifests()` — reads FeatureGate CR, returns enabled/disabled sets
- Add `getFeatureGatesFromInstallConfig()` — resolves `featureSet` + `featureGates` from install-config
- Remove the old `getFeatureGatesStatus()` function

**`pkg/cmd/render/render_test.go`:**
- Add `writeFeatureGateManifest()` test helper for the CR path
- Add `Test_getFeatureGatesFromInstallConfig` testing featureSet resolution, overrides
- Update test infrastructure to provide rendered manifest directories
- Update `TestTemplateDataWithCustomPKI` to provide ConfigurablePKI via FeatureGate CR

## Coordinated change

[openshift/installer#10679](https://github.com/openshift/installer/pull/10679) passes `--rendered-manifest-files` and `--payload-version` to the etcd-render invocation. **This CEO PR can land independently** — the fallback path handles the transition. Once the installer PR merges, a follow-up can remove the fallback.

## Test plan

- [x] `go test ./pkg/cmd/render/` — all tests pass (both primary and fallback paths)
- [x] `go vet ./pkg/cmd/render/` — clean
- [ ] CI

## Summary by CodeRabbit

* **New Features**
  * Rendering can now determine feature gate enablement from pre-rendered FeatureGate manifests supplied by the user.
  * Added CLI flags to specify rendered manifest locations and the payload version used during rendering.
* **Bug Fixes**
  * Bootstrap certificate secrets now consistently reflect enabled/disabled feature gates derived from the rendered manifests or the install configuration (including correct handling of configurable PKI selection).
* **Tests**
  * Updated the rendering/template test harness to generate a FeatureGate manifest at runtime and extended coverage for parsing feature gates from install configuration.

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -24,6 +24,7 @@ import (
 	configv1alpha1 "github.com/openshift/api/config/v1alpha1"
 	features "github.com/openshift/api/features"
 	"github.com/openshift/library-go/pkg/assets"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 	"github.com/openshift/library-go/pkg/pki"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -46,6 +47,9 @@ type renderOpts struct {
 	networkConfigFile    string
 	clusterConfigMapFile string
 	infraConfigFile      string
+
+	renderedManifestFiles []string
+	payloadVersion        string
 
 	delayedHABootstrapScalingStrategyMarker string
 	bootstrapIPLocator                      BootstrapIPLocator
@@ -86,6 +90,8 @@ func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&r.networkConfigFile, "network-config-file", "", "File containing the network.config.openshift.io manifest.")
 	fs.StringVar(&r.clusterConfigMapFile, "cluster-configmap-file", "", "File containing the cluster-config-v1 configmap.")
 	fs.StringVar(&r.infraConfigFile, "infra-config-file", "", "File containing infrastructure.config.openshift.io manifest.")
+	fs.StringArrayVar(&r.renderedManifestFiles, "rendered-manifest-files", nil, "Files or directories containing pre-rendered manifests, used to determine FeatureGate status.")
+	fs.StringVar(&r.payloadVersion, "payload-version", "", "Version that will eventually be placed into ClusterOperator.status. This normally comes from the CVO set via env var: OPERATOR_IMAGE_VERSION.")
 
 	// TODO(marun) Discover scaling strategy with less hack
 	fs.StringVar(&r.delayedHABootstrapScalingStrategyMarker, "delayed-ha-bootstrap-scaling-marker-file", "/assets/assisted-install-bootstrap", "Marker file that, if present, enables the delayed HA bootstrap scaling strategy")
@@ -279,7 +285,10 @@ func newTemplateData(opts *renderOpts) (*TemplateData, error) {
 			base64.StdEncoding.WithPadding(base64.NoPadding).EncodeToString([]byte(templateData.BootstrapIP)), templateData.BootstrapIP)
 	}
 
-	enabledFeatureGates, disabledFeatureGates := getFeatureGatesStatus(installConfig)
+	enabledFeatureGates, disabledFeatureGates, err := opts.getFeatureGates(installConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get feature gates: %w", err)
+	}
 
 	var pkiProfileProvider pki.PKIProfileProvider
 	if enabledFeatureGates.Has(features.FeatureGateConfigurablePKI) {
@@ -794,29 +803,136 @@ func getBootstrapScalingStrategy(installConfig map[string]any, delayedHAMarkerFi
 	return strategy, nil
 }
 
-// getFeatureGatesStatus returns the enabled and disabled feature gates.
-func getFeatureGatesStatus(installConfig map[string]any) (sets.Set[configv1.FeatureGateName], sets.Set[configv1.FeatureGateName]) {
-	enabled, disabled := sets.Set[configv1.FeatureGateName]{}, sets.Set[configv1.FeatureGateName]{}
+// getFeatureGates returns the enabled and disabled feature gate sets.
+//
+// Primary path: when --rendered-manifest-files and --payload-version are
+// provided, reads the pre-resolved FeatureGate CR from rendered manifests.
+// This is the standard operator render pattern used by CKASO and CKCMO.
+//
+// Fallback path: when rendered manifests are not available, resolves
+// feature gates from the install-config's featureSet and featureGates
+// fields directly. This fallback exists for transition until the
+// installer passes the new flags to the etcd-render invocation, and
+// should be removed once openshift/installer carries the change.
+func (r *renderOpts) getFeatureGates(installConfig map[string]any) (sets.Set[configv1.FeatureGateName], sets.Set[configv1.FeatureGateName], error) {
+	if len(r.renderedManifestFiles) > 0 && len(r.payloadVersion) > 0 {
+		return r.getFeatureGatesFromManifests()
+	}
+	klog.Warningf("--rendered-manifest-files or --payload-version not provided, falling back to install-config feature gate parsing")
+	enabled, disabled := getFeatureGatesFromInstallConfig(installConfig)
+	return enabled, disabled, nil
+}
 
-	// On bootstrap we might not be able to fetch a list of all feature gates
-	// Hardcode a list of necessary for bootstrap here
-	necessaryFeatureGates := []configv1.FeatureGateName{"ShortCertRotation"}
-	disabled.Insert(necessaryFeatureGates...)
+// getFeatureGatesFromManifests reads the FeatureGate CR from pre-rendered
+// manifests and returns enabled/disabled sets.
+func (r *renderOpts) getFeatureGatesFromManifests() (sets.Set[configv1.FeatureGateName], sets.Set[configv1.FeatureGateName], error) {
+	fg, err := r.readFeatureGate()
+	if err != nil {
+		return nil, nil, err
+	}
+	accessor, err := featuregates.NewHardcodedFeatureGateAccessFromFeatureGate(fg, r.payloadVersion)
+	if err != nil {
+		return nil, nil, err
+	}
+	featureGate, err := accessor.CurrentFeatureGates()
+	if err != nil {
+		return nil, nil, err
+	}
+	enabled := sets.New[configv1.FeatureGateName]()
+	disabled := sets.New[configv1.FeatureGateName]()
+	for _, name := range featureGate.KnownFeatures() {
+		if featureGate.Enabled(name) {
+			enabled.Insert(name)
+		} else {
+			disabled.Insert(name)
+		}
+	}
+	return enabled, disabled, nil
+}
 
-	featureGates, found := installConfig["featureGates"]
-	if !found {
-		return enabled, disabled
+// readFeatureGate scans rendered manifest files/directories for a FeatureGate CR.
+func (r *renderOpts) readFeatureGate() (*configv1.FeatureGate, error) {
+	for _, manifestPath := range r.renderedManifestFiles {
+		info, err := os.Stat(manifestPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to stat %s: %w", manifestPath, err)
+		}
+
+		var files []string
+		if info.IsDir() {
+			entries, err := os.ReadDir(manifestPath)
+			if err != nil {
+				return nil, fmt.Errorf("failed to read directory %s: %w", manifestPath, err)
+			}
+			for _, entry := range entries {
+				if entry.IsDir() {
+					continue
+				}
+				files = append(files, filepath.Join(manifestPath, entry.Name()))
+			}
+		} else {
+			files = []string{manifestPath}
+		}
+
+		for _, file := range files {
+			data, err := os.ReadFile(file)
+			if err != nil {
+				continue
+			}
+			jsonBytes, err := yaml.YAMLToJSON(data)
+			if err != nil {
+				continue
+			}
+			fg := &configv1.FeatureGate{}
+			if err := json.Unmarshal(jsonBytes, fg); err != nil {
+				continue
+			}
+			if fg.APIVersion == "config.openshift.io/v1" && fg.Kind == "FeatureGate" {
+				return fg, nil
+			}
+		}
+	}
+	return nil, fmt.Errorf("no FeatureGate manifest found in %v", r.renderedManifestFiles)
+}
+
+// getFeatureGatesFromInstallConfig resolves feature gates from the
+// install-config featureSet and featureGates fields. This is the fallback
+// for when rendered manifests are not available.
+// TODO: Remove this fallback once the installer passes --rendered-manifest-files
+// and --payload-version to the etcd-render invocation.
+func getFeatureGatesFromInstallConfig(installConfig map[string]any) (sets.Set[configv1.FeatureGateName], sets.Set[configv1.FeatureGateName]) {
+	enabled := sets.New[configv1.FeatureGateName]()
+	disabled := sets.New[configv1.FeatureGateName]()
+
+	featureSet := configv1.Default
+	if featureSetRaw, found := installConfig["featureSet"]; found {
+		if featureSetStr, ok := featureSetRaw.(string); ok {
+			featureSet = configv1.FeatureSet(featureSetStr)
+		}
+	}
+	if resolved := features.FeatureSets(0, features.SelfManaged, featureSet); resolved != nil {
+		for _, fg := range resolved.Enabled {
+			enabled.Insert(fg.FeatureGateAttributes.Name)
+		}
+		for _, fg := range resolved.Disabled {
+			disabled.Insert(fg.FeatureGateAttributes.Name)
+		}
 	}
 
-	for _, featureGate := range featureGates.([]any) {
-		key := strings.Split(featureGate.(string), "=")[0]
-		value := strings.Split(featureGate.(string), "=")[1]
-		if value == "true" {
-			enabled = enabled.Insert(configv1.FeatureGateName(key))
-			disabled = disabled.Delete(configv1.FeatureGateName(key))
-		} else if value == "false" {
-			enabled = enabled.Delete(configv1.FeatureGateName(key))
-			disabled = disabled.Insert(configv1.FeatureGateName(key))
+	if featureGatesRaw, found := installConfig["featureGates"]; found {
+		for _, entry := range featureGatesRaw.([]any) {
+			key, value, found := strings.Cut(entry.(string), "=")
+			if !found {
+				continue
+			}
+			name := configv1.FeatureGateName(key)
+			if value == "true" {
+				enabled.Insert(name)
+				disabled.Delete(name)
+			} else if value == "false" {
+				enabled.Delete(name)
+				disabled.Insert(name)
+			}
 		}
 	}
 
--- a/pkg/cmd/render/render_test.go
+++ b/pkg/cmd/render/render_test.go
@@ -13,15 +13,12 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
-
 	"github.com/ghodss/yaml"
 	configv1 "github.com/openshift/api/config/v1"
 	configv1alpha1 "github.com/openshift/api/config/v1alpha1"
 	"github.com/openshift/api/features"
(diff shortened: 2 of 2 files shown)


pkg/cmd/render/render.go
```diff
@@ -794,29 +803,136 @@ func getBootstrapScalingStrategy(installConfig map[string]any, delayedHAMarkerFi
...
+// getFeatureGatesFromInstallConfig resolves feature gates from the
+// install-config featureSet and featureGates fields. This is the fallback
+// for when rendered manifests are not available.
+// TODO: Remove this fallback once the installer passes --rendered-manifest-files
+// and --payload-version to the etcd-render invocation.
+func getFeatureGatesFromInstallConfig(installConfig map[string]any) (sets.Set[configv1.FeatureGateName], sets.Set[configv1.FeatureGateName]) {
+	enabled := sets.New[configv1.FeatureGateName]()
+	disabled := sets.New[configv1.FeatureGateName]()
+
+	if featureSetRaw, found := installConfig["featureSet"]; found {
```
hasbro17, line 908: I think this fallback path is still causing issues on non-techpreview installs. I'm seeing repeated panics in the bootkube logs for etcd-render.

```console
Jul 10 03:42:15 ip-10-0-97-204 bootkube.sh[5696]: W0710 03:42:15.634081       1 render.go:821] --rendered-manifest-files or --payload-version not provided, falling back to install-config feature gate parsing
Jul 10 03:42:15 ip-10-0-97-204 bootkube.sh[5696]: panic: feature "ShortCertRotation" is not registered in FeatureGates []
...
Jul 10 03:42:15 ip-10-0-97-204 systemd[1]: bootkube.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Jul 10 03:42:15 ip-10-0-97-204 systemd[1]: bootkube.service: Failed with result 'exit-code'.
Jul 10 03:42:15 ip-10-0-97-204 systemd[1]: bootkube.service: Consumed 20.897s CPU time, 1.3G memory peak.
Jul 10 03:42:20 ip-10-0-97-204 systemd[1]: bootkube.service: Scheduled restart job, restart counter is at 1.
```

Will share the review bot analysis below for more details.

hasbro17: 🤖 
The fallback path in `getFeatureGatesFromInstallConfig` panics on Default installs where `featureSet` is absent from the install-config (the standard case — `configv1.Default` is `""` and `omitempty` causes the
  installer to omit it entirely).

  When neither `featureSet` nor `featureGates` is present, the function returns empty enabled/disabled sets. These flow through `createBootstrapCertSecrets` → `createCertSecrets` (certs.go:58) →
  `NewEtcdCertSignerController` (etcdcertsignercontroller.go:159) → `NewHardcodedFeatureGateAccess([], [])`. The controller then calls `featureGates.Enabled(features.FeatureShortCertRotation)`, and since
  ShortCertRotation is in neither set, `featuregate.go:39` panics:

  ```go
  panic(fmt.Errorf("feature %q is not registered in FeatureGates %v", key, f.KnownFeatures()))
  ```

  The old `getFeatureGatesStatus` prevented this with `disabled.Insert(necessaryFeatureGates...)` where `necessaryFeatureGates = []configv1.FeatureGateName{"ShortCertRotation"}`. The new code drops that safety
  net without a complete replacement for the absent-featureSet case.

  **CI evidence:** The bootkube.log from [this job's artifacts](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-e2e-aws-ovn-serial-1of2/2075418387432869888/artifacts/e2e-aws-ovn-serial/ipi-install-install/artifacts/expanded-log-bundle/log-bundle-20260710040000/bootstrap/journals/bootkube.log)
 shows the `etcd-render` container crashing twice with the panic stack trace:

  ```console
  Jul 10 03:42:15 bootkube.sh[5696]: etcdcertsignercontroller.go:159 +0x528
  Jul 10 03:42:15 bootkube.sh[5696]: certs.go:58 +0xb49
  Jul 10 03:42:15 bootkube.sh[5696]: render.go:301 +0xbdb
  ...
  Jul 10 03:42:15 systemd[1]: bootkube.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
  Jul 10 03:42:15 systemd[1]: bootkube.service: Failed with result 'exit-code'.
  Jul 10 03:42:20 systemd[1]: bootkube.service: Scheduled restart job, restart counter is at 1.
  ```

  Then the same panic repeats at `03:42:26`. The job eventually passes because bootkube.service keeps restarting and on a later attempt the `api-render` step has already produced the FeatureGate CR manifest, so
  the primary `getFeatureGatesFromManifests` path succeeds. But the fallback panics every time it's hit, adding fragility and latency to bootstrap.

  The install-config for this job ([install-config.yaml](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-e2e-aws-ovn-serial-1of2/2075418387432869888/artifacts/e2e-aws-ovn-serial/ipi-install-install/artifacts/install-config.yaml)) confirms no `featureSet` field is present.

  **Suggested fix:** Always resolve the featureSet, defaulting to `configv1.Default` when absent:

  ```go
  featureSet := configv1.Default
  if featureSetRaw, found := installConfig["featureSet"]; found {
      if featureSetStr, ok := featureSetRaw.(string); ok {
          featureSet = configv1.FeatureSet(featureSetStr)
      }
  }
  if resolved := features.FeatureSets(0, features.SelfManaged, featureSet); resolved != nil {
      // populate enabled/disabled from resolved
  }
  ```

  This populates all known gates (including ShortCertRotation in disabled) for every featureSet, making the old hardcode genuinely unnecessary.

hasbro17: I know this goes away when https://github.com/openshift/installer/pull/10679 lands but this has to come first I think.

hasbro17 (changes requested):
See: https://github.com/openshift/cluster-etcd-operator/pull/1648/changes#r3561973119
**TLDR:** The fallback `getFeatureGatesFromInstallConfig` must resolve `configv1.Default` when `featureSet` is absent from install-config, otherwise the empty gate sets cause a panic at
  `featureGate.Enabled(ShortCertRotation)` during bootstrap. See [this job's
  bootkube.log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-e2e-aws-ovn-serial-1of2/2075418387432869888/artifacts/e2e-aws-ovn-serial/ipi-install-install/artifacts/expanded-log-bundle/log-bundle-20260710040000/bootstrap/journals/bootkube.log).

pkg/cmd/render/render.go
```diff
@@ -794,33 +823,62 @@ func getBootstrapScalingStrategy(installConfig map[string]any, delayedHAMarkerFi
...
+			if err != nil {
+				continue
+			}
+			jsonBytes, err := yaml.YAMLToJSON(data)
+			if err != nil {
+				continue
+			}
+			fg := &configv1.FeatureGate{}
+			if err := json.Unmarshal(jsonBytes, fg); err != nil {
+				continue
```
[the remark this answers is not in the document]
sanchezl: The `continue` is intentional — this function scans a directory of mixed manifests (e.g. Deployment, ConfigMap, FeatureGate). Files that are not valid YAML or do not unmarshal into a FeatureGate are expected and must be skipped. Returning an error here would break on the first non-FeatureGate manifest.

pkg/cmd/render/render.go
```diff
@@ -794,29 +803,136 @@ func getBootstrapScalingStrategy(installConfig map[string]any, delayedHAMarkerFi
...
+// fields directly. This fallback exists for transition until the
+// installer passes the new flags to the etcd-render invocation, and
+// should be removed once openshift/installer carries the change.
+func (r *renderOpts) getFeatureGates(installConfig map[string]any) (sets.Set[configv1.FeatureGateName], sets.Set[configv1.FeatureGateName], error) {
+	if len(r.renderedManifestFiles) > 0 && len(r.payloadVersion) > 0 {
+		return r.getFeatureGatesFromManifests()
+	}
+	klog.Warningf("--rendered-manifest-files or --payload-version not provided, falling back to install-config feature gate parsing")
+	enabled, disabled := getFeatureGatesFromInstallConfig(installConfig)
+	return enabled, disabled, nil
```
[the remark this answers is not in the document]
sanchezl: This fallback is transitional — it exists until the installer passes both flags (openshift/installer#10679). During the transition, having neither flag set is the expected case on bootstrap. Erroring on partial config would break real bootstrap scenarios for no safety gain.

pkg/cmd/render/render.go
```diff
@@ -794,29 +803,136 @@ func getBootstrapScalingStrategy(installConfig map[string]any, delayedHAMarkerFi
...
+				continue
+			}
+			name := configv1.FeatureGateName(key)
+			if value == "true" {
+				enabled.Insert(name)
+				disabled.Delete(name)
+			} else if value == "false" {
+				enabled.Delete(name)
+				disabled.Insert(name)
+			}
```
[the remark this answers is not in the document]
sanchezl: The `featureGates` field in install-config has a well-defined schema (`[]string` of `Key=Value` entries) enforced by the installer validation. This fallback path is also temporary and will be removed once the installer passes `--rendered-manifest-files`. Adding defensive checks for a schema the installer already validates adds complexity to code that is going away.

hasbro17:
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-pki-rsa-techpreview https://github.com/openshift/installer/pull/10595

pkg/cmd/render/render.go
```diff
@@ -86,6 +90,8 @@ func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&r.networkConfigFile, "network-config-file", "", "File containing the network.config.openshift.io manifest.")
 	fs.StringVar(&r.clusterConfigMapFile, "cluster-configmap-file", "", "File containing the cluster-config-v1 configmap.")
 	fs.StringVar(&r.infraConfigFile, "infra-config-file", "", "File containing infrastructure.config.openshift.io manifest.")
+	fs.StringArrayVar(&r.renderedManifestFiles, "rendered-manifest-files", nil, "Files or directories containing pre-rendered manifests, used to determine FeatureGate status.")
+	fs.StringVar(&r.payloadVersion, "payload-version", "", "Version that will eventually be placed into ClusterOperator.status. This normally comes from the CVO set via env var: OPERATOR_IMAGE_VERSION.")
```
tthvo, line 94: These flags `rendered-manifest-files` and `payload-version` are already defined as generic render flags in [openshift/library-go@pkg/operator/render/options/generic.go#L57-L59](
https://github.com/openshift/library-go/blob/795ac1a480b5d55a25f907afe7b64b676fb8fb38/pkg/operator/render/options/generic.go#L57-L59), right? I'm referencing [kube-controller-render](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/main/pkg/cmd/render/render.go), which also accepts the 2 flags.


 We may already have existing feature-gate funcs to help us here too, for [example](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/209d5915dc12d517be79430e4bf4e8b0f88e405b/pkg/cmd/render/render.go#L244) 👀

sanchezl: Yes, CKCMO and CKASO embed `genericrenderoptions.GenericOptions` which provides `rendered-manifest-files`, `payload-version`, and `FeatureGates()` out of the box.

The etcd render command does not use library-go's generic render options infrastructure at all, it has its own `renderOpts` struct with manually-defined flags. Refactoring it to embed `GenericOptions` would be a much larger change (the entire render pipeline would need to be restructured to match the pattern CKCMO uses with `ManifestOptions` + `GenericOptions`).

This PR takes the narrower approach doing the minimum needed to unblock the configurable PKI work. A full refactor to align with the generic render framework is worth doing but can be a followup effort.

sanchezl:
/testwith openshift/cluster-etcd-operator/main/e2e-aws-ovn-pki-rsa-techpreview https://github.com/openshift/installer/pull/10595

pkg/cmd/render/render.go
tthvo: > The etcd render command does not use library-go's generic render options infrastructure at all

Oh whoops, my bad, I didn't realize that at all! Thanks for the details :pray:

hasbro17:
Alright so this works now looking at the etcd-signer that's correctly configured per the PKI config to be RSA 4096 as the test does it.
https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-etcd-operator-1648-openshift-installer-10595-e2e-aws-ovn-pki-rsa-techpreview/2077020289362825216
```console
--- Checking: etcd-signer (openshift-etcd/etcd-signer) ---
...
        Subject: CN=openshift-etcd_etcd-signer@1784038543
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (4096 bit)
```

There's a typo bug in my test. The verification script checks for secret `etcd-metrics-signer` but the operator creates it as `etcd-metric-signer` (no trailing 's'):
```console
--- Checking: etcd-metrics-signer (openshift-etcd/etcd-metrics-signer) ---
Could not find certificate from <stdin>

  FAIL: Expected algorithm 'rsaEncryption', got ''
  FAIL: Expected 'Public-Key: (4096 bit)', got 'not found'
```

Will fix that separately. But this should be good to move forward.

pkg/cmd/render/render_test.go
```diff
@@ -598,10 +647,15 @@ func TestTemplateDataWithCustomPKI(t *testing.T) {
 		}
 	}
 
+	enabled, disabled := defaultFeatureGates()
+	enabled = append(enabled, features.FeatureGateConfigurablePKI)
```
hasbro17, line 651: The test appends `ConfigurablePKI` to the enabled list but doesn't remove it from disabled (which `defaultFeatureGates()` populates from the Default feature set). The resulting FeatureGate manifest has
  ConfigurablePKI in both sections.
Works only because `Enabled()` checks the enabled slice first. 
Fix: filter `ConfigurablePKI` out of the disabled slice when adding it to enabled.

hasbro17 (approved):
/approve

Just one nit

sanchezl:
## Verification Analysis

### What changed

1. **Primary path:** Added `--rendered-manifest-files` and `--payload-version` flags to the render command. When both are provided, reads the pre-resolved FeatureGate CR from rendered manifests (matching the CKASO/CKCMO pattern).
2. **Fallback path:** When flags are not provided, resolves feature gates from install-config's `featureSet` field, defaulting to `configv1.Default` when absent.
3. **Panic fix:** The fallback now always resolves all known feature gates for the featureSet. Previously, an absent `featureSet` (the standard case for Default installs) returned empty gate sets, causing `featureGate.Enabled(ShortCertRotation)` to panic.

### CI Results (9/9 passing)

| Check | Status | Build |
|-------|--------|-------|
| `unit` | PASS | [2076838282171256832](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-unit/2076838282171256832) |
| `verify` | PASS | [2076838282200616960](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-verify/2076838282200616960) |
| `verify-deps` | PASS | [2076674200155721728](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-verify-deps/2076674200155721728) |
| `images` | PASS | [2076674189242142720](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-images/2076674189242142720) |
| `okd-scos-images` | PASS | [2076756075503161344](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-okd-scos-images/2076756075503161344) |
| `configmap-scale` | PASS | [2076674188441030656](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-configmap-scale/2076674188441030656) |
| `e2e-agnostic-ovn` | PASS | [2077020709137158144](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-e2e-agnostic-ovn/2077020709137158144) |
| `e2e-gcp-operator` | PASS | [2077020724853215232](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-e2e-gcp-operator/2077020724853215232) |
| `e2e-aws-ovn-single-node` | PASS | [2077020725683687424](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-etcd-operator/1648/pull-ci-openshift-cluster-etcd-operator-main-e2e-aws-ovn-single-node/2077020725683687424) |

### Artifact-level evidence: Panic fix confirmed

The fallback path (`--rendered-manifest-files` not provided) was exercised during bootstrap in all three e2e jobs. Bootkube logs confirm the fix:

| Job | Before fix (build 2075418...) | After fix (build 2077020...) |
|-----|------|------|
| `e2e-agnostic-ovn` | **122 panics** — `feature "ShortCertRotation" is not registered in FeatureGates []` | **0 panics** |
| `e2e-aws-ovn-single-node` | panics (cluster never came up) | **0 panics** |

Both jobs logged `falling back to install-config feature gate parsing` confirming the fallback was hit, and bootstrap completed successfully without panics.

### Multi-PR payload job

The `/testwith` job testing this PR + installer#10595 (`e2e-aws-ovn-pki-rsa-techpreview`) failed due to a bug in the installer PR's verification script — it checks for a secret named `etcd-metrics-signer` (with 's') but the actual secret has always been named `etcd-metric-signer` (without 's') since its introduction in 2020. The etcd-signer check (same code path, same PKI provider) passed. Not caused by this PR.

### Caveats

- 6 e2e jobs are still pending (label-gated, not code issues)
- The primary manifest-reading path cannot be verified until installer#10595 lands and passes the new flags

sanchezl:
/verified by pull-ci-openshift-cluster-etcd-operator-main-e2e-agnostic-ovn,pull-ci-openshift-cluster-etcd-operator-main-e2e-aws-ovn-single-node,pull-ci-openshift-cluster-etcd-operator-main-e2e-gcp-operator

PR #1650, opened 2026-07-14
OCPEDGE-2707: Add TNF Pacemaker PrometheusRule alerts

fonta-rh:
## Summary

- Add 12 TNF-specific Prometheus alerting rules in a new `tnf-pacemaker.rules` group
- 2 cluster-level, 6 node-level, 4 resource-level alerts (6 critical + 6 warning)
- Alerts query gauge metrics exposed by the TNF metrics controller (#1634 / [OCPEDGE-2705](https://redhat.atlassian.net/browse/OCPEDGE-2705))
- New `jsonnet/tnf-alerts.libsonnet` keeps TNF alerts separate from etcd-mixin overrides in `custom.libsonnet`
- Add disruption tracker (`tnf_resource_disruption_total`) for API-blind-spot resource transitions

## Alert Overview

| Level | Alert | Severity | For |
|-------|-------|----------|-----|
| Cluster | `TNFNodeCountMismatch` | critical | 5m |
| Cluster | `TNFClusterInMaintenance` | warning | 2m |
| Node | `TNFNodeOffline` | critical | 2m |
| Node | `TNFNodeFencingUnavailable` | critical | 5m |
| Node | `TNFNodeFencingDegraded` | warning | 10m |
| Node | `TNFNodeUnclean` | critical | 5m |
| Node | `TNFNodeInMaintenance` | warning | 2m |
| Node | `TNFNodeStandby` | warning | 5m |
| Resource | `TNFResourceStopped` | critical | 5m |
| Resource | `TNFResourceFailed` | critical | 2m |
| Resource | `TNFResourceUnmanaged` | warning | 5m |
| Resource | `TNFResourceDisabled` | warning | 5m |

## Test plan

- [x] `hack/generate.sh` regenerates manifest with all 12 rules in `tnf-pacemaker.rules` group
- [x] `make build` passes
- [x] `make test-unit` passes (29 alert validation + 10 disruption tracker tests)
- [x] Deployed on dev-scripts TNF cluster with CI image — all 12 alerts and 25 metrics present
- [x] Scenario 1: TNFClusterInMaintenance — fired warning, cleared on revert ✓
- [x] Scenario 2: TNFNodeInMaintenance — fired warning scoped to correct node, cleared ✓
- [x] Scenario 3: TNFNodeStandby — fired warning scoped to correct node, cleared ✓
- [x] Scenario 4: TNFResourceUnmanaged — fired warning for both Etcd instances, cleared ✓
- [x] Scenario 5: TNFResourceDisabled (etcd) — circular dependency (see below)
- [x] Scenario 6: TNFNodeFencingUnavailable — fired critical for both nodes, cleared ✓
- [x] Scenario 7: TNFNodeOffline — fencing recovers node in < 2m (see below)
- [x] Scenario 8: TNFResourceStopped — circular dependency (see below)

### Etcd circular dependency (Scenarios 5, 7, 8)

Disabling the etcd Pacemaker resource or taking a node offline causes etcd quorum loss in a 2-node cluster. This brings down the API server, Prometheus (can't evaluate alerting rules), and the CEO pod (liveness probe → crash-loop). The alerts `TNFResourceDisabled`, `TNFResourceStopped`, and `TNFNodeOffline` **work correctly for non-etcd resources and non-quorum-breaking scenarios**, but cannot fire when etcd itself is affected because the observability stack depends on etcd.

The in-memory disruption counter also resets on crash-loop restarts. This is an inherent constraint of in-cluster monitoring — external monitoring (Pacemaker `crm_mon`, BMC-level alerts) is needed to cover etcd-level failures.

For `TNFNodeOffline` specifically: when fencing works correctly, it recovers the node in < 2 minutes, so the `for: 2m` duration is never satisfied — this is by design. The alert fires when fencing fails or takes longer than expected.

## Notes

- All alerts carry explicit `runbook_url`, `summary`, and `description` annotations
- `TNFNodeFencingDegraded` uses compound PromQL `and` to fire only when fencing is available but not fully healthy
- `TNFNodeUnclean` and `TNFNodeCountMismatch` alert on API-blind-spot metrics — expected to have zero practical firings in normal 2-node topology but included for completeness
- Disruption counter (`tnf_resource_disruption_total`) tracks resource started→stopped transitions as an in-memory Prometheus counter exposed on `:8443/metrics`

## Summary

- **New Features**
  - Added 12 TNF Pacemaker Prometheus alert rules (node count, maintenance/standby, offline/unclean states, fencing health, and TNF resource failures) with severity and runbook links.
  - Added disruption monitoring that counts TNF resources transitioning from started to stopped.

- **Bug Fixes**
  - Improved Pacemaker status watch/list handling and ensured disruption tracking uses the latest retrieved node statuses.

- **Tests**
  - Added coverage for disruption tracking behavior and validation of the generated TNF Pacemaker alert set (definitions, counts, severities, annotations).

### DIFF
--- a/jsonnet/tnf-alerts.libsonnet
+++ b/jsonnet/tnf-alerts.libsonnet
@@ -60,7 +60,7 @@
           },
           {
             alert: 'TNFNodeFencingDegraded',
-            expr: 'tnf_node_fencing_healthy == 0 and tnf_node_fencing_available == 1',
+            expr: 'tnf_node_fencing_healthy == 0 and tnf_node_fencing_available == 1 and on(node) tnf_node_in_service == 1',
             'for': '10m',
             labels: {
               severity: 'warning',
--- a/jsonnet/tnf-alerts.libsonnet
+++ b/jsonnet/tnf-alerts.libsonnet
@@ -86,7 +86,7 @@
           },
           {
             alert: 'TNFNodeInMaintenance',
-            expr: 'tnf_node_in_service == 0',
+            expr: 'tnf_node_in_service == 0 and tnf_cluster_in_service == 1',
             'for': '2m',
             labels: {
               severity: 'warning',
--- a/jsonnet/tnf-alerts.libsonnet
+++ b/jsonnet/tnf-alerts.libsonnet
@@ -1,0 +1,170 @@
+{
+  prometheusRules:: {
+    groups: [
+      {
+        name: 'tnf-pacemaker.rules',
+        rules: [
+          // Cluster-level alerts
+          {
+            alert: 'TNFNodeCountMismatch',
+            expr: 'tnf_cluster_node_count_as_expected == 0',
+            'for': '5m',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              summary: 'TNF cluster node count does not match expected topology.',
+              description: 'The TNF cluster reports an unexpected number of nodes. This may indicate a node was added or removed outside the expected two-node topology, potentially breaking quorum assumptions.',
+              runbook_url: 'https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/TNFNodeCountMismatch.md',
+            },
+          },
+          {
+            alert: 'TNFClusterInMaintenance',
+            expr: 'tnf_cluster_in_service == 0',
+            'for': '2m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'TNF cluster is in maintenance mode.',
+              description: 'The TNF Pacemaker cluster has been placed into maintenance mode. Automated failover and fencing are disabled. The cluster will not recover from failures until maintenance mode is cleared.',
+              runbook_url: 'https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/TNFClusterInMaintenance.md',
+            },
+          },
+          // Node-level alerts
+          {
+            alert: 'TNFNodeOffline',
+            expr: 'tnf_node_online == 0',
+            'for': '2m',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              summary: 'TNF node {{ $labels.node }} is offline.',
+              description: 'TNF node {{ $labels.node }} has been offline for more than 2 minutes. This may indicate a node failure, reboot, or network partition. The cluster is operating in a degraded single-node state.',
+              runbook_url: 'https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/TNFNodeOffline.md',
+            },
+          },
+          {
+            alert: 'TNFNodeFencingUnavailable',
+            expr: 'tnf_node_fencing_available == 0',
+            'for': '5m',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              summary: 'Fencing is unavailable for TNF node {{ $labels.node }}.',
+              description: 'No fencing devices are available for TNF node {{ $labels.node }}. Without fencing, the cluster cannot safely recover from a split-brain scenario. Check BMC reachability and credentials.',
+              runbook_url: 'https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/TNFNodeFencingUnavailable.md',
+            },
+          },
+          {
+            alert: 'TNFNodeFencingDegraded',
+            expr: 'tnf_node_fencing_healthy == 0 and tnf_node_fencing_available == 1',
+            'for': '10m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'Fencing is degraded for TNF node {{ $labels.node }}.',
+              description: 'Fencing devices for TNF node {{ $labels.node }} are available but not fully healthy. Fencing can still operate but with reduced redundancy. Investigate fence device health.',
+              runbook_url: 'https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/TNFNodeFencingDegraded.md',
+            },
+          },
+          {
+            alert: 'TNFNodeUnclean',
+            expr: 'tnf_node_clean == 0',
+            'for': '5m',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              summary: 'TNF node {{ $labels.node }} is in an unclean state.',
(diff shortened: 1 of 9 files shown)


pkg/tnf/pkg/pacemaker/disruption_test.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func getDisruptionCount(t *testing.T, node, resource string) float64 {
+	t.Helper()
+	val, err := testutil.GetCounterMetricValue(resourceDisruptionCounter.WithLabelValues(node, resource))
+	if err != nil {
+		return 0
+	}
+	return val
+}
```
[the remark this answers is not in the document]
fonta-rh: Won't fix — returning 0 on error is intentional. `GetCounterMetricValue` errors when a label combination hasn't been observed yet (counter not incremented), and returning 0 is the correct value for that case. Using `require.NoError` would break every test that asserts `count == 0` before the first disruption event.

pkg/tnf/pkg/pacemaker/healthcheck.go
```diff
@@ -892,3 +897,15 @@ func (c *HealthCheck) recordErrorEvent(errorMsg string) {
...
+	item, exists, err := c.pacemakerInformer.GetStore().GetByKey(PacemakerClusterResourceName)
+	if err != nil || !exists {
+		return
+	}
+	cr, ok := item.(*pacmkrv1.PacemakerCluster)
+	if !ok {
+		return
+	}
+	c.disruptionTracker.TrackResourceStates(cr.Status.Nodes)
+}
```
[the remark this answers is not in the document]
fonta-rh: Won't fix — both paths are structurally unreachable. Informer store `GetByKey` errors come from the underlying threadSafeMap (essentially impossible), `!exists` is expected during startup before the PacemakerCluster CR is created, and the type assertion is guaranteed by the typed informer. Adding error logging for impossible cases goes against codebase convention.

fonta-rh:
Addressed remaining CodeRabbit findings:

- **Test isolation (nitpick)**: Won't fix — tests already use unique label combinations per test case. Adding `Reset()` could mask real counter accumulation bugs and is a refactoring suggestion, not a correctness issue.
- **No-Sensitive-Data-In-Logs**: Won't fix — node names (`master-0`, `master-1`) are logged throughout the pacemaker package at V(2)/V(4) levels (`statuscollector.go`, `healthcheck.go`) and exposed as Prometheus metric labels. This is standard operator behavior, not sensitive data.
- **Test Structure And Quality**: Duplicate of inline finding on `getDisruptionCount` — addressed there.

pkg/tnf/pkg/pacemaker/disruption_test.go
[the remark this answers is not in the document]
fonta-rh: Fair point — you're right that `WithLabelValues()` creates the child on first access so the error path was never exercised for the unobserved-label case. Fixed in bb92325.

pkg/tnf/pkg/pacemaker/healthcheck.go
```diff
@@ -235,6 +248,8 @@ func (c *HealthCheck) sync(ctx context.Context, syncCtx factory.SyncContext) err
 		return nil
 	}
 
+	c.trackResourceDisruptions()
```
vimauro, line 255: In line 239, you are calling `getPacemakerStatus` which seems to do something fairly similar to `trackResourceDisruptions` but is not swallowing errors. Just checking if this is the intended behavior of those 2 functions in such short of time?

In any case, feel free to ignore, but maybe we can revisit `trackResourceDisruptions` and pass directly the `cr.Status.Nodes` to avoid recalling `c.pacemakerInformer.GetStore().GetByKey(PacemakerClusterResourceName)`

Nothing important here, so feel free to ignore

vimauro:
Overall looks good to me, nothing major to flag just a minor nit.

jsonnet/tnf-alerts.libsonnet
```diff
@@ -0,0 +1,170 @@
...
+            },
+            annotations: {
+              summary: 'TNF node {{ $labels.node }} is in an unclean state.',
+              description: 'TNF node {{ $labels.node }} is marked unclean by Pacemaker. This indicates a fencing, communication, or configuration issue that must be resolved before the node can rejoin the cluster.',
+              runbook_url: 'https://github.com/openshift/runbooks/blob/master/alerts/cluster-etcd-operator/TNFNodeUnclean.md',
+            },
+          },
+          {
+            alert: 'TNFNodeInMaintenance',
+            expr: 'tnf_node_in_service == 0',
```
dhensel-rh: When `pcs property set maintenance-mode=true` is run, Pacemaker propagates `maintenance="true"` to every individual `<node>` element in `pcs status xml`. The status collector reads this per-node attribute in `buildNodeConditions()` and sets `NodeInServiceConditionType = ConditionFalse` for each node independently of the cluster-level check.

This means a single cluster-wide maintenance action fires 3 alerts:

1. `TNFClusterInMaintenance` (from `tnf_cluster_in_service == 0`)
2. `TNFNodeInMaintenance` for node-1 (from `tnf_node_in_service == 0`)
3. `TNFNodeInMaintenance` for node-2 (from `tnf_node_in_service == 0`)

Suggested fix — guard `TNFNodeInMaintenance` to suppress when the cluster itself is in maintenance:

```diff
- expr: tnf_node_in_service == 0
+ expr: tnf_node_in_service == 0 and tnf_cluster_in_service == 1
```

This way `TNFClusterInMaintenance` handles the cluster-wide case alone, and `TNFNodeInMaintenance` only fires for per-node maintenance.

Note: the test fixture `cluster_maintenance.xml` does not set `maintenance="true"` on individual `<node>` elements, so the current tests do not exercise this overlap.

fonta-rh: Good catch — this is the same cascade class we fixed for the resource-level alerts. Applied the guard in 1ca0d8e and also fixed the cluster_maintenance.xml test fixture in 7494ed2 to set per-node maintenance="true" to match real Pacemaker propagation behavior.

pkg/tnf/pkg/pacemaker/healthcheck.go
fonta-rh: Good call — deduplicated in 431dc57. getPacemakerStatus now caches the CR's node statuses on the struct so trackResourceDisruptions reuses them instead of doing a second GetByKey.

vimauro (approved):

tjungblu:
let's override this, I know it's broken in other operator repos too.

/override ci/prow/e2e-agnostic-ovn-upgrade

PR #1664, opened 2026-07-29
OCPBUGS-100294: fix TNFNodeInMaintenance PromQL label mismatch

lucaconsalvi:
## Summary

The `TNFNodeInMaintenance` alert never fires because of a PromQL label mismatch in the `and` operator.

`tnf_node_in_service` has a `node` label, `tnf_cluster_in_service` does not. PromQL's `and` requires matching label sets, so the expression silently returns empty.

**Fix:** Add `on()` to ignore labels when joining: `and` → `and on()`.

This is the only affected alert — all other cross-metric `and` expressions correctly use `on(node)`.

## Verification

Tested side-by-side on a live TNF cluster (OCP 5.0 nightly):
- Buggy expression (`and`): returns empty — alert never fires
- Fixed expression (`and on()`): correctly returns `{node="master-1"}`
- Deployed a separate PrometheusRule with the fixed expression — alert fired within 2 minutes

## JIRA

[OCPEDGE-2848](https://redhat.atlassian.net/browse/OCPEDGE-2848)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the accuracy of the node maintenance alert by matching cluster service state with the correct label alignment.
  * Updated the related monitoring rule logic for `TNFNodeInMaintenance` while keeping the same alert intent and thresholds.
* **Tests**
  * Refreshed the expected alert data to reflect the updated rule formatting.

### DIFF
--- a/jsonnet/tnf-alerts.libsonnet
+++ b/jsonnet/tnf-alerts.libsonnet
@@ -86,7 +86,7 @@
           },
           {
             alert: 'TNFNodeInMaintenance',
-            expr: 'tnf_node_in_service == 0 and tnf_cluster_in_service == 1',
+            expr: 'tnf_node_in_service == 0 and on() tnf_cluster_in_service == 1',
             'for': '2m',
             labels: {
               severity: 'warning',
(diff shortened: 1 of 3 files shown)


lucaconsalvi:
Verified the fix on a payload-built cluster (`periodic-ci-openshift-release-main-nightly-4.22-e2e-metal-ovn-two-node-fencing-recovery`).

Put master-1 in node-level maintenance (`pcs node maintenance master-1`) — **TNFNodeInMaintenance fired correctly** (warning, node: master-1) after 2 minutes:

```json
{"alertname":"TNFNodeInMaintenance","alertstate":"firing","node":"master-1","severity":"warning"}
```

Without the fix, the same scenario returns empty and the alert never fires.

fracappa (approved):
/lgtm

lucaconsalvi:
Hi @tjungblu sorry to bother you another time, but due to the fact that I modified 2 files outside the tnf dir I need etcd team approve on this PR. Thanks in advance 🙏

PR #1666, opened 2026-07-30
CNTRLPLANE-3403: reduce default snapshot-count to 5000

andreacv98:
With this PR we reduce the default `--snapshot-count` from 10000 to 5000.

The value was selected after evaluating all five snapshot-count values across the full supported quota range (8–16 GiB) — full dataset and analysis in [CNTRLPLANE-3403](https://redhat.atlassian.net/browse/CNTRLPLANE-3403).
snapshot-count=5000 represents a good balance between WAL latency and memory peaks across all tested configurations, with particular benefit for clusters configured with quota sizes above the default 8 GiB.

## Summary by CodeRabbit

* **Improvements**
  * Added configurable etcd snapshot scheduling to improve data durability and recovery behavior.
  * Configured etcd to create snapshots after a defined number of operations.

### DIFF
--- a/pkg/operator/ceohelpers/podsubstitution.go
+++ b/pkg/operator/ceohelpers/podsubstitution.go
@@ -31,6 +31,7 @@ type PodSubstitutionTemplate struct {
 	ListenAddress       string
 	LocalhostAddress    string
 	LogLevel            string
+	SnapshotCount       uint64
 	EnvVars             []NameValue
 	BackupArgs          []string
 	CipherSuites        string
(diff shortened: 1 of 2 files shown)


pkg/operator/ceohelpers/podsubstitution.go
```diff
@@ -115,6 +116,7 @@ func GetPodSubstitution(
 		ListenAddress:       "0.0.0.0",   // TODO: this needs updating to detect ipv6-ness
 		LocalhostAddress:    "127.0.0.1", // TODO: this needs updating to detect ipv6-ness
 		LogLevel:            LoglevelToZap(operatorSpec.LogLevel),
+		SnapshotCount:       5000,
```
lance5890, line 119: should we consider setting `SnapshotCount` as a configurable param instead of a fixed number

andreacv98: @lance5890 it was something I considered, but after an internal discussion with @dusk125 we decided that it would have been out of scope for this particular feature, but it can definitely be considered in a future enhancement/feature.

PR #1675, opened 2026-08-10
OCPBUGS-104851: feat: add cert-watcher DaemonSet to restart etcd on CA bundle rotation

fracappa:
## Summary

- Adds a **cert-watcher DaemonSet** to TNF deployments that monitors CA bundle
  certificate files on disk and restarts etcd when they change, preventing
  `force_new_cluster` during CA rotation
- Adds a `watch-certs` subcommand to the `tnf-monitor` binary with fsnotify-based
  file watching and a 1-minute fallback poll
- Includes health-check serialization to prevent simultaneous restarts in the
  2-node etcd cluster

## Problem

In TNF deployments, etcd is managed by Pacemaker and runs as a podman container
outside of Kubernetes. When the etcd CA bundle is rotated (e.g., during
certificate rotation), etcd does not automatically reload the new CA certificates.

The kube-apiserver presents a client certificate signed by the **new** CA, but etcd
still trusts only the **old** CA — causing etcd to reject API server connections with
`remote error: tls: unknown certificate authority`. This results in kube-apiserver
entering CrashLoopBackOff and complete loss of API availability.

## Solution

A lightweight DaemonSet (`tnf-cert-watcher`) runs on each control-plane node and:

1. **Detects** CA bundle changes via `fsnotify` (with 2-second debounce) and a
   1-minute fallback poll for atomic directory replacements
2. **Serializes** restarts across the 2-node cluster by checking all etcd members
   are healthy before proceeding — preventing both nodes from restarting
   simultaneously and losing quorum
3. **Protects** against `force_new_cluster` by setting `restart_no_leave` on the
   local node via `crm_attribute` before restarting
4. **Restarts** etcd via `podman restart etcd` (SIGTERM preserves cluster membership)
5. **Recovers** by waiting for etcd health (up to 5 min) and cleaning up any
   Pacemaker failure state

### Key design decisions

| Decision | Rationale |
|----------|-----------|
| Mount stable parent dir, not leaf configmap dir | Kubernetes atomically replaces configmap dirs via symlink swap, invalidating leaf bind mounts |
| `podman restart` instead of `pcs resource restart` | `pcs resource restart` runs the OCF agent's stop then start actions. But that's problemativ: the OCF stop action stops the container, and during the stop→start transition, the OCF monitor on the peer node can fire, see the member is down, and mark it as FAILED. Additionally, it is a silent no-op when the resource is unmanaged |
| `restart_no_leave` on local node only | Multiple holders cause `force_new_cluster holders changed after decision` errors in the OCF agent |
| Defer restart if cluster unhealthy | Prevents cascading Pacemaker failures where both nodes end up FAILED |
| Keep old baseline on health timeout | Ensures retry on next poll cycle instead of silently accepting a bad state |

### DIFF
--- a/bindata/etcd/cluster-restore-tnf.sh
+++ b/bindata/etcd/cluster-restore-tnf.sh
@@ -108,10 +108,12 @@ function cleanup_podman_etcd_attributes() {
   crm_attribute --delete --name "standalone_node" || true
   crm_attribute --delete --name "learner_node" || true
   crm_attribute --delete --name "force_new_cluster" --lifetime reboot --node "${NODENAME}" || true
+  crm_attribute --delete --name "restart_no_leave" --lifetime reboot --node "${NODENAME}" || true
 
   peer_node_name=$(get_peer_node_name)
   if [ "$(echo "$peer_node_name" | wc -w)" -eq 1 ]; then
     crm_attribute --delete --name "force_new_cluster" --lifetime reboot --node "${peer_node_name}" || true
+    crm_attribute --delete --name "restart_no_leave" --lifetime reboot --node "${peer_node_name}" || true
   else
     echo "Warning: could not find peer node name. If restore fails, manually run on the peer node, and try again:" >&2
     echo "  crm_attribute --delete --name force_new_cluster --lifetime reboot --node \"\$(hostname)\"" >&2
--- a/pkg/tnf/certwatch/watcher.go
+++ b/pkg/tnf/certwatch/watcher.go
@@ -1,0 +1,290 @@
+package certwatch
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/json"
+	"encoding/xml"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/exec"
+	"github.com/openshift/cluster-etcd-operator/pkg/tnf/pkg/pacemaker"
+)
+
+const (
+	// FallbackPollInterval is a safety-net poll in case fsnotify misses an
+	// event (e.g. the directory is replaced atomically via symlink swap).
+	FallbackPollInterval = 1 * time.Minute
+	healthTimeout        = 5 * time.Minute
+	healthPollInterval   = 5 * time.Second
+	// debounceWindow coalesces rapid successive writes into a single restart.
+	debounceWindow = 2 * time.Second
+)
+
+// Run watches CA bundle files in certDir for changes and restarts etcd
+// on the local node when they change. It uses fsnotify for instant
+// detection with a periodic fallback. No Kubernetes API access is
+// needed — it reads cert files from a hostPath mount and acts via nsenter.
+func Run(ctx context.Context, certDir string) error {
+	hostname, err := getHostname(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get hostname: %w", err)
+	}
+	klog.Infof("Starting cert watcher on node %s, watching %s", hostname, certDir)
+
+	baseline, err := hashDir(certDir)
+	if err != nil {
+		return fmt.Errorf("failed to hash initial cert directory: %w", err)
+	}
+	if baseline == "" {
+		return fmt.Errorf("cert directory %s is empty, refusing to start with empty baseline", certDir)
+	}
+	klog.Infof("Recorded baseline cert hash: %s", baseline)
+
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return fmt.Errorf("failed to create fsnotify watcher: %w", err)
+	}
+	defer watcher.Close()
+
+	if err := watcher.Add(certDir); err != nil {
+		return fmt.Errorf("failed to watch %s: %w", certDir, err)
+	}
+
+	fallback := time.NewTicker(FallbackPollInterval)
+	defer fallback.Stop()
+
+	var debounce *time.Timer
+	debounceCh := make(<-chan time.Time)
+
+	for {
+		select {
+		case <-ctx.Done():
+			klog.Info("Cert watcher shutting down")
+			return nil
+
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return fmt.Errorf("fsnotify watcher closed")
+			}
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
+				klog.V(4).Infof("fsnotify event: %s", event)
+				if debounce == nil {
+					debounce = time.NewTimer(debounceWindow)
+					debounceCh = debounce.C
+				} else {
+					debounce.Reset(debounceWindow)
+				}
+			}
+
+		case watchErr, ok := <-watcher.Errors:
+			if !ok {
+				return fmt.Errorf("fsnotify error channel closed")
+			}
+			klog.Warningf("fsnotify error: %v", watchErr)
+
+		case <-debounceCh:
+			debounce = nil
+			debounceCh = make(<-chan time.Time)
+			baseline, err = checkAndRestart(ctx, certDir, baseline, hostname)
+			if err != nil {
+				klog.Errorf("Failed to handle cert change: %v", err)
+			}
+
+		case <-fallback.C:
+			baseline, err = checkAndRestart(ctx, certDir, baseline, hostname)
+			if err != nil {
+				klog.Errorf("Failed to handle cert change (fallback): %v", err)
+			}
+		}
+	}
+}
+
+// checkAndRestart compares the current cert hash against baseline and
+// restarts etcd if they differ. Each node restarts independently when it
+// detects the change. The health check guards against restarting while
+// the peer is already down, but does not guarantee strict serialization.
+// restart_no_leave is set on the local node only to prevent the OCF agent
+// from running force_new_cluster on the restart it is about to perform.
+func checkAndRestart(ctx context.Context, certDir, baseline, hostname string) (string, error) {
+	current, err := hashDir(certDir)
+	if err != nil {
+		return baseline, fmt.Errorf("failed to hash cert directory: %w", err)
+	}
+	if current == baseline || current == "" {
+		return baseline, nil
+	}
+
+	if !isClusterHealthy(ctx) {
+		klog.Infof("Cert change detected but etcd cluster not fully healthy, deferring restart to next poll")
+		return baseline, nil
+	}
+
+	klog.Infof("CA bundle changed (old: %s, new: %s), restarting etcd on %s", baseline, current, hostname)
+
+	if err := setRestartNoLeave(ctx, hostname); err != nil {
+		return baseline, fmt.Errorf("failed to set restart_no_leave: %w", err)
+	}
+	defer clearRestartNoLeave(ctx, hostname)
+
+	if err := restartEtcdContainer(ctx); err != nil {
+		return baseline, fmt.Errorf("failed to restart etcd container: %w", err)
+	}
+
+	if err := waitForEtcdHealthy(ctx); err != nil {
+		klog.Errorf("etcd did not become healthy after restart: %v", err)
+		return baseline, nil
+	}
+
+	if hasPacemakerFailcount(ctx) {
+		clearPacemakerFailcount(ctx)
+	}
+
+	klog.Infof("Updated baseline cert hash to: %s", current)
+	return current, nil
+}
+
+func hashDir(dir string) (string, error) {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return "", err
+	}
+
+	h := sha256.New()
+	var names []string
+	for _, entry := range entries {
+		if entry.IsDir() {
+			continue
+		}
+		names = append(names, entry.Name())
+	}
+	sort.Strings(names)
+
+	for _, name := range names {
+		data, err := os.ReadFile(filepath.Join(dir, name))
+		if err != nil {
+			return "", err
+		}
+		h.Write([]byte(name))
+		h.Write(data)
+	}
+
+	return fmt.Sprintf("%x", h.Sum(nil)), nil
+}
+
+func getHostname(ctx context.Context) (string, error) {
+	stdout, _, err := exec.Execute(ctx, "hostname")
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(stdout), nil
+}
+
+func setRestartNoLeave(ctx context.Context, hostname string) error {
+	cmd := fmt.Sprintf(`crm_attribute --lifetime reboot --node "%s" --name restart_no_leave --update true`, hostname)
+	if _, _, err := exec.Execute(ctx, cmd); err != nil {
+		return fmt.Errorf("crm_attribute failed: %w", err)
+	}
+	klog.Infof("Set restart_no_leave on %s", hostname)
+	return nil
+}
+
+func clearRestartNoLeave(ctx context.Context, hostname string) {
+	cmd := fmt.Sprintf(`crm_attribute --lifetime reboot --node "%s" --name restart_no_leave --delete`, hostname)
+	if _, _, err := exec.Execute(ctx, cmd); err != nil {
+		klog.Warningf("Failed to clear restart_no_leave on %s: %v", hostname, err)
+		return
+	}
+	klog.Infof("Cleared restart_no_leave on %s", hostname)
+}
+
+func restartEtcdContainer(ctx context.Context) error {
+	if _, _, err := exec.Execute(ctx, "podman restart etcd"); err != nil {
+		return fmt.Errorf("podman restart etcd failed: %w", err)
+	}
+	klog.Info("etcd container restarted via podman")
+	return nil
+}
+
+func hasPacemakerFailcount(ctx context.Context) bool {
+	stdout, _, err := exec.Execute(ctx, "pcs status xml")
+	if err != nil {
+		klog.Warningf("Failed to get pcs status xml: %v", err)
+		return false
+	}
+	var status pacemaker.PacemakerResult
+	if err := xml.Unmarshal([]byte(stdout), &status); err != nil {
+		klog.Warningf("Failed to parse pcs status xml: %v", err)
+		return false
+	}
+	for _, node := range status.NodeHistory.Node {
+		for _, rh := range node.ResourceHistory {
+			if rh.ID == "etcd" && rh.FailCount == "INFINITY" {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func clearPacemakerFailcount(ctx context.Context) {
+	klog.Info("Resetting etcd-clone failcount")
+	if _, _, err := exec.Execute(ctx, "pcs resource failcount reset etcd-clone"); err != nil {
+		klog.Warningf("Failed to reset etcd-clone failcount: %v", err)
+	}
+}
+
+type endpointHealth struct {
+	Endpoint string `json:"endpoint"`
+	Health   bool   `json:"health"`
+}
+
+func isClusterHealthy(ctx context.Context) bool {
+	stdout, _, err := exec.Execute(ctx, "podman exec etcd etcdctl endpoint health --cluster -w json")
+	if err != nil {
+		return false
+	}
+	var endpoints []endpointHealth
+	if err := json.Unmarshal([]byte(stdout), &endpoints); err != nil {
+		klog.Warningf("Failed to parse etcd health JSON: %v", err)
+		return false
+	}
+	if len(endpoints) == 0 {
+		return false
+	}
+	for _, ep := range endpoints {
+		if !ep.Health {
+			return false
+		}
+	}
+	return true
+}
+
+func waitForEtcdHealthy(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, healthTimeout)
+	defer cancel()
+
+	ticker := time.NewTicker(healthPollInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("timed out waiting for etcd health")
+		case <-ticker.C:
+			if isClusterHealthy(ctx) {
+				klog.Info("etcd cluster healthy after restart")
+				return nil
+			}
+			klog.V(4).Info("etcd health check not yet passing")
+		}
+	}
+}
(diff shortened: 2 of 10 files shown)


fonta-rh:
/hold
extra-protection for accidental merge-in. Can't do that before https://github.com/ClusterLabs/resource-agents/pull/2197/changes is in RHEL

bindata/etcd/cluster-restore-tnf.sh
```diff
@@ -108,10 +108,12 @@ function cleanup_podman_etcd_attributes() {
   crm_attribute --delete --name "standalone_node" || true
   crm_attribute --delete --name "learner_node" || true
   crm_attribute --delete --name "force_new_cluster" --lifetime reboot --node "${NODENAME}" || true
+  crm_attribute --delete --name "restart_no_leave" --lifetime reboot --node "${NODENAME}" || true
```
jaypoulz, line 111: @dhensel-rh do we have automated cluster-restore tests? If not, we should probably add those since I'm sure the changes I am adding for double graceful node shutdown would also interact with this.

pkg/tnf/pkg/jobs/lifecycle.go
```diff
@@ -52,6 +52,13 @@ var (
 	// retryStateMutex protects access to retryState map
 	retryStateMutex sync.Mutex
 
+	// configBaseline stores the initial config hash for drift-only jobs.
+	// These jobs should NOT run on first install — only when the config
+	// changes from this baseline. Map key is job name.
+	configBaseline = make(map[string]string)
```
jaypoulz: I have a similar idea in lifecycle part 3 where we store a config generation as part of the job name for the update-setup job. But I'm wondering why we need it here? Is the point of this to track which cert was installed so you don't re-trigger drift detection?

An alternative approach would be to allow the job to run, but detect if the cert it's trying to install is the same as the cert that's already there and not do anything. This way the job remains idempotent. We probably want both, to be honest.

pkg/tnf/pkg/jobs/lifecycle.go
```diff
@@ -204,7 +216,11 @@ func checkNodesReadinessAndSetCondition(ctx context.Context, nodes []*corev1.Nod
 // Detects drift and failures, updates retry state accordingly. Job deletion/recreation is handled
 // by ApplyJob via drift detection (NodeName changed), except for real config/node drift where
 // jobs are explicitly deleted before resetting state.
-func syncMultiNodeJobState(ctx context.Context, jobName string, schedulableNodesFunc SchedulableNodesFunc, affectedNodesFunc AffectedNodesFunc, jobConfigFunc JobConfigFunc, maxRetryAttempts int, kubeClient kubernetes.Interface, operatorClient v1helpers.StaticPodOperatorClient) error {
+//
+// When driftOnly is true, the job will NOT run on first install. Instead, the initial config
+// is stored as a baseline, and the job only fires when the config changes from that baseline.
+// This prevents jobs like etcd-restart from running during fresh installation.
+func syncMultiNodeJobState(ctx context.Context, jobName string, schedulableNodesFunc SchedulableNodesFunc, affectedNodesFunc AffectedNodesFunc, jobConfigFunc JobConfigFunc, maxRetryAttempts int, driftOnly bool, kubeClient kubernetes.Interface, operatorClient v1helpers.StaticPodOperatorClient) error {
```
jaypoulz: I'm not a fan of driftOnly. I would rather see it all or nothing. Either all jobs only run on drift and nil->something is considered drift (e.g. startup) or all jobs always run on startup and rely on being idempotent.

I like the former, but I don't know that it's needed for this patch. Seems more like a lifecyle refactor patch.

pkg/tnf/pkg/jobs/lifecycle.go
```diff
@@ -552,7 +616,9 @@ func RunNodeJobController(ctx context.Context, jobType tools.JobType, node *core
 }
 
 // RunClusterJobController starts job controller for cluster-wide job with round-robin retry logic.
-func RunClusterJobController(ctx context.Context, jobType tools.JobType, schedulableNodesFunc SchedulableNodesFunc, affectedNodesFunc AffectedNodesFunc, jobConfigFunc JobConfigFunc, retries int, controllerContext *controllercmd.ControllerContext, operatorClient v1helpers.StaticPodOperatorClient, kubeClient kubernetes.Interface, kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces, conditions []string) {
+// When driftOnly is true, the job will not run on first install — only when the config from
+// jobConfigFunc changes from the initial baseline (e.g., CA bundle rotation).
+func RunClusterJobController(ctx context.Context, jobType tools.JobType, schedulableNodesFunc SchedulableNodesFunc, affectedNodesFunc AffectedNodesFunc, jobConfigFunc JobConfigFunc, retries int, driftOnly bool, controllerContext *controllercmd.ControllerContext, operatorClient v1helpers.StaticPodOperatorClient, kubeClient kubernetes.Interface, kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces, conditions []string, extraInformers ...factory.Informer) {
```
jaypoulz: As noted above, I think driftOnly is a false choice.
All jobs want to behave driftOnly. If we start them otherwise, it's because we haven't migrated them to the driftOnly model yet. More of a lifecycle migration task than something to address in this patch.

jaypoulz (changes requested):
I think we need to make a decision on how we handle lifecycle refactor work as part of this. Personally, I would rather that be implemented outside this PR and focus this on the current working model/expectation of jobs in TNF.

bindata/etcd/cluster-restore-tnf.sh
```diff
@@ -108,10 +108,12 @@ function cleanup_podman_etcd_attributes() {
...
 
   peer_node_name=$(get_peer_node_name)
   if [ "$(echo "$peer_node_name" | wc -w)" -eq 1 ]; then
     crm_attribute --delete --name "force_new_cluster" --lifetime reboot --node "${peer_node_name}" || true
+    crm_attribute --delete --name "restart_no_leave" --lifetime reboot --node "${peer_node_name}" || true
```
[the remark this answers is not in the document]
jaypoulz: I don't understand:
When get_peer_node_name does not return exactly one name - the only options are you're running in single node mode, in which case there is no peer to clean up OR you've got exactly one peer.

Which case is missed here?

pkg/tnf/certwatch/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+			if err != nil {
+				klog.Errorf("Failed to handle cert change (fallback): %v", err)
+			}
+		}
+	}
+}
+
+// checkAndRestart compares the current cert hash against baseline and
+// restarts etcd if they differ. It verifies the whole cluster is healthy
+// to serialize restarts across nodes, then sets restart_no_leave on ALL
```
jaypoulz: this comment seems misleading since the restarts are done simultaeneously, right? Should be be serialized in alphabetical order or something?

I think insta-restart is probably fine BUT don't say we're serializing. If we're trying to maintain functionality and do need to serialize, let's coordinate the nodes better or add a simple delay to stagger them.

pkg/tnf/certwatch/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+	stdout, _, err := exec.Execute(ctx, "pcs resource failcount show etcd-clone")
+	if err != nil {
+		klog.Warningf("Failed to check etcd-clone failcount: %v", err)
+		return
+	}
+	if !strings.Contains(stdout, "INFINITY") && !strings.Contains(stdout, "fail-count") {
+		return
+	}
+	klog.Info("Clearing etcd-clone failure state")
+	if _, _, err := exec.Execute(ctx, "pcs resource cleanup etcd-clone"); err != nil {
```
jaypoulz: are we just doing this to insta-kick etcd to force a restart?
that's fine, but I want to doulbe check that nodes clear the restart_no_leave flag on start since it's not cleared here.

pkg/tnf/certwatch/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	if err := waitForEtcdHealthy(ctx); err != nil {
+		klog.Errorf("etcd did not become healthy after restart: %v", err)
+		return baseline, nil
+	}
+
+	cleanupEtcd(ctx)
+
+	klog.Infof("Updated baseline cert hash to: %s", current)
```
jaypoulz, line 151: should we be checking if restart_no_leave is still set and cleaning it here if it is?

pkg/tnf/certwatch/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+	if err := restartEtcdContainer(ctx); err != nil {
+		return baseline, fmt.Errorf("failed to restart etcd container: %w", err)
+	}
+
+	if err := waitForEtcdHealthy(ctx); err != nil {
+		klog.Errorf("etcd did not become healthy after restart: %v", err)
+		return baseline, nil
+	}
+
+	cleanupEtcd(ctx)
```
jaypoulz: why are we kicking the etcd resource again? we just restarted it? If we're going to kick it, why no set restart no leave and just call this immediately instead. That will restart the agents for both nodes, forcing a restart.

This feels like we restarting the container and then restarting the agent to restart it again. :)

jaypoulz: It feels to me like you are just doing this to ensure "start" runs and clears the condition. Just do that - then you don't need a restartEtcdContainer - you clean up waitForEtcdHealthy and then scan for any stale attributes and call it a day.

pkg/tnf/certwatch/watcher.go
jaypoulz: This comment seems to imply that ALL nodes get restart_no_leave, but it's only called once and with the current hostname, so that's not true. I don't know if all nodes are supposed to have it set, or if one is sufficient, but this should be addressed either in the comment or the code https://github.com/openshift/cluster-etcd-operator/pull/1675/changes#diff-aff7ec7af626500fa49169ab25f46838f5f0bab8916b8219a29ef8c8c3e328c1R128

pkg/tnf/operator/job_controllers.go
```diff
@@ -250,6 +250,13 @@ func startTnfJobcontrollers(
 		fencingJobConfigFunc := createFencingJobConfigFunc(lifecycleManager, kubeInformersForNamespaces)
 		jobs.RunClusterJobController(ctx, tools.JobTypeFencing, schedulableNodesFunc, fencingAffectedNodesFunc, fencingJobConfigFunc, 3, controllerContext, operatorClient, kubeClient, kubeInformersForNamespaces, jobs.DefaultConditions)
 
+		// Cert watcher DaemonSet: watches CA bundle files on disk and restarts
+		// the local etcd when they change. Runs independently of the operator
+		// and API server — prevents force_new_cluster during CA rotation.
+		if err := lifecycleManager.ensureCertWatcherDaemonSet(ctx); err != nil {
```
jaypoulz, line 256: I think we need to return this error. This is fatal since this is a required feature for handling cert-rotation.

pkg/tnf/certwatch/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+		klog.Warningf("Failed to cleanup etcd-clone: %v", err)
+	}
+}
+
+func isClusterHealthy(ctx context.Context) bool {
+	stdout, _, err := exec.Execute(ctx, "podman exec etcd etcdctl endpoint health --cluster")
+	if err != nil {
+		return false
+	}
+	return strings.Contains(stdout, "is healthy") && !strings.Contains(stdout, "is unhealthy")
```
jaypoulz: there's got to be a way to get this output as json and parse it against a firmer output value. This is fragile.

jaypoulz: not to mention there might be a way of getting this status directly from a helper in CEO. we have a container/controller whose job it is to monitor etcd in this exact way.

jaypoulz (changes requested):
Overall I like the direction, but I think we need to do one more push to clean up the loose ends. I would like to see proof of a payload job triggering two-node-fencing-etcd-certrotation to verify this patch.

pkg/tnf/certwatch/watcher.go
fracappa: Thanks for pointing this out. This is a stale comment from a previous implementation based on a Job that set that attribute on both nodes when starting, to avoid the peer to be restarted with a "force_new_cluster" approach.

pkg/tnf/certwatch/watcher.go
fracappa: `cleanupEtcd` (renaming it  to `clearPacemakerFailcount`) doesn't restart etcd, it just clears Pacemaker's failcount via `pcs resource cleanup`. Since etcd is managed by Pacemaker, when we do `podman restart etcd`, we're bypassing Pacemaker, so its monitor can fire during the brief downtime and increment the failcount (potentially to INFINITY, which would prevent Pacemaker from ever starting etcd on that node again). After we confirm etcd is back and healthy, we clear that stale failure state.

I've renamed it to `clearPacemakerFailcount` to make the intent clearer.

Maybe, for the sake of clarity I can move this [check](https://github.com/openshift/cluster-etcd-operator/pull/1675/changes/BASE..2a5ea0cabaa41b8cc8b3d6d3ef0f6eaf3cb146f1#diff-aff7ec7af626500fa49169ab25f46838f5f0bab8916b8219a29ef8c8c3e328c1R206) before calling the `cleanupEtcd` (being renamed as `clearPacemakerFailcount`) function.

pkg/tnf/certwatch/watcher.go
fracappa: We should, absolutely. The original idea was to restart etcd via pacemaker, so that we could have relied on the attribute cleanup when etcd is started (more details [here](https://github.com/ClusterLabs/resource-agents/pull/2197/changes#diff-2600c60d13bcbda50a7ac6542e535b71029aabd73ddd86e188ea997508e1e6e7R2720)). 

Since this approach was not working, we need to clean it up explicitly. 

We should probably reevaluate the upstream workflow in the resource-agents repo, since that's not used right now (we rely on `podman restart` instead)

pkg/tnf/certwatch/watcher.go
fracappa: Does the `pcs resource cleanup` command trigger a restart? My intention here is just to clear the failcount and to trigger a re-probe operation. If it does restart etcd, I should revisit the approach

pkg/tnf/certwatch/watcher.go
fracappa: The health helpers seem to happer in  `pkg/etcdcli/`, do you mean those ones? If so, they seem to use the Go etcd client library and require Kubernetes API access (informers, configmap listers, mounted TLS certs) to discover endpoints.

 The cert-watcher DaemonSet intentionally avoids Kubernetes API dependency,  it needs to work when kube-apiserver is down, which is the exact failure scenario it's designed to fix. 

I'd suggest we use `podman exec etcd etcdctl endpoint health --cluster -w json` instead, which should be less fragile as you said. What do you think?

jaypoulz: yeah - my main concern here was not trusting string parsing to give us consistent output over time.

pkg/tnf/certwatch/watcher.go
jaypoulz: IIUC (and I may be wrong) pcs resource clean does more than reset the fail-count. It also restarts the services. I used to use this to force etcd startup so I could monitor it in the logs. let's confirm either that I'm wrong OR if there is a way to reset the count without restarting if I'm right.

pkg/tnf/certwatch/watcher.go
jaypoulz: I think it does trigger a restart on top of resetting the fail-count. I could be wrong, but I've relied on it doing that in the past.

pkg/tnf/certwatch/watcher.go
jaypoulz: Addressed in https://github.com/openshift/cluster-etcd-operator/pull/1675/changes#diff-aff7ec7af626500fa49169ab25f46838f5f0bab8916b8219a29ef8c8c3e328c1R147

pkg/tnf/certwatch/watcher.go
jaypoulz: Here's the breakdown I got from chai-bot.
pcs resource cleanup
- Clears both the failcount and the entire operation history for the resource
- Forces Pacemaker to re-probe the resource to rediscover its current state
- Does not directly restart the resource, but the re-probe can trigger a restart — if Pacemaker's policies determine the resource should be started or moved after the slate is wiped clean
- This is the broader, heavier operation

pcs resource failcount reset
- Only resets the failcount counter (the number tracking how many times the resource failed on a node)
- Does not clear operation history or trigger a re-probe
- The resource's current state remains known to Pacemaker — no re-evaluation of placement happens beyond the failcount threshold check
- This is the narrower, safer operation

Are we sure we want the operational history and reprobe reset as well?

pkg/tnf/operator/job_controllers.go
jaypoulz: Addressed in https://github.com/openshift/cluster-etcd-operator/pull/1675/changes#diff-018ab763fef548981a51a09e7e202187d2649c25c2a241747793cb3161b345dfR257

pkg/tnf/certwatch/watcher.go
fracappa: I'd say it's an unecessary risk. We just restarted etcd ourselves and confirmed it's healthy. All we need is to clear the failcount so Pacemaker doesn't think the node has hit its failure threshold.

`pcs resource failcount reset` should be good enough here

pkg/tnf/certwatch/watcher.go
jaypoulz: @fracappa do we need to add restart-no-leave clearly before the early returns to ensure it's not left stale?

pkg/tnf/certwatch/watcher.go
```diff
@@ -0,0 +1,275 @@
...
+func restartEtcdContainer(ctx context.Context) error {
+	if _, _, err := exec.Execute(ctx, "podman restart etcd"); err != nil {
+		return fmt.Errorf("podman restart etcd failed: %w", err)
+	}
+	klog.Info("etcd container restarted via podman")
+	return nil
+}
+
+func hasPacemakerFailcount(ctx context.Context) bool {
+	stdout, _, err := exec.Execute(ctx, "pcs resource failcount show etcd-clone")
```
jaypoulz: Is there an json or XML version of this we can use to harden the output verification?

pkg/tnf/certwatch/watcher.go
```diff
@@ -0,0 +1,275 @@
...
+		return fmt.Errorf("crm_attribute failed: %w", err)
+	}
+	klog.Infof("Set restart_no_leave on %s", hostname)
+	return nil
+}
+
+func clearRestartNoLeave(ctx context.Context, hostname string) {
+	cmd := fmt.Sprintf(`crm_attribute --lifetime reboot --node "%s" --name restart_no_leave --delete`, hostname)
+	if _, _, err := exec.Execute(ctx, cmd); err != nil {
+		klog.Warningf("Failed to clear restart_no_leave on %s: %v", hostname, err)
```
jaypoulz, line 203: you need a return after this to prevent a warning + success log for the same operation.

pkg/tnf/certwatch/watcher.go
jaypoulz: Resolved here.

pkg/tnf/operator/job_controllers.go
jaypoulz: Resolved

pkg/tnf/certwatch/watcher.go
jaypoulz: Resolved

pkg/tnf/certwatch/watcher.go
jaypoulz: :+1: resolved

pkg/tnf/certwatch/watcher.go
jaypoulz: resolved

pkg/tnf/pkg/jobs/lifecycle.go
jaypoulz: out of date/fixed

pkg/tnf/pkg/jobs/lifecycle.go
jaypoulz: out of date/fixed

pkg/tnf/pkg/jobs/lifecycle.go
jaypoulz: out of date/fixed

pkg/tnf/certwatch/watcher.go
fracappa: That's a valid concern, we can `defer` it so that we're sure it doesn't run only on the happy path

pkg/tnf/certwatch/watcher.go
fracappa: it looks like `pcs resource failcount` show doesn't support JSON or XML output. But `crm_failcount` with `--quiet` returns just the raw value, which is trivially parseable. Maybe we can use that one, what do you think?

jaypoulz: I think we either leave it like you have it or see if it's in `pcs status xml` since we already have parsing for that in the status collector.

bindata/etcd/cluster-restore-tnf.sh
jaypoulz: chai filed a new epic for me, here: https://redhat.atlassian.net/browse/OCPEDGE-3032

jaypoulz (approved):
We are in micro-improvement territory now. I'd like to see if we can use pcs status xml instead of string parsing for the fail count, but otherwise this should be ready provided that tests pass for recovery(minus double-graceful shutdown), upgrade, and etcd-certrotation.

fracappa:
/payload-job periodic-ci-openshift-release-main-nightly-5.1-e2e-metal-ovn-two-node-fencing-etcd-certrotation

jaypoulz:
/payload-job periodic-ci-openshift-release-main-nightly-5.1-e2e-metal-ovn-two-node-fencing-upgrade periodic-ci-openshift-release-main-nightly-5.1-e2e-metal-ovn-two-node-fencing-recovery

jaypoulz:
/payload-job periodic-ci-openshift-release-main-nightly-5.1-e2e-metal-ovn-two-node-fencing-recovery

eggfoobar:
/payload-with-prs periodic-ci-openshift-release-main-nightly-5.1-e2e-metal-ovn-two-node-fencing-recovery openshift/origin#31530 openshift/origin#31597

eggfoobar:
/payload-job-with-prs periodic-ci-openshift-release-main-nightly-5.1-e2e-metal-ovn-two-node-fencing-recovery https://github.com/openshift/origin/pull/31530 https://github.com/openshift/origin/pull/31597

eggfoobar:
/payload-job-with-prs periodic-ci-openshift-release-main-nightly-5.1-e2e-metal-ovn-two-node-fencing-recovery openshift/origin#31530 openshift/origin#31597

fracappa:
/verified by https://pr-payload-tests.ci.openshift.org/runs/ci/6a853310-a6e2-11f1-8802-e6016b17a552-0 https://pr-payload-tests.ci.openshift.org/runs/ci/beaa7c60-a6ed-11f1-8c73-f26ddde154cf-0 https://pr-payload-tests.ci.openshift.org/runs/ci/6d35c160-a714-11f1-89be-1ee62417bac8-0

The upgrade failure it's a known widespread regression in 5.1 and it has nothing to do with this PR.

PR #1678, opened 2026-08-13
OCPBUGS-109740: Fix tnf_cluster_in_service during per-node maintenance

lucaconsalvi:
## Summary

- Fix `tnf_cluster_in_service` incorrectly staying `1` when all nodes are placed in maintenance individually via `pcs node maintenance --all`
- Expand `isClusterInMaintenance` to also detect when every node has its per-node maintenance flag set, since this does not set the cluster-level `maintenance-mode` property
- This ensures `TNFClusterInMaintenance` fires correctly instead of two noisy per-node `TNFNodeInMaintenance` alerts

## Test plan

- [x] Unit tests for `isClusterInMaintenance` covering: cluster-level property, all nodes individually in maintenance, single node in maintenance, no nodes, empty property
- [x] XML integration test with new `all_nodes_maintenance.xml` fixture verifying end-to-end cluster condition output
- [x] All existing TNF tests pass
- [ ] On-cluster verification: `pcs node maintenance --all` → check `tnf_cluster_in_service == 0`

Fixes: https://redhat.atlassian.net/browse/OCPEDGE-2901

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved maintenance status detection for clusters where all nodes are individually in maintenance mode.
  * Correctly preserves non-maintenance status for empty node lists.
  * Ensures clusters with every node in maintenance are reported as unhealthy and out of service.
* **Tests**
  * Expanded coverage for single-node, multi-node, empty-list, and combined maintenance scenarios.

### DIFF
--- a/pkg/tnf/pkg/pacemaker/statuscollector.go
+++ b/pkg/tnf/pkg/pacemaker/statuscollector.go
@@ -484,11 +484,21 @@ func getNodeAddresses(configNode ClusterConfigNode) []pacmkrv1.PacemakerNodeAddr
 	return addresses
 }
 
-// isClusterInMaintenance returns true if cluster-wide maintenance mode is enabled
-// (pcs property set maintenance-mode=true). Node-level maintenance (pcs node maintenance <node>)
-// is handled separately via buildNodeConditions.
+// isClusterInMaintenance returns true if the cluster is in maintenance, either
+// via the cluster-level property or because every node is individually in maintenance.
 func isClusterInMaintenance(result *PacemakerResult) bool {
-	return result.Summary.ClusterOptions.MaintenanceMode == "true"
+	if result.Summary.ClusterOptions.MaintenanceMode == "true" {
+		return true
+	}
+	if len(result.Nodes.Node) == 0 {
+		return false
+	}
+	for _, node := range result.Nodes.Node {
+		if node.Maintenance != "true" {
+			return false
+		}
+	}
+	return true
 }
 
 // processResourcesForState populates resourceState map. Fencing agents are handled separately.
--- a/pkg/tnf/pkg/pacemaker/statuscollector_test.go
+++ b/pkg/tnf/pkg/pacemaker/statuscollector_test.go
@@ -812,11 +812,6 @@ func TestBuildClusterConditions(t *testing.T) {
 }
 
 func TestIsClusterInMaintenance(t *testing.T) {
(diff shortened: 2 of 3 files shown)


lucaconsalvi:
## On-cluster verification

Tested on a cluster-bot TNF cluster (`5.1.0-0.nightly`), 2-node bare metal topology.

### Baseline

```
Cluster: InService=True (InService), Healthy=True (ClusterHealthy)
Node 0:  InService=True (InService), Healthy=True (NodeHealthy)
Node 1:  InService=True (InService), Healthy=True (NodeHealthy)
```

### After `pcs node maintenance --all`

Verified that `maintenance-mode` cluster property stays `false` (pacemaker only sets per-node flags):

```
Pacemaker Nodes:
 Maintenance: master-0 master-1

Cluster Properties:
  maintenance-mode=false (default)
```

After the next status collector run:

```
Cluster: InService=False (InMaintenance), Healthy=False (ClusterUnhealthy)
Node 0:  InService=False (InMaintenance)
Node 1:  InService=False (InMaintenance)
```

**PASS** — `tnf_cluster_in_service` correctly reports `0`, enabling `TNFClusterInMaintenance` to fire and suppressing per-node `TNFNodeInMaintenance` alerts via the `tnf_cluster_in_service == 1` guard.

### After `pcs node unmaintenance --all`

```
Cluster: InService=True (InService), Healthy=True (ClusterHealthy)
Node 0:  InService=True (InService), Healthy=True (NodeHealthy)
Node 1:  InService=True (InService), Healthy=True (NodeHealthy)
```

**PASS** — cluster recovers to healthy baseline.

fonta-rh (approved):
Reviewed the logic change in `isClusterInMaintenance`, test coverage, cross-file interactions with the TNFNodeInMaintenance alert guard, and edge cases.

- Empty-nodes guard prevents vacuous-truth on the `allMaintenance` check
- `!= "true"` correctly handles absent `maintenance` attribute (no nil deref risk)
- TNFNodeInMaintenance's `tnf_cluster_in_service == 1` guard now correctly suppresses during `pcs node maintenance --all`, which was the bug
- Healthcheck consumer at `healthcheck.go` handles the new behavior correctly
- Test fixture `all_nodes_maintenance.xml` covers the new path

Clean fix at the right level — expanding `isClusterInMaintenance` generalizes the mechanism rather than adding a special case.

/lgtm

PR #1683, opened 2026-08-18
OCPBUGS-111074: Register TNF taint alert for node-lost events

eggfoobar:
Graceful ACPI shutdown never emits STONITH, so select_fencing alone never taints. Add tnf-taint-lost-alert on select_nodes so the existing taint script runs on membership loss.

## Summary by CodeRabbit

* **New Features**
  * Added Pacemaker alerts for node-lost events.
  * Alerts now support selecting affected nodes and use a dedicated node-lost alert configuration.


eggfoobar:
closing in favor of a different approach

PR #1684, opened 2026-08-18
OCPBUGS-105882: Remove orphaned etcd members even during a revision rollout

tchap:
I've found this potential issue when working on operator status conditions. This simply narrows a window when things can go wrong since they are inconsistent.

---

When a master node is decommissioned, its etcd member should be removed from the cluster. If a revision rollout starts around the same time, the revisionStable guard introduced in OCPBUGS-74151 blocks the entire sync function, including removeMemberWithoutMachine. The orphaned member — one whose Node and Machine no longer exist — remains in etcd's member list for the duration of the rollout. EtcdEndpointsController mirrors that list into the etcd-endpoints configmap, so the dead IP persists and is picked up by API server configuration.

The revisionStable guard was intended to protect attemptToScaleDown and attemptToRemoveLearningMember, where members can temporarily appear unhealthy during a rollout because their pods are being reinstalled. It was applied as a blanket early return that also covered removeMemberWithoutMachine unintentionally.

removeMemberWithoutMachine only acts on members that have neither a Machine nor a Node resource and are reported as unhealthy — conditions that are independent of rollout state. Moving it before the revisionStable check ensures orphaned members are cleaned up promptly without compromising the safety properties OCPBUGS-74151 established for scale-down.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cluster member cleanup when associated Machine and Node resources are missing.
  - Ensured cleanup errors are retained and reported alongside revision stability and endpoint synchronization errors.
  - Improved validation ordering so Machine API availability is checked before subsequent synchronization checks.
  - Prevented errors from being lost when synchronization checks are skipped.

### DIFF
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
@@ -123,6 +123,21 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, syncCtx facto
 		return nil
 	}
 
+	// Only proceed if the machine API is functional.
+	if isFunctional, err := c.machineAPIChecker.IsFunctional(); err != nil {
+		return err
+	} else if !isFunctional {
+		return nil
+	}
+
+	// Remove members with no backing Machine or Node unconditionally: they are already
+	// non-functional and safe to clean up regardless of revision stability. Leaving them
+	// causes stale entries in the etcd-endpoints configmap.
+	var errs []error
+	if err := c.removeMemberWithoutMachine(ctx); err != nil {
+		errs = append(errs, err)
+	}
+
 	// Removing multiple members in the middle of a revision rollout can cause API unavailability
 	// when simultaneously deleting multiple machines with the controlplanemachineset "OnDelete" strategy,
 	// i.e we have multiple machines pending deletion and multiple new replacements added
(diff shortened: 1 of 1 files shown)


tchap:
/retitle OCPBUGS-23746: Remove orphaned etcd members even during a revision rollout

tchap:
/retitle OCPBUGS-105882: Remove orphaned etcd members even during a revision rollout

hasbro17:
This fixes the true-orphan case (both Machine and Node deleted out-of-band) so bringing `removeMemberWithoutMachine` above the revision gate cleans it up.

The gap is the **half-orphan** case (Node gone, but Machine still pending deletion due to the `preDrain` deletion hook blocking removal).

That is the case for the more common MachineHealthCheck(MHC) remediation / ControlPlaneMachineSet(CPMS) deletion workflow:

  - MHC deletes the old Machine, but our `EtcdQuorumOperator` PreDrain hook deliberately blocks the Machine from finalizing until etcd removes the member. So the Machine lingers as a tombstone and keeps its internal IP in `Status.Addresses`.
  - Even after the Node is gone, `getMachineForMember` falls back to `FindMachineByNodeInternalIP` and finds that tombstone → `machine != nil` → `removeMemberWithoutMachine` skips it (line 336).
  - Its only removal route is `attemptToScaleDown`, which this PR still leaves behind `IsRevisionStable`. So the deadlock persists for that member: the Machine can't finalize until the member is removed, and
  member removal is gated by revision stability, which the member's own endpoint churn will slow down.
  
Trying to confirm on https://redhat.atlassian.net/browse/OCPBUGS-105882 whether it was the half orphan case that brought this about since that will not be resolved by this.

Either way we may want to account for that by conditioning fast removal in the orphan case on **Node absence** rather than Machine absence
i.e "remove an **unhealthy** member whose **Node is gone** (Machine absent or pending deletion), regardless of revision stability."

hasbro17:
Also it's finally time to introduce unit tests for the `sync()` function since the ordering of the these gates is causing this bug and we only have unit tests for the helpers in isolation. 
Initially the sync was more straightforward but not anymore.

And currently nothing is testing these changes in sync() to validate we don't break the baseline cases based on revision rollouts and member topology.

Quick tally from claude:

Core cases (assert on resulting member list + returned error):

  | # | Revision | Member topology | Expected |
  |---|----------|-----------------|----------|
  | A | **in progress** | true orphan (no machine, no node, unhealthy) | **removed** — proves the fix |
  | B | **in progress** | voting member pending deletion, node+machine present, unhealthy, replacement present | **not removed** — #1540 guard (scale-down stays gated) |
  | C | **in progress** | learner pending deletion | **not removed** — learner path stays gated |
  | D | stable, endpoints match | voting member pending deletion + replacement, quorum ok | **removed** via scale-down — baseline intact |
  | E | stable, **endpoints lag** live | any removable member | **nothing removed, nil error** — `isEtcdEndpointsUpdated` gate |
  | F | any | orphan present, **MachineAPI not functional** | **nothing removed, nil** — `IsFunctional` gate |
  | G | any | **bootstrap not complete** | **nothing removed, nil** |
  | H | in progress | orphan whose `MemberRemove` errors | sync **returns the aggregated error** (not swallowed) — covers new aggregation |
  | I | in progress | no machine, no node, **reported healthy** | **current:** returns error (→Degraded). If Finding 2 adopted: no error, skipped |

  Two more  if we broaden the orphan predicate to "node gone + unhealthy member":

  | # | Revision | Member topology | Expected under Finding 0 |
  |---|----------|-----------------|--------------------------|
  | J | **in progress** | half-orphan: node gone, machine has `DeletionTimestamp`, unhealthy | **removed** (today: not removed) — the MHC fix |
  | K | **in progress** | node **present**, machine present, unhealthy (rollout-transient) | **not removed** — proves node-absence discriminator doesn't over-remove during pod reinstall |

hasbro17:
Closing in favor of https://github.com/openshift/cluster-etcd-operator/pull/1688 which addresses the half orphan case and targets the assisted-installer node deletion case.

PR #1688, opened 2026-08-24
OCPBUGS-105882: Remove dead etcd members during a revision rollout

hasbro17:
## What

Remove an etcd member whose backing Node is gone and that etcd reports
unhealthy, even while a static-pod revision rollout is in progress. This runs
ahead of the revision-stability gate; the quorum-sensitive scale-down and
learner-removal paths still wait for revisions to stabilize.

This supersedes #1684. That PR moved orphan removal ahead of the revision gate
but only for members with no Machine, so it missed the case where the Machine is
still present as a tombstone pending deletion. Keying on Node absence instead of
Machine absence covers both.

## Why

After a control-plane node is replaced, the stale (unhealthy) etcd member keeps
the `etcd-endpoints` configmap churning, which keeps rolling new revisions, which
keeps `IsRevisionStable()` false. `ClusterMemberRemovalController.sync()` bailed
out early whenever revisions were unstable, so the member was never removed and
the churn sustained itself. The stale member was both the cause of the
instability and the thing the instability blocked us from removing.

The fix keys removal on **Node absence** rather than Machine absence. A revision
rollout never deletes the Node object (it only reinstalls the static pod), so:

- a member with no Node has no running etcd pod and is genuinely dead, safe to
  remove regardless of revision state;
- a member that is only transiently unhealthy mid-rollout still has its Node and
  is left alone.

The scenario this addresses is the assisted-installer late-binding master
replacement flow: a manual bind/unbind of the host through the Kube API (the
Agent CR). The old Node and Machine are torn down out of band, leaving the
member as a true orphan (no Machine and no Node) or a half-orphan (Machine still
present as a tombstone pending deletion); both are now removed.

MachineHealthCheck / CPMS remediation is unaffected: there the etcd `PreDrain`
hook keeps the Machine and the Node alive until the member is first removed, so
the member still has a Node and continues to be handled by the gated scale-down
path.

Removing an orphan member only lowers etcd's quorum threshold and never removes
a healthy member, so it is always quorum-safe and needs no quorum check. The
Machine-API-functional precondition is retained (it is required to classify
whether a member's Machine still exists).

## Testing

- New `TestSync` table test covering the full gate ordering: dead members
  (true orphan and half-orphan) are removed during a rollout; scale-down and
  learner removal stay gated; machine-API / bootstrap / endpoints-lag gates; and
  that an error from the dead-member removal is surfaced (not swallowed) when the
  revision is unstable.
- `TestClusterMemberRemovalController` updated for the generalized predicate,
  with new cases for the half-orphan and the healthy-but-unbacked error.
- `make build` and `make verify` pass.

## Reference: assisted-installer late-binding unbind flow

`assisted-service` `internal/controller/controllers/agent_controller.go`:
- `unbindHost` — hub-side DB state only.
- `cleanUnboundSpokeNode` → `removeSpokeResources` (spoke-side):
  - deletes the Node from the spoke;
  - finds the Machine via the `machine.openshift.io/machine` annotation on the node;
  - annotates the Machine with `machine.openshift.io/delete-machine: true` and
    `machine.openshift.io/exclude-node-draining: true`;
  - for masters (no MachineSet): calls `Delete` on the Machine (and the BMH).

Assisted does not touch etcd membership/configmaps; it expects the etcd operator
to remove the member once the Node/Machine are gone. Our predicate (Node gone +
Machine absent-or-pending-deletion + member unhealthy) matches both end-states of
this flow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of unhealthy cluster members when their associated nodes are missing or being deleted.
  * Prevented unsafe member removal during revision rollouts and while etcd membership is out of sync.
  * Added live health checks before removing orphaned members.
  * Preserved accumulated errors to improve reliability and diagnostics.

* **Tests**
  * Expanded coverage for bootstrap states, pending deletions, learner members, quorum health, revision stability, and endpoint synchronization.

### DIFF
--- a/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
+++ b/pkg/operator/clustermemberremovalcontroller/clustermemberremovalcontroller.go
@@ -123,6 +123,29 @@ func (c *clusterMemberRemovalController) sync(ctx context.Context, syncCtx facto
 		return nil
 	}
 
+	// only attempt member removal if the machine API is functional. This is a precondition
+	// for classifying a member as dead: removeUnhealthyMemberWithoutNode decides whether a
+	// member's Machine still exists via the machine lister, which is only meaningful when the
+	// Machine API is functional.
+	if isFunctional, err := c.machineAPIChecker.IsFunctional(); err != nil {
+		return err
+	} else if !isFunctional {
+		return nil
+	}
+
+	var errs []error
+
+	// Remove members whose backing Node is gone and that are reported unhealthy, regardless of
+	// revision stability. A revision rollout never deletes the Node object (it only reinstalls the
+	// static pod), so "Node gone + unhealthy" reliably identifies a genuinely decommissioned member
+	// rather than one that is transiently unhealthy mid-rollout. Such a member is dead: its static
+	// pod cannot be running without a Node. Leaving it in the membership keeps churning the
+	// etcd-endpoints configmap, which keeps rolling revisions, which would otherwise block its own
+	// removal behind the revision gate below (OCPBUGS-105882).
+	if err := c.removeUnhealthyMemberWithoutNode(ctx); err != nil {
+		errs = append(errs, err)
+	}
+
 	// Removing multiple members in the middle of a revision rollout can cause API unavailability
 	// when simultaneously deleting multiple machines with the controlplanemachineset "OnDelete" strategy,
 	// i.e we have multiple machines pending deletion and multiple new replacements added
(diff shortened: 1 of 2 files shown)


hasbro17:
That PR description is making my head hurt ☹️  even though I know what's going on. Will need to prune that.
Opus 4.8 loves a good word salad.

hasbro17:
/hold

Need to verify via the e2e scaling tests

hasbro17:
/hold

It would be best to get e2e coverage from the assisted-installer side. 

The existing etcd vertical-scaling suite tests normal Machine deletion and replacement through CPMS or Machine API. Its pre-drain hook intentionally keeps the Node and Machine involved until the etcd member is removed.
Reproducing the assisted-installer state in the vertical-scaling suite would require provider-specific host termination and probably other convoluted and fragile steps.

This bug is triggered by the Agent late-binding unbind flow, which deletes the old host’s Node and Machine out of band while leaving its etcd membership behind.
An assisted-installer e2e can perform the actual Agent unbind flow and verify that the stale etcd member is removed afterward.

/cc @shay23bra

shay23bra:
> /hold
> 
> It would be best to get e2e coverage from the assisted-installer side.
> 
> The existing etcd vertical-scaling suite tests normal Machine deletion and replacement through CPMS or Machine API. Its pre-drain hook intentionally keeps the Node and Machine involved until the etcd member is removed. Reproducing the assisted-installer state in the vertical-scaling suite would require provider-specific host termination and probably other convoluted and fragile steps.
> 
> This bug is triggered by the Agent late-binding unbind flow, which deletes the old host’s Node and Machine out of band while leaving its etcd membership behind. An assisted-installer e2e can perform the actual Agent unbind flow and verify that the stale etcd member is removed afterward.
> 
> /cc @shay23bra

@hasbro17 
The operator unit tests cover the new gate ordering and orphan/half-orphan cases. Since the existing scaling E2E does not reproduce Assisted’s late-binding unbind behavior, adding a new cross-component E2E specifically for this fix seems disproportionate. I suggest validating the actual unbind workflow through Assisted Installer QE against the fixed operator, and tracking permanent Assisted-side regression coverage separately.

PR #1659, opened 2026-07-28
OCPBUGS-94106: Add APIServer informer to EnvVarController cache sync

tjungblu:
The EnvVarController reads cipher suites from observedConfig, which is populated by the ConfigObserver watching the APIServer CR. Without waiting for the APIServer informer cache to sync, the EnvVarController can race ahead and read empty observedConfig before the ConfigObserver has run.

Adding the APIServer informer to the EnvVarController's cachesToSync ensures the controller waits for the APIServer informer to complete its initial LIST before running sync(). This narrows the bootstrap timing race that causes intermittent "no supported cipherSuites found" errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment variable reconciliation during startup by ensuring API server configuration is fully synchronized before processing.
  * Environment variable updates now respond to relevant API server configuration changes.
  * Prevented potential timing-related issues that could result in incomplete or empty observed configuration.

PR #1689, opened 2026-08-27
OCPBUGS-94106: fall back to intermediate ciphers during etcd bootstrap

mkowalski:
## What

During bootstrap, `getCipherSuites()` (`pkg/etcdenvvar/etcd_env.go`) hard-errors when the operator's `observedConfig.servingInfo.cipherSuites` is still empty. That error sets `EnvVarControllerDegraded=True`, which blocks the InstallerController from creating the first etcd revision:

```
EnvVarControllerDegraded: no supported cipherSuites not found in observedConfig
InstallerControllerDegraded: missing required resources: [... secrets: etcd-all-certs-0]
StaticPodsAvailable: 0 nodes are active; 3 nodes are at revision 0
```

This is a chicken-and-egg race: the config observer cannot populate `servingInfo.cipherSuites` until the control plane is running, but etcd cannot start until this controller produces `ETCD_CIPHER_SUITES`. On slower platforms (observed on the **gcd** sovereign-cloud dev region) the window can exceed the installer's 45-minute bootstrap deadline, causing **intermittent install failures** (~20% of runs).

## Fix

Only at bootstrap (`LatestAvailableRevision == 0`), fall back to the same `TLSProfileIntermediateType` ciphers the render path already bakes into the bootstrap etcd member (`pkg/cmd/render/env.go:getTLSCipherSuites`). This is cryptographically identical to the cipher set the bootstrap etcd is already running, and is self-correcting — once `observedConfig` converges, the observed ciphers take over. Past bootstrap (`revision > 0`) an empty `observedConfig` is still a genuine convergence failure and continues to surface as degraded.

`getObservedTLSMinVersion()` already tolerates an empty `observedConfig` (`crypto.TLSVersion("")` returns the TLS 1.2 default); a clarifying comment was added.

## Why not #1659

The previously merged mitigation (#1659) added the APIServer informer to the `EnvVarController` cache sync. That only waits for the informer **cache** to sync, not for the config observer to **populate** `observedConfig`, so the race persists on slow environments. Waiting harder cannot fix this (the deadlock above); a bootstrap fallback is required. This PR revives the approach originally proposed in #1645, which stalled for process reasons (JIRA target-version validation, review latency) rather than technical ones.

## Testing

- `go test ./pkg/etcdenvvar/...` passes, including new table-driven `TestGetCipherSuites` (bootstrap fallback at revision 0 for nil/empty `observedConfig`; error preserved at revision > 0; observed ciphers used verbatim when present) and `TestGetObservedTLSMinVersionEmptyObservedConfig`.
- Still needed before merge: payload/e2e validation on `e2e-gcd-ovn`, `e2e-gcd-ovn-techpreview`, and `e2e-gcp-ovn-serial` (failure is intermittent ~20%, so several runs are required to demonstrate the fix).

## Evidence

Representative failing run: `periodic-ci-openshift-release-main-ci-5.0-e2e-gcd-ovn-techpreview/2092492579135295488` (payload `5.0.0-0.ci-2026-08-25-144245`, etcd-operator commit downstream of the #1659 merge): the cipher error fired ~1230× over ~12 min (06:51:48 → 07:03:38 UTC), etcd stuck at revision 0, bootstrap timed out at 07:08:19, then recovered a few minutes later.

---
This PR was drafted with AI assistance (opencode / Claude Opus 4.8). Please verify before merging.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLS bootstrapping when minimum TLS version information is unavailable by defaulting to TLS 1.2.
  * Added a safe fallback to intermediate-profile cipher suites during initial setup.
  * Preserved validation errors when cipher-suite configuration is missing after bootstrapping.
  * Corrected the cipher-suite configuration error message.
  * Improved bootstrap detection to ensure cipher settings update at the appropriate stage.

* **Tests**
  * Added regression coverage for TLS defaults, bootstrap detection, cipher-suite fallback behavior, and empty configuration handling.

### DIFF
--- a/pkg/etcdenvvar/etcd_env.go
+++ b/pkg/etcdenvvar/etcd_env.go
@@ -19,6 +19,7 @@ import (
 	"github.com/openshift/cluster-etcd-operator/pkg/operator/operatorclient"
 	"github.com/openshift/cluster-etcd-operator/pkg/tlshelpers"
 	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/library-go/pkg/operator/bootstrap"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 	"go.etcd.io/etcd/client/pkg/v3/tlsutil"
 
@@ -34,14 +35,15 @@ type envVarContext struct {
 	spec   operatorv1.StaticPodOperatorSpec
 	status operatorv1.StaticPodOperatorStatus
 
-	masterNodeLister        corev1listers.NodeLister
-	masterNodeLabelSelector labels.Selector
-	infrastructureLister    configv1listers.InfrastructureLister
-	networkLister           configv1listers.NetworkLister
-	configmapLister         corev1listers.ConfigMapLister
-	targetImagePullSpec     string
-	etcdLister              operatorv1listers.EtcdLister
-	featureGateAccessor     featuregates.FeatureGateAccess
+	masterNodeLister         corev1listers.NodeLister
+	masterNodeLabelSelector  labels.Selector
+	infrastructureLister     configv1listers.InfrastructureLister
+	networkLister            configv1listers.NetworkLister
+	configmapLister          corev1listers.ConfigMapLister
+	bootstrapConfigMapLister corev1listers.ConfigMapLister
+	targetImagePullSpec      string
+	etcdLister               operatorv1listers.EtcdLister
+	featureGateAccessor      featuregates.FeatureGateAccess
 }
 
 var FixedEtcdEnvVars = map[string]string{
@@ -295,11 +297,8 @@ func getObservedTLSMinVersion(envVarContext envVarContext) (tlsutil.TLSVersion, 
 		return "", fmt.Errorf("couldn't get minTLSVersion from observedConfig: %w", err)
 	}
 
-	// map tls version to string recognized by etcd.
-	// Note: an empty observedMinTLSVersion (e.g. during bootstrap before the
-	// config observer has converged) is tolerated - crypto.TLSVersion("")
-	// returns the default (TLS 1.2) without error, matching the bootstrap
-	// etcd member rendered by pkg/cmd/render/env.go.
+	// map tls version to string recognized by etcd. An empty observedMinTLSVersion
+	// (e.g. during bootstrap) is tolerated: crypto.TLSVersion("") returns TLS 1.2.
 	v, err := crypto.TLSVersion(observedMinTLSVersion)
 	if err != nil {
 		return "", fmt.Errorf("couldn't get minTLSVersion from observedConfig: %w", err)
@@ -335,27 +334,25 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
 	actualCipherSuites := tlshelpers.SupportedEtcdCiphers(observedCipherSuites)
 
 	if len(actualCipherSuites) == 0 {
-		// During bootstrap (revision 0) the observedConfig has not converged yet:
-		// the config observer cannot populate servingInfo.cipherSuites until the
-		// control plane (apiserver/etcd) is running, but etcd cannot start until
-		// this controller produces the ETCD_CIPHER_SUITES env var. Hard-erroring
-		// here degrades the operator and blocks the first etcd revision, which can
-		// exceed the installer's 45-minute bootstrap window on slower platforms
-		// (OCPBUGS-94106). Break the deadlock by falling back to the same
-		// TLSProfileIntermediateType ciphers the render path bakes into the
-		// bootstrap etcd member (see pkg/cmd/render/env.go:getTLSCipherSuites), so
-		// etcd runs with an identical cipher set until observedConfig converges.
-		if envVarContext.status.LatestAvailableRevision == 0 {
+		// While bootstrap is in progress, observedConfig may not be populated yet
+		// (config observer can't converge until etcd runs, but etcd needs these env
+		// vars to start). Fall back to the render path's TLSProfileIntermediateType
+		// ciphers instead of degrading, which would block the first revision past
+		// the installer's bootstrap window (OCPBUGS-94106). Once bootstrap completes,
+		// an empty observedConfig is a real failure and must still degrade.
+		bootstrapComplete, err := bootstrap.IsBootstrapComplete(envVarContext.bootstrapConfigMapLister)
+		if err != nil {
+			return nil, fmt.Errorf("unable to determine bootstrap status for cipherSuites fallback: %w", err)
+		}
+		if !bootstrapComplete {
 			profileSpec := v1.TLSProfiles[v1.TLSProfileIntermediateType]
 			fallbackCiphers := crypto.OpenSSLToIANACipherSuites(profileSpec.Ciphers)
 			actualCipherSuites = tlshelpers.SupportedEtcdCiphers(fallbackCiphers)
-			klog.Warning("getCipherSuites: observedConfig has no cipherSuites during bootstrap (revision 0), falling back to TLSProfileIntermediateType ciphers")
+			klog.Warning("getCipherSuites: observedConfig has no cipherSuites during bootstrap, falling back to TLSProfileIntermediateType ciphers")
 		}
 
-		// Past bootstrap (revision > 0) an empty observedConfig is a genuine
-		// convergence failure and must still surface as degraded.
 		if len(actualCipherSuites) == 0 {
-			return nil, fmt.Errorf("no supported cipherSuites found in observedConfig")
+			return nil, fmt.Errorf("no supported cipherSuites not found in observedConfig")
 		}
 	}
 
--- a/pkg/etcdenvvar/etcd_env.go
+++ b/pkg/etcdenvvar/etcd_env.go
@@ -295,7 +295,11 @@ func getObservedTLSMinVersion(envVarContext envVarContext) (tlsutil.TLSVersion, 
 		return "", fmt.Errorf("couldn't get minTLSVersion from observedConfig: %w", err)
 	}
 
-	// map tls version to string recognized by etcd
+	// map tls version to string recognized by etcd.
+	// Note: an empty observedMinTLSVersion (e.g. during bootstrap before the
+	// config observer has converged) is tolerated - crypto.TLSVersion("")
+	// returns the default (TLS 1.2) without error, matching the bootstrap
+	// etcd member rendered by pkg/cmd/render/env.go.
 	v, err := crypto.TLSVersion(observedMinTLSVersion)
 	if err != nil {
 		return "", fmt.Errorf("couldn't get minTLSVersion from observedConfig: %w", err)
@@ -331,7 +335,28 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
 	actualCipherSuites := tlshelpers.SupportedEtcdCiphers(observedCipherSuites)
 
 	if len(actualCipherSuites) == 0 {
-		return nil, fmt.Errorf("no supported cipherSuites not found in observedConfig")
+		// During bootstrap (revision 0) the observedConfig has not converged yet:
+		// the config observer cannot populate servingInfo.cipherSuites until the
+		// control plane (apiserver/etcd) is running, but etcd cannot start until
+		// this controller produces the ETCD_CIPHER_SUITES env var. Hard-erroring
+		// here degrades the operator and blocks the first etcd revision, which can
+		// exceed the installer's 45-minute bootstrap window on slower platforms
+		// (OCPBUGS-94106). Break the deadlock by falling back to the same
+		// TLSProfileIntermediateType ciphers the render path bakes into the
+		// bootstrap etcd member (see pkg/cmd/render/env.go:getTLSCipherSuites), so
+		// etcd runs with an identical cipher set until observedConfig converges.
+		if envVarContext.status.LatestAvailableRevision == 0 {
+			profileSpec := v1.TLSProfiles[v1.TLSProfileIntermediateType]
+			fallbackCiphers := crypto.OpenSSLToIANACipherSuites(profileSpec.Ciphers)
+			actualCipherSuites = tlshelpers.SupportedEtcdCiphers(fallbackCiphers)
+			klog.Warning("getCipherSuites: observedConfig has no cipherSuites during bootstrap (revision 0), falling back to TLSProfileIntermediateType ciphers")
+		}
+
+		// Past bootstrap (revision > 0) an empty observedConfig is a genuine
+		// convergence failure and must still surface as degraded.
+		if len(actualCipherSuites) == 0 {
+			return nil, fmt.Errorf("no supported cipherSuites found in observedConfig")
+		}
 	}
 
 	observedMinTLSVersion, err := getObservedTLSMinVersion(envVarContext)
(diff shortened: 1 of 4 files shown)


pkg/etcdenvvar/etcd_env.go
```diff
@@ -331,7 +335,28 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
...
+		// the config observer cannot populate servingInfo.cipherSuites until the
+		// control plane (apiserver/etcd) is running, but etcd cannot start until
+		// this controller produces the ETCD_CIPHER_SUITES env var. Hard-erroring
+		// here degrades the operator and blocks the first etcd revision, which can
+		// exceed the installer's 45-minute bootstrap window on slower platforms
+		// (OCPBUGS-94106). Break the deadlock by falling back to the same
+		// TLSProfileIntermediateType ciphers the render path bakes into the
+		// bootstrap etcd member (see pkg/cmd/render/env.go:getTLSCipherSuites), so
+		// etcd runs with an identical cipher set until observedConfig converges.
+		if envVarContext.status.LatestAvailableRevision == 0 {
```
tjungblu: hmm. I still feel like this is band-AId (hueheue), and we should find why the race actually happens and fix it. I can't believe other control plane won't have the same issue with this.

Or we just say: screw it and we always just fallback to the default ciphers whenever the input is not accessible. Whatever happens at runtime, accepting the security risk of the operator rolling out less secure ciphers in some error cases.

tjungblu: @redhat-chai-bot wdyt?

mkowalski: It is totally a quick band-AId. I don't like it as a long-term fix, I just wanted to have some PR that I can run 50 times and see if that even fixes the issue

mkowalski:
# Deep Review — Multi-Specialist Panel

## Disposition: 🔴 REQUEST_CHANGES

One BLOCKING finding (an unrelated error-string reword that breaks a CI symptom matcher). The core fix is sound — the **bugs**, **adversarial**, and **security** specialists all independently confirmed the bootstrap fallback is correct, bounded to revision 0, and **not** a net TLS downgrade — but there is one required change plus several high-value hardening suggestions, the strongest being a drift guard for the duplicated cipher derivation.

<details>
<summary><b>🐛 bugs</b> — 0 blocking, 0 suggestion, 3 note</summary>

- **NOTE**: No functional bugs. Verified `LatestAvailableRevision==0` is a reliable bootstrap discriminator (the CRD enforces `self >= oldSelf`, so it is monotonic within a CR lifetime; spec+status are read atomically via `GetStaticPodOperatorState`). Fallback output is byte-identical to the render path and to the observer's own default (`crypto.DefaultTLSProfileType == Intermediate`).
- **NOTE**: Fallback covers empty *and* all-unsupported cipher sets, but malformed/wrong-type `observedConfig` still errors even at rev0 (defensible; cannot occur in practice since the observer always writes a structured `[]string`).
- **NOTE**: TLS1.3 + fallback composes correctly (emits under the informational `CIPHER_SUITES` key).
</details>

<details>
<summary><b>⚔️ adversarial</b> — 0 blocking, 1 suggestion, 4 note</summary>

- **SUGGESTION**: `LatestAvailableRevision==0` is an imprecise proxy for "fresh bootstrap" — a delete/recreate of the Etcd CR or a DR/restore could show revision 0 on an established cluster and silently substitute Intermediate ciphers instead of going Degraded. Consider an explicit bootstrap gate (e.g. `len(status.NodeStatuses)==0` / bootstrap-complete configmap) in addition to `==0`.
- **NOTE**: Fallback triggers for any etcd-unsupported observed cipher set at rev0, discarding an admin's explicit selection (low reachability).
- **NOTE**: Substitution is signalled only by `klog.Warning` — no Event/condition.
- **NOTE**: Unblock reliably covers only revision 0; a transient degraded window remains if the observer still has not converged when revision advances to 1.
- **NOTE**: Second-guard error message misattributes cause to `observedConfig` if the in-code intermediate profile ever yields empty (render uses the distinct "no valid TLS ciphers found").
</details>

<details>
<summary><b>🔒 security</b> — 0 blocking, 0 suggestion, 2 note</summary>

- **NOTE**: **No downgrade vulnerability.** The render path already bakes Intermediate/TLS1.2 into the bootstrap etcd member unconditionally, and the observer sets `minTLSVersion`+`cipherSuites` together (defaulting to Intermediate). The fallback matches that, is bounded to revision 0, filters via `SupportedEtcdCiphers`, and logs no secrets. An admin's stricter profile is fully enforced once `observedConfig` converges.
- **NOTE**: Warning message is imprecise for the "converged-but-all-unsupported" case (observability nit).
</details>

<details>
<summary><b>🏛️ architecture</b> — 0 blocking, 2 suggestion, 2 note</summary>

- **SUGGESTION**: The `TLSProfiles[Intermediate] → OpenSSLToIANACipherSuites → SupportedEtcdCiphers` chain is now duplicated in `etcd_env.go` and `render/env.go:getTLSCipherSuites`, with parity enforced only by a comment. Extract a shared `tlshelpers` helper so the two cannot drift (that drift would re-introduce the OCPBUGS-94106 class of mismatch).
- **SUGGESTION**: `getCipherSuites` is a defensible layer, but bootstrap detection is a bare `== 0` literal; reuse the named `ceohelpers` idiom. Long-term: default the profile at the config-observer layer so `observedConfig` is never empty.
- **NOTE**: `getCipherSuites` (errors at rev>0) and `getTLSMinVersion` (silently defaults) now apply opposite policies to the same empty `observedConfig` — document or align.
- **NOTE**: The #1659 APIServer informer is complementary, not the fix — keep it, but do not let it read as the guard.
</details>

<details>
<summary><b>🧩 consistency</b> — 1 blocking, 3 suggestion, 1 note</summary>

- **🔴 BLOCKING**: The diff incidentally rewords the error `"no supported cipherSuites not found in observedConfig"` → `"...found in observedConfig"`. This is the literal `match_string` of the `EtcdBootstrapRev0CipherSuitesOCPBUGS94106` Sippy symptom — the reword silently breaks that detector. It is unrelated to the fix (avoidable scope creep). Keep the original string, or update the symptom in lockstep.
- **SUGGESTION**: Duplicated derivation vs render path (same as architecture); the new test re-derives it too (tautological).
- **SUGGESTION**: Test builds status via a raw struct literal instead of the package's `u.StaticPodOperatorStatus(u.WithLatestRevision(...))` helpers.
- **SUGGESTION**: Drop the manual `"getCipherSuites: "` prefix from the klog message to match package style.
- **NOTE**: `klog.Warning` (no-arg) is the correct idiom here despite the repo's Warningf skew.
</details>

<details>
<summary><b>🧪 qa</b> — 0 blocking, 6 suggestion, 1 note</summary>

- **SUGGESTION**: Untested — cipherSuites present but all-unsupported (→ empty) at rev0 vs rev>0.
- **SUGGESTION**: TLS1.3 branch untested; the current assertion (`got["ETCD_CIPHER_SUITES"]`) would *pass* on a dropped value. Assert the exact key (`CIPHER_SUITES` vs `ETCD_CIPHER_SUITES`).
- **SUGGESTION**: Fallback assertion is mirror-logic; pin the concrete 9-cipher literal and add a cross-path drift test vs `render.getTLSCipherSuites`.
- **SUGGESTION**: rev>0 error case asserts only `require.Error`, not the message.
- **SUGGESTION**: Malformed `observedConfig` not tested (must error, not fall back, even at rev0).
- **SUGGESTION**: **No controller-level test** proving `EnvVarControllerDegraded` stays False during a simulated bootstrap — that is the actual OCPBUGS-94106 regression test.
- **NOTE**: Render's `getTLSCipherSuites` is itself untested, weakening the drift argument.
</details>

<details>
<summary><b>📝 writer</b> — 0 blocking, 2 suggestion, 1 note</summary>

- **SUGGESTION**: The pre-existing comment at `envvarcontroller.go:101` now contradicts this fix — it overstates the #1659 informer sync as preventing the race. Downgrade to "partial mitigation" and cross-reference the fallback.
- **SUGGESTION**: The `getObservedTLSMinVersion` comment says the TLS 1.2 default matches "the bootstrap etcd member rendered by pkg/cmd/render/env.go" — but the render path sets **no** `ETCD_TLS_MIN_VERSION`. Reword (the default comes from `crypto.DefaultTLSVersion()`/etcd's built-in default, not render).
- **NOTE**: The #1659 PR-number reference could not be verified from the shallow clone (confirmed correct out-of-band).
</details>

## Panel Synthesis

- **Core fix is correct and safe.** Three independent specialists (bugs, adversarial, security) verified the bootstrap fallback unblocks the first etcd revision, is bounded to revision 0, and is cryptographically identical to what the bootstrap etcd already runs — no net TLS downgrade.
- **One BLOCKING regression**, and it is ironic: the error-string reword breaks the exact `EtcdBootstrapRev0CipherSuitesOCPBUGS94106` symptom that tracks this bug. Cheap to fix.
- **Strongest consensus suggestion** (3 specialists): eliminate the duplicated cipher derivation via a shared helper + drift test — this protects the fix's own correctness invariant.
- **Biggest test gap** (qa): there is no end-to-end controller test proving the operator no longer degrades at bootstrap — i.e. no direct regression test for the bug being fixed.

## Required Actions (before merge)

1. **Revert the error-string change** — restore `"no supported cipherSuites not found in observedConfig"` (or update the `EtcdBootstrapRev0CipherSuitesOCPBUGS94106` symptom in the same change). *(BLOCKING — consistency)*

## Optional Follow-ups (recommended, in priority order)

2. Extract a shared `tlshelpers.IntermediateEtcdCipherSuites()` used by both the fallback and `render.getTLSCipherSuites`; add a drift test asserting the two paths are byte-identical. *(architecture, consistency, qa)*
3. Add a controller-level regression test: empty `observedConfig` at revision 0 → `sync` returns nil, `EnvVarControllerDegraded=False`, fallback ciphers set. *(qa)*
4. Fix the two comment inaccuracies (getObservedTLSMinVersion render attribution; the stale `envvarcontroller.go:101` #1659 comment). *(writer)*
5. Strengthen unit tests: TLS1.3 key assertion, present-but-unsupported ciphers, malformed YAML, assert the rev>0 error message, pin the concrete cipher literal. *(qa)*
6. Consider a more precise bootstrap gate (`len(NodeStatuses)==0` alongside `==0`) and/or distinguish "no ciphers observed" from "observed but all-unsupported"; emit an Event/condition on substitution. *(adversarial, bugs, security)*
7. Minor: use `u.StaticPodOperatorStatus`/`u.WithLatestRevision` in tests; drop the klog prefix; distinct error when the profile itself is empty. *(consistency, adversarial)*

## Stats

- Specialists: 7 (all returned) · Reproducers: 0 (no BLOCKING runtime bug)
- Findings: **1 BLOCKING**, 14 SUGGESTION, 14 NOTE
- Diff: 1 commit, 2 files, +123/-2

---
_This review was generated using AI (opencode / Claude Opus 4.8) via a multi-specialist panel. Please verify before acting on it._

<sub>Generated by /code-review:deep-review</sub>

mkowalski:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-e2e-gcd-ovn-techpreview 30

pkg/etcdenvvar/etcd_env.go
tjungblu: @redhat-chai-bot 

> The deadlock is real and not a fixable "race".
The config observer can't populate servingInfo.cipherSuites until the apiserver is running → the apiserver needs etcd → etcd needs ETCD_CIPHER_SUITES from this controller → this controller reads from the config observer. That's a genuine bootstrap dependency cycle, not a timing race you can fix by waiting harder (as https://github.com/openshift/cluster-etcd-operator/pull/1659 demonstrated).

It's not a **deadlock** if it recovers itself after 10 minutes, right? Because the jira ticket mentions this self-corrects after a while.

So I suggest we either fix whatever sets this to set it **faster**, or when it already is set quickly, we are not reacting fast enough to the change in the observedCipherSuites.

And your observation is obviously wrong, because at that point, you already have a working etcd and apiserver running on the bootstrap node. Otherwise you wouldn't have a synchronized watch cache or the operator even running.

> Post-bootstrap (revision > 0), the error is preserved.

when I delete the static pod operator CRD from the apiserver, the revision will revert to the default 0 until the controller can reconcile from the static pod revisions again. During this time, you will regress the security by reverting to TLS default.

mkowalski:
## OCPBUGS-94106 — blast radius: concentrated on slow-bringup platforms (not fleet-wide)

Fleet data from BigQuery (`ci_analysis_us`, release **5.0**, **Jul 1 – Aug 28**). **Takeaway: after #1659 this is not a fleet-wide fire** — the residual install failures are isolated to **slow / resource-constrained** bringup, where the config observer converges slower than the installer's 45-minute window. Fast clouds are effectively unaffected today.

### Install-bootstrap failure rate by platform tier

| Platform / environment | Fresh-install bootstrap fail rate | Status |
|---|---:|---|
| Fast clouds — azure / gcp (default & serial) | **2–5%** | ✅ mitigated by #1659 |
| gcd (GCP sovereign-cloud) | **14–18%** | ⚠️ still failing |
| Bare-metal IPI (techpreview) | **14–27%** | ⚠️ still failing |
| IBM Z (s390x) / libvirt / multiarch | **18–19%** | ⚠️ still failing |
| OKD (okd-scos, aws) | **14–26%** | ⚠️ still failing |
| **Fleet aggregate — all 829 5.0 install jobs** | **2.9%** (1,558 / 53,633) | concentrated in the rows above |

<sub>\* gcp-disconnected operator-framework jobs run higher (52–60%) but may fold in disconnected-specific causes; a few misconfigured jobs show 100% and are excluded as unrelated. ~753 of the 1,558 failures come from 52 "race-shaped" (intermittent 10–95%) jobs.</sub>

### Why it's "isolated" now: #1659 collapsed the fast-path window (Jul 29)

Weekly count of the source-A cipher signature (`no supported cipherSuites … observedConfig`), fleet-wide:

```
Jun 29    1 ▏
Jul 13    3 █
Jul 20    3 █
Jul 27  170 ██████████████████████████████████   ← peak (pre-fix)
Aug 03   51 ██████████                            ← #1659 merged Jul 29
Aug 10    1 ▏
Aug 17    2 ▎
Aug 24    0
```

### Read
- **Current residual is isolated to slow/constrained bringup** (sovereign gcd, bare metal, IBM Z, OKD). Fast cloud installs are effectively unaffected.
- **Confirmed race-labeled occurrences: 441** (231 source-A cipher + 210 rev0-at-gather) — a **lower bound**; the source-A detector can't see slow-recovering variants (the gcd gap the new symptom now closes).
- **Prioritization caveat (honest framing):** the concentration exists *because* #1659 already closed the fast path. The underlying defect is a **timing race** that was fleet-wide in June (triggered by a library-go bump) and can re-widen on any future bootstrap slowdown — so the proper fallback (this PR) is still warranted; #1659 alone is not a durable fix.

<sub>Data: BigQuery `openshift-gce-devel.ci_analysis_us` (job_labels + junit), release 5.0, 2026-07-01→08-28. Generated with AI assistance (opencode / Claude Opus 4.8); please verify before acting.</sub>

pkg/etcdenvvar/etcd_env.go
tjungblu: @redhat-chai-bot do me a favor and update the corresponding JIRA issue with these findings. Then create another 5.1 release blocking bug to fix for the root cause of this.

PR #1710, opened 2026-09-16
CNTRLPLANE-4414: allow TLS 1.3 profiles without ciphers

kaleemsiddiqu:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLS cipher-suite validation across supported TLS versions.
  * TLS 1.3 configurations can complete with no explicitly provided cipher suites when permitted.
  * TLS 1.2 and older configurations continue to require valid cipher-suite settings and report appropriate errors.
  * Cipher-suite environment output now distinguishes informational TLS 1.3 values from required settings for other TLS versions.


bindata/etcd/quorum-restore-pod.yaml
```diff
@@ -93,7 +93,15 @@ ${COMPUTED_ENV_VARS}
       - |
         #!/bin/sh
         set -euo pipefail
-        
+
+        # TLS 1.3 profiles do not configure cipher suites, so ETCD_CIPHER_SUITES
```
ricardomaraschini: Wouldn't a `TLSv1.2` configuration with an empty Ciphers list cause the same issue ? i.e. I think this comment may be misleading.

bindata/etcd/quorum-restore-pod.yaml
```diff
@@ -102,7 +110,7 @@ ${COMPUTED_ENV_VARS}
           --client-cert-file=$(ETCDCTL_CERT) \
           --client-key-file=$(ETCDCTL_KEY) \
           --client-cacert-file=$(ETCDCTL_CACERT) \
-          --listen-cipher-suites=$(ETCD_CIPHER_SUITES)
+          ${CIPHER_SUITES_ARG}
```
ricardomaraschini: I find it interesting that we aren't passing the TLS version but only the Cipher Suites. How does `readyz` understand what TLS version to use ?

bindata/etcd/restore-pod.yaml
```diff
@@ -158,7 +166,7 @@ ${COMPUTED_ENV_VARS}
           --client-cert-file=$(ETCDCTL_CERT) \
           --client-key-file=$(ETCDCTL_KEY) \
           --client-cacert-file=$(ETCDCTL_CACERT) \
-          --listen-cipher-suites=$(ETCD_CIPHER_SUITES)
+          ${CIPHER_SUITES_ARG}
```
ricardomaraschini: The same questions I had on `quorum-restore-pod.yaml` file apply here as well.

pkg/etcdenvvar/etcd_env.go
```diff
@@ -331,6 +331,15 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
 		return nil, fmt.Errorf("couldn't get cipherSuites from observedConfig: %w", err)
 	}
 
+	observedMinTLSVersion, err := getObservedTLSMinVersion(envVarContext)
+	if err != nil {
+		return nil, fmt.Errorf("unable to compute ETCD_CIPHER_SUITES: %v", err)
```
ricardomaraschini: This error message seems misleading. We have failed to get (or compute, if you prefer) the Minimal TLS Version.

pkg/etcdenvvar/etcd_env.go
```diff
@@ -331,6 +331,15 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
...
+	}
+	if observedMinTLSVersion == tlsutil.TLSVersion13 {
+		// etcd does not allow cipher suites to be configured when only TLS 1.3 is enabled.
+		return nil, nil
```
ricardomaraschini: By returning earlier we have lost the part where we were setting `CIPHER_SUITES` for "informational purposes".

ricardomaraschini (changes requested):
I strongly believe that the best and least intrusive way of fixing this would be (pseudo code):

```go
ciphers := tlshelpers.SupportedEtcdCiphers(observedCipherSuites)
if minTLSVersion == "1.3" && empty(ciphers) {
    ciphers = defaultCiphersFor1.3 
}
```

It seems to me that this would hugely simplify this code as we would not need to change the YAML manifests. Let me know if my reasoning is off.

ricardomaraschini:
> I strongly believe that the best and least intrusive way of fixing this would be (pseudo code):
> 
> ```go
> ciphers := tlshelpers.SupportedEtcdCiphers(observedCipherSuites)
> if minTLSVersion == "1.3" && empty(ciphers) {
>     ciphers = defaultCiphersFor1.3 
> }
> ```
> 
> It seems to me that this would hugely simplify this code as we would not need to change the YAML manifests. Let me know if my reasoning is off.

It just occurred to me that we should be using the data from `observedCipherSuites` directly and not the result of `tlshelpers.SupportedEtcdCiphers`. In other words:

```go
if len(observedCipherSuites) == 0 {
    // set observedCipherSuites to the default according to the min tls version.
}
```

bindata/etcd/quorum-restore-pod.yaml
vincentdephily: It very probably uses the default minversion of TLS 1.2

ricardomaraschini: That would make it PQC incompatible even when using `Modern` profile.

vincentdephily: Sorry, bad wording on my part, I meant `minVersion=1.2 maxVersion=1.3`, which is PQC-compatible as 1.3 will be negociated.

ricardomaraschini: I see. Nonetheless that would be a misbehavior. `Modern` is supposed to set the minimal TLS Version to v1.3. My question still stands: How does `readyz` know what TLS version to use if we only pass a list of Cipher Suites?

vincentdephily: I agree this is a misbehavior, we should send `--listen-tls-minversion`.

pkg/etcdenvvar/etcd_env.go
```diff
@@ -331,6 +331,19 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
 		return nil, fmt.Errorf("couldn't get cipherSuites from observedConfig: %w", err)
 	}
 
+	observedMinTLSVersion, err := getObservedTLSMinVersion(envVarContext)
+	if err != nil {
+		return nil, fmt.Errorf("unable to compute the minimal TLS version: %v", err)
+	}
+
+	// TLS 1.3 profiles legitimately configure no cipher suites: etcd (and Go) do not
```
ricardomaraschini: For sake of accuracy: selecting the "Modern" profile does not cause this issue so it is exclusive of "Custom" profile as it does not accept a list of Ciphers. Can we adjust the comment ?

pkg/etcdenvvar/etcd_env.go
tjungblu, line 324: just going through the original code here:

``` 
func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
	var observedConfig map[string]any
	if err := yaml.Unmarshal(envVarContext.spec.ObservedConfig.Raw, &observedConfig); err != nil {
		return nil, fmt.Errorf("failed to unmarshal the observedConfig: %w", err)
	}
	observedCipherSuites, _, err := unstructured.NestedStringSlice(observedConfig, "servingInfo", "cipherSuites")
	if err != nil {
		return nil, fmt.Errorf("couldn't get cipherSuites from observedConfig: %w", err)
	}

	actualCipherSuites := tlshelpers.SupportedEtcdCiphers(observedCipherSuites)

	if len(actualCipherSuites) == 0 {
		return nil, fmt.Errorf("no supported cipherSuites not found in observedConfig")
	}

	observedMinTLSVersion, err := getObservedTLSMinVersion(envVarContext)
	if err != nil {
		return nil, fmt.Errorf("unable to compute ETCD_CIPHER_SUITES: %v", err)
	}

	envName := "ETCD_CIPHER_SUITES"
	if observedMinTLSVersion == tlsutil.TLSVersion13 {
		// When --tls-min-version is set to 'TLS1.3', etcd does not allow --cipher-suites to also be specified.
		// We still outout an env var for informational purposes.
		envName = "CIPHER_SUITES"
	}

	return map[string]string{
		envName: strings.Join(actualCipherSuites, ","),
	}, nil
}
``` 

I think it would be better if we scoped the length check that we introduced with #1689 to only work with TLS1.2. With 1.3 we do not seem to even expect any ciphers, so it makes no sense to try to assume them here.


I'll make a suggestion below.

pkg/etcdenvvar/etcd_env.go
```diff
@@ -331,6 +331,19 @@ func getCipherSuites(envVarContext envVarContext) (map[string]string, error) {
...
+	// allow selecting the cipher suites used by TLS 1.3. Fall back to the Modern
+	// profile's cipher suites so downstream consumers still have a value and the
+	// empty-list handling below does not incorrectly degrade the operator.
+	if observedMinTLSVersion == tlsutil.TLSVersion13 && len(observedCipherSuites) == 0 {
+		observedCipherSuites = crypto.OpenSSLToIANACipherSuites(v1.TLSProfiles[v1.TLSProfileModernType].Ciphers)
+	}
+
 	actualCipherSuites := tlshelpers.SupportedEtcdCiphers(observedCipherSuites)
 
 	if len(actualCipherSuites) == 0 {
```
tjungblu:
```suggestion
	if observedMinTLSVersion == tlsutil.TLSVersion12  && len(actualCipherSuites) == 0 {
```

and you can remove the default handling above

tjungblu:
/lgtm 
otherwise

pkg/etcdenvvar/etcd_env.go
kaleemsiddiqu: taken care of.

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
done using local cluster run with custom build with this fix.

